PR #22, opened 2017-03-26
Use lazy front-matter regex

reillysiemens:
Using a greedy regex could lead to unintended consequences, like the inability to syntax highlight a unified diff. This resolves issues like that by using `.*?` instead of `.*` so that the closing `+++` is matched as soon as possible.

### DIFF
--- a/src/front_matter.rs
+++ b/src/front_matter.rs
@@ -11,7 +11,7 @@ use errors::{Result, ResultExt};
 
 
 lazy_static! {
-    static ref PAGE_RE: Regex = Regex::new(r"^\n?\+\+\+\n((?s).*(?-s))\+\+\+\n?((?s).*(?-s))$").unwrap();
+    static ref PAGE_RE: Regex = Regex::new(r"^\n?\+\+\+\n((?s).*?(?-s))\+\+\+\n?((?s).*(?-s))$").unwrap();
 }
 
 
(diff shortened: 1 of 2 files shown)


reillysiemens:
Prior to making this modification attempting to build a Markdown page containing a unified diff would fail with an error like
```bash
[2:57:44] tucker@khanti:~/Projects/tuckersiemens.com git:(master*) $ gutenberg build
Building site
Failed to build the site
Error: Error when parsing front matter of file `/home/tucker/Projects/tuckersiemens.com/content/foobar.md`
Reason: expected a table key, found a plus at line 5
```

For reference, the unified diff contained in `foobar.md` was
```diff
--- foo.rs      2017-03-26 02:46:21.462163071 -0700
+++ bar.rs      2017-03-26 02:47:01.167966193 -0700
@@ -1,4 +1,4 @@
 fn main() {
-    let s = "Hello, Foo!";
+    let s = "Hello, Bar!";
     println!("{}", s);
 }
```

Keats:
Thanks! Can you add a test for the case that was failing before so it doesn't happen again?

reillysiemens:
Sure! Would you like the tested content to be specific to unified diffs or just contain some extra `+++`?

Keats:
The simplest test case that would have made it fail, so I guess some extra `+++`

reillysiemens:
I think that test does the trick. Let me know if the name doesn't cut it for you.

Keats:
That works, thanks!

PR #26, opened 2017-03-30
Add current_url and current_path

Keats:
To every templates, useful when doing navigation to know which link is currently
active


zoosky:
Great! I'll look into it asap.

Keats:
(merging for now, let me know if you something is missing in the issue)

PR #27, opened 2017-04-01
Copy only modified files in static directory

ecarrara:
Copy only static files with different modified time or size.

I created a function called `copy_if_modified` that copies the file only if the destination file has the different size or modification date of the source file.

_I'm learning rust now, all suggestions are welcome!_

### DIFF
--- a/src/site.rs
+++ b/src/site.rs
@@ -1,6 +1,6 @@
 use std::collections::{BTreeMap, HashMap};
 use std::iter::FromIterator;
-use std::fs::{remove_dir_all, copy};
+use std::fs::{remove_dir_all, copy, create_dir_all};
 use std::path::{Path, PathBuf};
 
 use glob::glob;
@@ -229,6 +229,9 @@ impl Site {
     pub fn copy_static_file<P: AsRef<Path>>(&self, path: P) -> Result<()> {
         let relative_path = path.as_ref().strip_prefix(&self.static_path).unwrap();
         let target_path = self.output_path.join(relative_path);
+        if let Some(parent_directory) = target_path.parent() {
+            create_dir_all(parent_directory)?;
+        }
         copy(path.as_ref(), &target_path)?;
         Ok(())
     }
--- a/src/site.rs
+++ b/src/site.rs
@@ -67,6 +67,7 @@ pub struct Site {
     pub tera: Tera,
     live_reload: bool,
     output_path: PathBuf,
+    static_path: PathBuf,
     pub tags: HashMap<String, Vec<PathBuf>>,
     pub categories: HashMap<String, Vec<PathBuf>>,
 }
@@ -91,7 +92,8 @@ impl Site {
             sections: BTreeMap::new(),
             tera: tera,
             live_reload: false,
-            output_path: PathBuf::from("public"),
+            output_path: path.join("public"),
+            static_path: path.join("static"),
             tags: HashMap::new(),
             categories: HashMap::new(),
         };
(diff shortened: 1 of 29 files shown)


src/site.rs
```diff
@@ -346,7 +340,6 @@ impl Site {
 
     /// Builds the site to the `public` directory after deleting it
     pub fn build(&self) -> Result<()> {
-        self.clean()?;
```
Keats: should keep that

src/utils.rs
```diff
@@ -58,6 +64,59 @@ pub fn find_content_components<P: AsRef<Path>>(path: P) -> Vec<String> {
...
+        if target_metadata.len() == source_metadata.len() &&
+            target_metadata.modified()? == source_metadata.modified()? {
+            return Ok(())
+        }
+    }
+
+    copy(&source, &target)?;
+    let accessed_time = source_metadata.accessed()?;
+    let modified_time = source_metadata.modified()?;
+    utime(target, &accessed_time, &modified_time)
```
Keats: why are the 3 lines above needed? We should be able to just return `copy(&source, &target)`

src/utils.rs
```diff
@@ -58,6 +64,59 @@ pub fn find_content_components<P: AsRef<Path>>(path: P) -> Vec<String> {
...
+    let source = source.as_ref();
+    let source_metadata = metadata(source)?;
+
+    if target.exists() {
+        let target_metadata = metadata(target)?;
+
+        if target_metadata.len() == source_metadata.len() &&
+            target_metadata.modified()? == source_metadata.modified()? {
+            return Ok(())
+        }
```
Keats: You need to remove the target file (`remove_file(&target)?;`) if the file exists and needs to be copied otherwise `copy` will error because the file already exists

Keats (changes requested):
Good idea, thanks!

src/site.rs
ecarrara: This line delete all files on _public_ directory. I want to keep these files and copy only files that have been modified on static directory.

src/utils.rs
ecarrara: We need to copy the modification date of original file. Without this the modification time of target file will be the current time (of copy operation).

src/site.rs
Keats: We want to be able to do a clean build too though. Maybe add an argument to `build` called `clean_first`?

src/utils.rs
Keats: I see, that might not be the best way to do it then if it requires libc and unsafe code.

We get the file path that triggered the notify in the `serve` command so I think a simple method `Site::copy_static_file<P: AsRef<Path>>(path: P) -> Result<()>` would be the easiest. Have to see how notify would handle things like copying several files at once but I believe it would work and be more efficient than going through each file

src/utils.rs
ecarrara: I do not think it's necessary. The `std::fs::copy` docs says 
> `pub fn copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q)`
> ...
> This function will overwrite the contents of to.
> ...
https://doc.rust-lang.org/std/fs/fn.copy.html

ecarrara:
@Keats Do I need to make this work on Windows? :worried:

Keats:
@ecarrara Yes, it needs to work on Windows.
I posted a comment on a outdated diff of a better way to do it imo but I'm not sure Github sent the notification as it was outdated:

>We get the file path that triggered the notify in the serve command so I think a simple method Site::copy_static_file<P: AsRef<Path>>(path: P) -> Result<()> would be the easiest. Have to see how notify would handle things like copying several files at once but I believe it would work and be more efficient than going through each file

That also removes all platform specific that would be an issue if we add more of them but would probably require starting this PR from scratch.

ecarrara:
@Keats sorry for replying too late!

I agree with you about copying only the changed file that triggered de change notification. But I think we should keep the logic "copy only modified files" in build command, like Hugo does[1], to speedup the build time when we have a large number of big files in static directory.

[1] - https://github.com/spf13/fsync/blob/master/fsync.go#L238

Keats:
> But I think we should keep the logic "copy only modified files" in build command, like Hugo does[1], to speedup the build time when we have a large number of big files in static directory.

I think that can be added later on if needed and should default to copying everything if it is added. Let's keep things simple to get started and since you can co-locate static files in the `content` directory, I expect the `static` folder to be quite small anyway compared to Hugo. Even then, I'm not sure it would make a big difference benchmark wise. Same for the `copy_if_modified`.

In short, if we can make that PR be only about the `copy_static_file` method and keep everything else for later if it becomes necessary that would be great.

src/site.rs
```diff
@@ -223,41 +225,37 @@ impl Site {
         html
     }
 
+    /// Copy static file to public directory.
+    pub fn copy_static_file<P: AsRef<Path>>(&self, path: P) -> Result<()> {
+        let relative_path = path.as_ref().strip_prefix(&self.static_path).unwrap();
+        let target_path = self.output_path.join(relative_path);
+        copy(path.as_ref(), &target_path)?;
+        Ok(())
```
Keats: does it work if you are copying a folder that didn't exist? Might need to test manually what happens and create the folders manually by splitting the path components if necessary

Keats:
lgtm after checking the comment!

Keats:
Thanks a lot!

PR #29, opened 2017-04-04
Add markdown and base64 filters

Keats:


Keats:
@sebasmagri does that work for you?

sebasmagri:
It does @Keats. Thanks!

PR #30, opened 2017-04-05
Try to make windows work + remove musl

Keats:


Keats:
Looks like the windows build that's failing is due to https://github.com/rust-lang-nursery/rustup.rs/issues/746

PR #41, opened 2017-04-21
Add more sublime syntaxes

Keats:
Also fix gutenberg serve that has been broken (not
re-rendering content) for quite some time now.


Keats:
@phil-opp is that what you had in mind?

.gitmodules
```diff
@@ -0,0 +1,21 @@
+[submodule "sublime_syntaxes/Packages"]
+	path = sublime_syntaxes/Packages
+	url = git@github.com:sublimehq/Packages.git
```
phil-opp: I think this `ssh` URL causes problems on travis and `cargo install`.

sublime_syntaxes/Assembly x86.sublime-syntax
```diff
@@ -0,0 +1,67 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: Assembly x86 (NASM)
+file_extensions:
+  - asm
+  - inc
```
phil-opp, line 7: Could we also add `nasm` as file extension here?

Keats: It's generated from the tmLanguage file. I can add it but it seems like it would be something that would break when we update.
An alternative would be to only keep the submodule for the official package repo and only have the `.sublime-syntax` files that can be updated manually so it's easier to spot differences like that

phil-opp: Hmm… So the best solution would be to send an upstream PR that changes the tmLanguage file?

I think the `.sublime-syntax` files would suffice since most syntax files are only updated infrequently. It would also make the build process easier by reducing the number of submodules. On the other hand, we might miss some syntax updates…

phil-opp:
I just tested it and it works great! I only miss highlighting for `toml` files and linker scripts. Could we add them too? I found [lmno/TOML](https://github.com/lmno/TOML) for `toml` and [jbw3/SublimeTextLinkerSyntax](https://github.com/jbw3/SublimeTextLinkerSyntax) for linker scripts.

sublime_syntaxes/Assembly x86.sublime-syntax
Keats: Syntax updates is the reason why the submodules are there yeah.

I think doing an upstream PR and keeping submodules is the best way to make it work but some syntaxes are not updated for years so it's hard to say whether it's worth it. I'll think about it this weekend.

phil-opp: I opened https://github.com/Nessphoro/sublimeassembly/pull/22 for this.

Keats:
TOML and Linker added!

phil-opp:
Awesome! TOML highlighting works great, but linker script are still not highlighted. Not sure what's the problem…

Keats:
that's an error in the syntax file:

`file_extentions: [ld]` should be `file_extensions: [ld]`

phil-opp:
Good catch, it works now! I opened https://github.com/jbw3/SublimeTextLinkerSyntax/pull/1 in the upstream repo for this.

Thanks so much for all your support! You're doing awesome work here!

Keats:
> Thanks so much for all your support! You're doing awesome work here!

Thanks a lot!

Keats:
Merging now, I'll update the linker syntax when your PR is merged

Issue #43, opened 2017-04-21
How to use `page.next`/`page.previous` fields?


phil-opp:
I still can't get the `next`/`previous` fields to work. All of the following lead to zero output:

```jinga2
{{ page.next }}
{{ page.next_in_section }}
{{ page.prev }}
{{ page.prev_in_section }}
```

The pages all have a `date` field in their front matter, e.g. `date = "2015-08-25"`. Is there something I'm doing wrong?

Keats:
Hmm I was sure I had tests for that but no so it might be broken. Going to write some now and see why it doesn't work.

I juste realised there might some ambiguity having both `next|prev` and `(next|prev)_in_section`: if you're rendering a section, you're expecting `next` and `prev` to refer to pages in the current section, and if you're rendering the index page for example you're expecting it to be global (I think). I can't think of a moment where you're rendering a section and want to access the global prev/next that might not be in the section?

Keats:
I think I'll do it at the same time as #14 
Curious to see your feedback on the last comment there too!

Keats:
That should be fixed in master!

phil-opp:
Thanks!

PR #47, opened 2017-04-24
Allow sorting pages by order and date

### LINKED ISSUE #14: Allow sorting by weight/order
Sections and some content don't have dates but we will want to order them somehow. Having weight in the front-matter might do the trick (lighter -> top, heavier -> bottom).

Sections should be able to define how they want to be sorted though. In the `_index.md` frontmatter, we can add a `sort_by` key that takes either `"date"` or `"weight"`.
An alternative is to add a `sort_by` filter that can handle that from the template itself.

Probably going to wait for some feedback before starting this.

Keats:
Closes #14, #43

Currently not working because of bug in the TOML crate for ser/de of enums

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Fix XML templates overriding and reloading
 - `title` and `description` are now optional in the front matter
 - Add GenericConfig, Vim syntax
+- Add `_index.md` for homepage as well
+- Allow sorting by `none`, `date` and `order` for sections
 
 ## 0.0.4 (2017-04-23)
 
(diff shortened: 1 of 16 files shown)


phil-opp:
Works great! My only problem is that it only supports descending ordering (e.g. order 0 comes after order 1). So one has to start at a high number if one wants to insert pages at the end later (e.g. order 100, 99, 98, …) or resort to the `reverse` template filter. It would be great if we could specify "ascending" or "descending" directly, or at least could use negative numbers for `order`.

**Edit**: I thought about this a bit more and I think the `reverse` template filter is the best solution for my problem, since I only want to _list_ the pages in reversed order but not insert additional pages before the first page.

Keats:
> I thought about this a bit more and I think the reverse template filter is the best solution for my problem, since I only want to list the pages in reversed order but not insert additional pages before the first page.

That was my thought when I added the sorting. If you want to order things but still be allowed to insert pages in the middle you can increment order by 10 for example. I *think* that's how people do with Hugo but I've only used date sorting myself.

PR #60, opened 2017-05-08
Sections

Keats:


Keats:
What's missing:

- Access section by name, so I can do `sections | find_section(name="posts")` for example. It can't be `sections.posts` as some section name might not be valid tera identifiers. Can be used with the new `{% set .. %}` tera feature. A end-user can iterate on sections instead if necessary instead but that feels a bit clunky

- Warnings for ignored pages: those are pages that can't be sorted in their section. The `build` and `serve` command should display the name/path of the problematic file. 

- Display the number of orphan pages when building/serving the site in the `-> Creating 11 pages and 5 sections` line

Keats:
First point can be made to work by allowing indexing in the Tera parser or adding a generic `get` filter

Issue #63, opened 2017-05-12
Add a highlight filter


Keats:
So we can make custom shortcodes like in http://diesel.rs/ homepage code blocks with a title

Keats:
Not sure if it's actually needed, closing for now

PR #88, opened 2017-06-19
v0.0.8

Keats:


Keats:
@reillysiemens Sass support is now in if you want to try it

I will have to test this version a bit more now and do #63 but otherwise it seems like it wll be shipping soon. Some breaking changes in that though, mostly the removal of the index page getting all the sections automatically: use the `get_section` global fn in the template instead

cc @phil-opp as you are using that feature in your index page

Keats:
Actually sass is not working properly since i did something stupid, should be working this weekend or so

Keats:
Also there seems to be an issue with osx and sass-rs: https://travis-ci.org/Keats/gutenberg/jobs/250763625 if someone has a mac handy

PR #106, opened 2017-07-27
Improve gutenberg init

Keats:


Keats:
@booyaa if you have time, I'm curious about your thoughts on that. Ideally there would be a default theme but I'd rather have easy theme supports right now

booyaa:
@Keats heya, thanks for asking! Can I have a play with this branch to see what the UX is like? I appreciate a theme system would future proof you, but I'll keep further discussion on the relevant issue.

Keats:
Go for it!

Keats:
@booyaa Any feedback? 
Imo it's not that great to have a white page but it's better than erroring like you experienced. Since Hugo still does the white page after so many years, I assume it's not a blocker

booyaa:
sorry i've been away, i'll give this PR a go and provide feedback today.

booyaa:
yup that's definitely better, looking forwards to see how the themes will fit into gutenberg.

PR #114, opened 2017-08-23
Add theme support

Keats:


Keats:
Still missing:

- [x] make a theme using that: https://github.com/Keats/gutenberg-hyde
- [x] rewrite themes extend AST to prepend the `{theme}/templates/` path: it might require a Tera change
- [x] merge theme.toml and config.toml extra variables
- [x] docs
- [ ] find a way to glob on `{html,xml}` instead of `html` for the tera themes: it might require using globset somehow. Globbing on `ml` would include the `theme.toml`

In the future when there are more themes, also investigate a `theme` command eventually that would list available themes and install them (via `git clone` I guess) but I think using `git clone` directly is easier

PR #116, opened 2017-08-26
Shell completions

### LINKED ISSUE #102: Add bash/zsh completions
`clap` supports this I believe

RoPP:
fix: #102 

First commit add `gutenberg completions {bash, zsh, fish, powershell}` command but as `clap_app!` macro doesn't accept `possible_values`, the second one rewrite `build_cli` function with builder pattern style.


Keats:
Thanks for that! I thought generating the completions requires some special package available but can't remember which one right now (the `build.rs` was failing on CI) which is why they were in the repo: would users run into panics?
Rustup does it the same way through a command but its has tons of commands/arguments. Gutenberg has only 3 commands and only one take one argument so I'm not sure it needs completions right now. I'm not planning to add more commands either.

Keats:
Going to close that in the end: if a package manager builds from source it will have access to the repo and we can just update the completion scripts when something in the CLI changes, which shouldn't happen very often. This keeps the CLI itself cleaner.

Thanks though

PR #132, opened 2017-10-18
i18n basic configuration and trans Tera tag

blfpd:
- [x] Config
- [x] trans(key, lang) template tag

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
 #[macro_use]
 extern crate clap;
 
-use clap::Shell;
+// use clap::Shell;
 
 include!("src/cli.rs");
 
--- a/components/config/src/lib.rs
+++ b/components/config/src/lib.rs
@@ -38,7 +38,7 @@ pub struct Config {
     /// Description of the site
     pub description: Option<String>,
     /// The language used in the site. Defaults to "en"
-    pub language_code: Option<String>,
+    pub default_language: Option<String>,
     /// Whether to generate RSS. Defaults to false
     pub generate_rss: Option<bool>,
     /// The number of articles to include in the RSS feed. Defaults to unlimited
@@ -49,6 +49,9 @@ pub struct Config {
     pub generate_categories_pages: Option<bool>,
     /// Whether to compile the `sass` directory and output the css files into the static folder
     pub compile_sass: Option<bool>,
+
+    /// Languages list and translated strings
+    pub translations: Option<HashMap<String, Toml>>,
 
     /// All user params set in [extra] in the config
     pub extra: Option<HashMap<String, Toml>>,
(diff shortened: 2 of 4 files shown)


components/config/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub struct Config {
     /// Description of the site
     pub description: Option<String>,
     /// The language used in the site. Defaults to "en"
-    pub language_code: Option<String>,
+    pub default_language: Option<String>,
```
Keats, line 41: can this be deprecated first? ie leave `language_code` for now and add `default_language` but print a deprecation warning if `language_code` is set

components/templates/src/global_fns.rs
```diff
@@ -8,6 +8,33 @@ use config::Config;
...
 
+pub fn make_i18n(config: Config) -> GlobalFn {
+    let languages = config.languages.unwrap();
+    let default_lang = config.default_language.unwrap();
+
+    Box::new(move |args| -> Result<Value> {
+        match args.get("key") {
+            Some(value) => match from_value::<String>(value.clone()) {
+                Ok(key) => {
+                    match args.get("lang") {
```
Keats: can simplify this by doing `args.get("lang").unwrap_or(default_lang)` (from memory, it might not be `unwrap_or`) to get rid of the match

Keats (changes requested):
Looks like a good start!
I'd rather not have i18n commits on master directly 
until it's somehow working though, I just pushed a `i18n-impl` branch, can you fix the comments and create a PR against that branch please?

blfpd:
Hi @Keats! I rebased from `master` and updated the PR with a compiling commit!
Are you OK with moving the _console_ as a __component__ ?

Keats:
Sorry, missed the email

I don't think it makes sense to have console as a component, components are only there to speed up compilation time but console is not really hurting much there.
Can you get rid of all the unrelated commits?

Keats:
ah nvm I'm silly i forgot about the i18n-impl branch.
Can you do a PR just for the `language_code` -> `default_language` in the config on the `next` branch? No need for deprecations there, it will be a 0.3.0 release so breaking changes are fine

blfpd:
Here the strip down version, with just the conf and the Tera utils function.

components/config/src/lib.rs
Keats: Note: next gutenberg version has breaking changes so it's fine

components/config/src/lib.rs
```diff
@@ -50,6 +50,9 @@ pub struct Config {
     /// Whether to compile the `sass` directory and output the css files into the static folder
     pub compile_sass: Option<bool>,
 
+    /// Languages list and translated strings
+    pub languages: Option<HashMap<String, Toml>>,
```
Keats: bikeshedding but what do you think about `translations` rather than `languages`? it would express that it's translation better I think but not 100% sure so only change if you think the same

blfpd: 👍 I don’t mind bikeshedding when we are on the same page.
I’ll also change i18n() to trans() as discussed on rfc. Closer to Django and easier to type.

build.rs
```diff
@@ -1,8 +1,6 @@
 #[macro_use]
 extern crate clap;
 
-use clap::Shell;
```
Keats, line 4: can you just comment it out rather than deleting it? Sorry for missing it before

Keats:
Thanks!

PR #145, opened 2017-10-21
<documentation/page.md: Fix a typo into page.md doc>

bobsaintcool:
I believe that there is an typo in the page.md documentation where an extra slash could be replaced by a dot.


Keats:
Thanks!

PR #152, opened 2017-10-23
Adding Twitch embed shortcode

iJos:


components/templates/src/builtins/shortcodes/twitch.html
```diff
@@ -0,0 +1,9 @@
+<div {% if class %}class="{{class}}"{% endif %}>
+    <iframe
+        src="https://player.twitch.tv/?channel={{channel}}"
```
Keats, line 3: it's possible to embed chat etc according to https://dev.twitch.tv/docs/v3/guides/embed-chat so maybe call this one `twitch_channel`?

https://discuss.dev.twitch.tv/t/can-you-embed-twitchs-chat-player-to-a-streamers-own-website-that-has-minimal-ads-sponsor-spots-live-samples-in-post/7380/2

Keats (changes requested):
Thanks!

Could you also add the docs for it in `docs/documentation/content/shortcodes.md`?

components/templates/src/builtins/shortcodes/twitch.html
Keats: Or maybe one shortcode that takes a param `chat=bool` to keep things simple and will either render the video or the chat

iJos: We can have 2 shortcodes, one for the chat and one for the stream (twitch_stream and twitch_chat), so they can be added independently to a page. What do you think @Keats?

I'll check it in a couple of hours, also I will add the docs.

Keats: I think a single shortcode is probably better with the following default: `twitch(channel="SOMETHING", chat=false)`. By default if called `twitch(channel="hey")` it will load the video but you can get the chat by doing `twitch(channel="hey", chat=true)`.

I don't think it's worth having 2 shortcodes for that

iJos: I think is done.

Maybe we can change the iframe class on sass file to prevent all iframes to be 100% width (so we can show both video and chat embeds inline?

components/templates/src/builtins/shortcodes/twitch.html
```diff
@@ -0,0 +1,19 @@
+<div {% if class %}class="{{class}}"{% endif %}>
+    <iframe
+        src="https://player.twitch.tv/?channel={{channel}}" 
+        frameborder="0" 
+        allowfullscreen="true"
+        height="{% if videoHeight %}{{videoHeight}}{% endif %}"
+        width="{% if videoWidth %}{{videoWidth}}{% endif %}">  
+    </iframe>
+    
+    {% if chat %}
```
Keats, line 10: i assume there might be time where you only want the chat so I believe it should be one or the other (if chat is true, then no video)

docs/content/documentation/content/shortcodes.md
```diff
@@ -169,3 +169,30 @@ Usage example:
...
+
+### Twitch
+Embed a Twitch stream with or without chat.
+
+The arguments are:
+
+- `channel`: the channel name (mandatory)
+- `chat`: shows the stream chat next to the video embed 
+- `class`: a class to add the `div` surrounding the iframe
+- `videoWidth`: the px **width** the video embed should have
```
Keats, line 181: standard is snake_case, not camelCased

docs/content/documentation/content/shortcodes.md
```diff
@@ -169,3 +169,30 @@ Usage example:
...
+
+The arguments are:
+
+- `channel`: the channel name (mandatory)
+- `chat`: shows the stream chat next to the video embed 
+- `class`: a class to add the `div` surrounding the iframe
+- `videoWidth`: the px **width** the video embed should have 
+- `videoHeight`: the px **height** the video embed should have
+- `chatWidth`: the px **width** the chat embed should have 
+- `chatHeight`: the px **height** the chat embed should have
```
Keats, line 184: when we only get one iframe, only one set of (height,width) will be required

Keats (changes requested):

docs/content/documentation/content/shortcodes.md
iJos: snake_case is ignored (another regex problem too?), we should create a new Issue for that

components/templates/src/builtins/shortcodes/twitch.html
iJos: Well I thought that the video is always visible, I will change that.

docs/content/documentation/content/shortcodes.md
iJos: Sure, I will change it

docs/content/documentation/content/shortcodes.md
Keats: It should be fixed on master

components/templates/src/builtins/shortcodes/twitch.html
iJos: @Keats I'm having a problem with the if statement inside the template... the if - else if being ignored (unless I use the true/false as a string like "true"} Do you know why?

Keats: I'm not sure what you mean, can you detail?

iJos: Sure, I think that with your current code, the shortcode parameters must be strings, I can't use a boolean as a variable because is ignored inside the template

Keats: ah yes it's parsed as strings :(

I'll merge that and make shortcodes take string/bool/numbers

Keats: Actually I pushed a fix to master, can you rebase and try it?

iJos: Should I compile from source or is updated in any package manager?

Keats: From source, its not released yet

iJos: Ok, I will try now

Keats:
Going to close that, the shortcode issues are fixed in the next release and Twitch is not that common to put in Gutenberg directly.

Issue #160, opened 2017-10-29
Markdown refactoring?


reillysiemens:
Hey, @Keats. I've been using Gutenberg for a couple of months to build my personal site and I'm really impressed with the architecture. While working on my site @RadicalZephyr and I encountered an issue. It seems that when arguments are present in a shortcode with a body the paragraph immediately following is not correctly emitted. Unfortunately we have not yet been able to produce a minimal reproduction yet.

This led us to dive into the Markdown parsing code. We found the code pretty challenging to understand and haven't been able to identify what the issue is. One thing we did notice is that it seems like the series of event handlers implicitly specify a state machine. We think it'd be easier to find the bug if we encode the behavior of the numerous mutable state variables as an explicit state machine (possibly multiple state machines). Would you be open to a pull request that restructures the `markdown_to_html` function to be less monolithic and hopefully easier to reason about?

Keats:
The markdown rendering is a pile of hacks, I'd be happy if you can make it more manageable!

Keats:
I released 0.2.2 so you can get started if you want

Keats:
It would be great to add testcases based on #165 when you get to it

reillysiemens:
@Keats: Sure thing! I'm planning to spend some time on this this afternoon. 😁

Keats:
It might be worth looking at https://github.com/kivikakk/comrak as well since pulldown-cmark doesn't really seem maintained. It's about 1.5x-2x slower but it might be ok.

See https://github.com/Keats/markdown-benches-rs for benches

reillysiemens:
@RadicalZephyr and I have noticed that it's hard to get responses from the pulldown-cmark maintainers, but it would be really worrisome if it was truly unmaintained as `rustdoc` just removed hoedown in favor of pulldown recently: https://github.com/rust-lang/rust/pull/48274

FWIW, we're continuing to chip away at this every other weekend or so, but it's been slow going because of just how difficult pulldown-cmark's internals are to deal with. The dream of extensible Markdown rendering in Rust hasn't been realized yet... :sweat_smile:

Keats:
> @RadicalZephyr and I have noticed that it's hard to get responses from the pulldown-cmark maintainers, but it would be really worrisome if it was truly unmaintained as rustdoc just removed hoedown in favor of pulldown recently: rust-lang/rust#48274

I noticed that as well, it got a new release for the Rust project but it doesn't look like other issues are addressed.

[...]

PR #162, opened 2017-10-30
Hack fix and new tests for body-shortcodes

RadicalZephyr:
This fix is a total hack, but it fixes the issue described in #160, without massively restructuring the code.

The essential problem is that shortcodes really _should_ be a "Gutenberg Flavored Markdown" extension but we don't get events that signal "start shortcode" and "end startcode" which would vastly simplify implementing the actual processing of shortcodes.

@reillysiemens, @fuginator, and I think this is a reasonable short term fix to include, but longer term (like maybe for 0.3.0?) the code should probably be restructured to make this way less hacky to implement.

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -70,13 +70,21 @@ pub fn markdown_to_html(content: &str, context: &Context) -> Result<(String, Vec
     // Defaults to a 0 level so not a real header
     // It should be an Option ideally but not worth the hassle to update
     let mut temp_header = TempHeader::default();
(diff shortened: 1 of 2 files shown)


Keats:
Up to you, we can do the cleanup mentioned in the other issue as I think I want to finish and upgrade the version of Tera in the next version. I'm going to publish the current master as 0.2.2 today or tomorrow so we're free to take time after that

RadicalZephyr:
Awesome.  I think this fix is worthwhile if you're doing a new release, and then we can take our time on doing the restructuring!

PR #163, opened 2017-10-30
Extract syntex highlighting module into a new component in workspace

RadicalZephyr:
This removes the dependency cycle between config and rendering that causes 4 packages to be recompiled every time a change is made.

I just want to code fast(er)!


reillysiemens:
This would really help make the dev cycle faster. 👍

Keats:
Very nice, thanks!

PR #172, opened 2017-11-09
Added installation method for Void Linux

memchk:
I have created a package (voidlinux/void-packages#9008) for Void Linux. This PR updates the install docs to show how to install gutenberg if you are a Void Linux user.

On a side note, I have been using gutenberg for the past couple of days and really like it!


Keats:
Glad to hear that and thanks for the package!

Issue #171, opened 2017-11-07
Github pages hates trailing slashes


leoyvens:
I tried hosting a Gutenberg website on github pages but it failed to load any css because Gutenberg appends a trailing slash, as in:
```html
<link rel="stylesheet" href="https://leodasvacas.github.io/poole.css/">
```
I had to manually remove the trailing slashes to get the site to work.

MarcAstr0:
I had a similar problem with Nginx. I had to create a rewrite rule to remove the trailing slash from links to files:

```
rewrite ^/(.*\..*)/$ /$1 permanent;
```

Keats:
Ha that's a bit annoying.
Is the problem only with assets urls or page urls as well? If it's only with the assets, are you generating the urls with the `get_url` global function?

cldershem:
I don't think it's specific to github pages, I assume `poole.css` is a file and therefore, the slash makes the webserver think it should find an `index.html` to serve up. I cannot, however, find where the trailing slash is added.

Keats:
It's coming from `Config::make_permalink` used by the `get_url` global fn: https://github.com/Keats/hyde/blob/master/templates/index.html#L15

That's the right thing to do for pages but not for assets. I guess a `trailing_slash` param to `get_url` could be added with a default of `true` but that you can turn off in the template. Would that solve the issue for everyone in the thread?

leoyvens:
I'm using hyde, if it's fixed there it's fixed for me.

piedoom:
Hi all, I was having a bit of trouble getting Rust to build Gutenberg from source (something about the windows compiler toolchain, nothing you have to worry about).  Anyways, I found a (very hacky) workaround.  Instead of

```
  <script src='{{ get_url(path="js/images.js") }}'></script>
```

you can write

```
  <script src='{{ get_url(path="js") }}images.js'></script>
```

Should hold out until the `next` branch releases.

Keats:
Fixed in 0.3.0 that is currently building in travis/appveyor

PR #173, opened 2017-11-10
Add trailing_slash opt. to get_url

memchk:
This PR addresses issue #171, adding a trailing_slash option as discussed in that thread, which defaults to true. I have updated the documentation and used the new fix on the docs site itself.

Please let me know if any adjustments are needed!


Keats:
Thanks a lot for that!

PR #174, opened 2017-11-13
WIP: Refactor markdown

RadicalZephyr:
Here's our start at rewriting the markdown extension as proposed in #160.  Feedback or suggestions are welcome!


components/rendering/src/markdown.rs
```diff
@@ -1,20 +1,121 @@
...
 use short_code::{SHORTCODE_RE, ShortCode, parse_shortcode, render_simple_shortcode};
 use table_of_contents::{TempHeader, Header, make_table_of_contents};
 
+struct GutenbergFlavoredMarkdownParser<'a, 'b, 'c> {
+    context: &'a Context<'a>,
+    errors: &'c mut Vec<Error>,
+    parser: Peekable<Parser<'b>>,
+}
+
+fn is_internal_link(link: &str) -> bool {
```
Keats, line 23: this is only used once so I don't think it needs to be a function

components/rendering/src/markdown.rs
```diff
@@ -1,20 +1,121 @@
...
+    link.starts_with("./")
+}
+
+impl<'a, 'b, 'c> GutenbergFlavoredMarkdownParser<'a, 'b, 'c> {
+    fn new(context: &'a Context<'a>, parser: Parser<'b>, errors: &'c mut Vec<Error>) -> GutenbergFlavoredMarkdownParser<'a, 'b, 'c> {
+        GutenbergFlavoredMarkdownParser {
+            context: context,
+            errors: errors,
```
Keats, line 31: no need to repeat `context` and `errors`, ` {context, errors, parser: parser.peekable()}` will work

components/rendering/src/markdown.rs
```diff
@@ -1,20 +1,121 @@
...
+                    match resolve_internal_link(href, self.context.permalinks) {
+                        Ok(url) => {
+                            Event::Start(Tag::Link(Owned(url), text.clone()))
+                        },
+                        Err(_) => {
+                            self.push_error(format!("Relative link {} not found.", href));
+                            self.empty_event()
+                        }
+                    }
+                } else{
```
Keats, line 57: space after else

components/rendering/src/markdown.rs
```diff
@@ -1,20 +1,121 @@
...
+                    Event::Start(Tag::Link(href.clone(), text.clone()))
```
Keats, line 58: not possible to return the `link` without cloning?

components/rendering/src/markdown.rs
```diff
@@ -1,20 +1,121 @@
...
+
+    fn next(&mut self) -> Option<Event<'b>> {
+        match self.parser.next() {
+            Some(event) => Some(self.process_event(event)),
+            None => None,
+        }
+    }
+}
+
+#[allow(unused_variables)]
```
Keats, line 90: is that just while it's a wip?

Keats (commented):
Looks like a good start!

Keats:
Ah can you do the PR on the `next` branch by any chance?

components/rendering/src/markdown.rs
RadicalZephyr: whoa! I had no idea! that's awesome

components/rendering/src/markdown.rs
RadicalZephyr: yes, to make fewer warnings.  We'll make sure to remove these and fix warnings before finally submitting.

components/rendering/src/markdown.rs
reillysiemens: Yes, it's only used once, but I think the value here is being explicit about what the parsing is looking for. When simply using `link.starts_with("./")` I think it's not as apparent that what we're really looking for is an internal link.

components/rendering/src/markdown.rs
reillysiemens: The current implementation clones as well, so we just left it while it's still a WIP, but we'll go back at the end and clean up as many clones as possible.

Keats:
I'm guessing this PR can be redone from scratch with the coming changes to the shortcode handling? Or at least significant changes

RadicalZephyr:
Closing this for now, since we're going to regroup a bit on our approach.  We'll open a new PR once we've rebased against the next branch.

PR #181, opened 2017-11-20
Add the detection of ChangeKind::Sass to tests

tvincent2:
Hello,

While looking for how to add a watcher for `config.toml` in `serve`, I noticed that `ChangeKind::Sass` is not tested in `can_detect_kind_of_changes`.
This PR intends to fix that.

Best regards,
Thomas


Keats:
Thanks!

PR #182, opened 2017-11-22
Added j1m.net as a site which uses Gutenberg

j1mc:
Thanks for making this software! I like it so far!


piedoom:
Nice - this isn't related to the PR but I don't know where else to contact - make sure your videos have the correct width, as on mobile they overflow and allow horizontal scrolling! https://photos.app.goo.gl/7hRgXmxhB0NEwrvt2

If you are using Hyde I might open an issue up for a minor css change.

j1mc:
Thanks, @piedoom - I'll take a look at that.

Also, this isn't related to the PR, either, but I'm going to write up a post about how I integrated it with gitlab pages and letsencrypt. Will also cover my docker build (which gets pulled-in by gitlab-ci).

Keats:
> If you are using Hyde I might open an issue up for a minor css change.

It's probably a Hyde issue yeah, surprised it's not covered by the original CSS though.
Looking forward to the post, I'm planning to add a deploy section in the docs but am only using Netlify so there's not much to write on it :)

Issue #183, opened 2017-11-23
Show something other than a blank page when no templates/themes configured


neunenak:
It's confusing that the default behavior when nothing is configured with Gutenberg is to show a blank page without any text at all, corresponding to the created <name>.md files in content. It would be good if the default behavior was at least to show that text with minimal formatting.

Keats:
I agree, I actually started something similar to the Phoenix/Django page when you first install but I wasn't satisfied with how it looks so I got rid of it.

I'm not entiretly sure if we want a single page for all pages where only the template to overwrite is different ("Create a page.html template in the templates directory to overwrite this template") or render the page content in a basic way? I'm afraid if we actually render the content, this will become basically a default theme that was people would use and want to improve.

If anyone is willing to work on that, the easiest way to add some default templates is to add builtins templates in https://github.com/Keats/gutenberg/tree/master/components/templates/src and register them before there https://github.com/Keats/gutenberg/blob/master/components/templates/src/lib.rs#L27
Lastly, you can remove the lines at https://github.com/Keats/gutenberg/blob/master/components/utils/src/templates.rs#L22-L24
No JS and all CSS should be in the `<head>` tag, no external files

BharatKalluri:
Hello @Keats! I would love to contribute to Gutenberg. How do you like the start page to look like?

I am thinking of making a simple page with a heading which says "Welcome to Gutenberg!" and a link to the docs and gitter room under that. Do you have any suggestions?

Keats:
The gitter room is not active but hopefully we'll get a discourse soon but otherwise yes that's pretty much it.

It would also say a sentence like 

> To modify this template, create a file named `blabla.html` in the templates directory. You will then be able to use the following variables: LINK TO PAGE IN THE DOCS

If you work on that, please do the PR on the `next` branch

PR #187, opened 2017-11-26
Gutenberg no longer shows blank page initially

BharatKalluri:
Addresses #183

### DIFF
--- a/components/templates/src/builtins/index.html
+++ b/components/templates/src/builtins/index.html
@@ -1,0 +1,21 @@
+<html>
+    <head>
+        <title>Gutenberg</title>
+    </head>
+    <body>
+        <div class="container">
+            <h1>Hello from gutenberg!</h1>
+            <p>
+                To modify this page, create a index.html file in the templates directory. 
+                More information on how to use gutenberg can be found in the <a href="https://www.getgutenberg.io/documentation/getting-started/installation/" target="_blank">Documentation</a>
+            </p>
+        </div>
+        <style>
+            .container {
+                font-family: "sans-serif";
+                text-align: center;
+                padding-top: 30vh;
+            }
+        </style>
+    </body>
+</html>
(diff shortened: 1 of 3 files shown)


components/utils/src/templates.rs
```diff
@@ -20,7 +20,9 @@ pub fn render_template(name: &str, tera: &Tera, context: &Context, theme: Option
     }
 
     if name == "index.html" || name == "section.html" || name == "page.html" {
-        return Ok(String::new());
+        return tera
+            .render("index.html", &{})
+            .map_err(|e| e.into());
```
Keats, line 25: In the end I think having one default file for each would be better as it allows us to give more descriptive message and we can remove those lines entirely. Right now it would say to modify `index.html` for sections and pages

components/templates/src/builtins/index.html
```diff
@@ -0,0 +1,21 @@
+<html>
+    <head>
+        <title>Gutenberg</title>
+    </head>
+    <body>
+        <div class="container">
+            <h1>Hello from gutenberg!</h1>
```
Keats, line 7: `Gutenberg` (capital G)

components/templates/src/builtins/index.html
```diff
@@ -0,0 +1,21 @@
...
+            <p>
+                To modify this page, create a index.html file in the templates directory. 
+                More information on how to use gutenberg can be found in the <a href="https://www.getgutenberg.io/documentation/getting-started/installation/" target="_blank">Documentation</a>
```
Keats, line 10: > You can find what variables are available in this template in the <a href="https://www.getgutenberg.io/documentation/templates/pages-sections/#section-variables">documentation</a>.

Keats (changes requested):

components/utils/src/templates.rs
BharatKalluri: so, another file in templates for section and pages?

Keats: Another one for each yes so a `section.html` and `page.html`. They can be mostly copy/pasted of the same with changing the template name and doc links

BharatKalluri:
I messed up with branches and did not know how to revert back. so I made another pull request. I'll close this one and open another. sorry about that.

PR #188, opened 2017-11-26
added section and pages templates

BharatKalluri:
Fix for #183.
Fixes for other default pages.

### DIFF
--- a/components/templates/src/lib.rs
+++ b/components/templates/src/lib.rs
@@ -5,12 +5,10 @@ extern crate tera;
 extern crate base64;
 extern crate pulldown_cmark;
 
-#[macro_use]
 extern crate errors;
 extern crate utils;
 extern crate content;
 extern crate config;
-extern crate taxonomies;
 
 pub mod filters;
 pub mod global_fns;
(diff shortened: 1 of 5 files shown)


components/utils/src/templates.rs
```diff
@@ -19,8 +19,23 @@ pub fn render_template(name: &str, tera: &Tera, context: &Context, theme: Option
...
+    if name == "section.html" {
+        return tera
+            .render("section.html", &{})
+            .map_err(|e| e.into());
+    }
+
+    if name == "page.html" {
+        return tera
+            .render("page.html", &{})
+            .map_err(|e| e.into());
```
Keats, line 38: Hmm I remember now why I didn't make them a built-in: to allow themes. That's a bit trickier to handle, I think I'll do it after merging

components/templates/src/lib.rs
```diff
@@ -27,6 +25,7 @@ lazy_static! {
             ("sitemap.xml", include_str!("builtins/sitemap.xml")),
             ("robots.txt", include_str!("builtins/robots.txt")),
             ("anchor-link.html", include_str!("builtins/anchor-link.html")),
+            ("index.html", include_str!("builtins/index.html")),
```
Keats, line 28: in the end, you can remove that line, sorry

Keats (changes requested):
Sorry I just realised it's a bit different than that in the end because of themes :(
I think I'll do it myself in the end as it's a bit different and probably easier that way.

Keats:
Added in https://github.com/Keats/gutenberg/pull/175/commits/129e693521a1a7fbf392d7516f0ce688d0e0194b

PR #191, opened 2017-12-04
Add --output-dir parameters to commands 'serve' and 'build'

verpeteren:
This is useful for example when you want to generate a site to a different document-root.


Keats:
> This is useful for example when you want to generate a site to a different document-root.

What are the reasons against using the default one?

verpeteren:
I'd like to create several themes that can work together with the same content.
Currently, I use `gutenberg build && mv public var/www/something`.

It would be convienient to render different variants by just changing the `theme = something` in the `config.toml` file, `gutenberg --base-url https://example/com/something --output-dir var/www/something`.

(Ideally, an additional '--config-file config_something.toml' parameter, would make this perfect.)

src/cmd/build.rs
```diff
@@ -5,8 +5,9 @@ use site::Site;
 
 use console;
 
-pub fn build(config_file: &str, base_url: Option<&str>) -> Result<()> {
+pub fn build(config_file: &str, base_url: Option<&str>, public: &str) -> Result<()> {
```
Keats: I would name `public` `output_dir` instead, public is not very explicit in the code

src/cmd/serve.rs
```diff
@@ -77,7 +77,7 @@ fn rebuild_done_handling(broadcaster: &Sender, res: Result<()>, reload_path: &st
     }
 }
 
-pub fn serve(interface: &str, port: &str, config_file: &str) -> Result<()> {
+pub fn serve(interface: &str, port: &str, public: &str, config_file: &str) -> Result<()> {
```
Keats: same as above

src/main.rs
```diff
@@ -43,7 +43,8 @@ fn main() {
         ("build", Some(matches)) => {
             console::info("Building site...");
             let start = Instant::now();
-            match cmd::build(config_file, matches.value_of("base_url")) {
+            let public = matches.value_of("output_dir").unwrap_or("public");
```
Keats: Is the `unwrap` needed since there is a default value in `clap`?

src/main.rs
```diff
@@ -54,8 +55,9 @@ fn main() {
         ("serve", Some(matches)) => {
             let interface = matches.value_of("interface").unwrap_or("127.0.0.1");
             let port = matches.value_of("port").unwrap_or("1111");
+            let public = matches.value_of("output_dir").unwrap_or("public");
```
Keats: same as above

docs/content/documentation/getting-started/cli-usage.md
```diff
@@ -33,9 +33,17 @@ You can override the config `base_url` by passing a new URL to the `base-url` fl
...
+This is useful for example when you want to deploy previews of a site to a dynamic URL, such as Netlify
 deploy previews.
 
+You can override the default output directory 'public' by passing a other value to the `output-dir` flag.
+
+```bash
+$ gutenberg build --output-dir $DOCUMENT_ROOT
+```
+
+This is useful for example when you want to generate a site to a different document-root.
```
Keats: This sentence seems a bit redundant, I would remove it

Keats (changes requested):
Few comments but looks good otherwise.
Can you change the PR to be on the `next` branch though please?

src/main.rs
verpeteren: I followed the same structure as :

### interface
https://github.com/Keats/gutenberg/blob/master/src/cli.rs#L38
https://github.com/Keats/gutenberg/blob/master/src/main.rs#L55

### port

https://github.com/Keats/gutenberg/blob/master/src/cli.rs#L43
https://github.com/Keats/gutenberg/blob/master/src/main.rs#L56

Please advice

Keats: I'm just not sure how the default works, I guess you can try removing the `unwrap_or` and see if it still works?

verpeteren: Without the `unwrap_or`, the type goes from `&str` to `Option<&str>`.
It could be made working with `if let Some(x) = ..{}`, but IMHO that would be less clearer then the `unwrap_or`.

What might be interesting, to make the `main.rs` more uniform, is make the same `unwrap_or` for `base_url` change:
* https://github.com/Keats/gutenberg/blob/master/src/main.rs#L46
* https://github.com/Keats/gutenberg/blob/master/src/cmd/build.rs#L10-L12

Please advice.

Keats: >  the type goes from &str to Option<&str>

I guess with a default, it can never be `None` right? So a normal `.unwrap()` would work

Keats:
Looks like tests are failing now. Can you also remove my commit from your branch?

Keats:
There's still my commit in the list. You can probably just merge all your commits and delete mine to fix it

verpeteren:
Sorry; my git knowledge is not sufficient.
That is why I [manually reverted your commit](https://github.com/Keats/gutenberg/pull/191/commits/0b1939223d49dd1064578563cc573ead9fb88689).
Could you guide me through the needed steps?

Keats:
So that would look something like:

```bash
$ git rebase -i HEAD~7
```
 (7 here is a number representing all the commits in the branch, the one at the top should be mine)

On my commit and the manual revert replace `pick` with `d`. On all other commits except your first one, replace `pick` with `s`. That should leave you with a single commit with all the changes that you can `git push -f`.
Another approach but which would require a new PR I believe would be to do `git rebase -i HEAD~6`, drop the manual merge in the interface and use `s` for all commits except the first one and copy that resulting commit hash.
On a new branch based on `next`, you can `git cherry-pick myhash` to move the commit onto that new branch.

Keats:
Ha I think the rebase merging got some earlier commits. Git  😨 
I'll add the changes locally myself and push to the branch directly

Keats:
I've added the changes in https://github.com/Keats/gutenberg/pull/175/commits/14edd2b875cf5ed437229313823656ee1f3aeb45
Thanks!

Issue #178, opened 2017-11-18
`serve` should also listen for changes in config.toml


tvincent2:
Hello,

The `serve` command currently listens for changes to _content_, _templates_, _static_ and _sass_ directories.
When making changes to _config.toml_, I need to manually restart `serve`, which is a bit annoying. Adding it to the list of watched files would be really convenient.

If you think this idea is interesting, I can try to implement it and propose a PR.

Thanks for making gutenberg!
Thomas

piedoom:
Similar to this request, is there a way to listen for changes while developing a theme? Right now I have to restart every time I make a change.

Keats:
For `config.toml`, I'm less against it than before so a PR implementing it properly would be accepted I think.

For themes, no need for that: just treat a theme as a site. For example https://github.com/Keats/hyde is actually a site on my disk where I'm just not committing markdown/config.toml (see the `.gitignore`). Live reload works as expected in that case, the only thing being different from a theme is the `theme.toml`

tvincent2:
Should the PR be for `master` or `next`?

Keats:
On `next` please

Keats:
@tvincent2 are you managing?

tvincent2:
Hello @Keats ,

Adding `config.toml` to the list of watched files was easy, but I ran into some issues for the rebuild part and I've had no time to work on it in the last month.

Hopefully I should have some time next week so expect either something to review or a call for help. :)

Thomas

Keats:
Let me know if i can help!

tvincent2:
Hello,

I have made some progress but:

- Event detection only works the first time I modify `config.toml`, afterwards nothing happens.
- The rebuild part seems ok but I had to duplicate a lot of code from `Site::new` (all the config/template merge) so that would need some refactoring.
- Livereload does not work, but I honestly did not spend time to look into how it is supposed to work and just copy-pasted blindly. :)

Once I have updated my branch with `next`, I will submit a WIP PR if it is ok. Probably tomorrow.

Keats:
Is there any reason you cannot just create a new `Site` instance? Live reload can't work by design as a change in config.toml will need to reload/re-render the whole site anyway

tvincent2:
Of course I just needed to create a new `Site` instance! Thanks a lot @Keats, I was trying to do overcomplicated things.

It all works a lot better now, all I need to do is to test it a little more and rebase my code with `next`.

tvincent2:
[...]

PR #206, opened 2018-01-09
Reload on config change

tvincent2:
Hello!

Here is finally my attempt at closing #178 . Let me know if anything needs to be changed.

Cheers,
Thomas

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -78,7 +78,7 @@ fn rebuild_done_handling(broadcaster: &Sender, res: Result<()>, reload_path: &st
     }
 }
 
-pub fn create_new_site(interface: &str, port: &str, output_dir: &str, config_file: &str) -> Result<(Site, String)> {
+fn create_new_site(interface: &str, port: &str, output_dir: &str, config_file: &str) -> Result<(Site, String)> {
     let mut site = Site::new(env::current_dir().unwrap(), config_file)?;
 
     let address = format!("{}:{}", interface, port);
@@ -207,7 +207,7 @@ pub fn serve(interface: &str, port: &str, output_dir: &str, config_file: &str) -
                             },
                             (ChangeKind::Config, _) => {
                                 console::info(&format!("-> Config changed. The whole site will be reloaded. The browser needs to be refreshed to make the changes visible."));
-                                site = rebuild::after_config_change(interface, port, output_dir, config_file).unwrap();
+                                site = create_new_site(interface, port, output_dir, config_file).unwrap().0;
                             }
                         };
                         console::report_elapsed_time(start);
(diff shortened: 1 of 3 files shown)


Keats:
Can you get rid of the first 3 commits? I'll have a look at the code itself a bit later today

src/cmd/serve.rs
```diff
@@ -196,6 +199,11 @@ pub fn serve(interface: &str, port: &str, output_dir: &str, config_file: &str) -
                                 console::info(&format!("-> Sass file changed {}", path.display()));
                                 rebuild_done_handling(&broadcaster, site.compile_sass(&site.base_path), &p);
                             },
+                            (ChangeKind::Config, _) => {
+                                console::info(&format!("-> Config changed {}", path.display()));
```
Keats: there is only one config file, no need to show the path. The message should say that the whole site will be reloaded and will need to refresh the browser

src/cmd/serve.rs
```diff
@@ -196,6 +199,11 @@ pub fn serve(interface: &str, port: &str, output_dir: &str, config_file: &str) -
...
+                                let path = env::current_dir().unwrap();
```
Keats: no need for that line to be here, it can be in `after_config_change` directly

src/cmd/serve.rs
```diff
@@ -249,6 +257,8 @@ fn detect_change_kind(pwd: &str, path: &Path) -> (ChangeKind, String) {
         ChangeKind::StaticFiles
     } else if path_str.starts_with("/sass") {
         ChangeKind::Sass
+    } else if path_str.starts_with("/config.toml") {
```
Keats: we can do `path_str == "/config.toml"` directly rather than `starts_with`

src/rebuild.rs
```diff
@@ -254,3 +254,21 @@ pub fn after_template_change(site: &mut Site, path: &Path) -> Result<()> {
         },
     }
 }
+
+/// What happens when the config is changed
+pub fn after_config_change(path: &Path, interface: &str, port: &str, config_file: &str) -> Result<Site> {
```
Keats: this looks suspiciously like the the beginning of the `serve` function, can this just be extracted from it and stay in `serve.rs`?

Keats (changes requested):

src/cmd/serve.rs
```diff
@@ -196,6 +205,10 @@ pub fn serve(interface: &str, port: &str, output_dir: &str, config_file: &str) -
                                 console::info(&format!("-> Sass file changed {}", path.display()));
                                 rebuild_done_handling(&broadcaster, site.compile_sass(&site.base_path), &p);
                             },
+                            (ChangeKind::Config, _) => {
+                                console::info(&format!("-> Config changed. The whole site will be reloaded. The browser needs to be refreshed to make the changes visible."));
+                                site = rebuild::after_config_change(interface, port, output_dir, config_file).unwrap();
```
Keats: i think you can just call `create_new_site` here without having anything in `rebuild.rs`

tvincent2: Indeed!

Keats:
thanks!

PR #209, opened 2018-01-12
added remark concerning include-not-in-templates

verpeteren:
See https://github.com/Keats/gutenberg/issues/208


Keats:
I think that note would be better in `documentation/themes/creating-a-theme.md`

Keats:
Thanks!

PR #211, opened 2018-01-15
Update LICENSE

coliff:


Keats:
Should it be a date range (2017-2018)? From the first reply on https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year

coliff:
Some people do and some people don't... A date range might be better though I agree - feel free to edit and merge :-)

PR #212, opened 2018-01-15
added reference to materialize theme and 'who is using'

verpeteren:


Keats:
Thanks!
Looking at the screenshot of the theme (https://github.com/verpeteren/gutenberg-materialize/blob/master/screenshot.png), you should probably add some kind of shadows on text that are meant to be on top of images. It's not very readable otherwise

PR #217, opened 2018-01-23
Proper Arch Linux spelling

svenstaro:
It's either Arch Linux or archlinux but never Archlinux or ArchLinux.


Keats:
woops, thanks!

Issue #221, opened 2018-01-27
Networking/Docker issues regarding interface & base_url


andrewzah:
When running in docker, I need to use --interface 0.0.0.0 to access the site from outside the container, but that also overrides base_url.

If you look at src/cmd/serve.rs - [create_new_site](https://github.com/Keats/gutenberg/blob/master/src/cmd/serve.rs#L84) - one overrides the other.

---

Ergo when running serve for example, it should be possible to set both the interface and the base url.

Something like `gutenberg serve --interface 0.0.0.0 --base-url 127.0.0.1` should be available, or some option within `config.toml`.

Keats:
Hmm I didn't think about that usecase but that's definitely something to fix.
Happy to get a PR for that or i'll take a look next week.

andrewzah:
I can take a crack at it.
Yesterday I [made a (testing) commit](https://github.com/azah/gutenberg/commit/4c8bdfd3e7297311f1fb5de336ebca0f634a7139) changing a few things and adding a `--base-url` flag to `serve`.

It made the generated links follow the base_url, but then in Docker it became unreachable.

andrewzah:
Actually nevermind, I just didn't change the base_url in create_new_site.

So it's reachable via docker (0.0.0.0), but links get set to 127.0.0.1.

andrewzah:
I'll work on a proper PR now - if there's anything I should keep in mind regarding this, let me know.

Keats:
Thanks for that! I believe only the serve command needs to be updated and if you the change should be documented in https://github.com/Keats/gutenberg/blob/master/docs/content/documentation/getting-started/cli-usage.md

azah:
So if I make base_url optional as a flag (in cli.rs), is it possible to read from the config.toml there?
Otherwise I would have to supply a default address, rendering the config.toml's base_url useless.

Keats:
It's in 0.3.1

PR #223, opened 2018-01-29
Separate interface and base url for gutenberg serve

andrewzah:
Implement #221 

You can now have a different base_url and listening interface.
The `gutenberg serve` cmd now accepts `-u`/`--base-url`, but it defaults to
the `base_url` field in `config.toml`.

In `serve.rs`, the websocket address also changes to the given base_url.

---

The CLI doc page was also updated.


andrewzah:
Actually, I made the mistake of only testing this with 127.0.0.1 & 0.0.0.0 as the base_url.

Now that I'm moving over to a sample production deploy w/ Docker,
Rust is now panicking if I change base_url to "andrewzah.com", so I'm going to investigate.

andrewzah:
```
web_1  | thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: WS Error <Io(Error { repr: Os { code: 99, message: "Address not available" } })>', src/libcore/result.rs:906:4
web_1  | note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

It looks like the websocket address fails, predictably, as it tries to listen to an external domain.
Tomorrow I'll just revert the websocket back. However, maybe the websocket port should be customizable also?

src/cmd/serve.rs
```diff
@@ -81,16 +81,21 @@ fn rebuild_done_handling(broadcaster: &Sender, res: Result<()>, reload_path: &st
...
+    };
     let address = format!("{}:{}", interface, port);
-    // Override the base url so links work in localhost
+
     site.config.base_url = if site.config.base_url.ends_with('/') {
-        format!("http://{}/", address)
+        format!("http://{}/", base_address)
     } else {
-        format!("http://{}", address)
+        format!("http://{}", base_address)
```
Keats, line 92: that change is going to break `serve` for everyone. You need to keep overriding the base url if there isn't one passed to the `serve` command otherwise every link will refer to the public domain and it becomes impossible to develop locally

src/cli.rs
```diff
@@ -54,6 +54,11 @@ pub fn build_cli() -> App<'static, 'static> {
                         .default_value("public")
                         .takes_value(true)
                         .help("Outputs the generated site in the given path"),
+                    Arg::with_name("base_url")
+                        .short("u")
+                        .long("base-url")
+                        .takes_value(true)
+                        .help("Force the base URL to be that value (default to the one in config.toml)"),
```
Keats: it defaults to 127.0.0.1 in `serve`

docs/content/documentation/getting-started/cli-usage.md
```diff
@@ -43,14 +43,21 @@ $ gutenberg build --output-dir $DOCUMENT_ROOT
 
 ## serve
 
-This will build and serve the site using a local server. You can also specify
-the interface/port combination to use if you want something different than the default (`127.0.0.1:1111`).
+This will build and serve the site using a local server. The default interface is `127.0.0.1:1111`, and the
+default base_url is the base_url you specified in `config.toml`.
```
Keats: As mentioned before, the `base_url` in `config.toml` is never used in `serve`

Keats (changes requested):

andrewzah:
@Keats yea not sure what exactly I was thinking there.
Remade those changes and the docs, let me know what you think.

Keats:
Looks good, do you have a Dockerfile around so I can try it? as well as #224

andrewzah:
Yeah the site itself is here: https://github.com/azah/personal-blog
That image is part of my overall docker server script: https://github.com/azah/andrewzah.com/tree/master/services/website

The image name is `andrewzah/rust_ms`. Essentially I just copy `public/` to `/www/public` and build Caddy to serve it.

I used gutenberg serve for testing the local docker containers.

Keats:
Thanks I'll have a look.
FYI https://andrewzah.com can't be reached on Chrome and gets a Secure Connection Failed on Firefox

andrewzah:
Yeah, unfortunately I'm moving to a different VPS provider right now. So it should be up by the end of today, hopefully.

Keats:
> I used gutenberg serve for testing the local docker containers.

Can you detail how you do that? The dockerfile just contains the generated files and I see no use of Gutenberg from inside the container

andrewzah:
Ah I updated the Dockerfile, here's what I was using:

```dockerfile
FROM alpine:edge

ARG PORT=1111
EXPOSE $PORT

RUN adduser -S gutenberg

COPY . /srv/gutenberg
RUN apk add --upgrade --no-cache --virtual .build-rust g++ git rust cargo file make cmake \
  && apk add --no-cache libstdc++ libgcc \
  && git clone -b feature/separate_interface_and_base_url https://github.com/azah/gutenberg.git \
  && cd gutenberg \
  && echo "Building gutenberg now." \
  && git submodule sync \
  && RUST_BACKTRACE=1 cargo build --release \
  && mv ./target/release/gutenberg /srv/gutenberg/gutenberg \
  && chown -R gutenberg /srv/gutenberg \
  && rm -rf ./target/ ~/.cargo/ \
  && apk del --purge .build-rust

USER gutenberg
WORKDIR /srv/gutenberg
CMD ./gutenberg serve -p $PORT -i 0.0.0.0 -u 127.0.0.1
```
This adds the various dependencies needed to build rust/gutenberg, builds it, then removes everything. Total size should be around 48MB.

Then you can docker run -p 1111:1111 or use docker-compose:

```yaml
version: '3.3'

services:
  web:
    restart: always
    build:
      context: .
    command: ./gutenberg serve -p $PORT -i 0.0.0.0 -u 127.0.0.1
    ports:
      - "$PORT:$PORT"
```

andrewzah:
Also, the site is back online now, at either [andrei.blue](https://andrei.blue) or [andrewzah.com](https://andrewzah.com).

src/cli.rs
```diff
@@ -54,6 +54,12 @@ pub fn build_cli() -> App<'static, 'static> {
                         .default_value("public")
                         .takes_value(true)
                         .help("Outputs the generated site in the given path"),
+                    Arg::with_name("base_url")
+                        .short("u")
+                        .long("base-url")
+                        .default_value("127.0.0.1")
+                        .takes_value(true)
+                        .help("Changes the base_url. Defaults to 127.0.0.1"),
```
Keats: clap should generate the default in the message automatically so no need for the text about the default

Keats:
Minor comment but looks ok otherwise

Keats:
Thanks!

Issue #100, opened 2017-07-14
Who's using it?


Keats:
- https://vincent.is
- https://adrien.is/

pka:
* http://t-rex.tileserver.ch (https://github.com/pka/t-rex-website/)

crawfxrd:
* https://os.phil-opp.com

reillysiemens:
- https://tuckersiemens.com (https://github.com/reillysiemens/tuckersiemens.com/)

17cupsofcoffee:
https://www.seventeencups.net (https://github.com/17cupsofcoffee/seventeencups.net) :confetti_ball:

Keats:
@17cupsofcoffee Nice! You mention some bugs in your article, is there anything not fixed compared to the next version (0.1.4) https://github.com/Keats/gutenberg/blob/master/CHANGELOG.md?

17cupsofcoffee:
@Keats I had a couple of instances of stuff acting weird with the live reload (might have been when I edited and saved before another build had finished?), but I've not dug into what the issue was yet - nothing that hitting save again/restarting the server couldn't fix though. Other than that, it was pretty much plain sailing! Looking forward to trying out the new version :)

Keats:
Yeah live reload is the part where it's most likely to have noticeable issues as I'm trying to only reload what's needed but I can't guess what's being used in the templates. Maybe it would make sense to default to a dumb reload (just rebuild the whole site - it will be fast enough for most sites) but have a way to enable the current smart reload in `config.toml` since if you have a few thousands pages, it's better than a full rebuild. 

#122 is related if someone wants to tackle it but I don't think it will be possible to write a perfect smart reload short of keeping track what variables/global functions are being used in the templates.

sirodoht:
* https://sirodoht.xyz/ (https://github.com/sirodoht/sirodoht.xyz)

Keats:
List is in the README and probably going to be in the site as well so this issue can be closed

PR #234, opened 2018-02-20
Add tuckersiemens.com to example sites

reillysiemens:
I see that #100 has been closed, but maybe you'd consider including this example site as well? If you figure you've got enough examples as it is don't worry about just closing this.


Keats:
I closed it because I'm going to move it to the website in #219 at some point but I'll still take any!

PR #240, opened 2018-02-25
Filter ignored content in page.rs.

PhilipDaniels:
Fixes https://github.com/Keats/gutenberg/issues/236

* Add ignored_content to the Config structure.
* Use the GlobSet crate to parse the glob patterns into a matcher, which
  is created once at program initialization. If there are no patterns in
  ignored_content, an empty globber is created, which excludes no files.
  This is consistent with the existing behaviour of Gutenberg, before
  this feature was added.
* Bail if there are any errors in the glob patterns.
* Add a call to the globber in page.rs to actually do the filtering.
* Update documentation.

A note on the Config structure
------------------------------
* I had to remove the PartialEq derive from the Config structure as it
  does not work for the GlobSet type. No harm is done, Config does not
  need to be PartialEq anyway, since there is no need to sort Configs.
* The implementation follows the pattern of the existing config settings
  in that it uses an Option<...>. This would appear unnecessary, in that
  an empty vec could be used as the default, but it appears to be needed
  by the TOML parsing. A better approach would be to use a separate
  SerializableConfig and map to/from a Config struct. This would also
  allow the elimination of most, if not all, of the other Options in
  the Config structure, but that ought to be another PR.

### DIFF
--- a/components/config/src/lib.rs
+++ b/components/config/src/lib.rs
@@ -115,14 +115,12 @@ impl Config {
         let mut glob_set_builder = GlobSetBuilder::new();
 
         if let Some(ref v) = config.ignored_content {
-            if v.len() > 0 {
-                for pat in v {
-                    let glob = match Glob::new(pat) {
-                        Ok(g) => g,
-                        Err(e) => bail!("Invalid ignored_content glob pattern: {}, error = {}", pat, e)
-                    };
-                    glob_set_builder.add(glob);
-                }
+            for pat in v {
+                let glob = match Glob::new(pat) {
+                    Ok(g) => g,
+                    Err(e) => bail!("Invalid ignored_content glob pattern: {}, error = {}", pat, e)
+                };
+                glob_set_builder.add(glob);
             }
         }
         config.ignored_content_globber = Some(glob_set_builder.build().expect("Bad ignored_content in config file."));
--- a/components/config/src/lib.rs
+++ b/components/config/src/lib.rs
@@ -5,6 +5,7 @@ extern crate toml;
 extern crate errors;
 extern crate highlighting;
 extern crate chrono;
+extern crate globset;
 
 use std::collections::HashMap;
 use std::fs::File;
@@ -13,6 +14,7 @@ use std::path::{Path, PathBuf};
 
 use toml::{Value as Toml};
 use chrono::Utc;
+use globset::{Glob, GlobSet, GlobSetBuilder};
 
 use errors::{Result, ResultExt};
 use highlighting::THEME_SET;
@@ -22,7 +24,7 @@ mod theme;
 
 use theme::Theme;
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Config {
     /// Base URL of the site, the only required config argument
     pub base_url: String,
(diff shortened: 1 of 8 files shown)


components/config/src/lib.rs
```diff
@@ -97,6 +106,27 @@ impl Config {
...
+
+        // Convert the file glob strings into a compiled glob set matcher. We want to do this once,
+        // at program initialization, rather than for every page, for example. We arrange for the
+        // globset matcher to always exist (even though it has to be an inside an Option at the
+        // moment because of the TOML serializer); if the glob set is empty the `is_match` function
+        // of the globber always returns false.
+        let mut glob_set_builder = GlobSetBuilder::new();
+
+        if let Some(ref v) = config.ignored_content {
+            if v.len() > 0 {
```
Keats: we can drop that `if` and get the same result no?

components/content/Cargo.toml
```diff
@@ -8,6 +8,7 @@ tera = "0.11"
 serde = "1"
 slug = "0.1"
 rayon = "1"
+globset = "0.3.0"
```
Keats: possible to put it in dev-dependencies as it's only used in tests I think

Keats:
A couple of comments but perfect otherwise!

>  A better approach would be to use a separate
SerializableConfig and map to/from a Config struct. This would also
allow the elimination of most, if not all, of the other Options in
the Config structure, but that ought to be another PR.

How would that work? You can have any combination of config so other than `Config` being a hashmap I don't see how to get rid of the options. The `ignored_content` is a bit different here as we need to build the glob.

components/content/Cargo.toml
PhilipDaniels: Good spot, changed.

components/config/src/lib.rs
PhilipDaniels: I don't follow: I need to get a ref to the Vec within the Option. Or do you mean I should consume the ignored_content?

Keats: you can skip the `len()` check

PhilipDaniels:
> How would that work? You can have any combination of config so other than Config being a hashmap I don't see how to get rid of the options. The ignored_content is a bit different here as we need to build the glob.

What I mean is, the fields in the Config struct are defined as Options, and they seem to need to be in order for the current TOML deserialization to work (because they *are* optional in the config file). But if you look at the `parse` function, in fact you set actual defaults for most if not all of them. For example, `compile_sass` is never `None`, it is always true or false, so it could be declared as `bool` rather than `Option<bool>` and then you wouldn't have to unwrap it everywhere.

It's a question of figuring out how to make Config look like that, but still get optional deserialization to work. My thought above was to use a second private struct which would still contain the Options, and use that as the TOML deserialization target, then you could simply construct the public Config with `unwrap_or(false)` or similar defaulting code.

Alternatively, it might be possible to get it to work by adding some `serde` field attributes to specify defaults: but I don't know, I have never used serde. The second-struct technique will definitely work.

components/config/src/lib.rs
PhilipDaniels: Ah, of course. Done.

Keats:
> It's a question of figuring out how to make Config look like that, but still get optional deserialization to work. My thought above was to use a second private struct which would still contain the Options, and use that as the TOML deserialization target, then you could simply construct the public Config with unwrap_or(false) or similar defaulting code

I see. I'm not entirely sure that's worth doing yet though but I like the idea

PR #243, opened 2018-02-28
Add andrei.blue to README

andrewzah:
Forgot to make this a while ago. 

Thanks for making Gutenberg by the way, it's been pretty nice.


Keats:
> Thanks for making Gutenberg by the way, it's been pretty nice.

Thank you!

I added the site in another branch as I already moved the examples somewhere else (see commit above)

PR #244, opened 2018-02-28
Add a bunch of highlight themes, light and dark.

PhilipDaniels:
Add some of what I consider the better (i.e. clearer and more readable) themes from the theme gallery linked to in the README.


Keats:
Thanks! That reminds I should have a look as well for nice themes

PR #247, opened 2018-03-07
Add comparison between SSG

Keats:


README.md
```diff
@@ -7,6 +7,29 @@ An opinionated static site generator written in Rust.
...
+| Search                   |     ✕     |    ✕   |   ✕  |    ✔    |
+| Sane template engine     |     ✔     |    ✔   |  ✕✕✕ |    ✔    |
+| Themes                   |     ✔     |    ✕   |   ✔  |    ✔    |
+| Shortcodes               |     ✔     |    ✕   |   ✔  |    ✔    |
+| Internal links           |     ✔     |    ✕   |   ✔  |    ✔    |
+| Table of contents        |     ✔     |    ✕   |   ✔  |    ✔    |
+| Automatic header anchors |     ✔     |    ✕   |   ✔  |    ✔    |
+| Aliases                  |     ✔     |    ✕   |   ✔  |    ✔    |
+| Pagination               |     ✔     |    ✕   |   ✔  |    ✔    |
+| Custom taxonomies        |     ✕     |    ✕   |   ✔  |    ✕    |
```
Keats, line 30: @epage anything I got wrong for Cobalt or a feature I missed?

README.md
```diff
@@ -7,6 +7,29 @@ An opinionated static site generator written in Rust.
...
 in the `docs/content` folder of the repository.
 
+## Comparisons with other static site generators
+
+|                          | Gutenberg | Cobalt | Hugo | Pelican |
+|--------------------------|-----------|--------|------|---------|
+| Single binary            |     ✔     |    ✔   |   ✔  |    ✕    |
+| Language                 |    Rust   |  Rust  |  Go  |  Python |
+| Syntax highlighting      |     ✔     |    ✔   |   ✔  |    ✔    |
+| Sass compilation         |     ✔     |    ✔   |   ✕  |    ✔    |
```
epage: `sass` is not yet enabled by default yet, see https://github.com/cobalt-org/cobalt.rs/issues/99

README.md
```diff
@@ -7,6 +7,29 @@ An opinionated static site generator written in Rust.
...
+| Assets co-location       |     ✔     |    ✕   |   ✔  |    ✔    |
```
epage: What do you mean by this?

README.md
epage: In case you are interested in linking to the issues
- https://github.com/cobalt-org/cobalt.rs/issues/395 for pagination / ToC
- https://github.com/cobalt-org/cobalt.rs/issues/392 for i18n
- https://github.com/cobalt-org/cobalt.rs/issues/301 for themes
- https://github.com/cobalt-org/cobalt.rs/issues/368 for custom taxonomies (if I understand that Hugo feature correctly)

Also could you clarify something for me.  I assume the only place worth doing pagination is in a ToC, so why are they listed separately?

README.md
```diff
@@ -7,6 +7,29 @@ An opinionated static site generator written in Rust.
...
+| i18n                     |     ✕     |    ✕   |   ✔  |    ✔    |
+| Image processing         |     ✕     |    ✕   |   ✔  |    ✔    |
+| Search                   |     ✕     |    ✕   |   ✕  |    ✔    |
+| Sane template engine     |     ✔     |    ✔   |  ✕✕✕ |    ✔    |
```
epage: What do you mean by "sane"?  Should this instead name the template syntax used with a link to it?

README.md
```diff
@@ -7,6 +7,29 @@ An opinionated static site generator written in Rust.
...
+| Themes                   |     ✔     |    ✕   |   ✔  |    ✔    |
+| Shortcodes               |     ✔     |    ✕   |   ✔  |    ✔    |
+| Internal links           |     ✔     |    ✕   |   ✔  |    ✔    |
+| Table of contents        |     ✔     |    ✕   |   ✔  |    ✔    |
+| Automatic header anchors |     ✔     |    ✕   |   ✔  |    ✔    |
+| Aliases                  |     ✔     |    ✕   |   ✔  |    ✔    |
+| Pagination               |     ✔     |    ✕   |   ✔  |    ✔    |
```
epage, line 29: Should there be a row for listing supported text formats like markdown?

What about data files and supported formats?

epage:
btw Thanks for putting this together!

This will also be helpful for guiding feature development.  What gets highlighted is an interesting insight into what people feel are important.

README.md
epage: - https://github.com/cobalt-org/cobalt.rs/issues/318 and https://github.com/cobalt-org/cobalt.rs/issues/321 for image processing though I was looking at it from a different angle than Hugo.

Keats: I'm not going to link to issues to keep it short.

> I assume the only place worth doing pagination is in a ToC, so why are they listed separately?

Hmm why would have pagination in a ToC? A ToC is like what I'm using for the left menu of https://tera.netlify.com/docs/installation/ for example. 
Pagination is for pages like a blog or sometimes for taxonomies like the `tags/some-tag` list pages where you list all pages.

README.md
Keats: Looking at the issue, I think you should be able to enable it without any issue but i'll change it to a cross in the meantime.

README.md
Keats: Being able to have a folder with images/other files next to the .md like https://github.com/Keats/gutenberg/tree/master/test_site/content/posts/with-assets

README.md
Keats: Mostly whether the template engine will drive you insane when you try to make a nice design. That line is very subjective but I will keep it like that, it's the initial reason of Gutenberg existence after all :)

README.md
Keats: > Should there be a row for listing supported text formats like markdown?

Good point I'll add that.

> What about data files and supported formats?

Do you mean https://gohugo.io/templates/data-templates/ ? I've never used that, I'll need to read it first.

README.md
epage: Ok, didn't look too closely at what a ToC is and assumed it was an index.

README.md
epage: This is exclusively how cobalt works.

I didn't even realize gutenberg supported this because I saw reference to a [`static` folder](https://www.getgutenberg.io/documentation/getting-started/directory-structure/) in the docs but saw no other reference to assets being able to sit next to everything else.

README.md
epage: > Do you mean https://gohugo.io/templates/data-templates/ ? I've never used that, I'll need to read it first.

Yup, [jekyll](https://jekyllrb.com/docs/datafiles/) has it too.

I've not used it yet either, so I'm too familiar with use cases for it.  I just saw it used elsewhere, saw it was trivial to implement with serde, and implemented it :).

README.md
Keats: Ah I missed that, I'll fix it.

Keats: How does it work in Cobalt if you each .md has a `screenshot.png` file with it? Can you put the .md in a folder or do you have to have a unique asset name across the site?

README.md
Keats: I have to say, even the examples in the Hugo and Jekyll docs make it look useless. I can see the `getJSON` or `getCSV` from Hugo being useful but it's orthogonal to the concept of data files. I'll add it for completeness though

Keats:
Thanks for the review!

README.md
epage: I'm not sure of the question which makes me wonder if there is a feature implicit in the question that gutenberg has that cobalt doesn't.

Assets do not need to have unique names.  They are copied straight over from the site source to the destination.

Keats: Have a look at https://github.com/Keats/gutenberg/tree/master/docs/content/themes

This is generated from a git repository and they are all called `screenshot.png`. You obviously can't put several `screenshot.png` in the same folder so you have 2 choices:

- rename them to have unique names
- put them in a subfolder (what I am doing there)

If you put them in a subfolder, you might run into a destination path issue depending on how cobalt is setup up. In the above I want the path to be `themes/after-dark/` for example, not `themes/after-dark/index` without having to edit the path manually in my front-matter. What would be the paths for those pages in Cobalt?

epage: Ok, so the concern is when having files assets with pages that use `path/` instead of `path.html`.

`themes/after-dark/index.md` would be written to disk as `themes/after-dark/index.html`.

cobalts default permalink is `/{{parent}}/{{name}}{{ext}}` and since we are appending the extension, we don't append a `/index.html` to it.

If someone overrides the permalink to not include an extension, then they will run into the problem you mentioned.

PR #255, opened 2018-03-16
Preserve folder structure of SCSS content

ErichDonGubler:
This lets one do something like a mirrored `assets` folder in BOTH `static` and `sass`:

```
-> <root>
 |-> static/
   |-> assets/
     |-> stuff.js
     |-> image.jpg
 |-> sass/
   |-> assets/
     |-> style.scss
```

...so you can keep more assets bundled together like so:

```
-> <root>
 |-> public/
   |-> assets/
     |-> stuff.js
     |-> image.jpg
     |-> style.css

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -528,11 +528,16 @@ impl Site {
         self.copy_static_directories()
     }
 
-    pub fn compile_sass(&self, base_path: &PathBuf) -> Result<()> {
+    pub fn compile_sass(&self, base_path: &Path) -> Result<()> {
         ensure_directory_exists(&self.output_path)?;
 
-        let base_path = base_path.to_string_lossy().replace("\\", "/");
-        let sass_glob = format!("{}/{}", base_path, "sass/**/*.scss");
+        let sass_path = {
+            let mut sass_path = PathBuf::from(base_path);
+            sass_path.push("sass");
+            sass_path
+        };
+
+        let sass_glob = format!("{}/**/*.scss", sass_path.display());
         let files = glob(&sass_glob)
             .unwrap()
             .filter_map(|e| e.ok())
@@ -542,13 +547,16 @@ impl Site {
         let mut sass_options = Options::default();
         sass_options.output_style = OutputStyle::Compressed;
         for file in files {
-            let name = file.as_path().file_stem().unwrap().to_string_lossy();
-            let css = match compile_file(file.as_path(), sass_options.clone()) {
-                Ok(c) => c,
-                Err(e) => bail!(e)
-            };
+            let css = compile_file(&file, sass_options.clone())?;
 
-            create_file(&self.output_path.join(format!("{}.css", name)), &css)?;
+            let path_inside_sass = file.strip_prefix(&sass_path).unwrap();
+            let parent_inside_sass = path_inside_sass.parent();
+            let css_output_path = self.output_path.join(path_inside_sass).with_extension("css");
+
+            if parent_inside_sass.is_some() {
+                create_dir_all(&css_output_path.parent().unwrap())?;
+            }
+            create_file(&css_output_path, &css)?;
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ErichDonGubler** [comment]
D'oh, I just realized I haven't done a `cargo test` yet -- I was in too big of a hurry to get this out before I left the office!

I don't know if CI will flunk this or not -- I'll be back in just a bit to make sure any testing concerns have been addressed. Would it be interesting to add new tests to ensure this new behavior?

**Comment by ErichDonGubler** [comment]
Hey, all passing! :)

File: components/site/src/lib.rs [hunk 1]
```diff
@@ -542,13 +547,19 @@ impl Site {
...
                 Ok(c) => c,
                 Err(e) => bail!(e)
             };
 
-            create_file(&self.output_path.join(format!("{}.css", name)), &css)?;
+            let file_name = file.strip_prefix(&sass_path).unwrap().with_extension("css");
+            let might_need_dir_creation = file_name.parent().is_some();
+            let file_name = self.output_path.join(file_name);
+
+            if might_need_dir_creation {
```
Keats: you can avoid creating the variable `might_need_dir_creation` and the `unwrap` below by using a `if let`

components/site/src/lib.rs
```diff
@@ -528,11 +528,16 @@ impl Site {
...
     }
 
-    pub fn compile_sass(&self, base_path: &PathBuf) -> Result<()> {
+    pub fn compile_sass(&self, base_path: &Path) -> Result<()> {
         ensure_directory_exists(&self.output_path)?;
 
-        let base_path = base_path.to_string_lossy().replace("\\", "/");
-        let sass_glob = format!("{}/{}", base_path, "sass/**/*.scss");
+        let sass_path = {
+            let mut sass_path = base_path.to_owned();
```
Keats: `to_owned()` -> `to_string()` just to be consistent with the rest of the code

Keats (changes requested):

Keats:
> Would it be interesting to add new tests to ensure this new behavior?

Definitely, you can add some subfolders in the `test_site/sass` folder and write tests in the `components/site/tests/site.rs` file.

Can you change the PR to be on the `next` branch as well please?

ErichDonGubler:
I will add some tests. :)

I will rebase onto `next`. Looks like it should rebase cleanly...

components/site/src/lib.rs
ErichDonGubler: This shouldn't be a `String` -- `to_owned` here returns a `PathBuf`.

To make it more clear, I've tried changing this to `PathBuf::from(...)`. What do you think?

components/site/src/lib.rs
ErichDonGubler: There's a reason that this has been done this way, and I realize it's not obvious -- maybe you can help me figure out how to make it more clear.

* My intent is to get the `parent` BEFORE we `join` the `file_name` onto the `output_path` because the `output_path` ALWAYS gives this a parent -- it's nice to avoid system calls.
* However, when we actually call `create_all_dirs`, we want the `join`ed path, because otherwise we'd be missing the `public/` segment.

As an alternative approach, I could always just get the `parent` of the `join`ed path. That means that we'd be unconditionally calling the `create_all_dirs` function, though.

ErichDonGubler:
I don't have time right this second to add tests, but I hope to get to it by the end of the day. Thanks for the rapid response! :)

components/site/src/lib.rs
Keats: woops sorry

components/site/src/lib.rs
ErichDonGubler: I just tried editing this a bit to make the variable names more clear. What do you think?

ErichDonGubler:
Tests have been added. Is there something else I need to fluff up? :)

Keats:
Looks good, thanks!

Issue #249, opened 2018-03-09
Some characters break shortcodes


NuLL3rr0r:
Let's say I have a shortcode named quote:

```
<blockquote>
    <p>
        {{ body }}
    </p>

    -- {% if href %}<a href="{{ href }}">{% endif %}{{ author }}{% if href %}</a>{% endif %}
</blockquote>
```

If I call this short code like this:

```
{% quote(author="C++ Standard Core Language Defect Reports and Accepted Issues, Revision 82, delete and user-written deallocation function") %}
...if the value of the operand of delete is the null pointer the operation has no effect.
{% end %}
```

I'll get this on the terminal when I try to run Gutenberg:

```
thread '<unnamed>' panicked at 'index out of bounds: the len is 1 but the index is 1', /checkout/src/liballoc/vec.rs:1551:10
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

If I replace those commas, with something else or remove them it works. It seems it parses those commas as separators for other arguments or something like that.

Now, If I call my shortcode like this:

```
{% quote(author="C++ Standard Core Language Defect Reports and Accepted Issues; Revision 82; delete and user-written deallocation function", href="http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#348") %}
...if the value of the operand of delete is the null pointer the operation has no effect.
{% end %}
```

The generated html code is going to be like this:

    <p>{% quote(author="C++ Standard Core Language Defect Reports and Accepted Issues; Revision 82; delete and user-written deallocation function", href="http://www.open-std.org/jtc1/sc22/wg21/docs/cwg</p><blockquote>
        <p>
        ...if the value of the operand of delete is the null pointer the operation has no effect.
        </p>
    
        -- <a href="http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#348">C++ Standard Core Language Defect Reports and Accepted Issues; Revision 82; delete and user-written deallocation function</a>
    </blockquote>

Which obviously should not output the part before opening blockquote tag. It happens due to the underscore character in the url. If I change cwg_defects.html to cwgdefects.html for example, it works.

It has been tested with v0.3.2.

Keats:
That's a known issue (dup of #165). There is an ongoing rewrite of the markdown/shortcode renderer by @reillysiemens and @RadicalZephyr that should fix that for 0.4

I might fix the faulty regex in the next minor version at least though

NuLL3rr0r:
[...]

PR #261, opened 2018-03-17
Add Axiomatic Semantics to examples

Libbum:
My site is now live, so just adding it to the examples.

The search isn't working at the moment, but I see you're actively working on building an index in #253.

Now that I have the layout all sorted and everything up and running, I can see that Data Files would be something quite useful for me. So when I've got the time I'll take a look at implementing something there.


Keats:
Thanks!
We discussed a bit data files in https://github.com/Keats/gutenberg/pull/247#discussion_r173852647 but I couldn't find a good usecase for it. Can you open an issue so we can discuss it?

Libbum:
After taking a look at that and thinking about my use case&mdash;you're probably right. Those in particular are pretty useless and my case is probably an edge anyhow. I'll probably be fine once #249 gets sorted out really. Will be on the lookout for another way to contribute. Thanks for building some great tools!

PR #269, opened 2018-03-25
Anchor tag for summary

Libbum:
A simple addition, which allows you to use something like the following on your index page:

```htmldjango
{% block content %}                                                                                                                                    
    {% for page in paginator.pages %}
        <h1>{{ page.title }}</h1>
        {{ page.summary | safe }}
        <a href="{{ page.permalink }}#more">Continue Reading</a>
    {% endfor %}
{% endblock content %}
```

The `#more` anchor link directs you to the position after the summary, rather than having to scroll to this point yourself.

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -170,7 +170,7 @@ impl Page {
             permalinks,
             anchor_insert
         );
-        let res = markdown_to_html(&self.raw_content, &context)?;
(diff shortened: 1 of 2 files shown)


Keats:
Good idea, I would just change the name: `more` is a common word enough that people might have it as a title. No clue what else to use though, maybe `continue-reading`?

Libbum:
Yeah, that makes sense. I thought about another naming idea for a while myself and couldn't think of anything better. `continue-reading` is probably the best alternative. c8dfc51 changes this.

Keats:
Thanks!

PR #291, opened 2018-04-20
Allow creation of specific html aliases

jamesmunns:
Hey there,

I wrote this change before realizing that the existing behavior of taking the alias `foo/bar/baz.html` and creating `foo/bar/baz.html/index.html` worked sufficiently.

I thought I would offer it up in case you would prefer the behavior of creating `foo/bar/baz.html` directly instead of `foo/bar/baz.html/index.html`, but if not, feel free to reject the PR.

Thanks for all the work with Gutenberg, porting my Jekyll based blog has been pretty painless so far!

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -604,13 +604,13 @@ impl Site {
 
                 // If the alias ends with an html file name, use that instead of mapping
                 // as a path containing an `index.html`
-                let page_name: String = match split.pop() {
-                    Some(part) if part.ends_with(".html") => part.to_string(),
+                let page_name = match split.pop() {
+                    Some(part) if part.ends_with(".html") => part,
                     Some(part) => {
                         split.push(part);
-                        "index.html".into()
+                        "index.html"
                     }
-                    None => "index.html".into()
+                    None => "index.html"
                 };
 
                 for component in split {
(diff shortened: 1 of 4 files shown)


Keats:
Would it be possible to add a test in some ways?

Keats:
And can you change the branch target to be `next` as well instead of `master` please?

jamesmunns:
@Keats I have updated the target branch, thanks for the heads up!

I'll also look into writing a test for this, that definitely sounds reasonable. Hopefully I'll get to that by this weekend.

Thanks for the

components/site/src/lib.rs
```diff
@@ -599,14 +599,27 @@ impl Site {
...
+                let page_name = match split.pop() {
+                    Some(part) if part.ends_with(".html") => part,
+                    Some(part) => {
+                        split.push(part);
+                        "index.html"
+                    }
+                    None => "index.html"
+                };
+
+                for component in split {
```
jamesmunns, line 616: Also @Keats, would it make sense to use [`create_dir_all()`](https://doc.rust-lang.org/std/fs/fn.create_dir_all.html) here, rather than iterating over the path components?

Keats: Go for it

Keats:
@jamesmunns Any chance you can update the docs as well to mention that?

jamesmunns:
@Keats yup, I will make sure to do that. I haven't gotten a chance to work on this yet, but I'd still like to.

jamesmunns:
Hey @Keats, I made the requested changes (sorry it took so long!), let me know what you think. My editor removed some trailing whitespace, let me know if you would like me to revert those changes.

Keats:
Perfect!

PR #296, opened 2018-04-26
Send Appropriate Response on Error

m-laniakea:
Users can now place custom error pages in static/error/<error_code>.html
e.g. static/error/404.html

If no custom page is found, a default plaintext describing the error is served.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -22,14 +22,16 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use std::env;
-use std::fs::remove_dir_all;
+use std::fs::{File, remove_dir_all};
 use std::path::Path;
 use std::sync::mpsc::channel;
 use std::time::{Instant, Duration};
 use std::thread;
 
 use chrono::prelude::*;
-use iron::{Iron, Request, IronResult, Response, status};
+use iron::{Iron, IronError, Request, IronResult, Response, status, AfterMiddleware};
+use iron::prelude::Chain;
+use iron::headers::ContentType;
 use mount::Mount;
 use staticfile::Static;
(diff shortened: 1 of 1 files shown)


Keats:
I think the file looked up should be `templates/404.html` but that means we need to ensure it's going to be generated, like the sitemaps etc

m-laniakea:
Would you prefer `templates/<error_code>.html` or `templates/error/<error_code>.html`?

So you want Gutenberg to check for error templates and generate them, is that right?

Keats:
Yes, so nginx/netlify etc can display the 404 if there is one. I'd prefer `templates/404.html`, I don't think there are other error codes to be added afaik.
Also if you can change the branch to do the PR on the `next` branch that would be great!

Serkan-devel:
Maybe something like 503.html ?

Keats:
would 503 be used for static sites?

Keats:
Note that I merged a PR to move from iron to actix-web as iron is unmaintained so this PR would need to be rewritten. I still want it in though

onelson:
Peeking at the actix-web middleware [intro](https://actix.rs/docs/middleware/), seems like the implementation would be virtually identical. Mostly the work would be to switch out the types for their actix counterparts.

As an aside, I host my gutenberg site on s3, and I only remember the option to provide a single generic "error" document. Wonder if we could emulate both behaviours by checking for a status code file, then falling back to check for a generic error file.

Keats:
Anyone wants to pick this up?

onelson:
I can probably take a look over the weekend if someone else doesn't get to it first.

onelson:
I have something compiling now, but need to test and polish. Not sure how best to handle the hand-off for the PR. Close and open a new one?

Keats:
Yes, i'll close this one as I don't think any part of the code can be re-used

Issue #227, opened 2018-02-10
Provide a "snap" as software package — for Linux systems


x80486:
Snaps are containerized software packages that are simple to create and install. They auto-update and are safe to run. And because they bundle their dependencies, they work on all major Linux systems without modification.

If you use snap packaging (Ubuntu's new universal packaging standard), we can submit this application to Ubuntu's Developer Portal and have it show up in the software store.

For more information, see [here](https://snapcraft.io/).

Keats:
I've never heard of snap before. Is it more common to have it as pre-built or build from source?

x80486:
You can go either way. I think depending on the app's nature it will be easier to decide. There is a "language guide" [here](https://docs.snapcraft.io/build-snaps/languages)...you can take a look.

Some snaps looks "pretty dense" — by looking at its `snapcraft.yaml` file.

s0kil:
Does this method work yet? Or still being considered?

Keats:
It currently doesn't work, the snapcraft builder seems to have some issues with rust (see #322 )

Keats:
It looks like it still doesn't work :(
Can anyone on Ubuntu try the snapcraft.yml on master and try to debug it?

musoke:
This is the result of running `snapcraft cleanbuild --debug`:
```
   Compiling bytes v0.4.10
   Compiling futures-cpupool v0.1.8
   Compiling mio v0.6.16
   Compiling resolv-conf v0.6.1
error: failed to run custom build command for `openssl-sys v0.9.39`
process didn't exit successfully: `/root/build_zola/parts/zola/build/target/debug/build/openssl-sys-2b71e6cab2b8c344/build-script-main` (exit code: 101)
--- stdout
cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR
cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR
cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR
cargo:rerun-if-env-changed=OPENSSL_DIR
[...]

PR #299, opened 2018-04-28
Add snacraft config

musoke:
This allows the easy generation of snaps as requested in #227.

This commit does not add automatic generation and publication of snaps
to the Snap Store, but they can easily be generated from source for 
local/personal use by

```
snapcraft
sudo snap install gutenberg_xxx.snap
```

I have not taken the liberty of registering or publishing the app in the Snap Store.  If that is done, updates can be pushed to the store automatically from travis.


Keats:
Thanks!

Should this live outside the repo though? I don't use Ubuntu so I'm unlikely to maintain it myself, I only maintain the AUR repo and I have a separate repo for that (it's required by AUR anyway)

Keats:
What do you think @musoke ?

musoke:
Sorry for the delay @Keats.  Personally I would keep it in the repo so anyone who wants a snap can easily build it themselves.  I don't foresee the need for any maintenance other than changing the version number.  Even that could be eliminated by setting it to something like `git`.

musoke:
Also: Canonical intends that snaps become ubiquitous on distros other than Ubuntu.  Personally I use them mostly to runs apps not yet on Debian stable.

Keats:
I don't mind letting in the repo if I don't really need to do anything, but how would snap knows how to build Gutenberg from the file in the PR or even find this snap file?
Basically what are the steps needed from this PR to a snapcraft user being able to install Gutenberg from snap?

musoke:
I believe that the simplest way to publish a snap to Ubuntu's store would be to [use their build service](https://build.snapcraft.io/).  It seemed easy when I tried it just now, but didn't finish as I didn't think it was appropriate for me to register the package name.  Builds would by default be marked as unstable, so you would not personally be making a stability guarantee to users whom install the snap.

Alternatively, one can [use Travis](https://docs.snapcraft.io/build-snaps/ci-integration). That may be less desirable 

- it looks like gutenberg's build times are already 30 minutes
- you would have to do a bit more than register the snap name
- may involve some maintenance

There is also Launchpad, which I use for a project of my own.  However, this would probably be more work for you.

Keats:
Thanks, I'll try their build service to see how that looks

Keats:
I found https://github.com/theoriginalbigbaddave/Cobalt/blob/master/snap/snapcraft.yaml which is very similar except for the `source` and `source-tag` info and the build-packages: are those not needed?

Sadly installing snapcraft didn't work for my distro so I'll have to test it directly via their build service (I registered the name), any chance you can confirm it builds locally with this snapcraft.yml?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,21 @@
...
+  gutenberg:
+    command: gutenberg
+    plugs:
+      - network
+      - network-bind
+
+parts:
+  gutenberg:
+    plugin: rust
+    rust-channel: nightly
```
Keats: Gutenberg builds on stable, nightly is not 100% safe

Keats (changes requested):

musoke:
I took a look at the snapcraft docs. It seems like `source` is only need when pulling in remote packages, so I don't understand why cobalt uses it.

I just pushed an update to use the stable rust channel.

I tested building this - works for me on Debian Stretch with snapcraft 2.42.1.

Keats:
Cool I'll merge it and try to build it with their service later today, thanks!

Keats:
It fails on the build server: https://build.snapcraft.io/user/Keats/gutenberg/239694

musoke:
I followed the instructions on that page. It starts compiling but runs into a problem with a custom build command for `sass-sys`.  Here is truncated output.

```
   Compiling term-painter v0.2.4
   Compiling rand v0.3.22
   Compiling bytes v0.4.6
   Compiling quote v0.5.1
   Compiling thread_local v0.3.5
   Compiling debug_unreachable v0.1.1
   Compiling syn v0.11.11
   Compiling log v0.3.9
   Compiling idna v0.1.4
   Compiling chrono v0.4.2
   Compiling mio v0.6.14
error: failed to run custom build command for `sass-sys v0.4.4`
process didn't exit successfully: `/root/build_gutenberg/parts/gutenberg/build/target/release/build/sass-sys-d9b94e18146f841e/build-script-build` (exit code: 101)
--- stderr
fatal: Not a git repository (or any of the parent directories): .git
thread 'main' panicked at 'error running make: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

warning: build failed, waiting for other jobs to finish...
error: failed to compile `gutenberg v0.3.3 (file:///root/build_gutenberg/parts/gutenberg/build)`, intermediate artifacts can be found at `/root/build_gutenberg/parts/gutenberg/build/target`

Caused by:
  build failed
Failed to run '/root/build_gutenberg/parts/gutenberg/rust/bin/cargo install -j4 --root /root/build_gutenberg/parts/gutenberg/install --path /root/build_gutenberg/parts/gutenberg/build' for 'gutenberg': Exited with code 101.
Verify that the part is using the correct parameters and try again.
Debug mode enabled, dropping into a shell
root@snapcraft-idolisingly-unbigamous-erich:~#
```

I guess we need to tell snapcraft to build libsass first?  _Should_ be relatively straightforward, but we will see.

musoke:
Actually, maybe snapcraft just didn't clone the libsass submodule.
Either way, I find it odd that `snapcraft cleanbuild --debug` fails although `snapcraft` works.

Keats:
Is there a forum somewhere where we can get support? It's a bit hard for me to debug without having snapcraft installed

musoke:
There is a forum: https://forum.snapcraft.io/categories

I just checked again, the pre-merge branch musoke/snap also fails to build with `snapcraft cleanbuild --debug`.  `snapcraft` is fine.

Keats:
Ok, I'll post something on the forum later today. Thanks!

Keats:
Having had a second look, it looks like `snapcraft cleanbuild` does not pull the submodules maybe? Which ends up with having an empty directory in libsass and git failing as it's missing the submodule

musoke:
Yes, I think it is exactly that.

Keats:
Is there no way to build the snap using `snapcraft` rather than cleanbuild?

musoke:
When I was testing locally I just ran `snapcraft` and it worked no 
probelem. However, it seems that snapcraft's build service uses 
`snapcraft cleanbuild`.

Keats:
Posted on https://forum.snapcraft.io/t/snap-of-gutenberg/6080

Issue #225, opened 2018-02-02
Picture & thumbnails support


vojtechkral:
Hi,
  I have recently started using Gutenberg and I like it very much (especially compared to other more famous static site generators that have ridiculously complex configuration).

One thing I've been misssing is the ability to create picture/photo galleries in a blogpost. And so I went ahead an added the implementation using the Piston/image library. You can have a look at the work in progress at my [pictures](https://github.com/vojtechkral/gutenberg/tree/pictures) branch. (It's just one or two evenings of hacking and so stuff like proper docs and tests is still missing unfortunatelly, but the implementation itself is present and seems to work well so far for me.) 

Feel free to let me know whether you think it is worth finalizing / if you'd consider integrating this in future.

A disadvantage of generating thumbnails is that it's time-consuming, and with Gutenber re-generating the site each time from scratch, this can become noticeable. Generating thumbs for some 30 photos takes some 10 seconds on my machine (each photo being about 6MB in size).

In default configuration there is no thumb generation, the feature is entirely opt-in.

Keats:
> I have recently started using Gutenberg and I like it very much (especially compared to other more famous static site generators that have ridiculously complex configuration).

Glad you like that!

I believe there's value in having Gutenberg handle image resizing but I'm not sure i would go with an approach of just doing it for all pictures. In my mind it would be more of a Tera filter like `image_url | img_resize(width=400, height=300)` and the filter would look for an already resized image at that path (for example a pic called "cat.png" in the filter above could be named "cat_400x300.png") and if not found perform the resizing. That should allow the thumbnail resizing to only happen once but it can happen anywhere in your site.

The downside being that adding a picture in a `.md` file would now require using a shotcode but I believe if you want thumbnailing, you will also probably want some kind of special CSS and use shortcode anyway.

What do you think?

vojtechkral:
@Keats Yes, I'm going to have to use shortcodes anyway.

Regarding template function, the thing is I'm not just inserting individual images, but whole galleries, ie. 10 to tens of pictures. In my previous solution I was using Pelican with a gallery plugin, the plugin feeds a list of pictures to the template, which uses PhotoSwipe to create a clickable gallery. So the problem is actually twofold:

- Collect a list of pictures for a page and pass it to the template in the context
- Create thumbnails for these pictures

In my implementation, this is done by matching page assets against a configurable regex, I guess you could configure the regex to only match files with a specific pattern...

OTOH I suppose I could also go with a simpler approach...

[...]

PR #300, opened 2018-04-29
 Add image resizing support #225

vojtechkral:
My image resizing branch. Ended up quite a large change.
Hopefully everything is ok, if not, let me know.

I finally put it together with some documentation.

I haven't written any tests, I'm not really sure how to go about it, since this is all quite data-dependent...


components/templates/src/global_fns.rs
```diff
@@ -1,35 +1,48 @@
...
                 Ok(u) => u,
                 Err(_) => return Err($err.into())
             },
             None => return Err($err.into())
-        };
+        }
+    };
+}
+
+macro_rules! optinal_arg {
```
Keats: `optional_arg`

components/utils/src/fs.rs
```diff
@@ -93,6 +93,33 @@ pub fn copy_directory(src: &PathBuf, dest: &PathBuf) -> Result<()> {
     Ok(())
 }
 
+/// Compares source and target files' timestamps and returns true if the source file
+/// has been created _or_ updated after the target file has
+// pub fn file_stale<PS, PT>(p_source: PS, p_target: PT) -> Result<bool> where PS: AsRef<Path>, PT: AsRef<Path> {
```
Keats: line not needed?

docs/content/documentation/content/image-resizing/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Arguments
+
+- `path`: The path to the source image relative to the `content` directory in the [directory structure](./documentation/getting-started/directory-structure.md).
+
+- `w` and `h`: Width and height in pixels of the resized image. Usage depends on the `op` argument.
+- `op`: Resize operation. This can be one of five choices: `"scale"`, `"fitwidth"`, `"fitheight"`, `"fit"`, or `"fill"`.
+  What each of these does is explained below.
+  This argument is optional, default value is `"fill"`.
+- `q`: JPEG quality of the resized image, in percents. Optional argument, default value is `75`.
```
Keats: can we use full words for aguments? IE `width`, `height`, `quality`. `op` is fine imo

Same for the struct and function calls of the `imageproc` subcrate

Keats (commented):
I only had a very quick read through but it looks really well done!
Thanks a lot for that, I'll do a more in-depth review when I get more time

vojtechkral:
@Keats :+1:

Keats:
I merged the markdown refactor in the `next` branch so that changes quite a few things (including always having the `config` in shortcodes)

vojtechkral:
@Keats I've not yet found the time to rebase the changes, but I'm on to it :-) Hopefully within next couple of days...

Keats:
Take your time!

vojtechkral:
Ok, it's rebased. Hopefully correctly (Edit: it's not, I forgot to fix tests, I'll do that..). I still had to propagate some more context down to markdown rendering.

The shortcodes page says that

> Note that if you want to have some content that looks like a shortcode but not have Gutenberg try to render it, you will need to escape it by using `{{/*` and `*/}}` instead of `{{` and `}}`.

I've found this to not be necessary in the image resizing doc page. I might've missed something.

Also, shortcodes still can't have some special chars in the filename (such as `_`), but this is to be expected, right?

Keats:
> I've found this to not be necessary in the image resizing doc page. I might've missed something.

I had a look at the diff and that's weird, it should definitely try to render those shortcodes. I'll have a look later

> Also, shortcodes still can't have some special chars in the filename (such as _), but this is to be expected, right?

They can, I just tried it locally

Is it possible you had some kind of stale crates somehow?

vojtechkral:
> Is it possible you had some kind of stale crates somehow?

Yeah, migth be. I'll just rebuild from scratch and try again...

components/content/src/page.rs
```diff
@@ -246,6 +267,9 @@ impl ser::Serialize for Page {
         state.serialize_field("next", &self.next)?;
         state.serialize_field("toc", &self.toc)?;
         state.serialize_field("draft", &self.is_draft())?;
+        let (assets, assets_imgs) = self.serialize_assets();
+        state.serialize_field("assets", &assets)?;
+        state.serialize_field("assets_imgs", &assets_imgs)?;
```
Keats: `assets_imgs` -> `images`?

components/errors/Cargo.toml
```diff
@@ -7,3 +7,4 @@ authors = ["Vincent Prouillet <prouillet.vincent@gmail.com>"]
 error-chain = "0.11"
 tera = "0.11"
 toml = "0.4"
+image = "0.18.0"
```
Keats: there's 0.19 on crates.io now

components/imageproc/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+authors = ["Vojtěch Král <vojtech@kral.hk>"]
+
+[dependencies]
+lazy_static = "1"
+regex = "0.2"
+tera = "0.11.0"
+image = "0.18.0"
+rayon = "0.9"
+twox-hash = "1.1"
```
Keats: Note to myself: this crate seems to depend on an old rand version, might be worth doing a PR to upgrade it

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+    Fill(u32, u32),
+}
+
+impl ResizeOp {
+    pub fn from_args(op: &str, width: Option<u32>, height: Option<u32>) -> Result<ResizeOp> {
+        use ResizeOp::*;
+
+        // Validate args:
+        match op {
+            "fitwidth" => if width.is_none() { return Err(format!("op=fitwidth requires a `width` argument").into()) },
```
Keats: can we have `fit_width` and `fit_height` to match the usual type of arguments?

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/// Holds all data needed to perform a resize operation
+#[derive(Debug, PartialEq, Eq)]
+pub struct ImageOp {
+    source: String,
+    op: ResizeOp,
+    quality: u8,
+    hash: u64,
+    collision: Option<u32>,
```
Keats: can you add a comment on what this value is in this struct?

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+
+        ImageOp { source, op, quality, hash, collision: None }
+    }
+
+    pub fn from_args(source: String, op: &str, width: Option<u32>, height: Option<u32>, quality: u8) -> Result<ImageOp> {
+        let op = ResizeOp::from_args(op, width, height)?;
+        Ok(Self::new(source, op, quality))
+    }
+
+    fn num_colli(&self) -> u32 { self.collision.unwrap_or(0) }
```
Keats: we have space for full length words! `num_colli` -> `num_collisions`?

vojtechkral:
@Keats Yeah, it was a silly mistake, I accidentaly used an old gutenberg binary

On a different topic: Do you think the `resize_image` shortcode I use in the documentation page should become a built-in? Right now only the Tera function is built-in. For usage from markdown, the shortcode (or a similar one) is needed.

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+                    // which should be cheaper than resizing and then cropping (smaller number of pixels to resize).
+                    let (crop_w, crop_h) = match fw < fh {
+                        true  => (img_w, (fw * h as f32).round() as u32),
+                        false => ((fh * w as f32).round() as u32, img_h),
+                    };
+                    let (off_w, off_h) = match fw < fh {
+                        true  => (0, (img_h - crop_h) / 2),
+                        false => ((img_w - crop_w) / 2, 0),
+                    };
+                    img.crop(off_w, off_h, crop_w, crop_h).resize_exact(w, h, RESIZE_FILTER)
```
Keats: `w`, `h`, `crop_w`, `crop_h` are pretty self-explanatory but can you lenghten fw, fh (I'm guessing `factor_w`?) and off_w, off_h (`offset_w`?)

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+/// All output is written in a subdirectory in `static_path`,
+/// taking care of file stale status based on timestamps and possible hash collisions.
+#[derive(Debug)]
+pub struct Processor {
+    content_path: PathBuf,
+    resized_path: PathBuf,
+    resized_url: String,
+    img_ops: HashMap<u64, ImageOp>,
+    // Hash collisions go here:
+    img_ops_colls: Vec<ImageOp>,
```
Keats: `colls` -> `collisions`?

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+            img_ops: HashMap::new(),
+            img_ops_colls: Vec::new(),
+        }
+    }
+
+    fn resized_url(base_url: &str) -> String {
+        match base_url.ends_with('/') {
+            true  => format!("{}{}", base_url, RESIZED_SUBDIR),
+            false => format!("{}/{}", base_url, RESIZED_SUBDIR),
+        }
```
Keats: maybe a bit clearer with a `if`/`else`?

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    pub fn source_exists(&self, source: &str) -> bool {
+        self.content_path.join(source).exists()
+    }
+
+    pub fn num_img_ops(&self) -> usize {
+        self.img_ops.len() + self.img_ops_colls.len()
+    }
+
+    fn insert_with_colls(&mut self, mut img_op: ImageOp) -> u32 {
```
Keats: same as before `colls` -> `collisions` and `colli` -> `collisions` in the body

docs/content/documentation/content/image-resizing/index.md
```diff
@@ -0,0 +1,123 @@
...
+    {{ resizeimage(path="documentation/content/image-resizing/gutenberg.jpg", width=150, height=150, op="fill") }}
+
+
+
+## Using `resize_image` in markdown via shortcodes
+
+`resize_image` is a built-in Tera global function (see the [Templates](./documentation/templates/_index.md) chapter),
+but it can be used in markdown, too, using [Shortcodes](./documentation/content/shortcodes.md).
+
+The examples above were generated using a shortcode file named `resizeimage.html` with this content:
```
Keats: would be called `resize_image.html` to work

docs/content/documentation/content/image-resizing/index.md
```diff
@@ -0,0 +1,123 @@
...
+
+```jinja2
+  <img src="{{ resize_image(path=path, width=width, height=height, op=op) }}" />
+```
+
+Note: Shortcode names don't currently support some special characters,
+that's why the above shortcode is named `resizeimage` rather than `resize_image`.
```
Keats: can probably be removed :+1:

Keats (changes requested):
Small nits, I'll have another read to understand the collisions better.

Something I'm not sure though is whether we need to pass the page/sections. In a way it's nice to have automatic access to them but you can't control the order, making it pretty useless if you care about that. It is only useful if you want to make a single gallery post and you're 100% sure you don't have images you don't want to appear in it in your folder.

vojtechkral:
> In a way it's nice to have automatic access to them but you can't control the order, making it pretty useless if you care about that.

I was kind of relying on my filesystem serving them in alphabetical order (I think, I don't think I added a sort call). Of course, that's an assupmtion that might not be true on other filesystems. Might be a good idea to sort them. Anyway, the idea is that you specify the order alphabetically with filenames.

> It is only useful if you want to make a single gallery post and you're 100% sure you don't have images you don't want to appear in it in your folder.

Well the example gallery code really is rather simplistic. If you wanted, I think you could use Tera functions like `ending_with` or `containing` to whitelist or blacklist pictures...

docs/content/documentation/content/image-resizing/index.md
vojtechkral: This one is done already in my local code :-)

docs/content/documentation/content/image-resizing/index.md
vojtechkral: Ditto

components/imageproc/src/lib.rs
vojtechkral: Yes, I'll try to add some explanation about the hash collision resolution thing...

components/imageproc/src/lib.rs
vojtechkral: Yes, there was also a bug on the next line (`fitwidth` was copypasted over).

components/imageproc/src/lib.rs
vojtechkral: Yes, `f` is for `factor`, I'll update...

vojtechkral:
Ok, hopefully I've addressed the comments. I refactored the collision resolution code somewhat removing the `Option` as it was pretty much useless. I added a bunch of comments and clarified variable names, hopefully.

components/imageproc/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "imageproc"
+version = "0.1.0"
+authors = ["Vojtěch Král <vojtech@kral.hk>"]
+
+[dependencies]
+lazy_static = "1"
+regex = "0.2"
```
Keats: there is regex 1.0 now

components/imageproc/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+/// A strcture into which image operations can be enqueued and then performed.
+/// All output is written in a subdirectory in `static_path`,
+/// taking care of file stale status based on timestamps and possible hash collisions.
+#[derive(Debug)]
+pub struct Processor {
+    content_path: PathBuf,
+    resized_path: PathBuf,
+    resized_url: String,
+    /// A map of a ImageOps by their stored hash.
+    /// Note that this cannot be a HashSet, because hashest handles collisions and we don't want that,
```
Keats: hashet -> hashset

docs/content/documentation/templates/pages-sections.md
```diff
@@ -37,10 +37,14 @@ previous: Page?;
 next: Page?;
 // See the Table of contents section below for more details
 toc: Array<Header>;
+// Paths of colocated assets, relative to the content directory
+assets: Array<String>;
+// Paths of colocated image assets, ie. files with an extension of "jpg", "jpeg", "png", "gif", or "bmp"
+images: Array<String>;
```
Keats: any chance that can be split into another PR? I'm not entirely sure this feature is worth having so it would be better to not have the discussion about it blocking the rest

docs/templates/shortcodes/gallery.html
```diff
@@ -0,0 +1,6 @@
+{% for img in page.images %}
+  <a href="{{ config.base_url }}/{{ img }}">
```
Keats: not possible to use the `get_url` global fn?

Keats (changes requested):
The thing I don't understand about the collisions: if the `source`, `op` and `quality` are the same, do we need to care about collisions?
If a target file with the same name exists, we probably don't need to do anything unless the target file is stale (that we can check with the function you added) in which case we just process it and overwrite the target file.

We could also hash the image content directly (https://github.com/abonander/img_hash but this would need to be updated) instead of checking for staleness but it would probably be slower

vojtechkral:
> The thing I don't understand about the collisions: if the `source`, `op` and `quality` are the same, do we need to care about collisions?

The collision resolution deals with cases where `source`, `op` or `quality` are _not_ equal and yet the hash is equal. That would result in two distinct images (for example) overwriting each other's thumbnail.

> We could also hash the image content directly (https://github.com/abonander/img_hash but this would need to be updated) instead of checking for staleness but it would probably be slower

Yes, I think so.

I'll have a look at the rest of the comments later on...

Keats:
> The collision resolution deals with cases where source, op or quality are not equal and yet the hash is equal. That would result in two distinct images (for example) overwriting each other's thumbnail.

Wouldn't collisions in 64-bit hash values be pretty much non-existent in the context of Gutenberg (how many images will need retouching in a site, 0-200 images maybe)? I'm also wondering whether https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html would be enough, it is slower but I don't think the hashing will be the bottleneck

vojtechkral:
> Wouldn't collisions in 64-bit hash values be pretty much non-existent in the context of Gutenberg (how many images will need retouching in a site, 0-200 images maybe)?

Well, for an individual site, the chance is extremely low. The algorithm has to work for _all_ Gutenberg sites though. Although even then the chance is still probably very low. I've ran some calculations based on [this SE thread](https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed) but I've only got very rough estimates riddled by assumptions...

In conclusion, honestly, I don't know whether it is really needed or not, I was merely coding defensively...

>  I'm also wondering whether https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html would be enough, it is slower but I don't think the hashing will be the bottleneck

Yes, I guess it could be used.

docs/content/documentation/templates/pages-sections.md
vojtechkral: Ok. Should I remove the gallery example as well then? I suppose I could also replace it with Tera code picking images from `assets`.

Keats:
> In conclusion, honestly, I don't know whether it is really needed or not, I was merely coding defensively...

No worries it's there let's keep it

docs/content/documentation/templates/pages-sections.md
Keats: I think so, it can be added later at any time if we want to and it won't be a breaking change

vojtechkral:
Okay, I update the code sometime during the next couple of days (I hope) to address the comments. I'll remove the `images` metadata for now, but I'd like to able to do something like that "in private" as that's a key usecase to me. I'll try if I can make-do with just `assets` and Tera features, that would be fine.

Keats:
@vojtechkral let me know if you don't have time and would like me to finish the PR

vojtechkral:
@Keats I am a bit snowed under, yes. How about this: I'll try to do it this week and if I can't make it I'll let you know, if that's ok...

Keats:
No worries, take your time

vojtechkral:
Updated the code. Sorry for the delay :-/

I just realized I'm not sure if I understood right your comment in `page.rs` - did you want to remove just `page.images` , or `page.assets` too? I removed `page.images` and refactored the gallery example to only use `page.assets`. Let me know if that's ok / not ok.

Also, one thing I'm not sure about is whether the `static/_resized_images` should be called that - maybe call it `static/_processed_images` for future-proofing if we later want to add some image operations other than resize?

Keats:
> Also, one thing I'm not sure about is whether the static/_resized_images should be called that - maybe call it static/_processed_images for future-proofing if we later want to add some image operations other than resize?

we can do with `_processed_images` yes and keep `page.assets`

vojtechkral:
@Keats Done!

Keats (approved):

Keats:
Thanks a lot for the amazing work!

Issue #298, opened 2018-04-27
Colocated images break when accessing page.content


cormac-obrien:
`content` folder:

```
content/
|-example-page/
  |-index.md
  \-some-image.jpg
```

`index.md`:
```
+++
title = "example"
date = 2018-04-26
template = "page.html"
+++

This is an example page

![An image][1]

Some more text

[1]: some-image.jpg
```

I modified the `hyde` theme slightly to show this, adding `{{ page.content }}` under the per-post date `<span>` tags.

On the actual `page-1` page, the image displays fine:

![screenshot 10](https://user-images.githubusercontent.com/5809282/39340432-c2df06e0-4993-11e8-8d32-12218f1849fd.png)

However, the generated `index.html` does not get the right link:

![screenshot 12](https://user-images.githubusercontent.com/5809282/39340508-228402a8-4994-11e8-8afc-76108ab91f10.png)

It seems to me like Gutenberg should resolve this link correctly regardless of where `{{ page.content }}` is accessed instead of only for the HTML generated for that markdown file. Is this possible?

Keats:
Hmm I'm not sure how in your markdown example it could prepend the permalink automatically unless we go and add it to every asset if it's relative

Keats:
Looking at it now, it also needs to be done for links, not only images but it should be doable

Issue #297, opened 2018-04-26
Can't have links in headers


clarfonthey:
Very simple example: `# [Rust](https://rust-lang.org)` gets rendered as just `<h1 id="rust">Rust</h1>`, when it should be `<h1 id="rust"><a href="https://rust-lang.org">Rust</a></h1>`.

I took a look at the code and it appears that links in headers are intentionally ignored in Gutenberg, despite being parsed (correctly) by `pulldown-cmark`. Why is that?

(link to line in question: https://github.com/Keats/gutenberg/blob/master/components/rendering/src/markdown.rs#L213)

Keats:
The markdown rendering is going to get rewritten for the next version, I can fix it at that point

Issue #270, opened 2018-03-27
allow array type for shortcodes


piedoom:
I'm creating an album shortcode that contains an arbitrary amount of images.  It'd be very nice to be able to just go something like:

```html
{{ album(paths=["file1.jpg", "file2.jpg", "file3.jpg"]) }}
```

Because arrays are ignored, the next best solution that I can imagine is something like defining manually different amounts for different shortcodes, i.e. `album2`, `album3`, `album4`, and so on.

Keats:
Yep definitely something I want, just need to fix the markdown renderer first!

asaaki:
Another workaround should be to provide a splittable string and an optionally overridable separator.

```rust
{{ album(paths="file1.jpg,file2.jpg,file3.jpg", sep=",") }}
```

But I'd also agree that we need self-definable arrays (not only in shortcodes, but supported by tera in general).

Jieiku:
Anyone know of an example Shortcode that takes an array as an argument?

I have a super neat idea for a shortcode, and passing an array to it will make things much easier.

Keats:
It should work as expected in the original comment?

Jieiku:
YEP! it works, I made an image hover comparison shortcode: [imghover](https://github.com/Jieiku/abridge/blob/master/templates/shortcodes/imghover.html)

https://abridge.netlify.app/overview-images/

PR #302, opened 2018-05-07
Markdown next

### LINKED ISSUE #165: Shortcode names and arguements.
It appears as though shortcode file names cannot have an underscore in them and that the value that argument values cannot contain special characters.

To reproduce the filename issue:
- create shortcode, `./templates/shortcodes/contact_info.html`.
```
{{ label }}
{{ info }}
```
- use shortcode
```
{{ contact_info(label="this is a label", info="this is the info") }}
```

To reproduce the argument issue:
- create shortcode, `./templates/shortcodes/contact.html`.
```
{{ label }}
{{ info }}
```
- use shortcode
```
{{ contact(label="this is a label", info="this [is the] info") }}
```
This is exacerbated because it appears the dev server will fall back to a previous state when it cannot build correctly...and then sometimes gets stuck and needs to be manually restarted.

Keats:
Closes #165 #298 #297 #270 #249 #160


Keats:
@jturner314 @RadicalZephyr @reillysiemens @cldershem @clarcharr @NuLL3rr0r @piedoom 

I believe this solves all current issues/missing features with markdown rendering/shortcodes if any of you want to give it a try

clarfonthey:
I'll try this out later today or tomorrow.

piedoom:
This is some awesome work!  Thanks.

Keats:
(merging into next as this affects other PRs but i'll still take feedback)

PR #305, opened 2018-05-16
Use proper Unicode word count; fixes #304

### LINKED ISSUE #304: Count words using Unicode instead of split_whitespace
This should be relatively simple, as it would only require using either the [`unic_segment`](https://docs.rs/unic-segment/0.7.0/unic_segment/struct.Words.html) or [`unicode-segmentation`](https://docs.rs/unicode-segmentation/1.2.1/unicode_segmentation/trait.UnicodeSegmentation.html#tymethod.unicode_words) crate.

clarfonthey:
This pulls in a dependency on the `unicode-segmentation` crate, which is relatively small. Now, the word count will be based off of the Unicode definition of word boundaries, rather than just whitespace, which will give a more accurate count for more languages. It also will give a more accurate count in languages with whitespace as well.

### DIFF
--- a/components/utils/Cargo.toml
+++ b/components/utils/Cargo.toml
@@ -6,8 +6,8 @@ authors = ["Vincent Prouillet <prouillet.vincent@gmail.com>"]
 [dependencies]
 errors = { path = "../errors" }
 tera = "0.11"
+unicode-segmentation = "1.2"
 walkdir = "2"
-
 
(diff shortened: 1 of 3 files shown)


Keats:
Nice thanks!
Can you change the target branch to `next` when you have time?

Keats:
Nevermind I apparently can do it myself!

PR #310, opened 2018-05-27
Fixes #279 switch from iron to actix-web.

### LINKED ISSUE #279: Move from Iron to another web library/framework
Iron isn't maintained anymore. It would be good to replace it.

onelson:
This implementation seems to match the behavior of the previous iron
server.

Static files are rendered as-is, and when a directory is requested, we
attempt to render a `index.html` found inside, or 404 if it's not
present.

The actix docs mention using channels to send a message to the server to
shut it down gracefully while running in another thread (as we're doing
here), but it seems like there would have to be some reorganization in
order to manage this effectively, perhaps holding the channel sender inside
`main.rs` so we can push a message through to the server when the call
to `cmd::serve()` finally returns.

For the time being, I left things without any careful attempts to
cleanup the server thread. This more or less matches the old iron
implementation as far as I can see.

The default static file handling in actix is _just_ a little off from what we'd want.
I left some comments in the source regarding why we can't just use their hook
for directory index redirection.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -23,15 +23,15 @@
 
 use std::env;
 use std::fs::remove_dir_all;
-use std::path::Path;
+use std::io;
+use std::path::{Path, PathBuf};
 use std::sync::mpsc::channel;
 use std::time::{Instant, Duration};
 use std::thread;
 
 use chrono::prelude::*;
-use iron::{Iron, Request, IronResult, Response, status};
-use mount::Mount;
-use staticfile::Static;
+use actix;
+use actix_web::{fs, server, App, HttpRequest, HttpResponse, Responder};
 use notify::{Watcher, RecursiveMode, watcher};
 use ws::{WebSocket, Sender, Message};
 use ctrlc;
@@ -59,8 +59,8 @@ enum ChangeKind {
 const LIVE_RELOAD: &'static str = include_str!("livereload.js");
 
 
-fn livereload_handler(_: &mut Request) -> IronResult<Response> {
-    Ok(Response::with((status::Ok, LIVE_RELOAD.to_string())))
+fn livereload_handler(_: HttpRequest) -> &'static str {
+    LIVE_RELOAD
 }
 
 fn rebuild_done_handling(broadcaster: &Sender, res: Result<()>, reload_path: &str) {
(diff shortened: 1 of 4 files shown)


Keats:
Nice, I'll have a look later thanks!

src/cmd/serve.rs
```diff
@@ -102,6 +102,24 @@ fn create_new_site(interface: &str, port: &str, output_dir: &str, base_url: &str
...
+/// `index.html` on a successful redirect (rare), which is unsightly.
+///
+/// Rather than deal with all of that, we can hijack a hook for presenting a
+/// custom directory listing response and serve it up using their
+/// `NamedFile` responder.
+fn handle_directory<'a, 'b>(dir: &'a fs::Directory, req: &'b HttpRequest) -> io::Result<HttpResponse> {
+    let mut path = PathBuf::from(&dir.base);
+    path.push(&dir.path);
+    path.push("index.html");
+    Ok(fs::NamedFile::open(path).respond_to(req).unwrap())
```
onelson: Looking back on the diff, this largely doesn't seem correct. `unwrap`ping then rewrapping should be harmful _and_ redundant. Looks like the error type that comes from `respond_to` isn't an `io::Error` either -- let me map it real quick!

onelson (commented):
Should be good now. It's a little surprising that the callback for `StaticFiles#files_listing_handler` demands an `io::Error` rather than their `failure` driven internal error types (else this would have been less wordy to write).

src/cmd/serve.rs
fafhrd91: you should create ticket for actix-web

onelson: @fafhrd91 will do. I'll write something up soon.

Keats: @fafhrd91 would it make sense to have actix-web be stand alone and re-export the bits it needs from actix so you only have 1 package to add to get started?

Keats:
That seems to work fine! Any chance you can resolve the conflict in Cargo.lock? I'll merge it after

src/cmd/serve.rs
fafhrd91: Re-export makes sense. I’ll add reexport in next release

onelson:
Seems like there will be some cleanup/code reduction possible with a subsequent actix-web release, but in the meantime I think this will do it.

This was my first look at actix, and it's pretty cool!

Keats:
Thanks!

PR #312, opened 2018-05-31
Fixup odd error handling in directory handler

onelson:
Quick follow-up to the actix-web adoption PR from the other day.

Got the devs to look at actix/actix-web#261 this morning and they figured out where the types were getting flipped around.

This rev removes my ham-fisted side-stepping of the compiler error.


Keats:
Thanks!

PR #313, opened 2018-06-02
Fix panicking handling content file without extension

unrealhoang:
Prevent panic when there're events for files without extension in `content` folder.

Somehow, when I update file using vim, it also generate events for some unknown number-only filename in that folder.
For example, after editing `my_blog.md`:

```
-> Content changed /Workspace/SideProject/blog/content/my_blog.md
Done in 56ms.

Change detected @ 2018-06-03 02:36:21
-> Content changed /Workspace/SideProject/blog/content/4913
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -253,7 +253,7 @@ fn is_temp_file(path: &Path) -> bool {
             }
         },
         None => {
-            path.ends_with(".DS_STORE")
+            true
         },
     }
 }


Keats:
Hmm I thought I got the temporary files sorted out but it looks like editors are deciding to create new kinds all the time..

I think handling it in https://github.com/Keats/gutenberg/blob/master/src/cmd/serve.rs#L235 would be better as this is not really a content change so it might fail later down the line. Returning `true` in https://github.com/Keats/gutenberg/blob/master/src/cmd/serve.rs#L255 in every case should work without misbehaving I believe

unrealhoang:
I fully agree it should be handled there. Return `true` for every empty extension case should work and also being more correct. Since the contract of `after_content_change` is extension can not be empty.

unrealhoang:
My other suggestion is that we can make the check for temp file per file kind, `is_temp_file(&Path)` will be changed to `is_temp_file(&ChangeKind, &Path)` for more granular control.

Keats:
I think we can do with the simple approach for now and just returning `true` for files without extension

unrealhoang:
@Keats Updated.

Keats:
Thanks!

PR #323, opened 2018-06-24
Custom 404 page

onelson:
Looking over the original discussion on #296, there was some talk about whether there should be a template or if this should be a plain static file. I haven't yet looked at how/where the default templates are, or how to add a new one, so to start I've opted to use the static file approach from the original PR.

I've adapted the original rev via rebase so as to not erase @m-laniakea's contribution, but I've omitted the "plug the status code into the target file name"  bit since I don't see the value in responding to much more than 404. There could be a case for 403 but there is a quick slide from here to "why doesn't gutenberg provide a production http server?"

I think 404 is the only sane default we can provide without slipping.

Lingering questions:

* Should the target page be a template? If so, how to introduce a new default template?
* If remaining as a static file, should there be a config option to specify the file rather than hardcoding the path?

EDIT: I should also call out that an earlier rev where `cargo update` happened broke the server. I tightened up the version spec for actix in 56033e37, but this is something to look out for moving forward (I guess). actix 0.6 appears to not be compatible with actix-web 0.6.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ term-painter = "0.2"
 url = "1.5"
 # Below is for the serve cmd
 actix = "~0.5"
-actix-web = "0.6"
+actix-web = "~0.6"
 notify = "4"
 ws = "0.7"
 ctrlc = "3"
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -526,6 +526,7 @@ impl Site {
         if self.config.generate_rss {
             self.render_rss_feed()?;
         }
+        self.render_404()?;
         self.render_robots()?;
         // `render_categories` and `render_tags` will check whether the config allows
         // them to render or not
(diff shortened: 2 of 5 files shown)


onelson:
Come to think of it, I tightened the version spec on actix, but I should really have also tightened the spec for actix-web as well. I'm sure the next actix-web will be targetting the actix I just pulled us back from.

Keats:
Sorry about that, I force-pushed a rebase of the next branch so you will have to cherry pick and force push on your end as well I think.

The target should be a template and should be added as `components/templates/src/builtins/404.html` and then added in the lib.rs of the same crate with `include_str!`, this should ensure that the template is always there as a default but easily overridable by creating a 404.html in the templates directory.

> If remaining as a static file, should there be a config option to specify the file rather than hardcoding the path?

I think keeping 404.html is simple enough, no need to configure that.

I ran into the same issue with actix a few minutes ago and had to switch back to 0.5 as well, ideally i shouldn't have to put actix manually at all, especially if minor updates can have breaking changes.

onelson:
Thanks for the template advice. OK great. I'll fix it all up and sort out the rebase.

~~The maintainer mentioned re-exporting actix (or the specific bits we need) from actix-web from actix itself. That could be a good way to skip having to dance around the drift -- I will check on the status of that upstream change.~~ (the re-export isn't there yet)

onelson:
Happy to address any further nits, but as far as I know this is ready to merge.

Keats:
Thanks!

Issue #332, opened 2018-07-13
Live demos of themes?


codesections:
Are there any live site demos of the varrious Gutenberg themes?  I looked on the [theme page](https://www.getgutenberg.io/themes/) and did not see any links.

If there are not, will would be happy to create a demo for each theme with lorum ibsum content and submitt a pull request to add links to the demo site to the theme page.

Keats:
As mentioned in a PR, that would definitely be something I'm interested in

codesections:
Thanks!  Demos of the current themes are avalible at <https://gutenberg-theme-demo.netlify.com/>.  I haven't publisized this yet because I'm waiting on the resolution of Keats/book#1—the live demo of the Book theme incorporates my changes to that theme, because I could not get the theme to work as it was written.  Once that issue is resolved, I'll share this externally.

After 0.4 is released, I'll check these demos for compatibility issues, resolve any issues, and then submitt a pull request to add the links to the main theme site in the documentation.

codesections:
Now that the issues with Book have been resolved, the theme demo site is ready to go.  As discussed, I'll hold off on submitting a pull request for links from the docs until after I've tested for compatibility with v0.4, but the site is live now:

https://gutenbergthemes.codesections.com/

Keats:
I added demos of themes directly to each theme repo through netlify, this way they get updated immediately and it only took me ~1 min per theme.
We could have better quality demos though in terms of content

codesections:
I've submitted a pull request to update Even with the demo content I put together.  keats/Even#5 If that looks good, I can submit similar pull requests with similar demo content for the other themes.

Keats:
There are live demos of themes now so closing this. Getting better default data can be an issues on individual theme repos

PR #333, opened 2018-07-13
Add a link to a live demo for the After Dark theme

codesections:
After hearing no objection to the plan I suggested in issue #332, I went ahead and put together a live site that demos the After Dark theme (I plan to do them all, and started with After Dark simply because it comes first alphabetically).  The demo site is available at <https://after-dark--gutenberg-theme-demo.netlify.com/>.

One reason I think that live demos would be helpful is that Hugo tends to have them (see, e.g, <https://themes.gohugo.io/theme/after-dark/>), and I found them to be much more helpful in getting a feel for a theme than a screenshot or description is.

Happy to make any changes to this commit or to the live demo if you would like.

### DIFF
--- a/docs/content/themes/after-dark/index.md
+++ b/docs/content/themes/after-dark/index.md
@@ -10,6 +10,7 @@ created = 2018-02-22T19:13:36+01:00
 updated = 2017-11-07T17:39:37+01:00
 repository = "https://github.com/Keats/after-dark"
 homepage = "https://github.com/Keats/after-dark"
+live_demo = "https://after-dark--gutenberg-theme-demo.netlify.com/"
 minimum_version = "0.2"
 license = "MIT"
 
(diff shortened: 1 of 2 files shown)


Keats:
The themes pages are actually generated from https://github.com/Keats/gutenberg-themes so the changes shouldn't be done in this repo. I believe adding a `live_demo` fields in `theme.toml` makes sense though, but I would wait till my themes are updated for gutenberg 0.4 as there is a big breaking change for taxonomies in the pipes

codesections:
Thanks, that makes sense.  I looked at how the themes are generated from <https://github.com/Keats/gutenberg-themes>, and it looks like there would still need to be one change in this repo: updating the `docs/templates/theme.html` template to render the link if the `live_demo` field is set.  Please let me know if I'm misundersanding something and that doesn't need to be changed in this repo.  I revised this pull request to only make that change.

Once this repo goes to 4.0, I'll submit a pull request to Keats/gutenberg-themes to add the links to the `theme.toml`.

By the way (and you may have seen this already), I'm building the demo sites by including the various themes as submodules, so the live demos should keep up with any changes as the themes are updated.

Keats:
Super nice, thanks!

PR #337, opened 2018-07-16
Add an external link checker

### LINKED ISSUE #115: Look into building a link checker
An optional tool that will check every http/https link in all posts to ensure there are no dead links at build time.
Any dead link would be reported to the console as a warning

Keats:
Close #115


Keats:
That seems to work, although there are some false negatives because of some sites behaving badly like crates.io (https://github.com/rust-lang/crates.io/issues/788) which always return 404. 
The workaround in the thread is to request html but I'm not going to go down the path of adding domain specific workarounds.

Issue #339, opened 2018-07-19
Sections are not automatically created as said in the docs.


itsfarseen:
According to docs:
> A section is automatically created when a folder is found in the content section, unless it only contains a index.md file and is actually a page with assets.

But a section is not created unless it contains a `_index.md`.

__Steps to reproduce__
```bash
gutenberg init test
cd test
mkdir content/section1
echo -e '+++\n+++\n#Hello World' >./content/section1/post1.md
gutenberg serve
```
__Expected output__
```bash
Building site...
-> Creating 1 pages and 1 sections
... snip ...
```

__Actual output__
```bash
Building site...
-> Creating 1 pages (1 orphan) and 0 sections
... snip ...
```

codesections:
It seems to me that the current behavior is correct, and that the docs ought to be updated to reflect that behaviour.  Do you agree?  

That is, I  think that a section should only be created if the folder contains an `_index.md`.  I think this has two advantages:
 * It creates better semantics for seeing if a folder is a section: if it has an `_index.md` file, then it's guarenteed to be a section; if it lacks an `_index.md` file, it is guarenteed *not* to be a section.
 *  It allows for easier creation of orphan pages, which might be desierable in some use-cases.  For example, an ecomerce site might want to have `/products/product1` and `products/product2` as pages without having `/products` as a section.  (I note that the user could achieve a similar effect by creating the `/products` section and then setting `render = false` in its front matter.  But it seems less intuative to create a section and then prevent it from rendering than to just not create it.  Plus, Gutenberg will correctly note that the section was created (but not rendered) and will thus not classify the pages within that section as orphans, which seems less useful).

If you agree that the current behaviour is correct, then I'd be happy to submit a pull request updating the docs.

Keats:
Looks like it's a documentation issue, only the index section is automatically created.
I'll take a PR for it, thanks!

PR #341, opened 2018-07-19
Edit docs to say that sections require an `_index.md` file

codesections:
This pull request provides a solution for #339.

The current version of the documentation states that sections are created automatically, but this is not correct: sections are created only if the folder contains an `_index.md` file.  This commit changes the documentation to reflect that reality and updates the surrounding
language for consistency and to explain the behaviour of `index.md files.


docs/content/documentation/content/page.md
```diff
@@ -6,6 +6,11 @@ weight = 30
 A page is any file ending with `.md` in the `content` directory, except files
 named `_index.md`.
 
+If a file ending with `.md` is named `index.md`, then it will generate a page
+with the name of the containing folder (for example, `/content/about/index.md` would
+create a page at `[base_url]/about`).  If the file is given any name *other* than
+`index.md`, then it will create a page with that name (without the `.md`).
```
Keats: `_index.md` no?

docs/content/documentation/content/section.md
```diff
@@ -3,19 +3,28 @@ title = "Section"
...
-contains a `index.md` file and is actually a page with assets.
+A section is created whenever a folder (or subfolder) in the `content` section contains an 
+`_index.md` file.  If a folder does not contain an `_index.md` file, no section will be 
+created, but markdown files within that folder will still create pages (known as orphan pages).
 
-You can add `_index.md` file to a folder to augment a section and give it some metadata and/or content.
-
-The index page is actually a section created automatically like any other: you can add metadata
-and content by adding `_index.md` at the root of the `content` folder.
+The index page is actually an automatically created section.  Thus, the `index.html` template
```
Keats: the first sentence feels a bit odd, maybe remove "automatically created" from it? Otherwise it feels like it is only automatic

Keats (changes requested):

codesections:
I updated the pull request to address your change request re: `section.md`.

On your other change request—the one about `page.md`—I don't believe I got things backwards.  That section of the docs is discussing what happens if the user creates an `index.md` file (*not* an `_index.md` file).  If the user creates an `about/index.md` file then a *page* will be created at `about`.  This is different than if the user creates an `_index.md` file, which would create a *section*.  In typing this out, I realized that the docs could do a better job of drawing that distinction, so I updated them to clarify.

Keats:
Thanks!

Issue #338, opened 2018-07-17
Sorting previous/next semantics for weight


Keats:
The current semantic of prev/next on page seem wrong when looking at weights

Given 3 posts with date `[2017-01-01, 2018-01-01, 2019-01-01]`. The pages will be sorted in a descending order (2019, 2018, 2017) and the pages will have the following next/prev:

- 2019 => next = None, previous = 2018
- 2018 => next = 2019, previous = 2017
- 2017 => next = 2018, previous = None

The next/previous make sense here since we are talking about the next blog post would be the most recent and the previous as the older one.

Enter `order` and we get what we would expect:

3 posts with order = [3, 2, 1]

Expectations:
- 3 => next = None, previous = 2
- 2 => next = 3, previous = 1
- 1 => next = 2, previous = None

Lastly, `weight`:

3 posts with weight = [1, 2. 3]

Expectations:
- 1 => next = 2, previous = None
- 2 => next = 3, previous = 1
- 3 => next = None, previous = 2

Currently, the previous and next are reversed for weight which makes sense since both date/order work that way but `weight` actually works the other way semantically: it's ascending.
I think it would make sense to inverse setting `next/previous`when `weight` is used to match the expectations.

2 questions:
- am I getting something wrong with my expectations about the weight ordering?
- would switching the next/previous for weight make sense to everyone?

cc @codesections

codesections:
Just to make sure I understand both the current situation and the proposed change, is the table below accurate for 3 posts with weight = [1, 2. 3]?

| post title |  *current* previous | *current* next | *proposed* previous | *proposed next* |
|:------------:|:--------------------------:|:--------------------:|:----------------------------:|:----------------------:|
|      1        |             2                 |              --        |                --               |         2                |
|      2        |             3                 |              1        |                1                |        3                |
|      3        |             --                |              2        |                2                |         --                |

Assuming that's right, I think that's a good change and would help with the sort of issues that come up in the Book theme: anytime there's a list of content that **isn't** a blog, it might make sense to talk about the `next` item in the list as the one *below* the current item, not the one above it.  The proposed change in the way weight defines previous and next would help address this non-blog use case.

[...]

PR #343, opened 2018-07-23
Expand documentation about sorting pages

codesections:
This pull request expands the documentation regarding sorting, as discussed in issue #338.

The biggest change is to pull the main discussion of sorting out of the `content/section` page and into its own page in the `templates` section.  My logic for this change is that sorting order really only comes up when iterating over pages/sections in a template.  I left enough detail in the `content/section` page for users to know how to set the `sort_by` variable and referred them to the new `templates/iterating and sorting` page for details.

Happy to discuss or make any changes you think would be appropriate.


Keats:
The text is good but I think it should be in `content` rather than `templates` imo, in the `sections.md` file like it was

codesections:
Let me explain why I moved it and then, if you still disagree, I'm happy to move it back to `sections.md`.

If you're writing content without writing templates (e.g., using someone else's themes without any files in your `templates/` directory), then you don't need to know about `reverse` filters, `page.next`/`page.previous`, or any of the rest.  After all, you *can't* call `page.next` from inside a content file—only from inside a template.  *All* the content writer needs to know is that the `_index.md` file can have set a `sort_by` variable to `date`, `order`, or `weight`.  And my view is that the `content` section of the docs should be limited (as much as possible) to what the content creator needs to know.  

The user who wants to make their own templates, on the other hand, needs to know all about the ways `order` and `weight` interact with `reverse` and how that impacts `page.next`.  I think that sort of advanced info is best kept specific to creating templates, which is the only place where it really comes up.

But, if you disagree, I'll move it back.

Keats:
> And my view is that the content section of the docs should be limited (as much as possible) to what the content creator needs to know

I agree on that but I think the different sort types are part of the content since that's where we explain all the options in the front-matter. I would find it weird to have a couple of options explained in a completely different part of the documentation. We can always move it back to templates later if other people prefer it

codesections:
Ok, moved it back.

PR #345, opened 2018-07-23
Revise netlify.toml example

codesections:
This is extremely minor, but the initial `netlify.toml` example confused me a bit because I didn't realize that it was using a *netlify* variable; I thought I needed to replace something with a variable from my own configuration.  I made a very minor edit to the docs to clarify.


Keats:
Thanks

Issue #348, opened 2018-07-25
Medthod for using non-release version on Netlify?


codesections:
I would like to use a non-release version of Gutenberg to build my site on Netlify (because there are several features that work fine in 0.4.0 that don't work for me in 0.3.3 and the release version of 0.3.4 won't work for me due to #344).

This seems like it should be simple—one of the big advantages of a Rust app is getting a compiled binary that should run easily, and Netlify claims that it should be a simple matter of "add[ing] the version of the [Gutenberg] binary you wanted to use to your repository. Since [Gutenberg] is a static binary, this is a very convenient solution if you want to manage it yourself."  ([source](https://www.netlify.com/blog/2017/04/11/netlify-plus-hugo-0.20-and-beyond/).

However, I can't get this to work, even after extensive attempts.  When I attempted to do so, I would get the following build error:
```
1:20:23 PM: Executing user command: ./gutenberg build
1:20:23 PM: ./gutenberg: error while loading shared libraries: libonig.so.5: cannot open shared object file: No such file or directory
```

Is there a good way to do this?  If there is, I'd be happy to add it to the docs so that other can deploy with their own version of Gutenberg, which seems like it should be one of Gutenberg's killer features.

codesections:
Update:  Better page in Netlify's docs explaining how this should work: <https://www.netlify.com/blog/2016/10/18/how-our-build-bots-build-sites/>

> If our build environment is missing something other than a ruby, node, or python dependency, you might need to add it yourself. Our build environment runs Ubuntu Linux v 14.04 LTS using x86_64 architecture, so you’ll want to bring a binary that can run there. Since you don’t have permissions to apt-get install anything, you’re limited to pre-compiled binaries that fit that architecture. A frequent use pattern is to get a copy of the binary we need to build your site, and add it to your repository so you can run it directly during your build. By default your build starts in the root of your repository, so it could be as easy as using ./my-binary as part of your build command.

Shouldn't `cargo build --release` produce a binary that runs on Ubuntu Linux v 14.04 LTS using x86_64 architecture?  Or am I missing something?

Keats:
I don't think it is possible, Netlify uses their own docker image so you can't really do whatever you want, it's limited to what's available in it.

> Shouldn't cargo build --release produce a binary that runs on Ubuntu Linux v 14.04 LTS using x86_64 architecture? Or am I missing something?

Yes but to compile Gutenberg and its dependencies you will need dependencies not available in the docker image.

[...]

PR #349, opened 2018-07-25
Expand Netlify doc page

codesections:
As discussed in #348, if users want to use a non-release version of Gutenberg, they need to manually deploy the files.  Netlify makes this fairly easy, but their docs aren't entirely clear on how to do so.  This pull request expands the Gutenberg doc page to explain the process for manually deploying to Netlify from the command line.


Keats:
Thanks!

PR #352, opened 2018-07-28
Remove `order` and add `heavier`/`later`

codesections:
This commit removes the option to sort by order and also removes `page.next` and `page.previous` variables.  Instead, pages can be sorted by two methods `date` and `weight`.  The Tera `reverse` filter will reverse either of those sorts, so the old `order` behavior can be achieved by using the `reverse` filter with `weight`.

In place of the `previous`/`next` variables, this commit adds the `page.earlier`/`page.later` variables (which are set when the page is sorted by date) and the `page.heavier`/`page.lighter` variables (which
are set when the page is sorted by weight).  These variables have the advantage of not having confusing semantics when the `reverse` filter is used.

**Note**: I cannot emphasize this enough—I genuinely do not know how to program in Rust.  This was cargo-cult programing of the worst sort; I was copying lines of code with modifications that seemed to fit the context without true understanding of the surrounding code.  All the tests passed, and things looked good in manual testing, but I genuinely did not know what I was doing and want to be clear about that.

### DIFF
--- a/components/content/src/sorting.rs
+++ b/components/content/src/sorting.rs
@@ -53,7 +53,7 @@ pub fn sort_pages(pages: Vec<Page>, sort_by: SortBy) -> (Vec<Page>, Vec<Page>) {
 
 /// Horribly inefficient way to set previous and next on each pages that skips drafts
 /// So many clones
-pub fn populate_previous_and_next_pages(input: &[Page], sort_by: SortBy) -> Vec<Page> {
+pub fn populate_siblings(input: &[Page], sort_by: SortBy) -> Vec<Page> {
     let mut res = Vec::with_capacity(input.len());
 
     // The input is already sorted
@@ -140,7 +140,7 @@ pub fn populate_previous_and_next_pages(input: &[Page], sort_by: SortBy) -> Vec<
 mod tests {
     use front_matter::{PageFrontMatter, SortBy};
     use page::Page;
-    use super::{sort_pages, populate_previous_and_next_pages};
+    use super::{sort_pages, populate_siblings};
 
     fn create_page_with_date(date: &str) -> Page {
         let mut front_matter = PageFrontMatter::default();
@@ -208,13 +208,13 @@ mod tests {
     }
 
     #[test]
-    fn can_populate_previous_and_next_pages() {
+    fn can_populate_siblings() {
         let input = vec![
             create_page_with_weight(1),
             create_page_with_weight(2),
             create_page_with_weight(3),
         ];
-        let pages = populate_previous_and_next_pages(&input, SortBy::Weight);
+        let pages = populate_siblings(&input, SortBy::Weight);
 
         assert!(pages[0].clone().lighter.is_none());
         assert!(pages[0].clone().heavier.is_some());
--- a/components/content/src/sorting.rs
+++ b/components/content/src/sorting.rs
@@ -92,8 +92,7 @@ pub fn populate_previous_and_next_pages(input: &[Page], sort_by: SortBy) -> Vec<
                         next_page.later = None;
                         new_page.later = Some(Box::new(next_page));
                     },
-                    SortBy::None => {
-                    }
+                    SortBy::None => {}
                 }
                 break;
             }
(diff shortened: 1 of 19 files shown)


components/content/src/sorting.rs
```diff
@@ -91,9 +80,21 @@ pub fn populate_previous_and_next_pages(input: &[Page]) -> Vec<Page> {
...
+                        next_page.lighter = None;
+                        next_page.heavier = None;
+                        new_page.lighter = Some(Box::new(next_page));
+                    },
+                    SortBy::Date => {
+                        next_page.earlier = None;
+                        next_page.later = None;
+                        new_page.later = Some(Box::new(next_page));
+                    },
+                    SortBy::None => {
```
Keats: you can do just do `SortBy::None => ()` rather than opening braces

components/content/src/sorting.rs
```diff
@@ -221,80 +186,49 @@ mod tests {
...
         let input = vec![
-            create_page_with_order(2, "a.md"),
-            create_page_with_order(3, "a.md"),
-            create_page_with_order(1, "a.md"),
+            create_page_with_weight(2),
+            create_page_with_weight(3),
+            create_page_with_weight(1),
         ];
         let (pages, _) = sort_pages(input, SortBy::None);
         // Should be sorted by date
```
Keats: this comment seems to be lying? probably me copy/pasting a bit too much

components/rebuild/tests/rebuild.rs
```diff
@@ -2,7 +2,7 @@ extern crate rebuild;
 extern crate site;
 extern crate tempfile;
 extern crate fs_extra;
-
+
```
Keats: remove whitespace

Keats (changes requested):

Keats:
It actually looks ok to me, I think you just did your first succesful Rust PR!

Only thing missing is an update to the docs but I can do it later myself if you want.

codesections:
That's great! 

I'm happy to update the docs; I intended to do so after confirming that the implementation was ok. It will probably be tomorrow before I do so, however.

codesections:
Ok, docs updated.  I updated the `section` and `page` pages, which I think are the only places that mentioned `order` (and were the only relevant pages using the word "order"), but let me know if you can think of other pages that need to be updated.

(I also merged in the changes to those pages from the master branch; the diff will show more changes than I actually made.)

Keats:
I believe there is also the page in templates about Page showing the variables available in templates

codesections:
Ok, I added that page and fixed the minor issues you mentioned in your change request. 

When revising the docs, I noticed something that seems like a bug to me, but that was beyond my current ability to fix.  Right now `populate_previous_and_next_pages` is called with the `sort_by` front-matter variable.  This means that `page.lighter` and `page.heavier` are *only* populated if the `sort_by` is set to `weight` (and the flip for `earlier` and `later`).

This doesn't seem ideal/intuitive.  It seems like someone should be able to set the `date` variable in their pages and get links to the previous/earlier pages even if they're not sorting by `date`.  I'm not really sure how we'd need to fix this, though, since we populate the variable based on adjacent pages in the sorted array.  Sorting the array multiple times could work, but that would involve loops through a sorting process that (according to comments in the code) is not very efficient.  (And, anyway, I'm not sure how to do that). 

Maybe it's not worth solving, since that use case may be rare, and it's mentioned in the docs.  But I thought I should flag it. 

Also, it's just now occurring to me while typing this comment that `populate_previous_and_next_pages` is no longer an accurate name for that function.  Would you like me to change it globally?

Keats:
`populate_previous_and_next_pages` -> `populate_siblings`?

> This doesn't seem ideal/intuitive. It seems like someone should be able to set the date variable in their pages and get links to the previous/earlier pages even if they're not sorting by date

I don't think that's a wanted behaviour: without sorting, earlier/later and lighter/heavier are meaningless since the sorting might skip pages and you will end up with a different order in the list of pages than when using earlier/later.

codesections:
Ok, name updated!

Keats:
Thanks!

PR #364, opened 2018-08-07
Section assets

### LINKED ISSUE #340: Cannot use asset colocation inside section content
Consider a site with the following directory structure:

```
└── content/
    ├── about.md
    └── blog/
        ├── _index.md
        ├── blog_header_image.jpg
        ├── post1.md
        └── post2.md
```

Based on the [documentation regarding asset colocation](https://www.getgutenberg.io/documentation/content/overview/#assets-colocation), I would expect to be able to access `blog_header_image.jpg` with a relative link from `_index.md`.  Instead, it seems that there is not a way to access `blog_header_image.jpg`.

paulcmal:
Fixes #340 

I would be tempted to let `ignored_content` be handled outside of section/page and just feed the function an array of rules (currently what we get from config). I think it could allow - if need be - to have different rules for sections and pages, or have specific per-content rules in the frontmatter. And would reduce code duplication across page and section. What do you think?

Also, don't hesitate to read through my changes carefully as i've close to zero experience with Rust. The code looks correct to me and the test passes. Before writing the test i made sure it actually worked with actual content... and it did! (maybe there's edge cases i missed?)

On a side note, thank you for writing such clear code :heart:. I feel like everything is logically separated as it should, and readable in a glimpse. It's the first time for me i don't feel completely confused when encountering a CMS/SSG codebase :+1:

### DIFF
--- a/components/content/src/section.rs
+++ b/components/content/src/section.rs
@@ -82,9 +82,17 @@ impl Section {
     pub fn from_file<P: AsRef<Path>>(path: P, config: &Config) -> Result<Section> {
         let path = path.as_ref();
         let content = read_file(path)?;
+        println!("Parsing from file {:?}", path);
         let mut section = Section::parse(path, &content, config)?;
 
-        if section.file.name == "_index" {
+        println!("filename {:?}", section.file.name);
+        // I don't see any reason why, but section.file.name always is "_index"! ← bug
+
+        let file_name = path.file_name().unwrap();
+
+        // Is this check really necessary? Should the else case happen at all?
+        if file_name == "_index.md" || file_name == "index.md" {
+            // In any case, we're looking for assets inside parent directory
             let parent_dir = path.parent().unwrap();
             let assets = find_related_assets(parent_dir);
 
--- a/components/content/src/section.rs
+++ b/components/content/src/section.rs
@@ -84,30 +84,26 @@ impl Section {
         let content = read_file(path)?;
         let mut section = Section::parse(path, &content, config)?;
 
-        if section.file.name == "_index" {
-            let parent_dir = path.parent().unwrap();
-            let assets = find_related_assets(parent_dir);
+        let parent_dir = path.parent().unwrap();
+        let assets = find_related_assets(parent_dir);
 
-            if let Some(ref globset) = config.ignored_content_globset {
-                // `find_related_assets` only scans the immediate directory (it is not recursive) so our
-                // filtering only needs to work against the file_name component, not the full suffix. If
-                // `find_related_assets` was changed to also return files in subdirectories, we could
-                // use `PathBuf.strip_prefix` to remove the parent directory and then glob-filter
-                // against the remaining path. Note that the current behaviour effectively means that
-                // the `ignored_content` setting in the config file is limited to single-file glob
-                // patterns (no "**" patterns).
-                section.assets = assets.into_iter()
-                    .filter(|path|
-                        match path.file_name() {
-                            None => true,
-                            Some(file) => !globset.is_match(file)
-                        }
-                    ).collect();
-            } else {
-                section.assets = assets;
-            }
+        if let Some(ref globset) = config.ignored_content_globset {
+            // `find_related_assets` only scans the immediate directory (it is not recursive) so our
+            // filtering only needs to work against the file_name component, not the full suffix. If
+            // `find_related_assets` was changed to also return files in subdirectories, we could
+            // use `PathBuf.strip_prefix` to remove the parent directory and then glob-filter
+            // against the remaining path. Note that the current behaviour effectively means that
+            // the `ignored_content` setting in the config file is limited to single-file glob
+            // patterns (no "**" patterns).
+            section.assets = assets.into_iter()
+                .filter(|path|
+                    match path.file_name() {
+                        None => true,
+                        Some(file) => !globset.is_match(file)
+                    }
+                ).collect();
         } else {
-            section.assets = vec![];
+            section.assets = assets;
         }
 
         Ok(section)
--- a/components/content/src/section.rs
+++ b/components/content/src/section.rs
@@ -8,7 +8,7 @@ use serde::ser::{SerializeStruct, self};
 use config::Config;
 use front_matter::{SectionFrontMatter, split_section_content};
 use errors::{Result, ResultExt};
-use utils::fs::read_file;
+use utils::fs::{read_file, find_related_assets};
 use utils::templates::render_template;
 use utils::site::get_reading_analytics;
 use rendering::{RenderContext, Header, render_content};
@@ -33,6 +33,8 @@ pub struct Section {
     pub raw_content: String,
     /// The HTML rendered of the page
     pub content: String,
+    /// All the non-md files we found next to the .md file
+    pub assets: Vec<PathBuf>,
     /// All direct pages of that section
     pub pages: Vec<Page>,
     /// All pages that cannot be sorted in this section
@@ -54,6 +56,7 @@ impl Section {
             components: vec![],
             permalink: "".to_string(),
             raw_content: "".to_string(),
+            assets: vec![],
             content: "".to_string(),
             pages: vec![],
             ignored_pages: vec![],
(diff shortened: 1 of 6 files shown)


Keats:
Looks good!
cc @codesections 

The main question I would say is whether this should mirror the way pages does co-located assets: in a folder with a fixed named of `_index`. 
The downside of loading the files in the directory is that they might not be next to each other in the file explorer since they will be sorted by name, and it clashes with having to create a folder for pages.

paulcmal:
> whether this should mirror the way pages does co-located assets: in a folder with a fixed named of _index.

Do you have a better suggestion? :)

>  The downside of loading the files in the directory is that they might not be next to each other in the file explorer since they will be sorted by name, and it clashes with having to create a folder for pages.

Sorry but i don't get how your first point is a problem, and don't understand your second point. It looks very concise and point-making but unfortunately i just don't understand ^^

Keats:
Sorry, let me explain it better..

Currently to have co-located assets you need to create a folder and have a `index.md` file in it like in https://github.com/Keats/gutenberg/tree/master/test_site/content/posts/with-assets (see https://www.getgutenberg.io/documentation/content/overview/#assets-colocation for full docs).

This PR loads all non-md files in the same folder as a `_index.md` file as this section's asset. When you start having a few .md files in a directory like in https://github.com/Keats/vincentprouillet/tree/master/content/blog, the assets files will be spread around the directory based on their filename instead of being neatly co-located with the related _index.md.file

My idea was to use the same system the pages use: if we see a `_index.md` file, we don't load any assets. However if we find a `_index` folder in the content directory, we assume that it's equivalent to the `_index.md` file with the difference it has some co-located assets and load the `index.md` inside this folder.

> On a side note, thank you for writing such clear code heart. I feel like everything is logically separated as it should, and readable in a glimpse. It's the first time for me i don't feel completely confused when encountering a CMS/SSG codebase +1

I forgot to mention that in my first post: thank you!

paulcmal:
> Sorry, let me explain it better..

Thank you for taking the time to clarify :)

> However if we find a _index folder in the content directory, we assume that it's equivalent to the _index.md file with the difference it has some co-located assets and load the index.md inside this folder.

I feel like both are consistent behavior in their own way. Couldn't we support both? In a section, if there is an _index folder, load content/assets from there. Otherwise, check for _index.md and assets in the section folder directly. But in my opinion not allow to mix from both.

That's very personal, but i would be really confused the logic you're proposing however consistent it is: i would never remember when to name my file _index.md, and when to name it index.md.

Also because i don't have so many pages and files, i would be more bothered to have to traverse one more folder every time i want to edit a section than i would be to have children pages, content and assets in the same directory. Especially since because i'll be using page assets every child section/page will have its own dedicated folder (so no .md files just hanging around) and my file manager separates folders from regular files (like most do). So in the end to me it's virtually the same as having simply index.md and related assets in a dedicated directory, but without the cognitive trouble of having to deal with an additional folder.

Sorry that was a long answer, but as a conclusion i would like both behaviors to be supported. Are you okay with this? Would you accept a PR of mine in this direction if i give it a try? :heart:

codesections:
I actually like the behavior that @paulcmal implemented, and would favor keeping it was implemented in the PR as is without supporting a multiple ways to accomplish the same thing (which seems likely to be more confusing in the long run, as we experienced with `order` and `weight`). 

I think the behavior in the PR could be explained to users in a way that makes sense.  Let me explain it here, and then if it makes sense and we decide to go with it, I can add to the docs.

Let's start with a simple site without any asset co-location.  It has a homepage, an `about` page, and a `blog` section with three posts.  It's directory structure looks like this:

```
└── content/
    ├── _index.md
    ├── about.md
    └── blog/
        ├── post1.md
        ├── post2.md
        └── post3.md
```

Then, the owner of the site decides to use asset co-location for the blog posts.  To do this, they replace each post Markdown file with a directory of the same name, and then add an `index.md` file to each directory, along with whatever assets they want to co-locate.  After this, the directory structure looks like this
```
└── content/
    ├── _index.md
    ├── about.md
    └── blog/
        ├── _index.md
        ├── post1/
             ├── index.md
             └── image_for_post1.jpg
        ├── post2/
             ├── index.md
             └── image_for_post2.jpg
        └── post3/
             ├── index.md
             └── image_for_post3.jpg
```
And, inside `post1/index.md`, they can reference `image_for_post1.jpg` with a relative link.

Now, say they want to have an asset that they use in their `blog` section.  Under the PR, they'd handle this in exactly the same way as they did to add asset colocation for pages: add an image file to the directory, and use a relative link from the correct markdown folder within the folder.  So, their directory structure would look like this:

```
└── content/
    ├── _index.md
    ├── about.md
    └── blog/
        ├── _index.md
        ├── image_for_blog.jpg
        ├── post1/
             ├── index.md
             └── image_for_post1.jpg
        ├── post2/
             ├── index.md
             └── image_for_post2.jpg
        └── post3/
             ├── index.md
             └── image_for_post3.jpg
```

This way, we always follow the same rules:
 *  Any directory within `content` can contain any of four items:
    - an asset file (any file that doesn't end in `.md` is an asset file).  An asset file is not rendered automatically, but can be referenced using a relative link
    - a page file (any file that ends in `.md` that *isn't* titled `index.md` or `_index.md`).  A page file is rendered automatically using the `page.html` template, and is available at the `[directory_path]/[file_name]` url.
    - an `index.md` file (a file titled `index.md`).  An `index.md` file is rendered automatically using the `page.html` template and is available at the `[directory_path]` url.  
    - an `_index.md` file (a file titled `index.md`).  An `_index.md` file is rendered automatically using the `section.html` template and is available at the `directory_path` url.
    - a subdirectory, which can in turn contain any of the five items (including additional subdirectories).
 *  From any page or section, the user can access asset files stored in the same directory with a relative link 

This seems to me like it's using the same rules for sections and content—but it seems like @keats views it differently.  Do you view this as inconsistent?  If so, could you say more about how this treats the two differently—it seems the same to me.

paulcmal:
Hey, if that's okay with you maybe the implementation/technical chat can take place here, and more generic feature/orientation chat on #340?

I've started messing around with the code last night to implement the _index folder for asset collocation in sections. The main change so far is (more-than-improvable) spaghetti code taking place when loading section_entries:

```
.partition(|entry| ( (entry.as_path().file_name().unwrap() == "_index.md" && !entry.as_path().parent().unwrap().join("_index/index.md").is_file()) || (entry.as_path().parent().unwrap().file_name().unwrap() == "_index" && entry.as_path().file_name().unwrap() == "index.md")));
```

I've stumbled upon a minor bug in Section::from_file: section.file.name is always _index.md (even when the actual filename is index.md) and i didn't take the time to figure out why. Also, after the partition took place to separate section_entries from page_entries, i'm wondering if the additional filtering ([1](https://github.com/Keats/gutenberg/pull/364/commits/c7156a84f0ca4132c3380d75ba993f379a967cb5#diff-c599700a7ed060748704dcdcf0ea2f30R196), [2](https://github.com/Keats/gutenberg/pull/364/commits/c7156a84f0ca4132c3380d75ba993f379a967cb5#diff-c599700a7ed060748704dcdcf0ea2f30R213)) makes any sense. If i'm reading correctly, the partitioning that took place before already did the job. And in any case, i had to remove the first one to let "index/_index.md" find its place in the list of sections :)

Last but not least, i don't understand when [this else is met](https://github.com/Keats/gutenberg/blob/c7156a84f0ca4132c3380d75ba993f379a967cb5/components/content/src/section.rs#L117). If from_file is called from Gutenberg, checking for index/_index should take place before within site crate and i don't think the else would ever be met. If it's called by another program, i don't see a reason to enforce a naming policy for asset collocation.

So now about half the work is done. I just need to figure out how to pass proper info to the loaded section, because for now it considers _index/ as part of it:
```
Ok(Section { file: FileInfo { path: "/home/user/Sites/Gutenberg/content/blog/_index/index.md", name: "_index", relative: "blog/_index/_index.md", parent: "/home/user/Sites/Gutenberg/content/blog/_index", grand_parent: Some("/home/user/Sites/Gutenberg/content/blog"), components: ["blog", "_index"] }, meta: SectionFrontMatter { title: Some("Bloggity blog"), description: None, sort_by: None, weight: 0, template: None, paginate_by: None, paginate_path: "page", insert_anchor_links: None, render: true, redirect_to: None, in_search_index: true, extra: {} }, path: "blog/_index/", components: ["blog", "_index"], permalink: "http://127.0.0.1:1111/blog/_index/", raw_content: "\n![](capitalism.jpg)\n", content: "", assets: ["/home/user/Sites/Gutenberg/content/blog/_index/capitalism.jpg"], pages: [], ignored_pages: [], subsections: [], toc: [] })
```

So the fields that need to be edited for the section are:
- file name (should be index.md not _index.md), relative (what should it be??), parent grandparent and components (should be up one folder)
- path, components and permalink

↑↑ Any idea where i should start taking a look to address this? I'll continue working on this tonight or tomorrow. I left some comments and debug messages in the commit so anyone trying it out can see what's happening within :)

Have a great day!

PS: Don't hesitate to tell me if i ask too many questions. I'm just trying to get my head around your project but i do understand it can be troublesome to answer every stupid question ® that people like me can come up with in their twisted minds ^^

Keats:
> Section::from_file: section.file.name is always _index.md 

It might be hardcoded as only files named `_index.md` are considered sections.

I warmed up to @codesections argument so I think the original behaviour of the code was fine

paulcmal:
> I warmed up to @codesections argument so I think the original behaviour of the code was fine

Amazing :)

I've reverted my commit for the _index folder and brought in other changes as separate (smaller) commits so you can cherry-pick what you see fit! :tada: 

PS: Am i wrong about to believe my two latest commits (removing filters and if) are harmless optimizations?

Keats:
I think the optimization commits look fine. I'll have a look at the whole PR another day when I get more time

Keats:
The code looks good, any chance you can update the docs as well?

codesections:
I'd be happy to update the docs if you'd prefer @paulcmal .  It will probably be a few days before I can, though.  Or feel free to handle; just thought I'd offer.

paulcmal:
> The code looks good, any chance you can update the docs as well?

I gave it a try. What do you think?

> Or feel free to handle; just thought I'd offer.

Thank you <3 but i've handled this time. I'm sure you'll find more docs to write anyway :D

PR #370, opened 2018-08-12
Update page variable list for custom taxonomies

17cupsofcoffee:
I love the new support for custom taxonomies! I got a little thrown off when updating my site, though, as the docs didn't specify how the page would expose the taxonomies to the template. Wasn't too hard to figure it out, but I figured I'd save the next person who tries it the effort :)


Keats:
Thanks!

PR #378, opened 2018-08-22
CSV and TOML loading global functions

kellpossible:
This pull request adds the ability to load TOML files and CSV files into templates using these global functions: `load_csv` and `load_toml`

In the future it would also be cool to add:

 - More options for the CSV parser
 - Markdown table shortcode/filter/function which can load from CSV file.


kellpossible:
closed because I did this on the wrong branch

Issue #376, opened 2018-08-19
Summaries break reference-style Markdown links


Freaky:
Because `<!-- more -->` handling is done on the raw Markdown document, this breaks:

```markdown
[Foo][foo]
<!-- more -->
[foo]: http://foo.example
```

The reference gets split off from the content when the summary is rendered, leaving the links rendering literally.

While it's far from a satisfying fix, this works:

```rust
--- components/content/src/page.rs
+++ components/content/src/page.rs
@@ -191,13 +191,8 @@ impl Page {
         ).chain_err(|| format!("Failed to render content of {}", self.file.path.display()))?;
         self.content = res.0;
         self.toc = res.1;
-        if self.raw_content.contains("<!-- more -->") {
-            self.summary = Some({
-                let summary = self.raw_content.splitn(2, "<!-- more -->").collect::<Vec<&str>>()[0];
-                render_content(summary, &context)
-                    .chain_err(|| format!("Failed to render content of {}", self.file.path.display()))?.0
-            })
-        }
+        self.summary = self.content.find("<p><a name=\"continue-reading\"></a></p>")
+                                   .map(|pos| self.content[0..pos].to_owned());

         Ok(())
     }
```

Keats:
What would be a satisfying fix for you? Looking at it, I don't see a much better solution than your patch

Freaky:
Pushing it into rendering seems more appropriate, seeing as it's basically all about how raw content is parsed and depends on how it's rendered.

Freaky:
As an aside, most of the `Owned()` bits in the Markdown parser should really be `Borrowed()`.  e.g. `Owned(String::new())` &rarr; `Borrowed("")`.

PR #380, opened 2018-08-24
Improve summary handling

Freaky:
As per #376, this moves `<!-- more -->` handling into the Markdown parser.

In addition to fixing reference-style links and being a bit tidier, it should also be a bit faster, reducing the amount of unnecessary string searching and only rendering the document once.

While I'm here, swap over a lot of Owned() to Borrowed() to reduce unnecessary String allocations.

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -185,19 +185,12 @@ impl Page {
 
         context.tera_context.add("page", self);
 
-        let res = render_content(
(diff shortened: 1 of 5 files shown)


Keats:
Looks good! Can you add a test with the issue that was happening before in #376 ?

Freaky:
Tests updated and added to.

Keats:
Thanks!

Issue #381, opened 2018-08-25
Break out check_external_links into a command-line flag or subcommand


Freaky:
`check_external_links` is very expensive, and isn't even recommended to be left on even in the docs.  I think it would be more practical as either a run-time argument to `build`, or ideally as its own subcommand.

It's also quite frustrating for it to just stop on the first error, necessitating running it repeatedly, particularly when using the command for the first time.

Freaky:
Also: `Reason: Link https://crates.io/crates/rand is not valid: 404 Not Found`

I think this was an `Accept-Content` thing.

Keats:
For crates.io: I mentioned it on the post for 0.4 but it behaves badly and always return 404. I'm tempted to replace my crates.io use with https://crates.rs/

> It's also quite frustrating for it to just stop on the first error, necessitating running it repeatedly, particularly when using the command for the first time.

I agree this should be improved, instead of stopping at one error it should check all links and report all the ones not working. The issue is that it is currently part of the rendering which stops when it sees an error.

I'm not sure about having a separate command for that but i like the flag idea

RealOrangeOne:
The separate subcommand has a couple benefits I can see. Primarily with the fact it's not run at build time, it means it can be run as part of a CI job, and won't affect build time at all!

If this became a generic `zola check` command, it could mean that additional checks out-of-band could be done in future. (Not that I can think of any right now!)

Keats:
+1 on `zola check`, that's a great idea

RealOrangeOne:
I'd expect this can be implemented fairly simply using something like https://github.com/programble/scraper. Running this against the output directory, grabbing anything which looks like a link, and running it. It could use this to not only check external links, but also internal ones (you know, just in case they were created manually or something)

Keats:
This approach won't work since you can have files like drafts etc that are not published yet and I think it would be better to do that without having to render all the templates as it can still be slow when you a have huge number of pages.

The code for markdown link checking is already there, I think extending that to support internal link checking (https://github.com/getzola/zola/issues/539) will be the tricky bit but it's doable. We can have a separate markdown parsing function that only cares about the links, since actually rendering the markdown + syntax highlighting will slow it down.
If anyone wants to grab that issue, go for it!

PR #384, opened 2018-08-25
link_checker: Set Accept header

Freaky:
As mentioned in #381, crates.io 404's any request without an Accept:
text/html header.  It 200's any request with one, but at least
false-successes don't prevent checking any other links.

This also makes it easier to add a custom User-Agent if desired.

rustfmt and fix a clippy nit (unnecessary return) while I'm here.


Keats:
Thanks!

PR #389, opened 2018-08-27
Update pagination for test_site posts

Libbum:
Noticed that the pagination updates from #338 hadn't been propagated through to the `test_site`. This is just a simple fix for that.


Keats:
Thanks, Github is somehow buggy on my machine and I can't seem to change the base branch, any chances you can change it to `next`? I'll fix the tests myself and I'm a bit surprised the tests on master are not currently failing :o

Libbum:
Yeah, I couldn't see why either... But it's updated now. Hopefully they go through fine with the newer tests.

PR #391, opened 2018-08-29
Updates docs to match proper toml syntax.

### LINKED ISSUE #390: Taxonomies documentation incorrect.
I believe the [documentation](https://www.getgutenberg.io/documentation/getting-started/configuration/) for the new taxonomies configuration to be incorrect. The docs and the blog post both state that the format in `config.toml` should be:

```rust
taxonomies = [
    # each tag will have its own RSS feed
    {name: "tags", rss: true},
    # 5 items per page for a term, you can also customise the pagination path
    # like for the rest of the paginated content
    {name: "categories", paginate_by: 5},
    # Basic definition: no RSS or pagination
    {name: "authors"},
]
```
However, this is not valid `toml` which requires `=` instead of `:` for key/value pairs.  You can see this reflected in `gutenberg`'s own [tests](https://github.com/Keats/gutenberg/blob/742987144c0b70137fe715a3b4a4d05d01c38220/test_site/config.toml#L10):

```rust
taxonomies = [
    {name = "categories", rss = true},
]
``` 
I can send a PR if this can be confirmed by anyone else.

cldershem:
Fixes #390 .

### DIFF
--- a/docs/content/documentation/getting-started/configuration.md
+++ b/docs/content/documentation/getting-started/configuration.md
@@ -43,9 +43,9 @@ rss_limit = 20
 # The taxonomies to be rendered for that site and their configuration
 # Example:
 #     taxonomies = [
-#       {name: "tags", rss: true}, # each tag will have its own RSS feed
-#       {name: "categories", paginate_by: 5},  # 5 items per page for a term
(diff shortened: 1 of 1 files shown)


Keats:
Thanks for fixing it in the docs and in my blog!

PR #405, opened 2018-09-07
add troubleshooting doc

zoosky:


Keats:
This is already mentioned in https://www.getgutenberg.io/documentation/templates/overview/ so I don't think it makes sense to repeat it.
However it reminded me that the current implementation of `__tera_context` was kinda buggy but it is already fixed in Tera so it will be in the next release.

PR #410, opened 2018-09-09
Populate sections when page content changes

### LINKED ISSUE #409: Page changes don't propogate to sections for re-rendering
When a page is changed, the corresponding section is re-rendered, but since `populate_sections()` isn't called, the section still contains the old `Page` object and so the template doesn't reflect the change when it's rendered.

porglezomp:
Fixes #409

### DIFF
--- a/components/rebuild/src/lib.rs
+++ b/components/rebuild/src/lib.rs
@@ -194,6 +194,8 @@ fn handle_page_editing(site: &mut Site, path: &Path) -> Result<()> {
         Some(prev) => {
             // Front matter didn't change, only content did
             if site.pages[path].meta == prev.meta {
+                // We need to update the page object in the section so it will render correctly
+                site.populate_sections();
                 // Other than the page itself, the summary might be seen
                 // on a paginated list for a blog for example
                 if site.pages[path].summary.is_some() {


porglezomp:
`populate_sections()` seems like it's relatively expensive, so I think ideally we want to make the `pages` list hold references to pages, rather that copies of pages. I'm not sure exactly how Tera handles rendering variables, so I don't know what types of indirection is supported there.

porglezomp:
Okay, I see it's based on serde serialization, so we could put all of the pages into `Arc`s, I guess?

components/rebuild/src/lib.rs
```diff
@@ -194,6 +194,8 @@ fn handle_page_editing(site: &mut Site, path: &Path) -> Result<()> {
         Some(prev) => {
             // Front matter didn't change, only content did
             if site.pages[path].meta == prev.meta {
+                // We need to update the page object in the section so it will render correctly
+                site.populate_sections();
                 // Other than the page itself, the summary might be seen
                 // on a paginated list for a blog for example
                 if site.pages[path].summary.is_some() {
```
porglezomp, line 201: Probably need to change this condition, since the summary may have changed from something -> nothing, or even the full text might be used on a paginated list (my previous blog did this).

Keats:
> populate_sections() seems like it's relatively expensive, so I think ideally we want to make the pages list hold references to pages, rather that copies of pages.

That's one of the oldest issue still open: https://github.com/Keats/gutenberg/issues/205
I haven't got around to it yet but would happily take a PR reducing the number of copies.

The copying is the reason the rebuild is a bit conservative on what it does currently and favours doing less. Most of the time you're editing the content of the page you're on the page itself so that is what gutenberg reloads.

I will be happy to add that (and more) once `populate_sections` is less expensive

Keats:
I spent the day benchmarking various parts of the gutenberg pipeline including `populate_sections` and it is probably fast enough to be called there actually.

Keats:
A big refactoring has happened and it now is fast enough to call `populate_sections` pretty much everywhere on rebuild, which it is doing.

Thanks!

PR #411, opened 2018-09-09
[WIP] Add a prototype of transparent sections

porglezomp:
This implements "transparent" sections, which copy their pages into the parent
section's page list. This is useful for implementing sections that shouldn't
influence pagination, for instance, year sections underneath a posts section on
a blog.


Keats:
Now that the code is in much better shape to add this, are you interested in updating this PR @porglezomp ? I will do it otherwise.

Keats:
I have pushed an updated version to the `next` branch

PR #426, opened 2018-09-13
[doc] Add required libs to compile

pand-app:
Hello,

Just adding information to compile from sources

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -40,7 +40,7 @@ $ xbps-install -S gutenberg
 
 ## From source
 To build it from source, you will need to have Git, [Rust and Cargo](https://www.rust-lang.org/)
-installed.
+installed (also few compiling tools & libraries : make, libsass, g++, libssl-dev)
 
 From a terminal, you can now run the following command:
 


docs/content/documentation/getting-started/installation.md
```diff
@@ -40,7 +40,7 @@ $ xbps-install -S gutenberg
 
 ## From source
 To build it from source, you will need to have Git, [Rust and Cargo](https://www.rust-lang.org/)
-installed.
+installed (also few compiling tools & libraries : make, libsass, g++, libssl-dev)
```
Keats, line 43: which dependency requires `libssl-dev`?

Keats: Also those are only the dependencies for OSX/Linux, Windows is a bit different. Maybe I should just point to the libsass installation guide

pand-app: True, I don't know for libssl, but from alpine image, it's not working without libssl (I've not checked which one was failing :/)

Keats:
I detailed installation a bit more in https://github.com/Keats/gutenberg/pull/398/commits/cf5f22135f645781e46402d6f0208214cdf644ac
`libsass` should not be needed though as `sass-sys` is bundling the code and building it

PR #429, opened 2018-09-13
Update tera dependency to 0.11.16

pand-app:
Hello,

Today, [after-dark](https://github.com/Keats/after-dark) theme is broken, because tera version 0.11.14 contain some issues about macro, fixed in 0.11.15 according to [changelog](https://github.com/Keats/tera/blob/master/CHANGELOG.md#01115-2018-09-09).

There's a quick Pull request to update tera dependency thanks to `cargo update -p tera`

Seems to be working after that.


Keats:
Thanks! It is already updated in the `next` branch which is where development happens.

pand-app:
Hello,
I've created a tag "next" on my docker image, and I can confirm that everything build successfully right now, thanks for your work !

Issue #397, opened 2018-09-03
Defining page templates within `_index.md`


piedoom:
Hi there,

I apologize if this is answered somewhere, but I couldn't find it - 
I am creating a site with a main set of static pages, along with a blog.  For instance, my directories look like this:

```
content
    _index.md
    some_page.md
        - blog
          some page.md
          _index.md
```

Currently, in `content/blog/_index.md` I specify the index template like this:

`content/blog/_index.md`
```
+++
template = "blog/posts.html"
+++
```

Is it possible to specify the template for all pages that are defined with the `_index.md` file?  Like: 

`content/blog/_index.md`
```
+++
template = "blog/posts.html"
page_template = "blog/post.html"
+++
```

Right now I am manually defining the template in each blog page.

`content/blog/2018-08-05.md`
```
+++
title = "My blog post."
template = "blog/post.html"
date = 2018-08-06T03:42:41.845Z
+++
```

Thank you!

Keats:
Duplicate of #159 I think.
That's something I'd like to have and would welcome a PR for

IvoWingelaar:
I can probably spend some time hacking at this, as I had the exact same problem.

IvoWingelaar:
I got a basic version working, but I'm wondering about expected behavior in nested directories. What should the `meta.template` field be of `content/blog/another_level/page.md` if I define a `meta.page_template` in `content/blog/_index.md`? Current behavior will only override the template on direct pages of a section, not indirect nested pages.

I also assume that we want to give priority to the `template` field in a page, and only set it to the `page_template` if the page doesn't define a template? This allows a user to override the rendering for a single page.

Keats:
Those are good questions!

> What should the meta.template field be of content/blog/another_level/page.md if I define a meta.page_template in content/blog/_index.md? Current behavior will only override the template on direct pages of a section, not indirect nested pages.

I *think* it makes sense to override the nested pages as well but I'm not entirely sure if people would even use that fact. Changing that later on would be a breaking change though. @piedoom @upsuper what do you think? 
Looking at the `book` theme (https://github.com/Keats/book) this is a case where there are many subsections and it would be nice to have the page template apply to it. This particular case also raise the question of whether to support themes per section but that's another problem!

> I also assume that we want to give priority to the template field in a page, and only set it to the page_template if the page doesn't define a template? This allows a user to override the rendering for a single page.

Yes, setting a template directly should override whatever setting we set in the section.

[...]

PR #434, opened 2018-09-18
Add page templates to sections

IvoWingelaar:
Following this #397, it appears this issue is quite more involved then I initially suspected. It appears that applying a `page_template` recursively is a worthwhile addition, but to keep this PR minimal this is **not** currently done.


Keats:
Can you add it to the documentation as well?

Keats:
ping @IvoWingelaar 
Was there any issue with the recursiveness? I don't mind having the PR be a bit bigger

IvoWingelaar:
> Can you add it to the documentation as well?

Yeah, of course I can add it to the documentation! I had some real-life obligations that interfered.

> Was there any issue with the recursiveness? 

None that are show stopping. The obvious place to add the recursion fails some tests for reasons that I currently do not understand. But to be fair, I didn't had the chance to properly debug yet.

IvoWingelaar:
> The obvious place to add the recursion fails some tests for reasons that I currently do not understand.

In `site::Site` we store a `sections: HashMap` that owns the `Sections`. That's why we need to make clones of them when we populate `content::Section::subsections`. This necessitates two passes over all sections to apply the themes correctly: the first one to link the section to all its pages, and a second one to apply the themes.

Keats:
I have a pretty huge refactor locally that removes all the cloning. It's not ready yet but should be soonish so probably worth waiting a bit

IvoWingelaar:
Ah, then I'll just prepare the documentation, and wait for the refactoring.

Keats:
The refactoring has been merged

Keats:
@IvoWingelaar are you planning to finish the PR or should I do it?

Keats:
@IvoWingelaar ping

Keats:
I have pushed an updated version that works recursively in the `next` branch so closing this

Issue #425, opened 2018-09-13
Add a --no-color CLI option


ianare:
In certain use cases the color formatting when running Gutenberg causes display problems in the terminal or for certain automated tools and processes (Jenkins, Python subprocess, awk, etc).

Please add an option to disable color output in the console (suggest the standard `--no-color`).

Thanks!

Keats:
I'm currently using https://github.com/LukasKalbertodt/term-painter which is not maintained anymore apparently. Anyone knows another cross-platform crate for that? 

There seems to also be some activities with NO_COLOR or CLICOLOR as environment variables to control that and I seem to remember a discussion about that in Rust somewhere.

sondr3:
https://github.com/BurntSushi/termcolor might be of interest, I came upon it while finding a library for coloring output myself too but I just needed something easy (I used [colored](https://github.com/mackwic/colored) but it doesn't support Windows).

Keats:
Looks like `termcolor` does what we need. If anyone want to work on that, please do it on the `next` branch. It should be fairly easy as all the term coloring happens in `src/console.rs`.
Maybe we can match what `ripgrep` does as cli flag but I have no strong feeling about it, an env variable does seem cleaner to me

PR #438, opened 2018-09-22
Allow to disable colors

greizgh:
This adds options to handle color in output, as mentioned in #425.
Two environment variables allow to alter the coloring:
* `CLICOLOR` defaults to 1 and can disable coloring
* `CLICOLOR_FORCE` defaults to 0 and bypass tty checking

This follows the logic from https://bixense.com/clicolors/ although there is no real standard ruling color options.

A new dependency was added: [atty](https://github.com/softprops/atty), to check if we are in a tty (and avoid coloring output when piped).

I don't find my current handling of coloring flag in error/success/info functions really elegant, do not hesitate to suggest a cleaner way to do it.

For now I did not replace term_painter, but it does not seems to play well with the `CLICOLOR_FORCE` option. Switching to termcolor is the next step.

TODO:
- [x] replace term_painter by termcolor

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -756,16 +756,18 @@ name = "gutenberg"
 version = "0.4.2"
 dependencies = [
  "actix-web 0.7.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "atty 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "content 0.1.0",
  "ctrlc 3.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "errors 0.1.0",
  "front_matter 0.1.0",
+ "lazy_static 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Keats:
Neat thanks!

Can you also add support for NO_COLOR (http://no-color.org/) so we cover all usecases?

@ianare are you fine with env variables rather than a flag?

greizgh:
I just replaced term_paint with termcolor. It now correctly handles the CLICOLOR_FORCE flag.
Let me know if something does not look good.

ianare:
Great, thanks! I'm OK with the environment variables, since by default the output will not be colored when piped.

Keats:
Nice! Can you also update the docs (getting-started/cli-usage.md)?

greizgh:
Oh sure, I'll take a look at the doc.

greizgh:
I added a paragraph in cli-usage page, let me know if there is anything else.

RealOrangeOne:
I feel like `get_color_choice` (and `has_color` for that matter) could be converted to a `static`? Whilst it's not necessarily heavily computational, it is done every time something is logged, when the value will never change at runtime?

greizgh:
The check for color preferences is now done only once. This adds a dependency on the [lazy_static](https://crates.io/crates/lazy_static) crate.

Keats:
Thanks!

Issue #446, opened 2018-09-25
subsections seem to be empty on first run


NyxCode:
I just wrote a template (used as both `index.html` and `section.html`) which should display the content of sub-sections:

```
{% for section in section.subsections %}
        {%- if section.subsections -%}
            {% if section.pages %}
                {{ section.pages | length }} Pages, {{section.subsections | length}} Sections
            {% else %}
                {{ section.subsections | length }} Sections
            {% endif %}
        {%- elif section.pages -%}
            {{ section.pages | length }} Pages
        {%- else -%}
            Empty
        {% endif %}
{% endfor %}
```

My current directory structure looks like this:
```
content/
   section1/
        section2/
        page1.md
```
When I open the root of the page in a browser, I would expect to see that `section1` contains 1 page and 1 sub-section. Strangely, this is not what happens. Gutenberg doesn't seem to recognize the sub-section `section2` until a random change is made in the template (a newline somewhere in the file will do) and the files are re-compiled. When that happens, the page correctly displays the content of the section.

I have absolutely no idea why this is happening and, since I'm new to Gutenberg, how to track the issue down. So any help would be appreciated!

#### To reproduce:
`git clone https://github.com/NyxCode/hyde`
`cd hyde`
`git checkout sub-sections`
`gutenberg serve`
`# go to localhost:1111 - you should see "1 Pages" below "some-section"`
`# now, add a newline in `templates/section.html` between line 72 and 73 to trigger a re-build`
`# refresh the page - you should see "1 Pages, 1 Sections" now`

Again, I'm sorry I haven't got the knowledge to find the cause of this bug.

Keats:
That's an interesting bug! Not sure what is happening, maybe a bit of https://github.com/Keats/tera/issues/342 but that doesn't explain everything

NyxCode:
I don't know if this has something to do with the issue, but it seems a bit strange to me how `Site` stores sections. When having this directory structure:

```
content/
   outer/
       inner/
```
`Site.sections` contains these 3 key/value pairs after running `create_new_site`:

```
/content/_index.md  =>  Section(path = /, subsections = [Section(path = outer/, subsections = [])])
```

```
/content/outer/_index.md  =>  Section(path = outer, subsections = [Section(path = outer/inner/, subsections = [])])
```

```
/content/outer/inner/_index.md  =>  Section(path = outer/inner/, subsections = [])  
```

Keats:
What is weird about that?

NyxCode:
[...]

PR #447, opened 2018-09-26
Fix populate_sections when using nested subsections

NyxCode:
Until now, `Site.sections` was incorrect when nested subsections were used. This PR fixes this behavior, together with #446.


Keats:
I moved it to the `next` branch, can you remove the commits from master?

NyxCode:
-__- let me just create a new one, sorry for the fuckups.

Issue #213, opened 2018-01-16
Add Proxy config variable


FreeMasen:
To allow for an easier time developing a solution that might use a static file server and REST endpoints, it would be nice to have the serve command proxy requests that don't match on the static route to a configurable port.

A good use for this would be for developing a contact page, that would need to handle a HTTP Post request with the form's contents. Currently these need to be debugged separately, to avoid CORS issues with the REST request. It would be nice to have a config variable or "proxy_port" that would forward requests/responses to/from that port on the localhost.

Keats:
Hmm that seems like it's easier to add CORS when needed on your server for localhost no?

FreeMasen:
In the past I have had a difficult time setting up CORS that works in Chrome/Firefox, it would also mean that the form `action` attribute would need to be modified from debug to release builds. I have spent a little time over the past few months trying to get some kind of proxy behavior out of hyper and the complexity seems much greater than the scope of this project.

Maybe a better route would be to have a `watch` sub-command that just builds on file change, essentially the same behavior as `serve` but no server needs to start and no script needs to be injected. While it is still two processes, it would reduce the need for making changes from debug to release.

PR #450, opened 2018-09-28
add watch command

FreeMasen:
In an effort to address #213 this pull request breaks the watch functionality out from the `serve` sub-command into its own sub-command. 

This would allow someone include `gutenbreg` in their development workflow when utilizing another server. While this doesn't add proxy functionality it does solve the problem that prompted the issue.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -80,7 +80,6 @@ fn main() {
             let watch_only = matches.is_present("watch_only");
             let output_dir = matches.value_of("output_dir").unwrap();
             let base_url = matches.value_of("base_url").unwrap();
-            println!("watch_only: {}", watch_only);
             console::info("Building site...");
             match cmd::serve(interface, port, output_dir, base_url, config_file, watch_only) {
                 Ok(()) => (),
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,11 +77,12 @@ fn main() {
                     ::std::process::exit(1);
                 }
             }
-
+            let watch_only = matches.is_present("watch_only");
             let output_dir = matches.value_of("output_dir").unwrap();
             let base_url = matches.value_of("base_url").unwrap();
+            println!("watch_only: {}", watch_only);
             console::info("Building site...");
-            match cmd::serve(interface, port, output_dir, base_url, config_file) {
+            match cmd::serve(interface, port, output_dir, base_url, config_file, watch_only) {
                 Ok(()) => (),
                 Err(e) => {
                     console::unravel_errors("", &e);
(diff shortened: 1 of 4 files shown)


Keats:
Hmm I am a bit hesitant to accept this as it adds a command for an issue that is not too hard to fix by adding CORS from the other end, which is only a couple of lines in Django for example.

FreeMasen:
So CORS does not feel like a solution to me.

Say I have a site built with gutenberg and I am going to serve it with nginx in production. I have Nginx configured to proxy pass the request for the route `/email/` to a custom server I have setup.

On one of the pages I would like to include a simple form for sending me an email, this form would have the attribute `action="/email/"` or something similar. To use `gutenberg serve` while working out the kinks in my email service, I would first need to add CORS to my email service which I only need during development. While Django might offer a great experience adding CORS, in a less featureful server framework adding CORS is not always trivial.  In addition, I would now need to setup a special configuration for my `gutenberg` template to handle the different address for these requests. This could be done with a conditional in the template like this

```jinja
<form action="{% if page.extra.debug %}http://localhost:5555/email/{% else %}/email/{% endif %}">

</form>
```

Then I would need to maintain the page variable for the front-matter on the `_index.html` for the contact page with a new property that includes `debug = false` when I push out my changes.

The other option would be to manually update this attribute, which again is not super great since it would be easy to forget.

Another situation where this command would be useful is using gutenberg in conjunction with other build tools that might offer similar function to the serve command. For example `webpack-dev-server` implements configurations that allow for not-only forwarding request but also automatically inserting headers, compression, custom before and after behavior, route re-writing, and much more.

FreeMasen:
One more thing I will add to my above comment.

The two examples I provided in my last comment are real world experiences that I have had using `gutenberg` to build [my own webiste](https://wiredforge.com).

Keats:
How about having a `gutenberg serve --no-server` like Jekyll  does rather than `gutenberg watch`?

Keats:
Or maybe `--watch-only`, whichever sounds the best

FreeMasen:
I can do that for sure. I am at Rust Belt Rust this weekend but will probably be able to tackle it this coming week

Keats:
No worries, take your time!

Keats:
@FreeMasen are you still planning on implementing that? Looking back at the options I think I prefer `--watch-only

FreeMasen:
Sorry I got a little side-tracked. I will get to work on this today and tomorrow.

src/main.rs
```diff
@@ -77,11 +77,12 @@ fn main() {
                     ::std::process::exit(1);
                 }
             }
-
+            let watch_only = matches.is_present("watch_only");
             let output_dir = matches.value_of("output_dir").unwrap();
             let base_url = matches.value_of("base_url").unwrap();
+            println!("watch_only: {}", watch_only);
```
Keats: this can be removed

Keats (changes requested):
One println to remove and it's ready to go, thanks!

src/main.rs
FreeMasen: Sorry about that I had removed that locally but apparently forgot to push

Keats:
Thanks!

PR #451, opened 2018-09-28
Fix populate_sections when using nested subsections

NyxCode:
Until now, `Site.sections` was incorrect when nested subsections were used. This PR fixes this behavior, together with #446.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -365,6 +365,7 @@ impl Site {
             section.ignored_pages = vec![];
         }
 
+        // Put pages into the 'pages' field of their section
         for page in self.pages.values() {
             let parent_section_path = page.file.parent.join("_index.md");
             if self.sections.contains_key(&parent_section_path) {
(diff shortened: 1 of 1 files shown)


Keats:
This would be a potentially pretty big slowdown for big sites AFAIK as the index section will now have every single page of the site every time. Might not be a big deal but I'll need to check first

NyxCode:
yeah, you're right. We need to think about a better solution than this.
How about storing the actual pages in a plain `vec` inside `Site` and then only storing references/indices?

Keats:
> How about storing the actual pages in a plain vec inside Site and then only storing references/indices?

A pretty big refactoring has landed doing exactly that. The main issue that remains is that sections are serialized with their subsections/pages to pass to the template engine so that is the bottleneck.
The easiest way to see if it is a big performance issue is the following:

- pull the `next` branch and do a `cargo build --release` of the current `next`
- generate all benches site in `components/site/benches` by running python gen.py
- check how long it takes to `cargo build` big-blog, huge-blog, huge-kb and the gutenberg docs
- add back your patch, you will need to move this to the library.rs file now
- another `cargo build --release`
- rebuild all the sites mentioned before and check if there are significant differences

Keats:
Going to close this in favour of #484

PR #453, opened 2018-09-29
Implement an opportunistic watcher for all optional directories

devurandom:
This allows Gutenberg to start with just a config file.  All other directories
 will be picked up once they are created.  Directories which are not necessary
 will thus not cause errors.

Closes: https://github.com/Keats/gutenberg/issues/445

I expect that this will require quite some work (and documentation, which is currently completely missing) until it can be accepted, because this is one of my first pieces of Rust code. Please do not hesitate to provide feedback!

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -29,10 +29,12 @@ use std::sync::mpsc::channel;
 use std::time::{Instant, Duration};
 use std::thread;
 
+use opportunistic_watcher::OpportunisticWatcher;
+
 use chrono::prelude::*;
 use actix_web::{self, fs, http, server, App, HttpRequest, HttpResponse, Responder};
 use actix_web::middleware::{Middleware, Started, Response};
-use notify::{Watcher, RecursiveMode, watcher};
+use notify::RecursiveMode;
 use ws::{WebSocket, Sender, Message};
 use ctrlc;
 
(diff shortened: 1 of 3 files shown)


src/cmd/serve.rs
```diff
@@ -147,31 +149,32 @@ fn handle_directory<'a, 'b>(dir: &'a fs::Directory, req: &'b HttpRequest) -> io:
     fs::NamedFile::open(path)?.respond_to(req)
 }
 
+const OPTIONAL_WATCH_DIRS: [&str; 4] = ["content", "sass", "static", "templates"];
```
Keats, line 152: `content` is not optional, there is nothing to do for gutenberg if it isn't there

Keats (changes requested):
That does seem like the fix is too complicated. I would have expected a ~10 lines diff as there is no need for an abstraction there imo

devurandom:
"Abstraction" refers to the `OpportunisticWatcher`? My intention was to do it in a way that provides a basis for an addition to the notify crate, so more users could benefit. Sadly I'm new to Rust, so I opted to start with something that fits Gutenberg and then expand it later.

Keats:
> "Abstraction" refers to the OpportunisticWatcher

Yep, it might make sense for the `notify` crate but in Gutenberg it would be overkill imo

Keats:
@devurandom Are you interested in adding the basic fix for that issue? IE doing it like it's done for the `static` folder

devurandom:
* I could try to simplify my current code (flatten the `struct`/`impl` into local variables and regular functions). I think this is superior to having a series of `if` blocks.
* I am currently very time constrained, probably until mid November, so the chances of this happening earlier than that are slim.

Keats:
Ok, thanks! I've pushed https://github.com/Keats/gutenberg/pull/398/commits/652196969ce77f260749e4d1cefbd546f5f74b27 in the meantime, let me know if notify integrates something similar later on

PR #454, opened 2018-09-30
Update reqwest to 0.9

ccoors:
Because reqwest 0.8 uses the openssl 0.9 crate which does not work with OpenSSL 1.1.x.


Keats:
I actually did that a couple of minutes ago on the next branch!

ccoors:
Oh, I didn't see that, sorry.

Keats:
No worries it was literally done a few minutes ago. Your PR made me realise I was doing it wrong actually so I updated the `next` branch if you want to have a look

PR #455, opened 2018-09-30
Fix some clippy warnings

ccoors:
This fixes some of the more trivial things clippy complains about. Should not contain any functional changes.


Keats:
Thanks!

PR #457, opened 2018-10-01
Replace hard-coded directory slash in serve output

### LINKED ISSUE #456: Serve uses a hard-coded "/" in CLI output
The `serve` command outputs a helpful line about which directories are being watched:

```
Listening for changes in /Users/(...)/{content, templates, config.toml, static, sass}
```

The last slash is hard-coded, which looks bad when working on a Windows host.

```
Listening for changes in C:\\(...)/{content, templates, config.toml, static, sass}
```

This can be fixed, by replacing the trailing slash with an OS-specific one.

vsupalov:
Add an OS-specific trailing slash to the `serve` CLI output, instead of a hard-coded "/". Fixes #456.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -24,7 +24,7 @@
 use std::env;
 use std::fs::{remove_dir_all, File};
 use std::io::{self, Read};
-use std::path::{Path, PathBuf};
+use std::path::{Path, PathBuf, MAIN_SEPARATOR};
 use std::sync::mpsc::channel;
 use std::time::{Instant, Duration};
 use std::thread;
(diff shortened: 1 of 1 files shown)


src/cmd/serve.rs
```diff
@@ -229,7 +229,9 @@ pub fn serve(interface: &str, port: &str, output_dir: &str, base_url: &str, conf
         watchers.push("sass");
     }
 
-    println!("Listening for changes in {}/{{{}}}", pwd.display(), watchers.join(", "));
+    let mut pwd_directory = pwd.clone();
+    pwd_directory.push(""); // add OS-specific trailing slash
```
RealOrangeOne: How difficult would it be to use something like https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html? Rather than cloning the data and mutating it? Not sure how crazy complicated that'd make the print formatting though?

Keats: +1 for this approach

vsupalov: Thanks for the hint! It's pretty straightforward, and doesn't look any more complicated. Added a corrected commit to the PR.

Keats:
Thanks!

PR #458, opened 2018-10-02
Feat: inject config in robots.txt context

greizgh:
### Use case

Creating a robots.txt referencing a sitemap:

```
User-agent: *
Sitemap: https://mydomain.tld/sitemap.xml
```

To achieve this, one could hardcode the base URL, but it feels kind of dirty.
A cleaner way would be to use the `config.base_url` value.

### Introduced change

The config object is now passed in the template context, allowing to write a template like this:

```
Sitemap: {{config.base_url}}/sitemap.xml
```

The documentation has been updated to reflect this change. Also a test case checks that the produced file does contains the configured base_url.

Let me now if there is something to change.


Keats:
Thanks!

PR #459, opened 2018-10-03
Slotmap refactor

Keats:
Ready for reviews!


components/library/src/content/section.rs
```diff
@@ -1,9 +1,8 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::result::Result as StdResult;
 
-use tera::{Tera, Context as TeraContext};
-use serde::ser::{SerializeStruct, self};
+use tera::{Tera, Context as TeraContext, Value};
+use slotmap::{Key};
```
RealOrangeOne, line 5: Curleys here are unnecessary.

Meta: `rustfmt`?

Keats: I will rustfmt the whole thing before releasing to avoid breaking every PR

PR #465, opened 2018-10-07
Add `get_json` function

RealOrangeOne:
Add a template function to request JSON data from remote endpoints. This is currently a very minimal implementation, which doesn't cache anything between runs (the cache lives purely in-memory). Future iterations may include a filesystem cache, and support for other datatypes (eg CSV). I added a fairly naive example to the docs site which tests both the response, and whether the response is correctly cached. 

Adding support for remote content like this is a controvertial feature in static site generators, because it both adds an external dependency on the upstream API value, but also means the site requires an active internet connection to build. I personally use it quite a lot to fetch fairly fixed metadata, or to build custom widgets for things which expose an API, but have fairly terrible embed support. If you don't actually use the function, then no requests are actually made.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2129,6 +2129,7 @@ dependencies = [
  "imageproc 0.1.0",
  "lazy_static 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "pulldown-cmark 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "reqwest 0.9.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "taxonomies 0.1.0",
  "tera 0.11.16 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


RealOrangeOne:
(Netlify is failing because it can't find the new function, which is to be expected)

Keats:
This looks like a duplicate of https://github.com/Keats/gutenberg/pull/379 ? Despite the name of the PR it also loads JSON now.

RealOrangeOne:
Not quite. By the looks of it, #379 loads data from the project itself. This requests it from a remote.

Keats:
See the discussion from https://github.com/Keats/gutenberg/pull/379#issuecomment-415892025

#379 is starting with local files but the goal is to add remote to it afterwards

RealOrangeOne:
aha, in that case yes, this is a duplicate. And that looks like a far more robust implementation anyway :+1:

PR #472, opened 2018-10-09
Allow inclusion of all items in RSS feeds.

### LINKED ISSUE #468: Docs: how to do "all" for RSS feed?
Reading the docs, the only info I see on setting number of items to include in the RSS feed is:

```toml
# The number of articles to include in the RSS feed
rss_limit = 20
```

In my case, I prefer never to truncate. Should the value be set to blank (null), 0, -1, an absurdly large number I'll never hit…?

(I'm happy to open a docs PR to clarify this.)

chriskrycho:
Switch to an `Option<usize>` for the serialized value of `rss_items`. This lets us just set a blank value in the configuration and thereby include *all* items.

This is a backwards-compatible change; it does not affect the behavior of existing configurations.

- Fixes #468.
- Closes #471. This alternative approach covers the same territory as that documentation change but in a better way in my view.

### DIFF
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -5,13 +5,13 @@ use std::path::{Path, PathBuf};
 
 use chrono::Utc;
 use globset::{Glob, GlobSet, GlobSetBuilder};
+use syntect::parsing::{SyntaxSet, SyntaxSetBuilder};
 use toml;
 use toml::Value as Toml;
-use syntect::parsing::{SyntaxSet, SyntaxSetBuilder};
 
-use theme::Theme;
-use highlighting::THEME_SET;
 use errors::{Result, ResultExt};
+use highlighting::THEME_SET;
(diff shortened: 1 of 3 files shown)


chriskrycho:
So the docs don't look so happy done this way:

null

It looks like what I should do is just comment it out, if you want to take this approach?

```toml
# The number of articles to include in the RSS feed. Will include all items if
# not set.
# rss_limit = 20
```

Keats:
Development is happening on the `next` branch, I switched the PR to it and it seems there are some conflicts? Idea and code look good otherwise

chriskrycho:
I’ll do a merge and fix the conflicts! Probably tonight or tomorrow.

chriskrycho:
@Keats I ended up just recreating this entirely as a single commit on top of `next`. Should be good to go now!

Keats:
Thanks!

PR #474, opened 2018-10-09
Fix content parser's 'text' rule to use the correct production

jwatt:


jwatt:
Hmm. So https://github.com/Keats/gutenberg/#contributing says to develop on the 'next' branch, but the infra isn't able to build that...or something?

jwatt:
Ah, my version of the 'next' branch was out of date. Hopefully that should go smoothly now.

Keats:
Can you add some tests to make sure it is fixing it?

jwatt:
Actually, I believe this bug doesn't break anything just now because of the ordering in the `content` rule:

    content = _{
        ignored_inline_shortcode |
        inline_shortcode |
        ignored_shortcode_with_body |
        shortcode_with_body |
        text
    }

Since `ignored_shortcode_with_body` and `shortcode_with_body` come before `text`, [ignored] shortcodes with a body will always be parsed fully before the faulty `text` rule gets a chance to swallow up their opening `{% ... %}`. If the `text` rule was to come before those two, then it would be a problem.

Keats:
The code in the PR does make more sense than what there is currently though so I'll merge it

PR #483, opened 2018-10-13
Add F# syntax highlighting.

### LINKED ISSUE #467: Unsupported syntax highlighting languages: F♯ and ReasonML
I'm about to start building a new version of my website, and I have an ongoing series *now* and may also in the future which involves Rust, Elm, F♯, and ReasonML. As such, it'd be great to have syntax highlighting for F♯ and ReasonML!

I'll happily just use the `extra_syntaxes` option in the meantime, of course.

chriskrycho:
Fixes #467.

### DIFF
--- a/sublime_syntaxes/FSharp.sublime-syntax
+++ b/sublime_syntaxes/FSharp.sublime-syntax
@@ -1,0 +1,499 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: fsharp
+file_extensions:
+  - fs
+scope: source.fsharp
+contexts:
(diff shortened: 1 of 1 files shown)


chriskrycho:
This is derived from the [Ionide TextMate grammar](https://github.com/ionide/ionide-fsgrammar/blob/master/grammar/fsharp.json); I had to simplify two of the comment definitions slightly because that grammar was using a `"while"` key that Sublime doesn't know about, but it's otherwise the same.

Keats:
Nice, thanks! Is there an easy way to translate JSON -> sublime-syntax or was it done by hand?

PR #488, opened 2018-10-18
prevent html tags from appearing in the toc

### LINKED ISSUE #487: tags in headers results in unexpected behaviors in table of contents
This is really two and a half problems.

Regardless of the tag, the header `id` (and the hash portion of the permalink) is truncated. Either to the content of the tag, if the tag is appears before before any text, or to the text preceding the first tag.

Certain tags appear in the toc titles, like `a` or `code` tags, while others, like `strong` and `em`, don't.

Bonus potential Tera bug, `striptags` doesn't work to strip the appearing tags.

bdjnk:
This resolves #487 by not including html tags while building header information for the table of contents.

It retains html tags in headers for the regular markdown output, as previously.

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -77,16 +77,11 @@ pub fn markdown_to_html(content: &str, context: &RenderContext) -> Result<Render
                     // Header first
                     if in_header {
                         if header_created {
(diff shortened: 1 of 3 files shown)


Keats:
Perfect, thanks!

PR #492, opened 2018-10-20
Ensure root output directory exists explicitly

maghoff:
Aliases that have no directory nesting sneakily avoid the code path that ensures the directories exist. A way to fix this is to call `ensure_directory_exists` explicitly. This is implemented in this pull request.

Test data has been updated so if the implementation change is not applied, tests will fail.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -623,6 +623,7 @@ impl Site {
     }
 
     pub fn render_aliases(&self) -> Result<()> {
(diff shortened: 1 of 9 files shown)


Keats:
I changed the commit to be on `next` which is where dev happens but it pulled all the commits from master as well :(
Can you cherry-pick your commit on  the next branch and force push?

maghoff:
Done, see #496

PR #494, opened 2018-10-20
Remote data

### LINKED ISSUE #476: Load data from remote endpoints
Originally discussed in https://github.com/Keats/gutenberg/pull/379, Gutenberg should have the ability to fetch data from remote endpoints, whether this be in JSON, CSV, TOML, or indeed plaintext. 

The API should be very simple, which would allow for a large amount of code reuse, as defined in https://github.com/Keats/gutenberg/pull/379#issuecomment-415892025. Simply swapping out the `path` parameter for `url` can switch between remove and local data file fetching. There's no reason we need anything more complicated than that. 

I'm more than happy to do the fix, porting the work I did in https://github.com/Keats/gutenberg/pull/465 to @kellpossible 's implementation. Creating this issue to track the feature request, and related discussions.

RealOrangeOne:
Fixes #476.
Re-implements #465
Extends #379 

Allow loading data from remote endpoints, and parsing through the existing formats `load_data` supports. 

This PR is intentionally designed to reuse as much of the code as possible when reading from data files or remote endpoints. This not only reduces the chance of odd bugs appearing, but means I don't have to write as much! Because I also moved things out into separate files, GitHub makes it looks like i've done way more work than I actually have.

- Support requesting remote data
- Allow loading data as plain text, with no parsing
- Implement a cache to reduce repeat API calls / parsing
- Move data loading into separate file to split it out, it's quite a lot of code!
- Send the current version of zola in the user agent. There's no reason for this, I just thought it might be nice!
- Rename `kind` argument `format`, as it's more appropriate
- If someone tries to load files from outside their site directory, throw an error. This _should_ also include symlinks

It would be possible to edit the headers when sending / receiving a response with the `format` argument. But i decided not to do that, and we / someone else can add it further down the line.

None of this code has been run through `rustfmt`, intentionally.

Note: Currently the test `can_load_remote_data` is failing, when it shouldn't be. This will fix itself once the repo move occurs.

cc @kellpossible, enjoy!


RealOrangeOne:
(Netlify is failing because it can't find the new function, which is to be expected)

components/templates/src/lib.rs
```diff
@@ -5,6 +5,13 @@ extern crate tera;
 extern crate base64;
 extern crate pulldown_cmark;
 extern crate csv;
+extern crate crypto_hash;
+extern crate chrono;
+extern crate reqwest;
+extern crate url;
+
+#[macro_use]
+extern crate clap;
```
Keats: why are we loading clap?

components/templates/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "templates"
-version = "0.1.0"
+# Version must match that of root Cargo.toml
```
Keats: why?

components/templates/Cargo.toml
```diff
@@ -12,6 +13,11 @@ toml = "0.4"
 csv = "1"
 serde_json = "1.0"
 error-chain = "0.12"
+crypto-hash = "0.3.1"
```
Keats: we don't need a crypto hash, we can use the one from std (`std::hash::Hash`) like the imageproc sub-crate does

components/templates/Cargo.toml
```diff
@@ -12,6 +13,11 @@ toml = "0.4"
...
+chrono = "0.4"
+reqwest = "0.9"
+url = "1.5"
+clap = "2"
```
Keats: I don't think clap is needed?

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+use csv::Reader;
+use std::collections::HashMap;
+use tera::{GlobalFn, Value, from_value, to_value, Result, Map};
+use std::ops::BitXor;
+
+static GET_DATA_ARGUMENT_ERROR_MESSAGE: &str = "`load_data`: requires EITHER a `path` or `url` argument";
+
+enum ProvidedArgument {
+    URL(Url),
+    PATH(PathBuf)
```
Keats: enum variants should be lowercase, `Url` and `Path`.
The enum name would be probably clearer as `DataSource` as well

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+        GET_DATA_ARGUMENT_ERROR_MESSAGE
+    );
+
+    let url_arg = optional_arg!(
+        String,
+        args.get("url"),
+        GET_DATA_ARGUMENT_ERROR_MESSAGE
+    );
+
+    if !path_arg.is_some().bitxor(url_arg.is_some()) {
```
Keats: I think it would be easier to read as `if path_arg.is_some() && url_arg.is_some()`
The case where both are empty will be taken care of at the end of the function

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+
+fn read_data_file(base_path: &PathBuf, full_path: PathBuf) -> Result<String> {
+    if !is_path_in_directory(&base_path, &full_path).map_err(|e| format!("Failed to read data file {}: {}", full_path.display(), e))? {
+        return Err(format!("{} is not inside the base site directory {}", full_path.display(), base_path.display()).into());
+    }
+    return read_file(&full_path)
+        .map_err(|e| format!("`load_data`: error {} loading file {}", full_path.to_str().unwrap(), e).into());
+}
+
+fn get_output_format_from_args(args: &HashMap<String, Value>, provided_argument: &ProvidedArgument) -> Result<String> {
```
Keats: This function should return an enum of the supported types to avoid checking strings later

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    if let Some(path) = path_arg {
+        let full_path = content_path.join(path);
+        if !full_path.exists() {
+            return Err(format!("{} doesn't exist", full_path.display()).into());
+        }
+        return Ok(ProvidedArgument::PATH(full_path));
+    }
+    else if let Some(url) = url_arg {
```
Keats: can be only an `if` as all branches in the original `if` return

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+        _ => Ok(String::from("plain"))
+    }
+}
+
+
+/// A global function to load data from a data file.
+/// Currently the supported formats are json, toml and csv
+pub fn make_load_data(content_path: PathBuf, base_path: PathBuf) -> GlobalFn {
+    let mut headers = header::HeaderMap::new();
+    headers.insert(header::USER_AGENT, format!("{} {}", crate_name!(), crate_version!()).parse().unwrap());
```
Keats: is clap added to the dependencies just for that? I would just hardcode `zola` and not have the version imo, it's not that important.

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mut cache = result_cache.lock().expect("result cache lock");
+        let response_client = client.lock().expect("response client lock");
+        if let Some(cached_result) = cache.get(&cache_key) {
+            return Ok(cached_result.clone());
+        }
+
+        let data = match provided_argument {
+            ProvidedArgument::PATH(path) => read_data_file(&base_path, path),
+            ProvidedArgument::URL(url) => {
+                let mut response = response_client.get(url.as_str()).send().and_then(|res| res.error_for_status()).map_err(|e| format!("Failed to request {}: {}", url, e.status().expect("response status")))?;
```
Keats: URL requests should add accept headers to the request since some sites send content following it.

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+                response.text().map_err(|e| format!("Failed to parse response from {}: {:?}", url, e).into())
+            },
+        }?;
+
+        let result_value: Result<Value> = match file_format.as_str() {
+            "toml" => load_toml(data),
+            "csv" => load_csv(data),
+            "json" => load_json(data),
+            "plain" => to_value(data).map_err(|e| e.into()),
```
Keats: I'm not sure about `plain`, what is it even loading?

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        result_value
+    })
+}
+
+/// load/parse a json file from the given path and place it into a
+/// tera value
+fn load_json(json_data: String) -> Result<Value> {
+    let json_content = serde_json::from_str(json_data.as_str()).map_err(|e| format!("{:?}", e))?;
+    let tera_value: Value = json_content;
```
Keats: you can fold that into the previous line like load_toml I think

components/utils/src/fs.rs
```diff
@@ -93,6 +100,18 @@ pub fn copy_directory(src: &PathBuf, dest: &PathBuf) -> Result<()> {
     Ok(())
 }
 
+pub fn get_file_time(path: &Path) -> Option<SystemTime> {
+    println!("{}", path.display());
```
Keats: can remove that

docs/content/documentation/templates/overview.md
```diff
@@ -186,7 +187,28 @@ template:
...
+
+
+```jinja2
+{% set response = load_data(url="https://api.github.com/repos/getzola/zola", format="json") %}
+{{ response }}
+```
+
+#### Data Caching
+
+Data file loading and remote requests are cached in memory memory during build, so multiple requests aren't made to the same endpoint. URLs are cached based on the URL, and data files are cached based on the files modified time. The format is also taken into account when caching, so a request will be sent twice if it's loaded with 2 differnet formats.
```
Keats: memory memory in the first line

Keats: differnet -> different in the last words

docs/templates/index.html
```diff
@@ -14,10 +14,11 @@
 
         <header>
             <nav class="{% block extra_nav_class %}container{% endblock extra_nav_class %}">
+                {% set gh_response = load_data(url="https://api.github.com/repos/getzola/zola", format="json") %}
```
Keats: I wouldn't use that in the template of the docs

Keats (changes requested):

components/templates/Cargo.toml
RealOrangeOne: ah, i somehow missed the implementation for `String` :+1:  And that we don't necessarily need an actual checksum, just some kind of hashed value would do

components/templates/Cargo.toml
RealOrangeOne: The reason I loaded `clap` into this package was for the `crate_version` macro for use in the header. Seeing as it's already in the binary, the additional overhead should be minimal.

components/templates/src/global_fns/load_data.rs
RealOrangeOne: `plain` is just the format which returns the data raw, with no additional processing. Either the request body or the file contents.

components/templates/src/global_fns/load_data.rs
RealOrangeOne: I can. I intentionally didn't modify these functions more than I had to, not sure how I missed this line in the cleanup!

components/templates/Cargo.toml
RealOrangeOne: See https://github.com/Keats/gutenberg/pull/494#discussion_r227265030

components/templates/src/lib.rs
RealOrangeOne: See https://github.com/Keats/gutenberg/pull/494#discussion_r227265030

components/templates/src/global_fns/load_data.rs
RealOrangeOne: Of course, that makes sense. It doesn't look like there's an existing library to handle converting extensions to mimetypes, but a simple `match` based on the output enum from https://github.com/Keats/gutenberg/pull/494#discussion_r227107910 will cover that easily enough!

components/templates/src/global_fns/load_data.rs
Keats: but when is it needed? I can't really think of a usecase where you would want that

components/templates/Cargo.toml
Keats: I would still remove the version from the header and keep it to just zola. Even if it is already in the bundle, it is pretty suprising to see clap as a dependency there

components/templates/src/global_fns/load_data.rs
RealOrangeOne: There probably are some use cases out there, and seeing as it was _basically free_ to implement, thought I might as well. It's probably something which may be critical to someone out there

RealOrangeOne:
@Keats I've made all the requested changes, and a couple bits of cleanup.

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [package]
 name = "zola"
+# Update templates Cargo.toml when bumping version
```
Keats: can be removed I think

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,403 @@
...
+enum DataSource {
+    Url(Url),
+    Path(PathBuf)
+}
+
+#[derive(Debug)]
+enum OutputFormat {
+    TOML,
+    JSON,
+    CSV,
```
Keats: I wouldn't uppercase them even if they are acronyms to be consistent with Rust style (https://doc.rust-lang.org/1.0.0/style/style/naming/README.html)

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,403 @@
...
+    Plain
```
Keats, line 33: I think I would still remove that for now, it can be added later if necessary

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,403 @@
...
+}
+
+impl Hash for OutputFormat {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.to_string().hash(state);
+    }
+}
+
+impl OutputFormat {
+    fn from(output_format: String) -> Result<Self> {
```
Keats: you could impl [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) instead

components/templates/src/global_fns/load_data.rs
```diff
@@ -0,0 +1,403 @@
...
+
+        if let Some(url) = url_arg {
+            return Url::parse(&url).map(|parsed_url| DataSource::Url(parsed_url)).map_err(|e| format!("Failed to parse {} as url: {}", url, e).into());
+        }
+
+        return Err(GET_DATA_ARGUMENT_ERROR_MESSAGE.into());
+    }
+}
+
+fn get_cache_key(data_source: &DataSource, format: &OutputFormat) -> u64 {
```
Keats: Making this part of the impl of DataSource would be cleaner imo

docs/content/documentation/templates/overview.md
```diff
@@ -186,7 +187,28 @@ template:
...
+#### Remote content
+
+Instead of using a file, you can load data from a remote URL. This can be done by specifying a `url` parameter to `load_data` rather than `file`.
+
+```jinja2
+{% set response = load_data(url="https://api.github.com/repos/getzola/zola") %}
+{{ response }}
+```
+
+By default, the response body will be returned with no parsing. This can be changed by using the `format` argument as above.
```
Keats, line 201: Hmm I guess we need `plain` because of that? How about making the `format` argument mandatory?

Keats (changes requested):
A couple more comments but looking good!

components/templates/src/global_fns/load_data.rs
RealOrangeOne: I did think of that, but because it's possible for it to fail (ie the string doesnt match one of the enum values), it's unsafe, and `TryFrom` is only in nightly right now.

docs/content/documentation/templates/overview.md
RealOrangeOne: I could make it mandatory, but the fact it bases it off the file extension is a really nice touch for local data files. Having to specify that my `data.json` file is a JSON file feels unnecessary when it's so easy to guess from the file extension, and change it by default if required.

components/templates/src/global_fns/load_data.rs
RealOrangeOne: I was toying with where to put it, as it's also equally valid on `OutputFormat`,  but as most of the computation varys for `DataSource`, I guess there makes the most sense of the 2.

components/templates/src/global_fns/load_data.rs
Keats: `FromStr` returns a `Result` though? `FromStr` and `TryFrom` look to be identical for this usecase

RealOrangeOne: Aha, I found my confusion. I was looking at `From` rather than `FromStr`. Yes you're right, `FromStr` would be ideal to implement here!

docs/content/documentation/templates/overview.md
Keats: I'm not sure myself what is the best choice there. I kind of like the nice UX for ending in JSON but in practice for remote data you will need to specify the format so it feels a bit like it could surprise the end user when something works for local data but not remote.

I would tend towards removing the special casing but not 100% convinced.

RealOrangeOne: I think the fact it will throw errors at build time complaining for a `kind` argument should eliviate this. Although perhaps a very descriptive error message / docs piece about it would be better?

Keats:
The 2 failing tests need to be fixed but I can merge it after

RealOrangeOne:
I've fixed the tests, and they run fine locally, however seem to be doing odd things on Travis.

Keats:
For `global_fns::load_data::tests::fails_when_request_404s`, you can make up a random URL to avoid issues like that.
For the other, you can use a fake public api like https://jsonplaceholder.typicode.com/todos/1 to avoid 403 and various issues

RealOrangeOne:
Tests passing :tada:  (besides netlify, but that doesn't look related to this PR?)

Keats:
Netlify is not going to work for a while with Zola sadly, not related to your PR :)

Thanks!

kellpossible:
Nice one @RealOrangeOne !!

components/templates/src/global_fns/load_data.rs
ryanmcgrath: Just a heads up - I actually would make use of `plain` to load CSS and shove it inline at build-time, or to load a base64-encoded value that's ugly in code but needs to exist. They're rare use cases but they do exist.

PR #496, opened 2018-10-22
Ensure root output directory exists explicitly.

maghoff:
This PR obsoletes #492.

Aliases that have no directory nesting sneakily avoid the code path that ensures the directories exist. A way to fix this is to call ensure_directory_exists explicitly. This is implemented in this pull request.

Test data has been updated so if the implementation change is not applied, tests will fail.

However, tests currently fail to build with the following unrelated error message:

```
error[E0061]: this function takes 1 parameter but 0 parameters were supplied
   --> components/site/tests/site.rs:205:10
    |
205 |     site.enable_live_reload();
    |          ^^^^^^^^^^^^^^^^^^ expected 1 parameter
```

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -565,6 +565,7 @@ impl Site {
     }
 
     pub fn render_aliases(&self) -> Result<()> {
+        ensure_directory_exists(&self.output_path)?;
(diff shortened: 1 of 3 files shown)


Keats:
Thanks! I'll fix the test quickly on my side

PR #506, opened 2018-10-31
Update reqwest to 0.9

cljoly:
According to the [changelog][chg] of reqwest, Header types are removed only
temporally. Nevertheless, in the meantime, we need to adapt.

I committed `Cargo.lock`, don’t know if I did this right, as I’m fairly new to rust.

[chg]: https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#breaking-changes


Keats:
Thanks for that. However the dev branch is `next` and already contains that change

cljoly:
Oops, my fault, should have checked.

PR #517, opened 2018-11-12
Add info about chocolatey package

gsomix:
Chocolatey is popular package manager for Windows as well as `scoop`.

https://chocolatey.org/packages/gutenberg/


Keats:
Thanks!
Do keep in mind the next version of `gutenberg` will be called `zola` so we will need to update chocolatey as well!

gsomix:
@Keats OK, thank you!

PR #523, opened 2018-11-16
Load data from `data` directory

RealOrangeOne:
Load data from a `data` directory rather than `content/`. This means the content directory is actually only used for content. 

It's still possible to load data from the content directory, by prefixing the `path` argument with `../content/`.


Keats:
cc @kellpossible

We did talk about where to put files in https://github.com/Keats/gutenberg/pull/379#issuecomment-416522879

I want to release 0.5 tomorrow or sunday so there isn't much time to decide :)

RealOrangeOne:
Having data files related to content is a usage i'd not considered, but would actually be incredibly useful. I'm not sure of the right implementation for it though. Searching relative to multiple locations isn't very explicit, which I don't like. 

oooh, here's a thought. What if we allow `load_data` to be given a `page` argument, which switches the load location from relative to the data directory, to searching the page assets only? (Obviously this would be ignored given a `url` argument?

Thoughts?

Keats:
What about doing the same approach as Hugo?
Any file in a `data` directory is automatically loaded so you only need to call `load_data` if you want to load a file outside of it or a URL.

RealOrangeOne:
That could work, although would be a much larger change than this, as it'd require a whole new global variable added to scope.

Making the data path configurable could be another way to do it?

Keats:
We can reserve the data variable  in 0.5.0 and implement it later. Or put
it on config.data instead

RealOrangeOne:
It doesnt really work on `config`, as it's not really to do with config. Theoretically once we have auto-loading setup, we can attach data relative to pages direct to the page itself!

I'll close this, and consider an alternative implementation.

Keats:
Yeah the config was just for convenience since it's already passed
everywhere. I'll mention the data variable name is reserved in the breaking
changes anyway just so it isn't a breaking change if we go that way.

kellpossible:
So what ended up happening @Keats @RealOrangeOne ?

Keats:
I'll just mark that `data` is reserved in the breaking changes and we can see later!

PR #527, opened 2018-11-19
Fix snapcraft build

caemor:
For #227 

I added the missing pkg-config for finding openssl to snapcraft.yaml and it works for me on Ubuntu 18.04 for now with the following command: `snapcraft cleanbuild --debug`


Keats:
Thanks!

PR #528, opened 2018-11-19
Updated install instructions

caemor:
I updated the install instructions (snap, scoop, chocolatey) and added the ones where Gutenberg already got renamed to Zola also to the top of the document.

Should yaourt be replaced by one of the other aur-helpers (pikaur, yay,...) as the development for it is stalled/discontinued for quite some time now although it is still working?

Scoop: https://github.com/lukesampson/scoop/blob/master/bucket/zola.json
Chocolatey: https://chocolatey.org/packages/zola
Snap: https://snapcraft.io/zola

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -6,12 +6,36 @@ weight = 1
 Zola provides pre-built binaries for MacOS, Linux and Windows on the
 [GitHub release page](https://github.com/getzola/zola/releases).
 
+## Linux
+
 ## Arch Linux
 
 Use your favourite AUR helper to install the `zola-bin` package.
 
 ```bash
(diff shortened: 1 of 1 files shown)


Keats:
> Should yaourt be replaced by one of the other aur-helpers (pikaur, yay,...) as the development for it is stalled/discontinued for quite some time now although it is still working?

I was kind of waiting for someone to open an issue to say using `yaourt` is a bad idea :p We can use `yay` instead

caemor:
> > Should yaourt be replaced by one of the other aur-helpers (pikaur, yay,...) as the development for it is stalled/discontinued for quite some time now although it is still working?
> 
> I was kind of waiting for someone to open an issue to say using `yaourt` is a bad idea :p We can use `yay` instead

Most of the time I am still using it myself too :-D

PR #541, opened 2018-11-29
Update snap installation instructions

bw-yuhogg:
If you don't install with classic confinement, you'll get permission errors trying to create sites with `zola init`.

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -21,7 +21,7 @@ $ yay -S zola-bin
 Zola is available on snapcraft:
 
 ```bash
-$ snap install --edge zola
+$ snap install --edge --classic zola
 ```
 
 ## Windows


Keats:
Thanks!
Does it mean that with the default install we cannot do any IO? Is it something we can change in the snapcraft file?

bw-yuhogg:
> Does it mean that with the default install we cannot do any IO?

Pretty much. You won't be able to write to files/folders in the user's home directory.

> Is it something we can change in the snapcraft file?

Yes.

One option is to switch from `strict` to `classic` confinement in snapcraft.yml. This will inform people installing from the command line that they have to install with classic confinement.

The other option is to try and find all the interfaces you need and add them to the `plugs` list. `home` is the most likely one here, but I'm not 100% sure if this will work.

See https://blog.ubuntu.com/2018/11/01/a-guide-to-snap-permissions-and-interfaces , https://docs.snapcraft.io/supported-interfaces/7744 , and https://blog.ubuntu.com/2017/01/09/how-to-snap-introducing-classic-confinement

bw-yuhogg:
Confirmed that adding the `home`  plug also fixes the permission errors. I'll open a separate PR for that if you'd like.

Keats:
> I'll open a separate PR for that if you'd like.

Yes please!

PR #550, opened 2018-12-07
add rendering for 404.html when running zola serve

antonok-edm:
Sanity check:

* [ ✔ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

## Code changes

* [ ✔ ] Are you doing the PR on the `next` branch?

* [ ~ ] Have you created/updated the relevant documentation page(s)?
The documentation already outlines the expected behavior.

## Notes
Modifying `templates/404.html` when running `zola serve` previously did nothing. It now re-renders the file, as suggested by the documentation.

Note that this only re-renders the previous version of `404.html` on a new save, e.g. it takes *two* saves to show a single change. I tried browsing through the code to find where that caching occurs, but I couldn't find it. I'd be happy to add on to this admittedly tiny PR if you have any clues as to where the issue might be hiding.

### DIFF
--- a/components/rebuild/src/lib.rs
+++ b/components/rebuild/src/lib.rs
@@ -352,6 +352,7 @@ pub fn after_template_change(site: &mut Site, path: &Path) -> Result<()> {
             site.render_orphan_pages()
         }
         "section.html" => site.render_sections(),
+        "404.html" => site.render_404(),
         // Either the index or some unknown template changed
         // We can't really know what this change affects so rebuild all
         // the things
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

> it takes two saves to show a single change

I believe that's because live reload is not added to the 404 template (or sitemap etc).
If you go in `components/site/lib.rs` and use `inject_livereload` in the `render_404` (check the rest of the file for examples), I think it *should* work.

antonok-edm:
Easy enough! It still requires a manual refresh of the page though, unfortunately.

Keats:
>  It still requires a manual refresh of the page though, unfortunately.

Even for CSS changes? Livereload can't reload HTML so that's normal to need to refresh after a template change.

antonok-edm:
Ah, yes that's my mistake -- it seems to work as expected then.

Keats:
Thanks!

PR #555, opened 2018-12-11
Mark required Rust version as 1.29

bw-yuhogg:
690b3f923511c1a updated the required version in Cargo.lock and in the CI config files, but not in the installation instructions.

Sanity check:


Keats:
The downside of having this information in many places.. thanks!

PR #568, opened 2018-12-30
id continue reading

toidiu:
page.content includes the `<p><a></a></p>` tag. This will allow one to hide the `p` tag via css. 

Also the `a` tag is currently empty.

**IMPORTANT: Please do not create a Pull Request adding a new feature without discussing it first.**

The place to discuss new features is the forum: <https://zola.discourse.group/>
If you want to add a new feature, please open a thread there first in the feature requests section.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [ ] Have you created/updated the relevant documentation page(s)?

PR #577, opened 2019-01-04
add id to continue reading p tag

toidiu:
`page.content` includes the `<p><a></a></p>` html which shows up as extra white space. This adds an id to the `p` tag which will allow one to hide the p tag via css.

Sanity check:

## Code changes

### DIFF
--- a/docs/content/documentation/content/page.md
+++ b/docs/content/documentation/content/page.md
@@ -104,4 +104,4 @@ available separately in the
 
 An anchor link to this position named `continue-reading` is created so you can link
 directly to it if needed for example:
-`<p  id="zola-continue-reading"><a href="{{ page.permalink }}#continue-reading">Continue Reading</a></p>`
+`<a href="{{ page.permalink }}#continue-reading">Continue Reading</a>`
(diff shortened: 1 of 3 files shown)


toidiu:
@Keats here is a fresh PR with the changes you requested in https://github.com/getzola/zola/pull/568

docs/content/documentation/content/page.md
```diff
@@ -104,4 +104,4 @@ available separately in the
 
 An anchor link to this position named `continue-reading` is created so you can link
 directly to it if needed for example:
-`<a href="{{ page.permalink }}#continue-reading">Continue Reading</a>`
+`<p  id="zola-continue-reading"><a href="{{ page.permalink }}#continue-reading">Continue Reading</a></p>`
```
Keats: I don't think you need to wrap that one

toidiu: do you mean wrapping the `a` tag with a `p` tag?

the `p` tag is the one with the id since and I wanted to expose the behavior in the docs somewhere.

Keats: This is an example of a link to the place after the continue-reading tag though, not an example of how the continue reading tag looks

toidiu: ah mis understood that piece of documentation. corrected it.

Keats:
Thanks!

PR #580, opened 2019-01-10
Add Dracula to Themes List

ryanmcgrath:
As noted in prior pull request, adds [dracula](https://draculatheme.com/) to the available syntax highlighting themes. Also added a link in the documentation to the theme.


Keats:
Looks like it still has some of the commits in master in the history?

PR #587, opened 2019-01-14
Add current_path and current_url in 404.html

### LINKED ISSUE #586: current_path and current_url not available in 404.html
# Bug Report

## Environment

Zola version: 0.5.1

## Expected Behavior
`current_url` and `current_path` should be available in 404.html.

## Current Behavior
According to this [documentation][1] `current_url` and `current_path` should be available in every template excluding RSS and sitemap.xml. Which is not true for 404.html

## Step to reproduce
Use `{{ current_url }}` in a 404.html template

[1]: https://www.getzola.org/documentation/templates/overview/

shalzz:
Only config variable was available in the context
for 404.html templates

This uses `base_url/404.html` as the url for all 404's. This is similar to what's done by Jekyll for it's 404 page.

fixes #586

Sanity check:

## Code changes

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -691,8 +691,12 @@ impl Site {
     /// Renders 404.html
     pub fn render_404(&self) -> Result<()> {
         ensure_directory_exists(&self.output_path)?;
+        let current_path = "404.html";
+        let current_url = self.config.base_url.clone() + "/" + current_path;
         let mut context = Context::new();
         context.insert("config", &self.config);
+        context.insert("current_path", &current_path);
+        context.insert("current_url", &current_url);
         let output = render_template("404.html", &self.tera, &context, &self.config.theme)?;
         create_file(&self.output_path.join("404.html"), &self.inject_livereload(output))
     }


Keats:
The main reason it is not set is because people might expect current_url/current_path to be the URL that 404'd. I don't think this fix is that useful as the 404 template is unique and if someone wants those value, it is easy to do it from the template.

shalzz:
The 404 template is unique but most of the time and in many sites you'd want to show the site header/headerbar and footer so the user can easily navigate to the homepage etc.
For me that's defined in index.html and therefore I extend that in 404.html as well to have the same look and feel of the whole website and rather than just a generic error page.

So for me this in-congruency breaks the build since I have {{ current_url }} set in the index.html as a canonical link tag. Sure I could workaround that by if's but that'd just be a workaround and another thing that I'd have to be conscious about. (Also having a 404 page URL hardcoded anywhere in your site would actually hurt your SEO.)

Since Jekyll is the most used SSG and it follows this pattern I'd say that most people would be surprised otherwise.

Keats:
But you don't want a canonical link tag to a 404 do you?
If `current_path`/`current_url` are added, they should refer to the current URL that is 404ing, not the 404 template

shalzz:
> But you don't want a canonical link tag to a 404 do you?

Yes to the 404 template path. I'm just introducing the same behaviour as Jekyll and not trying to reinvent the wheel.

But yes the point is to atleast have these variables in 404.html

JadedBlueEyes:
I agree with @Keats here. Zola isn't meant to be a copy of Jekyll, and you shouldn't have canonical links to 404 templates. 404 is a special case, and could have any URL. Best to treat as if it had no URL.

Keats:
I don't think the behaviour is going to change, however the docs needs to be update to mention the 404 page doesn't get those variables

PR #590, opened 2019-01-16
Make the theme template files render if they are present

### LINKED ISSUE #571: 404 page provided by the theme is not used
# Bug Report

## Environment

Zola version: zola 0.5.1

## Expected Behavior
I wanted to add the same 404 error page to the hyde theme that the original jekyll theme has. After creating the correct template in the hyde theme (it works for the theme itself, see https://github.com/Areiser/hyde) I used this theme for my blog.

I would expect the 404 page from the theme to be used unless I create my own one in the local templates folder

## Current Behavior
The default 404 page is used

## Step to reproduce
Create a zola page that uses the hyde theme from https://github.com/Areiser/hyde and call a broken link. Compare the page to a broken link when directly running https://github.com/Areiser/hyde

n-pochet:
Make sure that a `404.html` template in a theme can be rendered.

Fixes #571 

## Sanity check:

## Code changes


Keats:
How do you override a theme template if this PR gets in?

n-pochet:
In that case, you cannot.
What would then be the proper solution to have the `404.html` in the theme rendered instead of the default one?

n-pochet:
@Keats I modified the commit to follow this logic to render the templates:

1. Check the templates (with the bare template name)
2. Check if the template is part of a theme
3. Fallback to the default templates (in `ZOLA_TERA` and with the last `match`)

That made me also modify `ZOLA_TERA` static variable to prepend `default/` in front of the `404.html` template name. Would it be useful to also do that for the other elements in `ZOLA_TERA`?

Keats:
Can you add a test for the original issue? You probably need to edit something in the `test_site` directory and in components/site/tests/site.rs

Trying it myself it looks like a more generalised bug than just 404 so doing the same thing for the other builtin templates would be a good idea. The shortcode rendering in `rendering/src/shortcode.rs` needs to be updated to call that function as well so the defaults can be overwritten by a theme. 

The only change I would make is `default/` -> `__zola_builins/` to avoid any potential clashes

n-pochet:
I made the changes and rebased.
After the rebase, one test is failing: `can_split_lines_shortcode_body`.
At first sight, I didn't introduce any change that could cause that failing test.

Keats:
Don't worry the failing test is something I added and need to fix

components/site/tests/site.rs
```diff
@@ -629,3 +629,13 @@ fn can_apply_page_templates() {
     assert_eq!(child.meta.template, Some("page_template_child.html".into()));
     assert_eq!(child.meta.title, Some("Local section override".into()));
 }
+
+// https://github.com/getzola/zola/issues/571
+#[test]
+fn can_build_site_with_404_from_theme() {
```
Keats: can you remove the 404 from the title? Something like `can_build_site_custom_builtins_templates_from_theme`

components/rendering/src/shortcode.rs
```diff
@@ -111,12 +111,22 @@ fn render_shortcode(
...
+        if context.tera.templates.contains_key(&tpl_name) {
+            context
+                .tera
+                .render(&tpl_name, &tera_context)
+                .map_err(|e| Error::chain(format!("Failed to render {} shortcode", name), e))?
+        } else {
+            context
+                .tera
+                .render(&default_tpl_name, &tera_context)
+                .map_err(|e| Error::chain(format!("Failed to render {} shortcode", name), e))?
```
Keats: is it possible to call the render_template in utils instead of duplicating the logic here?

Keats (changes requested):

components/site/tests/site.rs
n-pochet: Function changed with the appropriate name.

components/rendering/src/shortcode.rs
n-pochet: Done.

Keats:
Thanks!

PR #591, opened 2019-01-16
Replace slugify to percent-encoding

### LINKED ISSUE #570: Use percent encoding for header id
# Bug Report

## Environment

Zola version: 0.5.1

## Expected Behavior

https://pbzweihander.github.io/arch-linux-installation

I think the id of header `설치 전 할 일` has to be `설치-전-할-일` or `%EC%84%A4%EC%B9%98-%EC%A0%84-%ED%95%A0-%EC%9D%BC`.

## Current Behavior

Zola uses [slug](https://docs.rs/slug/0.1.4/slug/) for making id, so zola makes it as `sseoci-jjeo-hal-il` (as you can see in https://pbzweihander.github.io/arch-linux-installation)

How about using [percent-encoding](https://docs.rs/percent-encoding/1.0.1/percent_encoding/) instead of slug when making id for headers?

pbzweihander:
Resolves #570.

`slug` crate makes Unicode characters very awkward, but `percent-encode` crate gives us very reasonable way to slugify characters.


pbzweihander:
I think the percent encoding is not best option. But I really need to replace that `slug` to use `zola` in my language.

PR #602, opened 2019-01-29
Replace trim_{left, right} with trim_{start, end}

peng1999:
trim_{start, end} is introduced in rust 1.30.0 and trim_{left, right} is deprecated since 1.33.0.

---

Sanity check:

Code changes


Keats:
Thanks!

PR #609, opened 2019-02-04
Handle csv parsing error when encountering rows with different lengths

mziter:
Current code is calling unwrap on each row that results from csv crate parsing logic. This has been changed to match and map an error result into the expected Teraform error with chaining. Associated unit test has been added.

### DIFF
--- a/components/templates/src/global_fns/load_data.rs
+++ b/components/templates/src/global_fns/load_data.rs
@@ -183,7 +183,7 @@ impl LoadData {
     pub fn new(content_path: PathBuf, base_path: PathBuf) -> Self {
         let client = Arc::new(Mutex::new(Client::builder().build().expect("reqwest client build")));
         let result_cache = Arc::new(Mutex::new(HashMap::new()));
-        Self {content_path, base_path, client, result_cache}
+        Self { content_path, base_path, client, result_cache }
     }
 }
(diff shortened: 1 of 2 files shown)


kellpossible:
looks good!

components/templates/src/global_fns/load_data.rs
```diff
@@ -291,7 +291,16 @@ fn load_csv(csv_data: String) -> Result<Value> {
...
-            let record = result.unwrap();
+            let record = match result {
+                Ok(r) => r,
+                Err(e) => {
+                    return Err(tera::Error::chain(
+                        String::from("Error encountered when parsing csv records"),
+                        e,
+                    ));
+                }
+            };
```
Keats, line 302: you want to use zola errors, not tera ;)

I believe you can replace this line by 

```rust
let record = result..map_err(|e| Error::chain("Error encountered when parsing csv records", e))?;
```

mziter: Sure, sorry about the oversight. I'll take a look soon and update PR.

Keats: Sorry actually, I just realised this is a Tera function so using the Tera error makes sense/is required!

PR #610, opened 2019-02-05
Revert "Update snap installation instructions"

leoschwarz:
This reverts commit 5fd7bf7e61c87320b1c5a9f358df2d82525ea465.

Apparently it is not just no longer necessary to use classic
confinement, but actually impossible, i.e. snap emits an error that the
zola snap is not compatible with `--classic`.

Snap version: 2.37.1 (Debian 9)

Sanity check:


Keats:
Thanks!

Issue #600, opened 2019-01-25
Fix Tera globbing on Windows in next branch


Keats:
https://ci.appveyor.com/project/Keats/zola/builds/21890445/job/jmjhuj8qvn1j3yva

Most likely a / vs \ case but I don't have a Windows machine for quick testing.

If anyone can help, that would be great!

Keats:
So the actual issue is that the globbing of templates when loading the site returns nothing even after fixing it to have the proper glob pattern.: all the files are seen but ignored.
Not sure whether this is a `globwalk` bug or not, need to dig a bit deeper.

Keats:
That was a bug in globwalk: https://github.com/Gilnaa/globwalk/pull/22

Keats:
Fixed in the next branch

PR #614, opened 2019-02-16
Powershell syntax highlighting

### LINKED ISSUE #613: Add support for PowerShell syntax highlighting
It would be nice to have support for PowerShell syntax highlighting.

https://github.com/PowerShell/EditorSyntax should make it easy.

lzybkr:
Add support for PowerShell syntax highlighting.

Fix #613 

Sanity check:

## Code changes


.gitmodules
```diff
@@ -37,3 +37,6 @@
 [submodule "sublime_syntaxes/Swift-for-f-ing-sublime"]
 	path = sublime_syntaxes/Swift-for-f-ing-sublime
 	url = https://github.com/colinta/Swift-for-f-ing-sublime.git
+[submodule "sublime_syntaxes/PowerShell-EditorSyntax"]
```
Keats, line 40: I think we don't need the submodules since you have converted the `tmLanguage` into `sublime-syntax` and we do end up not using anything from this repo?

lzybkr: Maybe not necessary, but I thought I was following examples, e.g. the submodule `Sublime-VimL` only has a `tmLanguage` file.

lzybkr:
I took a quick peek at the CI failures - it looks like in Travis, the rendering for some `python` changed from the expected rendering, maybe the `Packages` submodule was updated but `newlines.packdump` wasn't updated?

The appveyor failures didn't look like they were related to syntax, so probably not related to my change?

.gitmodules
Keats: Ah missed that one, I will need to check which submodules are needed

Keats:
CI failures are not related to your changes: https://github.com/getzola/zola/issues/600 which is caused by https://github.com/getzola/zola/blob/next/components/rendering/src/shortcode.rs#L117 I am guessing?

I need to find a Windows machine/start a VM to test it, I tried a bit earlier today but waiting 30min for each build on appveyor is not fun :(

Keats:
Looks like the travis failure however are related. It might be that the python syntax highlighting has been updated, I'll have a look

Keats:
Not sure what happened with the test there but I pushed the PowerShell syntax to the `next` branch myself and it seems to work?

Thanks!

Issue #562, opened 2018-12-20
Be able to split sitemaps


corentin-cres:
In case of large site generation (100k pages), sitemaps get huge and robots (for example, Google crawler) is not able to parse it all.

From the [Google guidelines](https://support.google.com/webmasters/answer/183668?hl=en),

> All formats limit a single sitemap to 50MB (uncompressed) and 50,000 URLs. If you have a larger file or more URLs, you will have to break your list into multiple sitemaps. You can optionally create a sitemap index file (a file that points to a list of sitemaps) and submit that single index file to Google.

It would be great that Zola limits the maximum number of URLs in a single sitemap (for example, defaulted to 20k) and splits the sitemap in multiple sitemaps if we reach this value ; and maybe be able to configure this value somewhere.

-----------

Another solution would be to give the `pages`, `sections`, `tags`, `categories` (https://www.getzola.org/documentation/templates/sitemap/) to any page named `sitemap*.xml` (with the wildcard, and not only `sitemap.xml`), so that we can implement some logic ourselves to be able to count the number of URLs displayed in each sitemap.
This solution has the advantage to allow custom logic, and not only the maximum number of URLs, to be implemented. However, this logic has to be written for every project, and is not implemented by default.

Keats:
I didn't know this was a thing :o

I'm wondering if there is some overlap with https://github.com/getzola/zola/issues/365 and your second solution.
The logic to implement that in Zola automatically would be fairly simple I think and has the advantage of doing the right thing automatically.

corentin-cres:
Thanks for your reply !

Agreed, I think the first solution is better too.
Should I create "New issue" from this post or something ?

Keats:
Do you mean to create a pull request? Fork the repository, select the `next` branch and start adding some code

JadedBlueEyes:
Can I just add, whilst we’re here, that the sitemap should be automatically added to (at least the default) robots.txt file.
https://www.sitemaps.org/protocol.html#submit_robots

Keats:
Ah, didn't know that. It makes the sitemap splitting a bit more complex then

JadedBlueEyes:
No, just the root one that links to the others. Just add `Sitemap: {{ get_url(path="/sitemap.xml") }}` to the start of the default robots.txt template, or something more robust probably.

paulcmal:
> I'm wondering if there is some overlap with #365 and your second solution.

[...]

PR #619, opened 2019-02-20
Split sitemap

sebastien-mariaux:
As discussed on issue [#562](https://github.com/getzola/zola/issues/562), this addition opens the possibility to split the sitemap when the number of pages exceed Google limitation for SEO. 

In that case a main sitemap is created with links to the sub-sitemaps.

The max number of urls in a given sitemap should be set in the .toml config file (relevant section is added in the documentation).


components/config/src/config.rs
```diff
@@ -134,6 +134,10 @@ pub struct Config {
 
     /// Set automatically when instantiating the config. Used for cachebusting
     pub build_timestamp: Option<i64>,
+
+    /// Maximum number of urls in the sitemap.xml file. 
+    /// If number of pages exceed that limit, several sitemaps site are created, referenced in a main sitemap.xml
+    pub sitemap_limit: usize,
```
Keats: Does that need to be user defined or can we just always break on ~30k?

components/rebuild/src/lib.rs
```diff
@@ -366,6 +366,8 @@ pub fn after_template_change(site: &mut Site, path: &Path) -> Result<()> {
     match filename {
         "sitemap.xml" => site.render_sitemap(),
         "rss.xml" => site.render_rss_feed(site.library.read().unwrap().pages_values(), None),
+        "multi_sitemap.xml" => site.render_sitemap(),
+        "main_sitemap.xml" => site.render_sitemap(),
```
Keats: maybe change the names to be more explicit,  `split_sitemap.xml` and `split_sitemap_index.xml`.

components/site/src/lib.rs
```diff
@@ -864,9 +864,44 @@ impl Site {
...
+                all_sitemap_entries.append(&mut terms);
+            }
+            
+            // Slice the vector in chunks and create sitemap file for each
+            let mut xml_files = Vec::new();
+            for (i, chunk) in all_sitemap_entries.chunks(self.config.sitemap_limit).enumerate() {
+                let mut chunk_context = context.clone();
+                chunk_context.insert("chunk", &chunk);
+                let sitemap = &render_template("multi_sitemap.xml", &self.tera, chunk_context, &self.config.theme)?;
+                let file_name:String = format!("sitemap{}.xml", i+1);
```
Keats: You don't need those types, Rust will infer them

components/site/src/lib.rs
```diff
@@ -864,9 +864,44 @@ impl Site {
...
+            all_sitemap_entries.append(&mut sections);
+            for terms in taxonomies {
+                let mut terms = terms;
+                all_sitemap_entries.append(&mut terms);
+            }
+            
+            // Slice the vector in chunks and create sitemap file for each
+            let mut xml_files = Vec::new();
+            for (i, chunk) in all_sitemap_entries.chunks(self.config.sitemap_limit).enumerate() {
+                let mut chunk_context = context.clone();
```
Keats: that is VERY expensive right there

components/site/src/lib.rs
```diff
@@ -864,9 +864,44 @@ impl Site {
...
+                chunk_context.insert("chunk", &chunk);
+                let sitemap = &render_template("multi_sitemap.xml", &self.tera, chunk_context, &self.config.theme)?;
+                let file_name:String = format!("sitemap{}.xml", i+1);
+                create_file(&self.output_path.join(&file_name), sitemap)?;
+                let mut xml_link:String = self.config.make_permalink(&file_name);
+                xml_link.pop(); // Remove trailing slash
```
Keats: why?

components/site/src/lib.rs
Keats: Instead of doing that, can you delay inserting into the context?
In short, do an early return of the fn if the number of entries is lower than the `sitemap_limit` where you can insert into the context.

Then you can do a big vec of sitemap entries and chunk that, only creating a context when you need to render one.

I'm also thinking whether we can do a breaking change and have only one entry in context `entries` that would simplify things. since there's no need to differentiate between pages/sections/taxonomies there afaik. But that's something I would do afterwards, don't worry about it.

components/site/src/lib.rs
```diff
@@ -864,9 +864,44 @@ impl Site {
...
+                xml_files.push(xml_link);
+            }
 
-        create_file(&self.output_path.join("sitemap.xml"), sitemap)?;
+            // Create main sitemap
+            let mut main_context = context.clone();
+            main_context.insert("xml_files", &xml_files);
```
Keats: Not sure about `xml_files`, how about `sitemaps`?

Keats (changes requested):

components/config/src/config.rs
sebastien-mariaux: I think it could be hard-coded if you are looking to keep it simple. 
The all point is to comply with Google requirements (50000 urls and max 50Mb), so 30K seems safe enough (With unrealistically long urls I get to 35Mb).

components/rebuild/src/lib.rs
sebastien-mariaux: Yes,I will do that.

components/site/src/lib.rs
sebastien-mariaux: xml_link is a file url, and the trailing slash makes it looks like a folder, therefore Google can't read (and index) it.
eg: https://sitemap1.xml/

There might be a more elegant and safe way to do it though..

components/site/src/lib.rs
sebastien-mariaux: there is a "sitemap" variable just above so it might be confusing: 
`let sitemap = &render_template("split_sitemap.xml", &self.tera, context, &self.config.theme)?;`
Maybe "sub_sitemap".

components/site/src/lib.rs
sebastien-mariaux: I gave it a try, did I understand you right?
thanks for your comments.

components/config/src/config.rs
Keats: It seems unlikely people would want to actually configure that so I would just hardcode it.

components/site/src/lib.rs
```diff
@@ -867,12 +863,48 @@ impl Site {
...
+        let sitemap_limit = 30000;
+
+        if total_number < sitemap_limit {
+            // Create one sitemap with all pages, sections and taxonomies
+            let mut context = Context::new();
+            context.insert("pages", &pages);
+            context.insert("sections", &sections);
+            context.insert("taxonomies", &taxonomies);
+            let sitemap = &render_template("sitemap.xml", &self.tera, context, &self.config.theme)?;
+            create_file(&self.output_path.join("sitemap.xml"), sitemap)?;
```
Keats, line 884: you can return that directly and skip the indentation in `else`

components/site/src/lib.rs
```diff
@@ -867,12 +863,48 @@ impl Site {
...
+        // Count total number of urls to include in sitemap
+        let total_number = pages.len() + sections.len() + taxonomies.len();
+        let sitemap_limit = 30000;
+
+        if total_number < sitemap_limit {
+            // Create one sitemap with all pages, sections and taxonomies
+            let mut context = Context::new();
+            context.insert("pages", &pages);
+            context.insert("sections", &sections);
+            context.insert("taxonomies", &taxonomies);
```
Keats: you need to have the same context variables here than when rendering a chunked sitemaps, otherwise we cannot use the same template.

components/site/src/lib.rs
```diff
@@ -867,12 +863,48 @@ impl Site {
...
         }
 
-        context.insert("taxonomies", &taxonomies);
-        context.insert("config", &self.config);
-
-        let sitemap = &render_template("sitemap.xml", &self.tera, context, &self.config.theme)?;
-
-        create_file(&self.output_path.join("sitemap.xml"), sitemap)?;
+        // Count total number of urls to include in sitemap
+        let total_number = pages.len() + sections.len() + taxonomies.len();
```
Keats: I would just concatenate those variables basically

Keats (changes requested):

Keats:
Thanks!

PR #645, opened 2019-03-28
use travis "minimal" vm in the documentation

pandark:
If not specified, Travis use a system with ruby, which we don't need. We can use the "minimal" one instead.


Keats:
Thanks!

Issue #626, opened 2019-02-27
a paragraph is created inside each markdown list element


pandark:
# Bug Report

## Environment

Zola version: 0.5.1

## Expected Behavior

```
- 1
- 2
  - 2.1
```

should render

```
<ul>
<li>1</li>
<li>2
<ul>
<li>2.1</li>
</ul>
</ul>
```

## Current Behavior

Instead, it renders

```
<ul>
<li>
<p>1</p>
</li>
<li>
<p>2</p>
<ul>
<li>
<p>2</p>
</li>
</ul>
</ul>
```

## Step to reproduce

Make a markdown list in a page/section.

Keats:
We don't do anything fancy with lists when rendering so that's likely coming from `pulldown-cmark`. They are working on a new version so hopefully it will be working fine when it is released.

Keats:
I just tried with the next version of pulldown-cmark and that works as expected!

pandark:
I'm not sure what triggers this, but it's not actually solved…

If I have only one list, it's fine, but if I have two lists, or if I use a shortcode in one of the elements, the `<p>`s are back :confused: 

@Keats Should I open a new issue or keep using this one?

Keats:
If you can create a new issue with some samples that would be great

pandark:
Ok, so it was actually two different things : #647 (shortcode adding a newline) and #648 (consecutive lists rendering a single messed-up one). :smiley:

PR #660, opened 2019-04-10
strip wrapping whitespace from newline outward from shortcodes

### LINKED ISSUE #647: The ending newline in shortcode files isn't ignored
Most text editors automatically add a newline at the end of each file (and don't show it).
In shortcodes, it is treated as a regular newline (usually displayed as an additional space in the resulting HTML but can be more problematic in a list or in preformatted text, for example).
Maybe #462 was fixed too well? :stuck_out_tongue: 

It is possible to circumvent the problem by specifically removing the ending newline in shortcode files, but that newline should probably be ignored by zola to make things simpler.

bdjnk:
This resolves #647.
It also resolves the bug aspect of #626, which occurs when a list item is the output of a shortcode.

### DIFF
--- a/components/rendering/src/shortcode.rs
+++ b/components/rendering/src/shortcode.rs
@@ -16,6 +16,7 @@ pub struct ContentParser;
 
 lazy_static! {
     static ref MULTIPLE_NEWLINE_RE: Regex = Regex::new(r"\n\s*\n").unwrap();
+    static ref OUTER_NEWLINE_RE: Regex = Regex::new(r"^\s*\n|\n\s*$").unwrap();
 }
 
 fn replace_string_markers(input: &str) -> String {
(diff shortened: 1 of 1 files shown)


components/rendering/src/shortcode.rs
```diff
@@ -411,4 +414,20 @@ Some body {{ hello() }}{%/* end */%}"#,
         let res = render_shortcodes("Body\n {% youtube() %}\nHello \n World{% end %}", &tera);
         assert_eq!(res, "Body\n Hello \n World");
     }
+
+    #[test]
+    fn outer_newlines_removed_from_shortcodes_with_body() {
```
Keats, line 419: the titles from those tests doesn't really seem to match what they are doing no? They strip outer newlines whitespaces, not the newlines themselves

bdjnk: I couldn't think of a concise way to describe the behavior as a test name.

It removes all outer newlines (and the whitespace between them), but the parenthetical was, to my mind, just a necessary byproduct.

If you have better test names in mind I'll gladly change them.

Keats:
Thanks!

Issue #539, opened 2018-11-26
`check_external_links` should check anchors


ralfbiedert:
When running with `check_external_links`, it should also check whether the given anchor targets are present. That is, when I link 

`https://test.com/page#link`

Zola should check that the page 

- `https://test.com/page` exists and 
- has an  `<a id="link">` tag

Keats:
That's a bit trickier as we are rendering markdown concurrently and anchors are generated while rendering. It would have to be done in a post-rendering phase somehow

Keats:
I have it locally but it relies on Zola being aware of all the anchors obviously and it won't parse HTML links for anchors (at least for now). You are defining them like so:

```md
### <a name="data_structures"></a> Data Structures
```

which would have to change to 

```md
### Data Structures {#data_stuctures}
```

since named anchors is also coming for Zola 0.8. What do you think?

ralfbiedert:
For internal links this is perfect, I didn't know of the `{#data_stuctures}` syntax, but it's nicer than injecting `<a name>`.

Long term, having the ability to check for external anchors (e.g., from shortcode-templated HTML to external HTML) would still be nice.

Keats:
> For internal links this is perfect, I didn't know of the {#data_stuctures} syntax, but it's nicer than injecting <a name>.

It's still only in the `next` branch, that's why :p 
The only question left is whether this check should be by default or only in`zola check`...
Since we can't catch every anchors, it could fail valid sites since there's no way to turn it off. I do wonder if having links to anchors generated in shortcodes or raw HTML in the markdown is common enough though

ralfbiedert:
How would default checking work with `serve` during development?

I'd be fine having `build` only do build stuff and focus on speed, with an option `check_on_build` that can be enabled. Then `check` can do more expensive checks by default (again, probably with TOML options for better control), which can be run before deployment.

Keats:
Anchor checking is done after the markdown for all pages is rendered (since we need that info to know their ToC). I don't think it would cause a noticeable slowdown like the external links, it's pretty much just creating one path from a string + some hashmaps lookup. Nothing like a HTTP request.
At the same time, since Zola does check internal links by default, I think it makes sense to check the anchors by default as well in build/serve...

I don't know if you've seen it but there is a new command in the upcoming version `zola check` that does all the verification without actually writing anything to disk. This is the new way to check for external links validity, rather than an option in the config.

ralfbiedert:
[...]

PR #663, opened 2019-04-15
Add check subcommand

marcusklaas:
This addresses issue https://github.com/getzola/zola/issues/381.

Sanity check:

## Code changes

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -4,6 +4,7 @@ use syntect::easy::HighlightLines;
 use syntect::html::{
     start_highlighted_html_snippet, styled_line_to_highlighted_html, IncludeBackground,
 };
+use rayon::prelude::*;
 
 use config::highlighting::{get_highlighter, SYNTAX_SET, THEME_SET};
 use context::RenderContext;
@@ -65,7 +66,7 @@ fn is_colocated_asset_link(link: &str) -> bool {
         && !link.starts_with("mailto:")
 }
 
-fn fix_link(link_type: LinkType, link: &str, context: &RenderContext) -> Result<String> {
+fn fix_link(link_type: LinkType, link: &str, context: &RenderContext, external_links: &mut Vec<String>) -> Result<String> {
     if link_type == LinkType::Email {
         return Ok(link.to_string());
     }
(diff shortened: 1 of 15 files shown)


marcusklaas:
Ok, I believe this is ready for a review now. The PR now also adds concurrent link checking.

I haven't been able to test the impact on link checking time since my machine doesn't have enough memory to compile the project in release mode (:crying_cat_face:)

marcusklaas:
I found a machine with enough memory :sweat_smile: 

Checking my personal blog went from 7.3s -> 4.8s. It could probably be accelerated a bit more by spawning a thread for every request I think. The current solution is limited by the number of CPU threads. But it's a start.

components/rendering/src/markdown.rs
```diff
@@ -283,3 +283,25 @@ pub fn markdown_to_html(content: &str, context: &RenderContext) -> Result<Render
         })
     }
 }
+
+fn check_external_links(links: &[String]) -> Result<()> {
+    let mut err_msg = links.par_iter().filter_map(|link| {
```
Keats, line 288: does the parallelism really help here? It is already rendering markdown in parallel so I'm not sure this has a big impact

marcusklaas: It appears so. If it encounters multiple links on a page, it otherwise goes through them one-by-one, most of time waiting for IO. The speedup comes from waiting on multiple requests to return at the same time.

marcusklaas: (for clarity, parallelism was previously only a page level, this adds parallelism to a link level)

Keats:
I'll try it locally this week and merge it if everything is ok!
A small tweak to make it faster as well would be to disable syntax highlighting in `check` since it's going to take a significant amount of time.

marcusklaas:
That's a good insight. I'll try to make it so :-)

Keats:
I have just tested it quickly but it doesn't quite solve the main usecase of `check` imo: the build will still fail at the first page that has a dead link, although now it will show all the links that failed in that page.

`zola check` should wait for all links of all pages to be checked and output a result for each failure rather than the first one. Same if we want to add https://github.com/getzola/zola/issues/539 but this one can be added later on.

Keats:
@marcusklaas are you planning to continue that PR?

marcusklaas:
Oh, yes! I had completely forgotten about this PR. I will work on this tomorrow!

marcusklaas:
Superseded by https://github.com/getzola/zola/pull/683.

Issue #632, opened 2019-03-12
Is there a way to get image size?


stusmall:
It would be useful if I had a method I could use to get the dimensions of images.   A gallery library I'm trying to use requires that I provide the dimensions of each image.

Keats:
There is no way to do that currently.
I assume you mean a JS gallery library? I've never seen on requiring the dimensions, do you have a link to it?

piedoom:
Generally I've found it useful when lazyloading images in a specific frame - e.g. have a thumbnail image click into a super hi res version that takes a bit to load in a modal. The nice part would be knowing how big to make the modal before the image is actually loaded in. I was fairly sure lightbox did something like this, but not too sure.  I can see the use case though for just general usage. 

Another (slightly more niche) use I can think of is to display fullres image dimensions similar to how wikipedia has them labeled along with their data size.

stusmall:
I'm using photswipe.  Here is their explanation of why they require the size: http://photoswipe.com/documentation/faq.html#image-size

stusmall:
I started on a PR for this.  I might have some questions for you once I get a little further along.

stusmall:
@Keats I made a first pass at implementing this.  I need to update the docs but I wanted to check in that this looks good before I update docs and get a PR together.

Keats:
Thanks! I'll have a look a bit later this week

Keats:
I think I would rather have a single Tera function to get image metadata (name to be bikeshedded), which returns a struct/hashmap of things like `{height,width, etc...}` so we can easily extend it later if we want to read EXIF data or other image metaadata for example

stusmall:
Smart.  I can make that happen.  I'll start on those changes

stusmall:
@Keats I've updated the branch with the feedback.  Let me know if that looks any better.  If it does I can prep docs for it.

Keats:
Sorry, I forgot about that. It will be easier to review as a PR but that seems ok

Issue #669, opened 2019-04-22
resize_image does not generate images when called from a template


callahad:
Zola version: 0.6.0

Steps to Reproduce:

1. Use `resize_image` inside a page or section template.

Observations:

1. No images are generated or saved to disk.
2. Filename hashes *are* calculated correctly
3. The same exact function invoked from a Markdown shortcode works as expected.

Minimal test site at https://github.com/callahad/zola-testcases/tree/master/resize-image-in-template

Keats:
Anyone interested in looking at it before I do?

Thanks for the reproduction!

Keats:
This should be fixed now in the `next` branch

tforgione:
I had the problem previously, I tried the next branch on my website and I can confirm it works for me.

Thanks!

**EDIT** Actually it's quite weird. The first time I run `zola build`, the images are not generated, but the second time I run it, they are. This is what I get from running @callahad's example

``` sh
$  zola build # First build
Building site...
-> Creating 0 pages (0 orphan), 0 sections, and processing 1 images
Done in 34ms.
$ ls public # No processed images
404.html  index.html  robots.txt  sailboat.png  sitemap.xml
$ zola build # Second build
Building site...
-> Creating 0 pages (0 orphan), 0 sections, and processing 1 images
Done in 26ms.
$ ls public # Processed images are there
processed_images  404.html  index.html  robots.txt  sailboat.png  sitemap.xml
```

Keats:
@tforgione I cannot reproduce the issue in the edit? I always get `processed_images` after `rm -rf public/ && zola build`

tforgione:
Hum... I cannot reproduce it either..., maybe something weird about my config... or maybe I made some weird mistake. Sorry for the inconveniance!

tforgione:
Ok, so after more attempts, I think I get what is happening.

If I clone a zola repository, let's say like this
```
git clone --recurse-submodules https://github.com/tforgione/zola-academic-example
```

and if I run `zola build` once, the processed images appear in `static/processed_images`. If I run `zola build` a second time, they also appear `public/processed_images`.

My hypothesis is that the processed images are generated into the `static/processed_images` directory, but after the `static` directory has been copied into the `public` directory.

If my hypothesis is correct, you should be able to reproduce the bug by removing `static/processed_images` and `public`, and then re-running `zola build`.

I hope this helps.

Thank you!

Keats:
[...]

PR #670, opened 2019-04-23
Add get_image_metadata

stusmall:
This PR is for #632


components/templates/src/global_fns/mod.rs
```diff
@@ -140,6 +142,39 @@ impl TeraFn for ResizeImage {
...
+            String,
+            args.get("path"),
+            "`get_image_meta` requires a `path` argument with a string value"
+        );
+        let src_path = self.content_path.join(&path);
+        if !src_path.exists(){
+            return Err(format!("`get_image_meta`: Cannot find path: {}", path).into());
+        }
+        let img = image::open(&src_path)
+            .map_err(|e| Error::chain(format!("Failed to process image: {}", path), e))?;
```
Keats, line 168: Failed to process image -> Failed to open image maybe? This doesn't do any processing

components/templates/src/global_fns/mod.rs
```diff
@@ -140,6 +142,39 @@ impl TeraFn for ResizeImage {
...
+        let mut map = tera::Map::new();
+        map.insert(String::from("height"), Value::Number(tera::Number::from(img.height())));
```
Keats, line 170: I think you can do `to_value(img.height()).unwrap()`, same for the width

docs/content/documentation/content/image-processing/index.md
```diff
@@ -145,3 +145,17 @@ Here is the result:
 <small>
   Image attribution: Public domain, except: _06-example.jpg_: Willi Heidelbach, _07-example.jpg_: Daniel Ullrich.
 </small>
+
+
+# Get image metadata
```
Keats: The Tera fns are explained in ` docs/content/documentation/templates/overview.md` currently so this doc should be there. A link to it can still be added in this image processing doc.

docs/content/documentation/content/image-processing/index.md
```diff
@@ -145,3 +145,17 @@ Here is the result:
...
+
+Another useful included function is `get_image_metadata`.  This allows you to access various bits of metadata about the file.
+Today the only supported keys are `width` and `height`.
+
+
+To get the image dimensions, all you need to do is:
+
+```jinja2
+  {% set meta = get_image_metadata(path="...") %}
+  Our image is {{ meta['width'] }}x{{ meta['height'] }}
```
Keats: More common usage would be `meta.width` and `meta.height`

Keats (changes requested):
A couple of things to change but seems good otherwise!

stusmall:
Thanks for the review.  I'll get these changes fixed quickly, probably by the weekend.

Keats:
ping @stusmall

stusmall:
Sorry I haven't been able to find some time to make those changes.  I probably will be able to in the coming week.

stusmall:
So the thing I'm struggling with right now is for some reason that I haven't figured out my images on my test project have stopped processing.  I don't think its related to my changes.  I'll push my test projects up to github,  I think I'll need a hand on figuring it out.

Keats:
Might be the same as https://github.com/getzola/zola/issues/669

Ping me when you push it and I'll have a look

stusmall:
That's exactly it.   Thanks!

Keats:
I'll have a look at that issue next, hopefully it's something silly that is easily resolved.

Keats:
It should be fixed now

stusmall:
I updated with all the requested doc changes.  I looked at changing to use to_value as requested but it looks like it accomplishes the same thing but adds an unwrap.  It might be best to leave as is even though I can't see a way to trigger than panic.  Let me know if you disagree and I'll happily change it.

Keats:
Thanks!

PR #676, opened 2019-04-28
Add sublime-syntax for Nix expression language

nuxeh:
This is based on [1] and converted to sublime-syntax using [2].

[1] https://github.com/wmertens/sublime-nix
[2] https://github.com/aziz/SublimeSyntaxConvertor


Keats:
Thanks!

PR #678, opened 2019-04-29
0.8.0

Keats:


phil-opp:
I found another breaking change. The following code renders the content as HTML with Zola 0.7, but as plain text (with HTML tags) with Zola 0.8:

```
{{ page.content | safe | replace(from="<!-- toc -->", to=macros::toc(toc=toc)) }}
```

The fix is to apply the `safe` operator again at the end:

```
{{ page.content | safe | replace(from="<!-- toc -->", to=macros::toc(toc=toc)) | safe }}
```

Also, it would be nice if we could get a warning or an error for old-style internal links `./`. Now they're silently treated as relative links, which makes them invalid (linking to `index.md` files). Manually grepping all files is cumbersome and error-prone.

Keats:
Yeah the `safe` one is (hopefully) the only breaking change from Tera. You can do `{{ page.content | replace(from="<!-- toc -->", to=macros::toc(toc=toc)) | safe }}` instead, no need to repeat.

I thought about having warning for old style internal links but forgot to implement it :(

phil-opp:
> Yeah the `safe` one is (hopefully) the only breaking change from Tera.

I guess the changes in https://github.com/Keats/tera/commit/a9871d557a3b0b74a27851812291c91de787cf9f cause this? Maybe we should mention this in the Changelogs.md because it seems like it might affect quite a few people.

>  You can do `{{ page.content | replace(from="<!-- toc -->", to=macros::toc(toc=toc)) | safe }}` instead, no need to repeat.

That's good to know, thanks!

> I thought about having warning for old style internal links but forgot to implement it :(

Do you have plans to add it later? It would be really useful for converting existing projects.

Keats:
I've updated the changelog to mention it.

> Do you have plans to add it later? It would be really useful for converting existing projects.

I won't have time next week :( Maybe the week after.

phil-opp:
Thanks!

> I won't have time next week :( Maybe the week after.

No worries! There's no need for me to update now, so I can wait. I just wanted to know whether it is still planned. Thanks for all the great work!

Keats:
It's now in the `next` branch and it picked up a link that wasn't changed 👍

phil-opp:
Awesome, thanks a lot!

PR #685, opened 2019-05-17
Allow manual specification of header IDs

### LINKED ISSUE #433: Custom anchor links
While it is very useful to get automatic anchor links from the title, in some cases it is nice to choose anchor links that give nice URLs.

Given suitable configuraton in Hugo, a header can be annotated with `{#foobar}` to insert as an anchor. This feels quite simple when authoring content for internal linking IMHO.

chris-morgan:
Justification for this feature is added in the docs.

Precedent for the precise syntax: Hugo.

Hugo puts this syntax behind a preference named headerIds, and automatic header ID generation behind a preference named autoHeaderIds, with both enabled by default. I have not implemented a switch to disable this.

My suggestion for a workaround for the improbable case of desiring a literal “{#…}” at the end of a header is to replace `}` with `&#125;`.

The algorithm I have used is not identical to [that which Hugo uses][0], because Hugo’s looks to work at the source level, whereas here we work at the pulldown-cmark event level, which is generally more sane, but potentially limiting for extremely esoteric IDs.

Practical differences in implementation from Hugo (based purely on reading [blackfriday’s implementation][0], not actually trying it):

- I believe Hugo would treat `# Foo {#*bar*}` as a heading with text “Foo” and ID `*bar*`, since it is working at the source level; whereas this code turns it into a heading with HTML `Foo {#<em>bar</em>}`, as it works at the pulldown-cmark event level and doesn’t go out of its way to make that work (I’m not familiar with pulldown-cmark, but I get the impression that you could make it work Hugo’s way on this point). The difference should be negligible: only *very* esoteric hashes would include magic Markdown characters.

- Hugo will automatically generate an ID for `{#}`, whereas what I’ve coded here will yield a blank ID instead (which feels more correct to me—`None` versus `Some("")`, and all that).

In practice the results should be identical.

Fixes #433.

[0]: https://github.com/russross/blackfriday/blob/a477dd1646916742841ed20379f941cfa6c5bb6f/block.go#L218-L234

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -33,11 +33,12 @@ struct HeaderRef {
     start_idx: usize,
     end_idx: usize,
     level: i32,
+    id: Option<String>,
 }
 
 impl HeaderRef {
     fn new(start: usize, level: i32) -> HeaderRef {
-        HeaderRef { start_idx: start, end_idx: 0, level }
+        HeaderRef { start_idx: start, end_idx: 0, level, id: None }
     }
 }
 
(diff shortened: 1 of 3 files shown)


zayenz:
Nice and clean implementation IMHO.

I have a question on if spaces should be allowed around an anchor. That is, if a user writes an anchor like `{# my-anchor }`, then the anchor `my-anchor` could be generated. Perhaps not important, but might solve a few head-aches for some people.

chris-morgan:
I’m inclined to say no to allowing internal whitespace:

- the precedent provided in Hugo does not look to;
- the division of `{#foo}` is more naturally `#foo` surrounded by `{}`, given how links to the section will end in `#foo`, so if anything it’d be `{ #foo }` rather than `{# foo }`;
- the syntax is already uncomfortably close to the template language’s syntax (`{# comment #}`, with spaces inside the markers being conventional), I feel like adding whitespace plays to that confusion—though you could argue that it’s instead desirable consistency.

I see no compelling reason to allow additional whitespace.

chris-morgan:
This naturally warrants a changelog entry, but #678 should probably be merged before I add it. I’d be content to count it a non-breaking change.

zayenz:
Good reasons to disallow white spaces, I’m on board with the choice.

zayenz:
A follow up musing, should the tags perhaps be validated then to be slugs. IIUIC, the text is currently set as the anchor unconditionally.

chris-morgan:
> the text is currently set as the anchor unconditionally.

This is a feature. I’m actually very slightly unhappy with it that Markdown characters won’t be permitted in it, e.g. various patterns of IDs with asterisks won’t work. Ideally it’d allow *anything* in there.

Here, we trust the user to know what they’re doing.

Keats:
The PR looks very good!

It would be nice indeed to allow every characters: it might be easier with the current pulldown_cmark version, it was quite manual before.
I *think* we can count that as a non-breaking change but I can also bump the next version to be 0.8.0 rather than 0.7.1 as well, not a big deal.

chris-morgan:
Yeah, I’m unfamiliar with pulldown_cmark, didn’t look at its API at all beyond its CowStr, just did the minimum that could work in the current structure. If this can be done at the source level so that it binds tighter than normal Markdown formatting, that would probably be good.

What do you reckon: should there be a switch for this, or are you (like me) content to just leave it forcibly enabled, with the `&#125;` workaround for the exceptionally improbable case that the user needs a literal `{#…}` ending to their header?

(P.S. The consistent use of the word “header” instead of “heading” has really been throwing me off. Headers are the things at the start of the HTTP message and at the top of the page. I stuck to the convention, but would *really* like to rename it.)

Keats:
> but would really like to rename it. 

Go for it!

> should there be a switch for this, or are you (like me) content to just leave it forcibly enabled

No need for a switch imo, it's unlikely to happen in practice.

> Yeah, I’m unfamiliar with pulldown_cmark, didn’t look at its API at all beyond its CowStr, just did the minimum that could work in the current structure. If this can be done at the source level so that it binds tighter than normal Markdown formatting, that would probably be good.

I'm not too familiar with the updated version either (yet), I would need to have a look as well but that can be done after this is merged.

Keats:
I was renaming all the Header to Heading but realised `pulldown_cmark` uses `Header` in its AST https://github.com/raphlinus/pulldown-cmark/blob/master/src/parse.rs#L41 :/

Issue #691, opened 2019-05-18
Pagination not available for section


bkhl:
# Bug Report

## Environment

Zola version:

Tried 0.6.0, 0.7.0 and a build from the repository as of today.

## Expected Behavior

In the `section.html` template, with paging enabled for the section, `section.pages` and `section.paginator` should be available.

https://www.getzola.org/documentation/content/section/ seems to say this is an existing feature.

## Current Behavior

If in `_index.md` for a section, I set `paginate_by` to non-zero, in `section.html`, the `section.pages` is an empty list, and `section.paginator` dos not exist.

## Step to reproduce

1. Create a section by creating a file like `content/my_blog/_index.md` containing for example

```
+++
title = "My blog"
slug = "my_blog"
sort_by = "date"
paginate_by = 10
+++
```

2. In the `section.html` template, put `{{ section.pages }}` ⇒ in the page this expands to `[]`

3. In the `section.html` template, put `{{ section.paginator }}` ⇒ error message while building:

```
         | Error: Failed to render pager 7
         | Reason: Failed to render 'section.html'
         | Reason: Variable `section.paginator` not found in context while rendering 'section.html
```

Keats:
The `paginator` variable is not in the `section` variable but a standalone one: https://www.getzola.org/documentation/templates/pagination/

The doc above is not very clear though and has one sentence that doesn't make any sense...

bkhl:
That explains it, I'll just do this then:

```
{% if paginator %}
{% set pages = paginator.pages %}
{% else %}
{% set pages = section.pages %}
{% endif %}
```

Since I'm actually already using `{% for page in paginator.pages %}` in my main `index.html` template for another site I guess I should have been able to work it out, but yes, it would have helped me if that documentation page was a bit more verbose.

Keats:
Fix by updating the documentation.

PR #692, opened 2019-05-19
Elaboration and imrovement in docs on paging

bkhl:
Added and clarified a few things that would probably have saved me from
raising getzola/zola#691.


Keats:
Thanks!

PR #706, opened 2019-06-16
Round up reading time

elbaro:

Sanity check:

## Code changes

Medium has no '0 min read', so it should be `max(1, read_time)` or `ceil(read_time)`. This PR rounds up.

### DIFF
--- a/components/utils/src/site.rs
+++ b/components/utils/src/site.rs
@@ -9,7 +9,7 @@ pub fn get_reading_analytics(content: &str) -> (usize, usize) {
 
     // https://help.medium.com/hc/en-us/articles/214991667-Read-time
     // 275 seems a bit too high though
-    (word_count, (word_count / 200))
+    (word_count, ((word_count + 199) / 200))
 }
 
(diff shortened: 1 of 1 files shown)


components/utils/src/site.rs
```diff
@@ -86,11 +86,18 @@ mod tests {
         assert!(res.is_err());
     }
 
+    #[test]
+    fn reading_analytics_empty_text() {
+        let (word_count, reading_time) = get_reading_analytics("  ");
+        assert_eq!(word_count, 0);
+        assert_eq!(reading_time, 0);
```
Keats, line 93: shouldn't that be one according to your PR comment?

elbaro: Didn't realize Medium allows a title-only post. Just tested in Medium and it is 1 min read, although I think 0 min read makes more sense. What do you prefer?

Keats: 0 makes more sense yes.

Keats:
Thanks!

PR #718, opened 2019-06-23
chore: Update toml to 0.5 everywhere

igor-raits:
Seems it was missed in last dependency update. Not sure if intentionally though.


Keats:
That wasn't intentional no :(

Issue #710, opened 2019-06-20
Can zola hard link files from static/?


yacoob:
It's a feature request, not a bug :) I'm considering zola for a site that's graphic-heavy - right now it's about 1GB of images. Preferably, I'd like to keep them in `static/`. My current bespoke generator hardlinks the images to save time, as most of the time my input and output directories are siblings on the same filesystem. Is there any chance zola could hardlink files as well when asked for in `config.toml`?

My plan B would be to run `zola build` and hardlink the images outside of zola.

Thanks for considering this!

yacoob:
FWIW, with ~1GB of images, first build with `zola serve` took ~15s, subsequent builds ~2s. Disk cache? :D Funnily enough, this optimization is important primarily for local development - Netlify (which I use) keeps images in LFS and doesn't check them out during build, so the speedup would be minimal there.

Keats:
What would be the configuration? Specify which files/directories to hard link? It seems a bit niche to me

yacoob:
I'd see it as `hard_link_static = true` in `config.toml`. It'd make a hard link in `public` for every file in `static`, leaving the directories as they were. There'd need to be a same fs check, or good failure handling. Yup, it's a bit niche, in a sense it'd only be useful for people with non-trivially sized `static/` directory. It'd help preserve the speed of the generator though :)

yacoob:
This looks straightforward enough - I can take a stab at a PR. Are you interested, or do you consider this too niche for inclusion?

PR #723, opened 2019-06-24
Add an option to hard link files from static/ instead of copying.

yacoob:
As discussed in https://github.com/getzola/zola/issues/710, I've tried to add an optional hardlink for static files. Worked for me 

The way I've tested this:
```
$ cd test_site
$ dd if=/dev/zero of=static/foo bs=100m count=6
$ cargo run -- build; ls -l static/foo public/foo
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `/Users/yacoob/workarea/zola/target/debug/zola build`
Building site...
-> Creating 22 pages (1 orphan), 10 sections, and processing 0 images
Done in 1.6s.

-rw-r-----  1 yacoob  staff  629145600 24 Jun 23:17 public/foo
-rw-r-----  1 yacoob  staff  629145600 24 Jun 23:17 static/foo
$ # added hard_link_static = true to config.toml
$ cargo run -- build; ls -l static/foo public/foo
   Compiling zola v0.8.0 (/Users/yacoob/workarea/zola)
    Finished dev [unoptimized + debuginfo] target(s) in 9.42s
     Running `/Users/yacoob/workarea/zola/target/debug/zola build`
Building site...
-> Creating 22 pages (1 orphan), 10 sections, and processing 0 images
Done in 723ms.

-rw-r-----  2 yacoob  staff  629145600 24 Jun 23:17 public/foo
-rw-r-----  2 yacoob  staff  629145600 24 Jun 23:17 static/foo
```
With `hard_link_static = true` the build was faster as expected, and the hardlink count in `ls` output has increased. 

I don't have a good idea how to add a meaningful test here. If you do, I'm all ears. Maybe just enable it for the `test_site`, to make sure the codepath gets executed during integration test?

---

Sanity check:

## Code changes

### DIFF
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -111,7 +111,7 @@ pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf, hard_link: boo
     Ok(())
 }
 
-pub fn copy_directory(src: &PathBuf, dest: &PathBuf, hardlink: bool) -> Result<()> {
+pub fn copy_directory(src: &PathBuf, dest: &PathBuf, hard_link: bool) -> Result<()> {
     for entry in WalkDir::new(src).into_iter().filter_map(std::result::Result::ok) {
         let relative_path = entry.path().strip_prefix(src).unwrap();
         let target_path = dest.join(relative_path);
@@ -121,7 +121,7 @@ pub fn copy_directory(src: &PathBuf, dest: &PathBuf, hardlink: bool) -> Result<(
                 create_directory(&target_path)?;
             }
         } else {
-            copy_file(entry.path(), dest, src, hardlink)?;
+            copy_file(entry.path(), dest, src, hard_link)?;
         }
     }
     Ok(())
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -111,7 +111,7 @@ pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf, hardlink: bool
     Ok(())
 }
 
-pub fn copy_directory(src: &PathBuf, dest: &PathBuf) -> Result<()> {
+pub fn copy_directory(src: &PathBuf, dest: &PathBuf, hardlink: bool) -> Result<()> {
     for entry in WalkDir::new(src).into_iter().filter_map(std::result::Result::ok) {
         let relative_path = entry.path().strip_prefix(src).unwrap();
         let target_path = dest.join(relative_path);
@@ -121,7 +121,7 @@ pub fn copy_directory(src: &PathBuf, dest: &PathBuf) -> Result<()> {
                 create_directory(&target_path)?;
             }
         } else {
-            copy_file(entry.path(), dest, src, false)?;
+            copy_file(entry.path(), dest, src, hardlink)?;
         }
     }
     Ok(())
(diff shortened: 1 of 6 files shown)


yacoob:
As for CI: two builds that have failed look like timeouts before zola was even built. It has failed similarily on `next` branch before my changes, so maybe it's expected. No experience with Travis on this side :D

components/utils/src/fs.rs
```diff
@@ -95,19 +95,23 @@ pub fn find_related_assets(path: &Path) -> Vec<PathBuf> {
 
 /// Copy a file but takes into account where to start the copy as
 /// there might be folders we need to create on the way
-pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf) -> Result<()> {
+pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf, hardlink: bool) -> Result<()> {
```
Keats: is it called `hardlink` instead of `hard_link` because of shadowing? You can just use `std::fs::hard_link` in the function body if that's the case instead.

docs/content/documentation/getting-started/configuration.md
```diff
@@ -41,6 +41,12 @@ generate_rss = false
 # not set (the default).
 # rss_limit = 20
 
+# Whether to copy or hardlink files in static/ directory. Useful for sites
+# whose static files are large. Note that for this to work, both static/ and
+# output directory need to be on the same filesystem. Also, theme's static/
+# files are always copies, regardles of this setting. False by default.
```
Keats, line 47: hmm that seems to have a few pitfalls but nothing too bad.

docs/content/documentation/getting-started/configuration.md
```diff
@@ -41,6 +41,12 @@ generate_rss = false
...
+# hard_link_static = true
```
Keats: this doc shows the default so it should be false here

Keats (changes requested):

components/utils/src/fs.rs
yacoob: Not that much experience with Rust here, thanks for the hint :>

docs/content/documentation/getting-started/configuration.md
yacoob: Changed.

Keats:
Code looks good.
A quick additional question: what's the workflow with big static folders: do you put everything in git? How is deploy handled?

yacoob:
I keep the big files (photos) in `git-lfs`. The repo itself is on [github](https://github.com/yacoob/gcu/tree/master/static/photos/full), the blobs are pushed to Netlify's [LFS backend](https://github.com/yacoob/gcu/blob/master/.lfsconfig). You can set LFS backend per remote, so I have yet another copy of those photos pushed to my gitea instance. Netlify skips the LFS checkout during the build, so all they see are the files with hashes. They use these to swap in actual blobs from LFS on publish. As a result, hardlinking is not really useful for Netlify deployment - it's useful whenever I test things locally at home :D 

LFS is, of course, optional. Some people are OK with storing large objects in git directly, and if the blobs are not changing too often, it might be a sensible approach.

Keats:
Thanks for the info!

yacoob:
Thank you for the merge! :>

Are you ok with no explicit test for this functionality?

Keats:
Hmm forgot about that. The logic seems simple enough and it uses std so it should be fine (famous last words).

PR #736, opened 2019-07-02
added https://uwe-arzt.de to examples

uwearzt:
Hi,

i changed my website from cobalt to zola. It can be mentioned in the examples, full source is available.


Keats:
Thanks!

PR #739, opened 2019-07-04
Add --open flag to serve command, to open URL in default browser

timvisee:
> **IMPORTANT: Please do not create a Pull Request adding a new feature without discussing it first.**
>
> The place to discuss new features is the forum: <https://zola.discourse.group/>
> If you want to add a new feature, please open a thread there first in the feature requests section.

I noticed this message after adding this feature. Since it's quite small I decided to initiate the PR anyway. Please tell me if a discussion for this is still desired. In that case I'll be sure to follow up to that.

This PR adds an `--open` flag to the `serve` command, to open the URL the page is locally hosted at in the default browser of the user. While working with `zola` I commonly wanted to have such functionality  built-in, thus I decided to add this.

---

Sanity check:

## Code changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1622,6 +1622,14 @@ version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "open"
+version = "1.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "openssl"
 version = "0.10.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


zimmi:
Just a casual observation: Maybe opening the browser should be the default when running `zola serve` and there should be something like a `--no-open` flag (if at all)?

timvisee:
> Just a casual observation: Maybe opening the browser should be the default when running `zola serve` and there should be something like a `--no-open` flag (if at all)?

I can imagine that to become annoying, even though it's a simple flag. I've had various cases where I had to re-invoke the command constantly due to theme, config or taxonomy changes. Just my thought of course.

awulkan:
I'd rather not have the website be opened by default. But I would like to have it as an optional flag.

Keats:
Looks good, thanks!
This definitely needs to be opt-in.

zimmi:
> I can imagine that to become annoying, even though it's a simple flag. I've had various cases where I had to re-invoke the command constantly due to theme, config or taxonomy changes. Just my thought of course.

Yeah, makes sense!

PR #741, opened 2019-07-10
docs: Add installation on FreeBSD

ghost:
Only just committed it to the FreeBSD ports tree.  The package should be available in a few days.

https://www.freshports.org/www/zola


Keats:
Neat, thanks!

PR #743, opened 2019-07-11
Fix broken links on the "Sections and Pages" documentation page

genderquery:
It looks like two links on this page didn't get updated with the change from `./` to `@/` in ec61a57, resulting in broken links. This PR fixes that.


Keats:
Thanks, I definitely need to add some warnings for that in 0.8.1

PR #744, opened 2019-07-11
Fix clippy warnings

traviscross:
Clippy is returning some warnings.  Let's fix or explicitly ignore
them.  In particular:

- In `components/imageproc/src/lib.rs`, we implement `Hash` explicitly
  but derive `PartialEq`.  We need to maintain the property that two
  keys being equal implies the hashes of those two keys are equal.
  Our `Hash` implementations preserve this, so we'll explicitly ignore
  the warnings.

- In `components/site/src/lib.rs`, we were calling `.into()` on some
  values that are already of the correct type.

- In `components/site/src/lib.rs`, we were using `.map(|x| *x)` in
  iterator chains to remove a level of indirection; we can instead say
  `.copied()` (introduced in Rust v1.36) or `.cloned()`.  Using
  `.copied` here is better from a type-checking point of view, but
  we'll use `.cloned` for now as Rust v1.36 was only recently
  released.

- In `components/templates/src/filters.rs` and
  `components/utils/src/site.rs`, we were taking `HashMap`s as
  function arguments but not generically accepting alternate `Hasher`
  implementations.

- In `src/cmd/check.rs`, we use `env::current_dir()` as a default
  value, but our use of `unwrap_or` meant that we would always
  retrieve the current directory even when not needed.

- In `components/errors/src/lib.rs`, we can use `if let` rather than
  `match`.

- In `components/library/src/content/page.rs`, we can collapse a
  nested conditional into `else if let ...`.

- In `components/library/src/sorting.rs`, a function takes `&&Page`
  arguments.  Clippy warns about this for efficiency reasons, but
  we're doing it here to match a particular sorting API, so we'll
  explicitly ignore the warning.

### DIFF
--- a/components/errors/src/lib.rs
+++ b/components/errors/src/lib.rs
@@ -31,10 +31,9 @@ impl StdError for Error {
     fn source(&self) -> Option<&(dyn StdError + 'static)> {
         let mut source = self.source.as_ref().map(|c| &**c);
         if source.is_none() {
-            match self.kind {
-                ErrorKind::Tera(ref err) => source = err.source(),
-                _ => (),
-            };
+            if let ErrorKind::Tera(ref err) = self.kind {
+                source = err.source();
+            }
         }
 
         source
--- a/components/imageproc/src/lib.rs
+++ b/components/imageproc/src/lib.rs
@@ -129,6 +129,7 @@ impl From<ResizeOp> for u8 {
     }
 }
 
+#[allow(clippy::derive_hash_xor_eq)]
 impl Hash for ResizeOp {
     fn hash<H: Hasher>(&self, hasher: &mut H) {
         hasher.write_u8(u8::from(*self));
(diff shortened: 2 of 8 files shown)


Keats:
In which version was `copied` added? I'll have a look at the CI failures for sass...

traviscross:
`.copied` went in (along with the updated `clippy` warning) in Rust v1.36.

https://github.com/rust-lang/rust/pull/60333

Keats:
Ah I would probably remove that change then, I usually target a couple of versions older than the latest stable (currently 1.34). I'll add that `copied` change in a couple of months

traviscross:
That works; I've updated the branch to use `.cloned()` instead for now, which still resolves the warnings without depending on a recent version of `rustc`.

traviscross:
Doing a clean build revealed some other clippy warnings that this branch now also fixes.

Keats:
Thanks!

PR #759, opened 2019-07-22
Allow ignored_content to support markdown files

### LINKED ISSUE #682: Ignore README.md in git submodule
## Environment

Zola version: 0.5.1

## Problem

We have a submodule in our content directory, which has its own README, so:

```
zola/
  content/
    submodule/
      README.md
      _index.md
```

The README.md file does not, and should not, have TOML front-matter, so zola build fails. Adding README.md to `ignored_content` in the config does nothing, because zola attempts to parse the file before potentially ignoring it.

## Proposed Solution

There should be a way to ignore markdown files with an `.md` extension that don't have TOML front-matter.

A hack that works for now is to name `README` to `README.markdown`, so Github sees it as a markdown file, but zola ignores it.

pkova:
Fixes #682.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ notify = "4"
 ws = "0.8"
 ctrlc = "3"
 open = "1.2"
+globset = "0.4"
 
 site = { path = "components/site" }
 errors = { path = "components/errors" }
--- a/components/rebuild/src/lib.rs
+++ b/components/rebuild/src/lib.rs
@@ -306,7 +306,17 @@ pub fn after_content_rename(site: &mut Site, old: &Path, new: &Path) -> Result<(
         old.to_path_buf()
     };
     site.library.write().unwrap().remove_page(&old_path);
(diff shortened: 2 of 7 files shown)


Keats:
Thanks!
I am a bit worried about the speed impact of checking a glob on every page, have you done some benchmarks on that? You can generate some big sites in components/site/benches

pkova:
I now ran `cargo +nightly bench` with the huge blog and and an `ignored_content` of 300 randomly generated strings of the format `*/<random 20 char string>.md`. This commit had no discernible performance impact.

Note that this is my first time doing anything in Rust so buyer beware!

Keats:
Looks ok from a local testing, can you add a test for that in components/site/tests?

pkova:
Done!

Let me know if you want something more rigorous.

Keats:
Thanks!

flowerornament:
Thanks for coming through on this @pkova!

PR #763, opened 2019-07-31
Allow relative paths in new_file()/new_section()

Freaky:
These functions expect that file_path can have base_path stripped from it, but during reloading they can be given relative paths.  Maybe this behaviour varies between the notify backends?

This fixes two `zola serve` panics on FreeBSD (poll backend).


Keats:
Thanks!

PR #772, opened 2019-08-12
Add trailing slash to multilingual section path

hrektts:
Other than the default language that is set at `config.toml`, `section.path` ends without a trailing slash in multilingual sites.
This behaviour seems to have been introduced with [this commit](https://github.com/getzola/zola/commit/260c413de406dd176605eb1db3a75d17c5418e64#diff-0ba0bdecfa60be201d2919df49632079L95-R95).
I think it would be reasonable for `section.path` to end with a slash, regardless of the language settings.

Sanity check:

## Code changes


Keats:
Nice catch, thanks

PR #774, opened 2019-08-12
Fix ToC generation for heading levels > 3 (bugfix)

southerntofu:
Previously, 4-level headers were inserted as children of the last 2nd level header because the code didn't go peek deep enough into the children of children (etc).

I wrote a simple recursive helper function that conveniently takes an Option<Header> to potentially insert a new Header into.

The documentation doesn't need to be edited.

TODO: i need to write some tests

### DIFF
--- a/components/rendering/src/table_of_contents.rs
+++ b/components/rendering/src/table_of_contents.rs
@@ -45,14 +45,13 @@ fn insert_into_parent(potential_parent: Option<&mut Header>, header: &Header) ->
                 // We have a direct child of the parent
                 parent.children.push(header.clone());
                 return true;
-            } else {
-                // We need to go deeper
-                if !insert_into_parent(parent.children.iter_mut().last(), header) {
-                    // No, we need to insert it here
-                    parent.children.push(header.clone());
-                }
-                return true;
             }
+            // We need to go deeper
+            if !insert_into_parent(parent.children.iter_mut().last(), header) {
+                // No, we need to insert it here
+                parent.children.push(header.clone());
+            }
+            return true;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


components/rendering/src/table_of_contents.rs
```diff
@@ -27,33 +27,66 @@ impl Default for Header {
...
+        None => {
+            // No potential parent to insert into so it needs to be insert higher
+            return false;
+        },
+        Some(parent) => {
+            let diff = header.level - parent.level;
+            if diff <= 0 {
+                // Heading is same level or higher so we don't insert here
+                return false;
+            } else {
```
Keats: I tend to prefer early return with no matching `else` to avoid indentation

components/rendering/src/table_of_contents.rs
```diff
@@ -27,33 +27,66 @@ impl Default for Header {
...
+                if diff == 1 {
+                    // We have a direct child of the parent
+                    parent.children.push(header.clone());
+                    return true;
+                } else {
+                    // We need to go deeper
+                    match insert_into_parent(parent.children.iter_mut().last(), header) {
```
Keats: can be simplified like

```rust
if !insert_into_parent(parent.children.iter_mut().last(), header) {
    // No, we need to insert it here
    parent.children.push(header.clone());
}
return true;
```

Keats (commented):
Weird I thought my tests was good enough :/
This PR should be ok once tests are added

components/rendering/src/table_of_contents.rs
southerntofu: good suggestion! thanks

southerntofu:
>  This PR should be ok once tests are added

Done!

Keats:
Thanks!

Issue #768, opened 2019-08-08
Underscores in slugs get converted to hyphen-minus and other slugification issues


chris-morgan:
I have a content file named `rust-cfg_attr.md`  on my Zola 0.8.0 site.

Zola ignores what I deliberately typed as the slug, and turns the `_` into `-`, yielding `rust-cfg-attr`. This makes me unhappy.

I expect Zola to make no modifications at all to the slug that I specify, so that I can do things how I want to, e.g. `Do_things_like_Wikipedia` if I really want to.

Keats:
It looks like the "slug everything" approach is too strong. It should instead only remove some characters like `&` or `/` from the slug. This will give us UTF-8 slugs as well, which are transformed into ASCII right now.

Anyone interested in doing that?

chris-morgan:
Why change anything? `&` is perfectly fine in URLs, and you won’t encounter `/` in filenames.

I’m interested in fixing this.

Keats:
From https://en.wikipedia.org/wiki/Percent-encoding I would prefer to stick with the unreserved characters + UTF8 since otherwise things like & needs to be encoded for the sitemap. I had a quick test with Hugo and the resulting slugs are exactly what I expected so it might be as simple as porting their slug code from Go to Rust.

chris-morgan:
Yes, ampersand will need to be encoded everywhere, just as would already happen; I fail to see why that would be a problem.

I see no reason why we should alter what the user has deliberately chosen.

Keats:
It will fail to work in some tools and users might not be aware of encoding, no reason not to remove reserved characters imo: generated URLs by Zola should work everywhere out of the box.

The Go code for Hugo is https://github.com/gohugoio/hugo/blob/master/helpers/path.go#L136 which raised the issue of what to do with accents in paths. I think we want to remove them (at least I wouldn't want them in a French page for example) but need to be careful to not remove other languages significant "accents" like `ビ` in Japanese and replace it with the correct ASCII version, `ä` in German should become `ae`, not `a`. 
I don't know whether an option like `RemovePathAccents` in Hugo is required though, it feels like the default should be `true` and I'm pretty sure the option is only there for backwards compatibility. 

A related issue: https://github.com/getzola/zola/issues/570

chris-morgan:
What tools do you have in mind where things might fail?

[...]

PR #781, opened 2019-08-17
maybe_slugify() only does simple sanitation if config.slugify is false

southerntofu:
If this PR fits the problem, i will add some tests.

Discussion in #768


components/library/src/slugs.rs
```diff
@@ -1,7 +1,7 @@
 pub use slug::slugify;
 
 pub fn quasi_slugify(s: &str) -> String {
-    s.replace("#", "").replace("/", "")
+    s.replace("#", "").replace("/", "").replace("?", "").replace("\n", "")
```
Keats: @southerntofu something I just remembered if that we should also removed invalid Windows characters in filenames like `:`.
It is already an issue in the page filename (https://www.getzola.org/documentation/content/page/): 

> The page name will be anything after _/- so a filename like 2018-10-10-hello-world.md will be available at [base_url]/hello-world. Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.

The full list: https://gist.github.com/doctaphred/d01d05291546186941e1b7ddc02034d3#file-ntfs-filenames-txt-L9-L17
Also needs to `trim_end` apparently: https://gist.github.com/doctaphred/d01d05291546186941e1b7ddc02034d3#file-ntfs-filenames-txt-L33

The full RFC3339 was already a mistake imo so let's not repeat it for actual pages/sections and be fully compatible with every OS.

southerntofu:
> I want slugify_paths to stick with true/false

I've been thinking: we've been changing our minds so much over this question already. slugification is a process that cannot be changed in the future without breaking existing URLs. So i can see two paths going forward:
- either wait for feedback from users to ensure we're not going against their expectations
- or make the list of stripped chars user-configurable, while retaining a simple true/false switch for most setups (false = NTFS-compatible filenames, true = backwards-compatible slugifiy)

I was experimenting with the second path here. Do you think it's so inappropriate?

Keats:
I would prefer the first path. Removing NTFS stripping would mean their site is not buildable at all on Windows (changing that setting would not be backward compatible so if you want to use Windows later you need to use WSL/VMs or lose all permalinks with those characters) and is adding complexity to Zola.
I would like to be able to have eg `:` in a path but that's imo the downside of using a static site engine. It might be possible to percent-encode those instead but that seems messy.

southerntofu:
> I would like to be able to have eg : in a path

That's a valid request and there's no reason a SSG could not support that if your filesystem and webserver both support it. Making sure Zola works out-of-the-box for everyone is important though.

IMO configuring the forbidden charset as proposed here adds very little complexity because the introduced complexity is transparent to the user who can rely on the simple true/false switch for slugification as previously proposed (the extra complexity is handled by serde).

Of course if you have a definitive strong opinion on this matter i will revert those changes. I just tried that approach because we've changed our minds at least 5 times already about which characters should be allowed or forbidden, so i can see how this question would come up again in the future :)

Keats:
> Of course if you have a definitive strong opinion on this matter i will revert those changes. I just tried that approach because we've changed our minds at least 5 times already about which characters should be allowed or forbidden, so i can see how this question would come up again in the future :)

That's a fair point but so far I believe only 3 persons including you raise the issue with unicode characters. I'd rather start more conservative so people actually have to ask for it and this way we know that it is needed rather than wondering.

southerntofu:
> this way we know that it is needed rather than wondering.

That makes sense, thanks.

components/library/src/slugs.rs
southerntofu: I think this is solved now. Also i updated the docs.

docs/content/documentation/content/page.md
```diff
@@ -3,29 +3,48 @@ title = "Page"
...
+
+```
++++
+title = "Le mouvement des Femmes Libres, à la tête de la libération kurde"
+slug = "femmes-libres-libération-kurde"
++++
+This is my article.
+```
+
+This frontmatter will output the article to `[base_url]/zines/femmes-libres-libération-kurde` with `slugify` disabled, and to `[base_url]/zines/femmes-libres-liberation-kurde` with `slugify` enabled.
```
Keats: slugify_path

docs/content/documentation/content/page.md
```diff
@@ -3,29 +3,48 @@ title = "Page"
...
+
+### Path from filename
+
+When the article's output path is not specified in the frontmatter, it is extracted from the file's path in the content folder. Consider a file `content/foo/bar/thing.md`. The output path is constructed:
+- if the filename is `index.md`, its parent folder name (`bar`) is used as output path
+- otherwise, the output path is extracted from `thing` (the filename without the `.md` extension)
+
+If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
+
+The output path extracted from the file path is then slugified or not depending on the `slugify` config, as explained previously.
```
Keats: slugify_path

docs/content/documentation/content/page.md
```diff
@@ -3,29 +3,48 @@ title = "Page"
...
+Creating an `about.md` file is exactly equivalent to creating an `about/index.md` file.  The only difference between the two methods is that creating the `about` folder allows you to use asset colocation, as discussed in the [Overview](@/documentation/content/overview.md#assets-colocation) section of this documentation.
+
+## Output paths
+
+For any page within your content folder, its output path will be defined by either:
+
+- its `slug` frontmatter key
+- its filename
+
+Either way, these proposed path will be sanitized before being used. If `slugify_paths` is enabled in the site's config, paths are [slugified](https://docs.rs/slug/0.1.4/slug/fn.slugify.html). Otherwise, a simpler sanitation is performed, outputting only valid NTFS paths. The following characters are removed: `<`, `>`, `:`, `/`, `|`, `?` `*` `#` `\\`, as well as newlines and tabulations. Additionally, trailing whitespace and dots are removed.
```
Keats: I would link to wikipedia https://en.wikipedia.org/wiki/Clean_URL#Slug rather than the slug crate docs

test_site/config.toml
```diff
@@ -4,6 +4,7 @@ highlight_code = true
 compile_sass = true
 generate_rss = true
 theme = "sample"
+slugify = true
```
Keats: slugify_path

Keats (changes requested):

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Breaking
 
-- Pages with draft=true are now only loaded/rendered in `zola serve`
+- Add `--drafts` flag to `build`, `serve` and `check` to load drafts. Drafts are never loaded by default anymore
+- [Slugification](https://en.wikipedia.org/wiki/Slug_(web_publishing)#Slug) of page paths is now optional. By default, non-ASCII characters (international characters i.e. UTF8) will be retained in the generated pages. If you are currently using non-ASCII characters in your content file paths and want to keep the URLs Zola previously generated, set `slugify = true` in your config
```
Keats: slugify_path

Keats:
Closed in favour of https://github.com/getzola/zola/pull/875 since I needed to rebase everything

PR #786, opened 2019-08-24
Add anchor existence checking to link_checker component

### LINKED ISSUE #539: `check_external_links` should check anchors

wyze:
This will check to make sure an anchor exists on the page when it is present in the url that is being checked.

Close #539


components/link_checker/Cargo.toml
```diff
@@ -6,3 +6,6 @@ authors = ["Vincent Prouillet <prouillet.vincent@gmail.com>"]
 [dependencies]
 reqwest = "0.9"
 lazy_static = "1"
+scraper = "0.10"
```
Keats: Is there a more lightweight scraper? This seems like it would add quite a few new dependencies.

Would it be OK to just do a simple `string.contains("id='the-id'") | string.contains(#"id="the-id'"#)`?

components/link_checker/src/lib.rs
```diff
@@ -91,4 +126,22 @@ mod tests {
         assert!(res.code.is_none());
         assert!(res.error.is_some());
     }
+
+    #[test]
+    fn can_validate_anchors() {
+        let url = "https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect";
```
Keats, line 136: can we skip the remote access and just test `check_page_for_anchor` for both tests?

Keats (changes requested):

wyze:
Thanks for the feedback. Changes have been pushed and ready for another review.

Keats:
Is the naive approach good enough in practice? Have you tried it with a site?

wyze:
I did test it on cheats.rs and had the same results using `scraper` dependency and this approach.

components/link_checker/src/lib.rs
```diff
@@ -70,9 +81,32 @@ pub fn check_url(url: &str) -> LinkResult {
...
+        },
+        None => false,
+    }
+}
+
+fn check_page_for_anchor(url: &str, body: reqwest::Result<String>) -> Result<()> {
+    let body = body.unwrap();
+    let index = url.find('#').unwrap();
+    let anchor = url.get(index + 1..).unwrap();
+    let checks: [String; 2] = [format!(" id='{}'", anchor), format!(r#" id="{}""#, anchor)];
```
Keats: I had completely forgotten but anchors can also be `name=` rather than id so there are actually 4 checks

Keats: I've pushed it directly to the PR

wyze:
Okay, thanks! I just got back from vacation last night and was going to try to get to this today or tomorrow.

Keats:
No problem, hope you had great vacations!

PR #792, opened 2019-08-29
Add support for SVG files in get_image_metadata (closes #769)

### LINKED ISSUE #769: Support get_image_metadata for SVG
# Bug Report

My blog mostly uses SVG as an image format. This way, images are crisp while still being small. @killercup and me recently noticed that posts get redrawn when an image gets loaded on the page. [Here is an example](https://matthias-endler.de/2019/maybe-you-dont-need-kubernetes/). Although a minor annoyance, it's not very visually pleasing.

One idea was to set the width and height of the image and there is the handy `get_image_metadata` builtin, but it does not support SVG yet.

## Environment

Zola version: 0.8.0

## Expected Behavior

The following shortcode should work for SVG:

```
{% set meta = get_image_metadata(path=page.path~src) %}
<img src="{{ src }}" height="{{ meta.height }}" width="{{ meta.width }}" />
```

## Current Behavior

Error: Failed to render content of snippet.
Reason: The Decoder does not support the image format `Image format image/"svg" is not supported.`

## Suggestion

For the case where the SVG file has a `viewBox` attribute set, we could use that extract the width and height from it. In case the attribute is not set, we could throw and error with a hint about the missing `viewBox`.

## Alternative approach

Previously I used [lqip](https://matthias-endler.de/2017/image-previews/) on my blog, but I did not port it when I moved from [cobalt.rs](https://github.com/cobalt-org/cobalt.rs) to zola.
I could dig that up again, but I was hoping to avoid that if possible and rather go with a more general solution.

mre:
This adds support for SVG files to get_image_metadata.
Since the documentation did not mention the supported file formats I did not modify that.
After all it does not change existing behavior.

Sanity check:

## Code changes


Keats:
It looks like the changes to Cargo.lock pull down some older dependencies than what is currently in the branch, causing Tera to be in an older version than the currently used and having failing tests.

mre:
Really? Must have been my rebase then. I'll pull the version from `next` again when I'm back on a computer. 
Any comments on the code or are you happy so far?

components/templates/src/global_fns/mod.rs
```diff
@@ -164,14 +166,27 @@ impl TeraFn for GetImageMeta {
...
     }
 }
+// Try to read the image dimensions for a given image
+fn image_dimensions(path: &PathBuf) -> Result<(u32, u32)> {
+    if let Some("svg") = path.extension().and_then(OsStr::to_str) {
+        let img = svg::Metadata::parse_file(&path)
+            .map_err(|e| Error::chain(format!("Failed to process image: {}", path.display()), e))?;
+        let h = img.height().ok_or("Undefined height")?;
+        let w = img.width().ok_or("Undefined width")?;
+        Ok((h as u32, w as u32))
```
Keats, line 188: is that casting a string to u32? If we can't read height/width it's probably better to have an error instead.

Keats: Nevermind, completely forgot you can use `?` on `Option`.

mre:
@Keats resolved the conflicts now. Please have another look. :)

Keats:
Argh there are some issues with the commits, can you rebase your commit on top of next?

Keats:
Seeing the list of commits, it might be easier to just cherry-pick on another branch actually

Keats:
Closing for now, please opn an updated PR if you want that merged!

PR #793, opened 2019-09-01
Fix crash of trans() function called on absent translation key

zdenek-crha:
## Description

The `trans()` Tera function used by Zola crashes when attempting to translate key that does not exist in `config.toml` file. The crash does not even contain information *which* key is missing from config, making fixing config impossible.

The root cause of the crash was un-handled error path when retrieving translations and keys from Config struct. 

The PR adds new method for `Config` struct for translation retrieval. The reason for addition to `Config` is mainly encapsulation of the config data access and reusability by other parts of Zola code.

Then updates implementation of `Trans::call()` function that provides `trans()` Tera function to use new way for translation retrieval and chain any errors with additional error message for better readability.

I have not removed existing unit tests nor `pub` from `Config::translation` field. It is something that 'might' be desirable, to ensure that nobody will try to mess with `Config struct` internals without proper error checks again.

## Sanity check

## Code changes

  * no documentation to update as I was fixing crash

### DIFF
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -8,6 +8,7 @@ use toml;
 use toml::Value as Toml;
 
 use errors::Result;
+use errors::Error;
 use highlighting::THEME_SET;
 use theme::Theme;
 use utils::fs::read_file_with_error;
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

zdenek-crha:
Do you want me to cleanup the `Config` class interface too? Remove `pub` on `translations` field and remove old style unit tests?

Keats:
Yes please

PR #801, opened 2019-09-15
docs(examples): Remove dead links in Zola example sites listing

klausi:
The first 2 links in the example doc are dead, I guess they should be removed?


Keats:
Thanks, I'll need to update my site in that list :p

PR #803, opened 2019-09-21
Don't scale images up when resizing with fit

Nukesor:
These are the changes as discussed in the following forum thread: https://zola.discourse.group/t/new-image-resizing-mode-shrink-fit/203

Documentation changes have been made and tested.


Keats:
cc @vojtechkral what do you think?  It makes sense to me

vojtechkral:
Cool, I'll have a look either tonight or tomorrow...

vojtechkral:
Generally, this looks good to me.

My original purpose was picture/photo galleries where the previous behaviour _might in theory_ make sense to make the thumbnails uniform... However, (1) this can be solved on the frontend with css/js, and (2) people are pretty unlikely to have pictures in galleries smaller then the thumbnail, so, this is probably moot.

Anyway, it's fine by me. I just hope the change doesn't break anyone's workflow, but I find it unlikely.

Thanks!

Keats:
Yeah I think it makes sense to me to not upscale since this is never really what you want. We'll see if it breaks for someone and if so we can add a shrink_fit and revert the changes to fit.

Thanks, both of you!

PR #812, opened 2019-10-05
Skip anchor checking for URL with prefix in config

### LINKED ISSUE #805: External link checker does not find GitHub anchors
I just updated my `blog_os` project from zola 0.7.0 to zola 0.9.0. It worked great and it even caught several invalid anchors in internal links. Thanks a lot for your work!

I noticed a problem when checking external links using `zola check`: For some URLs (mostly GitHub URLs), it reports an anchor as 'not found' even though it seems to exist. Some examples:

```
Error: Dead link in /home/runner/work/blog_os/blog_os/blog/content/first-edition/extra/cross-compile-libcore.md to https://github.com/japaric/rust-cross#the-target-triple: Anchor `#the-target-triple` not found on page
Dead link in /home/runner/work/blog_os/blog_os/blog/content/first-edition/extra/cross-compile-libcore.md to https://github.com/japaric/xargo#dependencies: Anchor `#dependencies` not found on page
Dead link in /home/runner/work/blog_os/blog_os/blog/content/first-edition/extra/naked-exceptions/03-returning-from-exceptions/index.md to https://github.com/rust-lang/rust/blob/c772948b687488a087356cb91432425662e034b9/src/librustc_back/target/mod.rs#L194-L214: Anchor `#L194-L214` not found on page
```

I think the reason is that GitHub uses IDs like `user-content-the-target-triple` instead of `the-target-triple` for its generated anchors. I'm not sure why the anchor still works without the `user-content` prefix. Maybe it's done in Javascript. Either way, I think that it might be worth to handle somehow, given that GitHub is a common link target.

TjeuKayim:
Fixes #805

`zola check`

The anchors of external URLs will not be checked if it starts with one of the configured prefixes.
So false positives can be put on the ignore list, like https://github.com/rust-lang/rust/blob/c772948b687488a087356cb91432425662e034b9/src/librustc_back/target/mod.rs#L194-L214: Anchor `#L194-L214 and https://caniuse.com/#feat=css-filters.

```toml
# Link Checker: skip anchor checking for external URLs that start with these prefixes
link_checker_skip_anchor_prefixes = [
    "https://caniuse.com/",
]
```

Sanity check:

## Code changes

### DIFF
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -86,7 +86,7 @@ impl Default for Taxonomy {
     }
 }
 
-type TranslateTerm  = HashMap<String, String>;
+type TranslateTerm = HashMap<String, String>;
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(default)]
@@ -317,9 +317,16 @@ impl Config {
             Error::msg(format!("Translation for language '{}' is missing", lang.as_ref()))
         })?;
 
-        terms.get(key.as_ref()).ok_or_else(|| {
-            Error::msg(format!("Translation key '{}' for language '{}' is missing", key.as_ref(), lang.as_ref()))
-        }).map(|term| term.to_string())
+        terms
+            .get(key.as_ref())
+            .ok_or_else(|| {
+                Error::msg(format!(
+                    "Translation key '{}' for language '{}' is missing",
+                    key.as_ref(),
+                    lang.as_ref()
+                ))
+            })
+            .map(|term| term.to_string())
     }
 }
 
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -7,8 +7,8 @@ use syntect::parsing::{SyntaxSet, SyntaxSetBuilder};
 use toml;
 use toml::Value as Toml;
 
-use errors::Result;
 use errors::Error;
+use errors::Result;
 use highlighting::THEME_SET;
 use theme::Theme;
 use utils::fs::read_file_with_error;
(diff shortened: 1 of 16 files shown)


components/config/src/config.rs
```diff
@@ -152,6 +153,8 @@ pub struct Config {
     #[serde(skip_serializing, skip_deserializing)] // not a typo, 2 are need
     pub extra_syntax_set: Option<SyntaxSet>,
 
+    pub link_checker: LinkChecker,
```
Keats, line 168: I would like all config options to be in the config component directly.

Keats (changes requested):
Just a small change, everything looks great otherwise.

components/config/src/config.rs
TjeuKayim: Okay, I can change that. Thanks for the feedback. 

What do you think about the LinkChecker config struct? This way it will be easier to add new options later, like the user agent. Should `check_url()` convert to a method? Or keep adding arguments while the link checker evolves?

components/config/src/config.rs
```diff
@@ -152,6 +152,8 @@ pub struct Config {
     #[serde(skip_serializing, skip_deserializing)] // not a typo, 2 are need
     pub extra_syntax_set: Option<SyntaxSet>,
 
+    pub link_checker_skip_anchor_prefixes: Vec<String>,
```
Keats: Sorry, I meant having the struct

```rust
#[derive(Clone, Debug, PartialEq)]
pub struct LinkChecker {
    /// Skip anchor checking for these URL prefixes
    pub skip_anchor_prefixes: Vec<String>,
}
```
in this file and use it in the config

components/link_checker/src/lib.rs
```diff
@@ -46,39 +46,56 @@ impl LinkResult {
...
+        LinkChecker { skip_anchor_prefixes: Vec::new() }
+    }
+}
+
 lazy_static! {
     // Keep history of link checks so a rebuild doesn't have to check again
     static ref LINKS: Arc<RwLock<HashMap<String, LinkResult>>> = Arc::new(RwLock::new(HashMap::new()));
 }
 
-pub fn check_url(url: &str) -> LinkResult {
```
Keats, line 54: this would mean that this stays a function that is `pub fn check_url(url: &str, config: &LinkChecker) -> LinkResult`. Adding new options to the LinkChecker makes it available automatically

Keats (changes requested):

components/link_checker/src/lib.rs
TjeuKayim: So create link_checker should depend on crate config?

components/config/src/config.rs
TjeuKayim: Oh, I misunderstood your wording then.

Keats:
Great, thanks!

Issue #814, opened 2019-10-07
`zola init` fails in existing empty directory 


demurgos:
# Bug Report

## Environment

Zola version: Github master branch
OS: Linux

## Expected Behavior
> Tell us what should have happened.

Running `zola init` in an empty directory should initialize the directory.

## Current Behavior
> Tell us what happens instead of the expected behavior. If you are seeing an
> error, please include the full error message and stack trace. You can get the
> stacktrace of a panic by adding `RUST_BACKTRACE=1` when running a `zola` command.

Example run:
```
$ RUST_BACKTRACE=1 ../zola/target/debug/zola init
Welcome to Zola!
Please answer a few questions to get started quickly.
Any choices made can be changed by modifying the `config.toml` file later.
> What is the URL of your site? (https://example.com): 
> Do you want to enable Sass compilation? [Y/n]: 
> Do you want to enable syntax highlighting? [y/N]: 
> Do you want to build a search index of the content? [y/N]: 
Failed to create the project
Error: File exists (os error 17)
```
As you see, the initialization fails with a "File exists" error. The error message could be improved to mention which file exists: it took some digging in the issues to figure out why it failed. (Running `zola init mydir` works, it's initializing the current empty dir that fails).

## Step to reproduce
> Please provide the steps to reproduce the issue.
> If the issue is hard to reproduce, please provide a sample repository or sample
> that triggers the bug.

Create an empty directory, go inside and run `zola init`.

## Related issues

- https://github.com/getzola/zola/pull/406
- https://github.com/getzola/zola/pull/777

Keats:
I should have tried that before releasing 🤦‍♂ 
In theory a very simple bugfix if anyone wants to try

Keats:
@riginding fixed it in the `next` branch 🎊

Jumanjii:
Awesome!
@Keats Do you know when will be the next release for newcomers to not meet this issue when launching Zola for the first time? :smiley:

Keats:
No timeline yet, I want to have the non slugification for paths (on phone,
can't find the issue) in before releasing. I'll try to work on that next
week.

ebkalderon:
@Keats Would it be possible to have this fix released sooner? Without it, the out-of-the-box user experience for new users looking to initialize a new Zola project inside an existing Git repo is very poor. Thanks for all the effort in creating and maintaining this project!

Keats:
[...]

PR #815, opened 2019-10-08
fix(init):  handle already existing path

riginding:
This should fix the bug mentioned in #814 .

I wanted to write test for it but I had some problems testing the create_project_function.
I think to test this properly I would need to use dependency injection and pass at_least input to `ask_url` and `ask_bool` function. I wanted to ask first if this is the prefered way to do it before I start changing more of the code.

Sanity check:

## Code changes


Keats:
The easiest way to test that would be to extract the actual creation into an another function that takes the result of the questions so it can be tested easily

Keats:
@riginding ping

riginding:
Sorry, for taking such a long time. I totally forgot that I opened this pull request.
I've now added the necessary tests.

Keats:
Perfect, thanks!

Issue #694, opened 2019-05-22
Render full section toc in page context


veldsla:
Absolute Zola beginner here.
In version 0.5 I used the following loop to generate a full TOC for the entire section on every page:
```html
{% set section = get_section(path="docs/_index.md") %}
{% for p in section.pages %}
<div class="bd-toc-item">
	<a class="bd-toc-link" href="{{ p.permalink }}">{{ p.title }}</a>
	<ul>
		{% for s in p.toc %}
		<li class="bd-sidenav">
			<a href="{{ s.permalink }}">{{ s.title }}</a>
		</li>
		{% endfor %}
	</ul>
</div>
{% endfor %}
```

But in 0.6+ only the toc for the current page is available as a global variable. Is there a way to generate the full table of contents in a page?

Keats:
> Is there a way to generate the full table of contents in a page?

No. I didn't think anyone would use the toc like that - I looked at quite a few sites/templates from various SSG before making the change but never saw a ToC outside of the page itself.

veldsla:
Ok no problem, I'm just starting anyway. I only have a few pages with a few items each so in my case it fit nicely. I'll rethink the design.

Thanks for the quick reply.

Keats:
Hmm I just realised when trying to update one of my site from Gutenberg to Zola that I was using almost exactly the sample you have in the issue 🤦‍♂  : the side bar of https://tera.netlify.com/docs/installation/ is made like that. I should have checked my own sites 🤦‍♂ 
That's a totally valid usecase, I assume yours is similar - I'll think about it.

maghoff:
I'm looking for this as well. In my case I get to page variables from a global context via `get_taxonomy`. I'd like to be able to present the subheadings at this level. Something like `get_toc(page)` would do the trick for me, but I don't know much about the constraints on the API design here.

(I'm attempting to rewrite https://magnushoff.com/articles/asmtut/ with Zola, dynamically populating the list from posts tagged #asmtut rather than manually updating the overview page)

Keats:
FYI this is done and will be in 0.10

PR #818, opened 2019-10-10
Let toc is visable through Page & Section variables in templates

nottxy:
Sanity check:

Yes

## Code changes

Yes

Yes

### DIFF
--- a/components/library/src/content/page.rs
+++ b/components/library/src/content/page.rs
@@ -290,7 +290,6 @@ impl Page {
         context.insert("current_path", &self.path);
         context.insert("page", &self.to_serialized(library));
         context.insert("lang", &self.lang);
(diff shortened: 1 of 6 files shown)


Keats:
Is that specifically for https://github.com/getzola/zola/issues/694 or did you need that yourself?

The PR itself looks good, although it might be worth it removing the current `toc` variable from page/section to avoid serializing it twice.

nottxy:
Yes, kind of some case like #694 .

It is better that the `toc` could be removed from page & section, but there would be a break change.

Keats:
> It is better that the toc could be removed from page & section, but there would be a break change.

That's fine, next version is planned to have breaking changes.

Keats:
Thanks!

PR #820, opened 2019-10-21
Improve init command

vctibor:
Command `zola init` fails if you don't provide folder name with non specific message "file exists". This adds message that should hopefully be little more helpful in what is missing.


Keats:
Thanks but that's a duplicate of #815

vctibor:
Ah, I missed that. Sorry about it.

PR #824, opened 2019-10-24
Do not attempt to serve on "well-known" ports

homeworkprod:
`zola serve` tries to find a port in the range 1000 and up.

Ports 0–1023 are known as [well-known ports](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports) and require root privileges on at Linux and likely macOS to bind a socket to them.

This (an excerpt from HTML served by `zola serve`) using port 1024 might be no coincidence:
port1024

Thus, just ignore those and start right after them (in the "registered ports" range).

### DIFF
--- a/components/utils/src/net.rs
+++ b/components/utils/src/net.rs
@@ -1,7 +1,9 @@
 use std::net::TcpListener;
 
 pub fn get_available_port(avoid: u16) -> Option<u16> {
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

PR #825, opened 2019-10-24
Properly escape `&` in injected live-reload script tag

homeworkprod:
Unescaped [Ampersands](https://en.wikipedia.org/wiki/Ampersand) used as query string separators in an URL must be escaped when that URL is used as value for an HTML (or XML, for that matter) attribute.

> […] `&` normally indicates the start of a character entity reference or numeric character reference; writing it as `&amp;` […] allows `&` to be included in the content of an element or in the value of an attribute.

From: https://en.wikipedia.org/wiki/HTML#Character_and_entity_references

And here's a screenshot (as seen in #824) in which the red coloring added by Firefox's source code viewer indicates an issue with the markup:
port1024

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -633,7 +633,7 @@ impl Site {
             return html.replace(
                 "</body>",
                 &format!(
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

PR #830, opened 2019-10-27
Add title and body to alias page

spk:
Sanity check:

## Code changes

This ensure the redirect page have a title and a body with a link in case the browser does not handle refresh redirects


Keats:
Hmm why? Which browser does not support the `http-equiv="refresh"`?

spk:
With firefox this can be disabled with ` accessibility.blockautorefresh = true` currently we have a blank page. A good practice is to have a title and body to show what we are trying to do and give more information to the user blocking auto refresh

Keats:
Ok, i'm guessing these don't need to be translated as it will be a very niche situation. Thanks!

PR #845, opened 2019-11-14
fixes #844 duplicate entries in doc

### LINKED ISSUE #844: Duplicate entries in Documentation for Syntax Highlighting
# Documentation issue

## Summary
The documentation for Syntax Highlighting contains conflicting/duplicate entries for TOML
https://www.getzola.org/documentation/content/syntax-highlighting/

Example : 
```
- TOML -> ["toml", "tml"]
...
- TOML -> ["toml", "tml", "Cargo.lock", "Gopkg.lock", "Pipfile"]
```

## Proposed solution
* I would keep only one entry for TOML syntax highlighting
* Another suggestion: reorder all the examples alphabetically to ease readability and detection of duplicates ?

SansGuidon:
fixes #844 
* Keeps only one entry for TOML syntax highlighting
* Reorders all the examples alphabetically to ease readability and detection of duplicates


Keats:
Hey, thanks for that but that list is actually generated by syntect automatically when rebuilding the syntaxes in https://github.com/getzola/zola/blob/master/components/config/examples/generate_sublime.rs#L30-L34 so it would need to be solved there

SansGuidon:
Thanks @Keats , I'll check that.

SansGuidon:
@Keats I've checked the code. The solution I came with is not what I really wanted, i.e it's a naive implementation which : 
* sorts the list 
* compare the entries for same language, and keep the one with the biggest list of file_extensions

I wanted instead to merge the duplicate entries, i.e detect when two of more entries refer to the same languages, and merge their file_extensions, but then I had to fight a battle vs Rust compiler. My understanding of this language is quite limited for now (had to learn Rust to work on this PR) while the compiler can be quite annoying / strict. 

Examples online were not always relevant or clear enough. So if you have examples of clean solutions for this, I would be glad to update the PR. Otherwise feel free to do so or close it / accept it depending on the requirements :)

components/config/examples/generate_sublime.rs
```diff
@@ -26,8 +29,24 @@ fn main() {
             builder.add_from_folder(package_dir, true).unwrap();
             let ss = builder.build();
             dump_to_file(&ss, packpath_newlines).unwrap();
-
-            for s in ss.syntaxes() {
+            let syntaxes_sorted = ss
```
Keats: you don't need itertools, you can just create a hashmap and append the extensions to a set and finally iter on the sorted hashmap

SansGuidon: I also tried to use HashMap, and it was not too difficult to iter on it nor sort it. 
The main difficult part for me about Rust syntax is really understanding of languages rules / syntax and I'm stuck at understanding how to create/update a file_extensions vector from existing ones.
And to be honest, I need a rest after hours of non succeeding in Rust so many times :) (without getting too much value from it -> I mean I still want to focus on my zola blog content :D ... way easier than succeeding at Rust)

SansGuidon: I came with a solution that do not need itertools and instead I'm making use of HashSet and BTreeMap (which ensures order, while HashMap was not helping here). I'll check how to fix the order, using something like sort_by, as for now the default order is not what I want when iterating on resulted collections. 

I'll push once I'm done with research

Keats: You don't really care about the insertion order like BTreeMap offers. A very easy way to do the sorting is to get all the keys of the hashmap into a vec, sort that and iterate on that vec grabbing the value from the hashmap. I can't remember if there is an easier way to do it.

Something like:

```
let mut keys = syntaxes.keys().collect::<Vec<_>>();
keys.sort();
for key in keys {
...
}
```

SansGuidon: @Keats I've applied the suggestions. As you will notice, I got rid of useless dependencies, however code might probably be improved.
That's the best I could produce for now with my basic knowledge of Rust, which I've only started learning about for this pull request :)

components/config/examples/generate_sublime.rs
```diff
@@ -26,10 +29,31 @@ fn main() {
             builder.add_from_folder(package_dir, true).unwrap();
             let ss = builder.build();
             dump_to_file(&ss, packpath_newlines).unwrap();
-
-            for s in ss.syntaxes() {
-                if !s.file_extensions.is_empty() {
-                    println!("- {} -> {:?}", s.name, s.file_extensions);
+            let mut syntaxes: HashMap<String, SyntaxReference> = HashMap::new();
```
Keats: I think Rust should be able to infer the type?

components/config/examples/generate_sublime.rs
```diff
@@ -26,10 +29,31 @@ fn main() {
...
+            for s in ss.syntaxes().iter() {
+                if !syntaxes.contains_key(&s.name) {
```
Keats: HashMaps have something called the entry api: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html

The code would look like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=10318e535286386fc4577c213b6fe0f1 if I had to write it I think

components/config/examples/generate_sublime.rs
SansGuidon: Indeed. Thank you !

components/config/examples/generate_sublime.rs
SansGuidon: Interesting :-) ! I'll use that definitely

Keats:
Thanks!

SansGuidon:
Thank you @keats 😁, this was a great opportunity to start learning about rust. Maybe not my last contribution.

Keats:
> Maybe not my last contribution.

I hope it isn't!

PR #850, opened 2019-11-23
Section extra -> SitemapEntry

### LINKED ISSUE #842: Sections in sitemaps don't get their [extra]
https://zola.discourse.group/t/sitemap-xml-no-extra-access-for-sections/240/2

Very easy to fix if someone wants to get a first try at Zola!

stanistan:
## General summary:

- Fixes #842 
- `Section` has parity with `Page` extra (using a `tera::Map` instead of `std::collections::Map`)
- `SitemapEntry` gets passed the `section.meta.extra`
- Added regression tests to make sure that the sitemap has the extra of the section
- unrelated formatting changes are because I ran `cargo fmt` on the project
- Question: There are some changes in modules structure (making `sitemap` `pub`) in order to make it easy to do a regression test there. Is there a better place for this?

## Code changes

### DIFF
--- a/components/front_matter/src/section.rs
+++ b/components/front_matter/src/section.rs
@@ -1,11 +1,9 @@
-use std::collections::HashMap;
-
-use tera::Value;
+use tera::{Map, Value};
 use toml;
 
-use errors::Result;
-
 use super::{InsertAnchor, SortBy};
+use errors::Result;
+use utils::de::fix_toml_dates;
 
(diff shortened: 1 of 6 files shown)


Keats:
> Question: There are some changes in modules structure (making sitemap pub) in order to make it easy to do a regression test there. Is there a better place for this?

In that specific case it's not super easy to test since you need a full library so it's fine making it public. This is an application after all, not a library.

PR #852, opened 2019-11-25
Add example site

trevordmiller:
- Add example site
- Update formatting to work with the new example which is now the longest string

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls)?

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1,28 +1,29 @@
 # Example sites
 
-| Site                                                               |                   Source Code                        |
-|:-------------------------------------------------------------------|:----------------------------------------------------:|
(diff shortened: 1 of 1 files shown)


Keats:
Can you remove the .swp and not change the other lines? I'm guessing vim is converting some space to tab or something

trevordmiller:
Yeah no problem on removing the .swp (my bad). The other lines were changed with the width of the table - since the new example is now the longest "cell".

trevordmiller:
@Keats Ok I fixed the .swp. I could remove the diffs from the other lines, but then the table will no longer be aligned

```
# Before
:----------------------------------------------------:|
# After
:--------------------------------------------------------:|
```

Keats:
Tthanks!

PR #855, opened 2019-11-26
Write files using `create_new` option to avoid accidental overwrites

### LINKED ISSUE #366: Page path collisions
While implementing section assets, i encountered an edge case where two pages try to render to the same path and one overwrites the other silently. I believe there should be some kind of runtime warning in such situations even though that should not happen often.

Otherwise, we should add in the content docs explanations about when collisions may happen (i'm guessing it's because of slugification?) and how to avoid them. Maybe it's already somewhere in the docs? I just didn't see ^^

Content tree:
```
└── content/
    ├── _index.md
    └── blog/
        ├── _index.md
        ├── capitalism.jpg
        ├── _index/
             ├── index.md
             └── capitalism.jpg ← (not the same one as in blog/, i was trying to force collisions)
        ├── index/
             ├── index.md
```

Here _index/index.md gets rendered to /blog/index and index does not get rendered (or at least not copied to public). Renaming either _index/ or index/ enables both to get rendered properly.

Debug info:

```
Parsing page from file "/home/user/Sites/Gutenberg/content/blog/_index/index.md"
Parsing page from file "/home/user/Sites/Gutenberg/content/blog/index/index.md"
Now pages:
[Ok(Page { file: FileInfo { path: "/home/user/Sites/Gutenberg/content/blog/_index/index.md", name: "index", relative: "blog/_index/index.md", parent: "/home/user/Sites/Gutenberg/content/blog", grand_parent: None, components: ["blog"] }, meta: PageFrontMatter { title: Some("Bloggity blog"), description: None, date: Some("2018-03-01"), draft: false, slug: None, path: None, taxonomies: {}, order: None, weight: None, aliases: [], template: None, in_search_index: true, extra: {} }, raw_content: "\n![](capitalism.jpg)\n", assets: ["/home/user/Sites/Gutenberg/content/blog/_index/capitalism.jpg"], content: "", slug: "index", path: "blog/index/", components: ["blog", "index"], permalink: "http://127.0.0.1:1111/blog/index/", summary: None, earlier: None, later: None, lighter: None, heavier: None, toc: [] }), Ok(Page { file: FileInfo { path: "/home/user/Sites/Gutenberg/content/blog/index/index.md", name: "index", relative: "blog/index/index.md", parent: "/home/user/Sites/Gutenberg/content/blog", grand_parent: None, components: ["blog"] }, meta: PageFrontMatter { title: Some("fucking page"), description: None, date: Some("2018-03-01"), draft: false, slug: None, path: None, taxonomies: {}, order: None, weight: None, aliases: [], template: None, in_search_index: true, extra: {} }, raw_content: "", assets: [], content: "", slug: "index", path: "blog/index/", components: ["blog", "index"], permalink: "http://127.0.0.1:1111/blog/index/", summary: None, earlier: None, later: None, lighter: None, heavier: None, toc: [] })
```

phil-opp:
The previously used `File::create` silently overwrites existing files, which causes problems e.g. when two pages accidentally specify the same `path`. By opening all files with the `create_new` option, we can ensure that no files are overwritten.

Fixes #366

### DIFF
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -1,4 +1,4 @@
-use std::fs::{copy, create_dir_all, read_dir, File};
+use std::fs::{copy, create_dir_all, read_dir, File, OpenOptions};
 use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::time::SystemTime;
@@ -19,8 +19,7 @@ pub fn is_path_in_directory(parent: &Path, path: &Path) -> Result<bool> {
 
 /// Create a file with the content given
 pub fn create_file(path: &Path, content: &str) -> Result<()> {
-    let mut file =
-        File::create(&path).map_err(|e| Error::chain(format!("Failed to create {:?}", path), e))?;
+    let mut file = OpenOptions::new().write(true).create_new(true).open(&path).map_err(|e| Error::chain(format!("Failed to create {:?}", path), e))?;
     file.write_all(content.as_bytes())?;
     Ok(())
 }


phil-opp:
This approach is a bit more far-reaching than only checking for path collisions, but I think it is a good sanity check. The error message that occurs for duplicate files is still ok, e.g.:

```
Error: Failed to create "public/freestanding-rust-binary/index.html"
Reason: File exists (os error 17)
```

Keats:
That works but I would prefer having a full report on why it is failing, eg which files are conflicting.
It would basically take the permalinks of each page/section + their alias and associate it to a Set<Key> and if the len(set) > 1, we have some duplicates that we can report with the exact file info.

phil-opp:
I understand. However I still think that the `create_new` option should be set to catch future mistakes (e.g. in other features). If there is any reason for Zola to overwrite files in the future, we can still provide an explicit way to do this. But currently, an overwritten file indicates a (configuration) bug so we should error in this case.

phil-opp:
To clarify: I think there should be both: A nice error report and a sanity check that no files are overwritten.

phil-opp:
Ok, from the CI tests it seems like there are a few cases where Zola does indeed overwrite files:

```
Err(Error { kind: Msg("Failed to create \"/tmp/.tmpvtbHLn/test_site_i18n/public/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) })
thread 'can_rebuild_after_editing_in_colocated_asset_folder_with_language' panicked at 'assertion failed: res.is_ok()', components/rebuild/tests/rebuild.rs:269:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- can_rebuild_after_deleting_file stdout ----
thread 'can_rebuild_after_deleting_file' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpQjzN5k/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_renaming_colocated_asset_folder stdout ----
thread 'can_rebuild_after_renaming_colocated_asset_folder' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpYAUV59/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_renaming_page stdout ----
thread 'can_rebuild_after_renaming_page' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpI4nlAs/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_renaming_non_md_asset_in_colocated_folder stdout ----
thread 'can_rebuild_after_renaming_non_md_asset_in_colocated_folder' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmp6gybnn/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_renaming_section_and_deleting_file stdout ----
thread 'can_rebuild_after_renaming_section_and_deleting_file' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmp3vozll/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_renaming_section_folder stdout ----
thread 'can_rebuild_after_renaming_section_folder' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpxWVPw2/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_simple_change_to_page_content stdout ----
thread 'can_rebuild_after_simple_change_to_page_content' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpATtdAu/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_sort_change_in_section stdout ----
thread 'can_rebuild_after_sort_change_in_section' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpwLg15l/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_transparent_change stdout ----
thread 'can_rebuild_after_transparent_change' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmpsfR6R1/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5

---- can_rebuild_after_title_change_page_global_func_usage stdout ----
thread 'can_rebuild_after_title_change_page_global_func_usage' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to create \"/tmp/.tmphmHRgy/test_site/public/posts/2018/transparent-page/index.html\""), source: Some(Os { code: 17, kind: AlreadyExists, message: "File exists" }) }', src/libcore/result.rs:1084:5
```

I guess we could fix those by deleting the existing content before recreating it (i.e. once before zola starts writing files). This would also ensure that we have no stale files left over from previous runs. Let me know if I should try something like this or if you prefer to close this PR!

Keats:
> However I still think that the create_new option should be set to catch future mistakes (e.g. in other features)

Very good point. But in that specific case of rebuilding, we do want to overwrite files otherwise `zola serve` doesn't work. We could use a different function in build and serve but that seems like a recipe for disaster.

phil-opp:
I see. Seems like this approach isn't the right on then.

PR #856, opened 2019-11-26
Compute canonical path before adjusting parent path

### LINKED ISSUE #854: page.translations contains unrelated pages
# Bug Report

I tried getting the translated versions of a page through:

```
{% for translation in page.translations %}
    {{ translation.title }} {{ translation.lang }}<br>
{% endfor %}
```

Even though the page has only a single translation, the array contains also contains completely unrelated (untranslated) pages. This makes it impossible to find the correct translation.

## Environment

Zola version: 0.9.0

## Expected Behavior
The `translations` array only has a single entry: The translated page. (Alternatively, it could also contain the default untranslated page for the default language.)

phil-opp:
Fixes #854

### DIFF
--- a/components/library/src/content/file_info.rs
+++ b/components/library/src/content/file_info.rs
@@ -56,6 +56,7 @@ impl FileInfo {
         let file_path = path.to_path_buf();
         let mut parent = file_path.parent().expect("Get parent of page").to_path_buf();
         let name = path.file_stem().unwrap().to_string_lossy().to_string();
+        let canonical = parent.join(&name);
         let mut components =
             find_content_components(&file_path.strip_prefix(base_path).unwrap_or(&file_path));
         let relative = if !components.is_empty() {
@@ -78,7 +79,7 @@ impl FileInfo {
             path: file_path,
             // We don't care about grand parent for pages
             grand_parent: None,
-            canonical: parent.join(&name),
+            canonical,
             parent,
             name,
             components,
--- a/components/library/src/content/file_info.rs
+++ b/components/library/src/content/file_info.rs
@@ -136,7 +136,7 @@ impl FileInfo {
         }
 
         self.name = parts.swap_remove(0);
-        self.canonical = self.parent.join(&self.name);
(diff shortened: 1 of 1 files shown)


phil-opp:
I think we also need to strip the language tag from the filename when generating the canonical path, otherwise translations are not found. Do you want me to add this change to this PR or do you prefer when I open a new one?

Keats:
I mean that's weird that it was not doing it, I swear I had code to handle that.. :(
Can you add some tests while you're there? At least a regression test with a link to the GH issue.

phil-opp:
> I mean that's weird that it was not doing it, I swear I had code to handle that.. :(

Maybe you mean this code? https://github.com/getzola/zola/blob/6d7c95e8cb6c60336119a9aeb53c1d30f1d5d2df/components/library/src/content/file_info.rs#L114-L143

This line seems to have the same problem:

https://github.com/getzola/zola/blob/6d7c95e8cb6c60336119a9aeb53c1d30f1d5d2df/components/library/src/content/file_info.rs#L138

You're using the adjusted `parent` field to create the canonical path again.

phil-opp:
I pushed a second commit that also fixes the issue in `find_language`.

phil-opp:
I added some basic regression tests for the two commits.

components/library/src/content/file_info.rs
```diff
@@ -254,4 +255,28 @@ mod tests {
...
+    fn correct_canonical_after_find_language() {
+        let mut config = Config::default();
+        config.languages.push(Language { code: String::from("fr"), rss: false, search: false });
+        let mut file = FileInfo::new_page(
+            &Path::new("/home/vincent/code/site/content/posts/tutorials/python/index.fr.md"),
+            &PathBuf::new(),
+        );
+        let res = file.find_language(&config);
+        assert!(res.is_ok());
+        assert_eq!(file.canonical, Path::new("/home/vincent/code/site/content/posts/tutorials/python/index"));
```
Keats, line 280: I think the reason for those bugs is that in my mind the canonical path was `/home/vincent/code/site/content/posts/tutorials/python`.
Anyway, thanks a lot!

PR #863, opened 2019-11-28
Add path to `TranslatedContent`

### LINKED ISSUE #859: How to access fields of translated pages from section?
I have a section page that prints the `summary` for each page. When translating the section, I want to print the translated summaries. However, the `TranslatedContent` structure as documented [here](https://www.getzola.org/documentation/templates/pages-sections/#translated-content) only gives access to the translated title, not any other fields of the translated page such as the summary. Is there a way to access these fields for translated pages?

phil-opp:
This makes it possible to retrieve the translated page through the `get_page` function.

Fixes #859

### DIFF
--- a/components/library/src/content/ser.rs
+++ b/components/library/src/content/ser.rs
@@ -1,5 +1,6 @@
 //! What we are sending to the templates when rendering them
 use std::collections::HashMap;
+use std::path::Path;
 
 use tera::{Map, Value};
 
@@ -12,6 +13,9 @@ pub struct TranslatedContent<'a> {
     lang: &'a str,
     permalink: &'a str,
     title: &'a Option<String>,
+    /// The path to the markdown file; useful for retrieving the full page through
+    /// the `get_page` function.
+    path: &'a Path,
 }
 
 impl<'a> TranslatedContent<'a> {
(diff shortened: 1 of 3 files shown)


components/library/src/content/ser.rs
```diff
@@ -12,6 +13,9 @@ pub struct TranslatedContent<'a> {
     lang: &'a str,
     permalink: &'a str,
     title: &'a Option<String>,
+    /// The path to the markdown file; useful for retrieving the full page through
+    /// the `get_page` function.
+    file_path: &'a Path,
```
Keats: it's called `path` in the other context so it should be renamed to `path`.

Keats (changes requested):

components/library/src/content/ser.rs
phil-opp: Done!

Keats:
Just for sanity check, did you test it locally to ensure it does what you need? I think we could add a call to get_page in https://github.com/getzola/zola/blob/master/test_site_i18n/templates/page.html and check it in the components/sites/tests/i18n files since there is no other easy to test it.
I should have mentioned the first time around, sorry :(

phil-opp:
Yes, I tested it locally.



>  I think we could add a call to get_page in https://github.com/getzola/zola/blob/master/test_site_i18n/templates/page.html and check it in the components/sites/tests/i18n files since there is no other easy to test it.

Done!

> I should have mentioned the first time around, sorry :(

No worries! It's definitely a good idea to add a test for this.

Keats:
Thanks!

PR #873, opened 2019-12-09
Update links to Tera documentation

Enet4:
I noticed that the links to the Tera docs are a bit different at the moment, probably since version 1.0 came out. This PR should make them work again.


Keats:
Thanks!

PR #882, opened 2019-12-16
tera to v1.0.0

erasin:
update tera to 1.0.0


Keats:
Thanks for that but that's already done in the `next` branch!

Issue #878, opened 2019-12-14
Consider adding tree-sitter support


remexre:
It'd be nice to have support for tree-sitter grammars; they can be generated from existing grammars used in parser generators, and it looks like [Atom currently supports it](https://github.blog/2018-10-31-atoms-new-parsing-system/) and [neovim's supporting it soon](https://github.com/neovim/neovim/pull/11113#issuecomment-559102311).

This probably can't yet replace the current `tmlanguage`-based system, but I'd find it personally useful /shrug. Maybe I'll finally get around to https://github.com/getzola/zola/issues/563#issuecomment-450591459 for a PoC :P

Keats:
What's the appeal of tree-sitter? It doesn't look like it supports semantic highlighting so the result is worse than what syntect offers.

remexre:
The main advantage to my eyes is that tree-sitter uses GLR parsing instead of the more ad hoc approach the textmate grammars use; this allows tree-sitter grammars to be generated from the grammars provided to parser generators.

Do you mean the sort of semantic highlighting like https://github.com/atom-archive/tree-sitter-syntax/pull/1 does? To be honest I don't see the utility of that for short code snippets, but it seems possible to do in the highlighter in a language-independent way, from that PR.

Keats:
That's not going to happen anytime soon, if ever.

Issue #563, opened 2018-12-21
Literate Programming Support


remexre:
It'd be nice to support literate programming in some form, something like:

    +++
    [literate]
    line = "--$"
    block = ["{-$", "$-}"]
    +++
    
    Blah blah this goes above the code
    
    ```haskell
    --$ First some convenience imports.
    import Foo (Bar(..), frob)
    import Baz (Quux(..))
    
    --$ One may think of frobbing the bars using quuxes; bazzing them is important
    --$ to do first, because of Frobenius' Collorary to Cooper's Theorem.
    performXyzzification :: Traversable t => t Quux -> Quux
    performXyzzification = undefined -- you get the point
    ```
    
    Blah blah this goes below the code

Ideally this would generate some HTML such that it'd be easy to write CSS that would either display the code and comments aligned side-by-side or stacked, alternating between code and prose (like [here](https://yager.io/CPU/CPU1.html)). I think it should be possible to do the appropriate styling with flexbox and `<div class="literate-code">` vs. `<div class="literate-prose">`, but I haven't tried it.

Keats:
That does look like it would need a completely custom markdown renderer :/

Probably unlikely to happen anytime soon

remexre:
Hm, I might try adding an "external command as a filter" and see if I can
implement this like that.

remexre:
okay, external command implementation is in as https://github.com/remexre/zola/commit/11ff0a12b8d58f24414aef0a5356e550eb9becc0; as noted in the commit message, testing for this is blocked on https://github.com/getzola/zola/issues/879.

does this look OK to you, though? (if so, I'll add docs+tests)

---


**EDIT**: Okay, I can build in an Alpine docker ctr, and it seems to work there.

Keats:
Not going to happen. I've looked at tree-sitter again recently but sadly still too few syntaxes and adding one is not trivial :(

PR #883, opened 2019-12-18
Adds external highlighters for code blocks.

remexre:
Sanity check:

TODOs:

* [ ] Rename to `external_renderer`
* [ ] Avoid files getting deleted ([ref](https://zola.discourse.group/t/230))

---

This is close enough to being a solution to #563 that I'd be fine with closing it if this were to be merged. I would also consider this to be a reasonable temporary solution for #878 (though I do think that in the long term, `tree-sitter` would be preferable to `syntect`). This provides GraphViz support easily (ref https://zola.discourse.group/t/269), and should provide PlantUML support too (though I haven't used that tool, so I can't say for certain).

I'd like to be able to get this working with LaTeX with all the proper fallbacks, so MathML falls back to MathJax falls back to a static image. Unfortunately, the static image bit proves to be the problem -- see below / the commit message.

---

This feature is activated by specifying shell commands in config.toml.
For example, an external "highlighter" that renders GraphViz to SVG
could be defined as:

    # ...
    [external_highlighters]
    dot = "echo -n '<p>'; dot -Tsvg | tail -n +4; echo '</p>'"
    # ...

A code block like:

    ```dot
    digraph G {
	A -> B;
	B -> C;
	A -> C;
    }
    ```

then results in the HTML output by piping the contents of the code block
to the command specified as external_highlighters.dot.

The following environment variables are passed to external highlighter
commands:

- `ZOLA_BASE_URL`: The base URL in the config
- `ZOLA_CURRENT_PAGE`: The permalink URL to the current page
- `ZOLA_OUTPUT_DIR`: The directory being output to

The last one would be more useful if the output directory weren't
cleared before the content is translated from Markdown (ref
https://zola.discourse.group/t/230). This would be particularly useful
for e.g. generating images from LaTeX, and replacing the code blocks
with `` tags.


uwearzt:
Great feature, better name it `external_renderer`. I would like to add https://plantuml.com/ .
Are you still working on this?

remexre:
> Great feature, better name it `external_renderer`.

Thanks! And `renderer` does seem like a more accurate name; I'll add that as a TODO up top, pending a broader response; I'm more or less waiting for a :+1:/:-1: response to https://zola.discourse.group/t/external-highlighters-for-code-blocks/283 before I mark this "ready to review."

> I would like to add plantuml.com .

I don't believe PlantUML support should require code changes to this pull request -- it should be possible to add a block like:

```
[external_highlighters]
plantuml = "java -jar /path/to/plantuml.jar -p -tsvg"
```

to your `config.toml`. I don't actually use PlantUML (and don't have a working JVM at the moment), so this may need adjustment.

> Are you still working on this?

This is still maintained, and I intend for it to be merged. I'm using this branch for my own site, though I probably won't rebase it until it's ready to be merged or the next release is ready.

remexre:
ping @Keats -- thoughts?

Keats:
As mentioned in https://zola.discourse.group/t/external-highlighters-for-code-blocks/283 I'm worried how this might conflict with improvements to the default syntax highlighter so this is unlikely to be considered anytime soon in this form.

remexre:
I don't really see how it'd conflict; other than overloading the same Markdown syntax, the two should be wholly orthogonal. You can see [in the (updated) patch](https://github.com/remexre/zola/commit/944d7445ca5766e0398ecf754c1d7f2940216869#diff-60b9e60d90d18c3f13add48d803d1bb3) that I override the entirety of the code-block syntax for the languages specified -- any changes to how languages render by default should have no effect.

(Though I am realizing now that the `context.config.highlight_code` tests should be after the corresponding tests this patch adds.)

Keats:
Ah sorry I thought I mentioned it somewhere but it doesn't look that way. Adding https://github.com/ivanceras/svgbob is on the wishlist of a few people and it would use the ```bob approach as well but in that case it wouldn't an external highlighter nor a highlighted syntax which would mean 3 different things done by the same markdown syntax.

I don't know what would be the best way to solve that, I think a shortcode is neater because you can pass arguments if wanted rather than defining a new external tool for each set of argument. You do lose syntax highlighting in the editor though :/

remexre:
Hm, it feels like a more general mechanism might be nice in the code... From the user side, the overloading of the block syntax doesn't seem significantly more complex, especially since it needs to be enabled in a config file (so it won't mysteriously be active).

wrt syntax highlighting -- literate programming tools, markdown autoformatters, etc. also won't recognize shortcodes as code blocks unless they're programmed to, which is a big workflow breaker for me.

Issue #871, opened 2019-12-08
zola should warn/error on empty links


Keats:
When writing a post about something that isn't live yet, I typically write something like `[this site]()` and forget to fill the link 100% of the time.

This kind of link is pretty much never wanted so it is likely it should be an error, even in `zola build`. This forces people to actually put something (even `TODO`) that gets picked up by `zola check`.

It could also be only a warning but we do not really have this kind of content warning right now. Is there anyone using `[..]()` by choice?

zbrox:
I quickly looked into this. I wasn't sure what the preferred approach would be or how to return a warning per se. 

Would something like this be ok?

https://github.com/zbrox/zola/commit/e17bfff10a8941be6a4d0b09c39a3a0de58199e8

aslowwriter:
I see there is a merged PR for this issue, should it still be open? if so what still needs work?

Keats:
That should be fixed yes

PR #884, opened 2019-12-19
Detect empty links on markdown rendering and issue an error

zbrox:
Referencing #871

Sanity check:

## Code changes

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -242,6 +242,10 @@ pub fn markdown_to_html(content: &str, context: &RenderContext) -> Result<Render
 
                         Event::Start(Tag::Image(link_type, src, title))
                     }
+                    Event::Start(Tag::Link(link_type, link, title)) if link.is_empty() => {
+                        error = Some(Error::msg("There is a link that is missing a URL"));
+                        Event::Start(Tag::Link(link_type, "#".into(), title))
+                    }
                     Event::Start(Tag::Link(link_type, link, title)) => {
                         let fixed_link = match fix_link(
                             link_type,
--- a/components/rendering/tests/markdown.rs
+++ b/components/rendering/tests/markdown.rs
@@ -852,3 +852,24 @@ fn leaves_custom_url_scheme_untouched() {
 
     assert_eq!(res.body, expected);
 }
+
+#[test]
+fn stops_with_an_error_on_an_empty_link() {
+    let content = r#"[some link]()"#;
+
+    let tera_ctx = Tera::default();
+    let config = Config::default();
+    let permalinks_ctx = HashMap::new();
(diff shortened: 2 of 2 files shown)


components/rendering/src/markdown.rs
```diff
@@ -242,6 +242,10 @@ pub fn markdown_to_html(content: &str, context: &RenderContext) -> Result<Render
 
                         Event::Start(Tag::Image(link_type, src, title))
                     }
+                    Event::Start(Tag::Link(link_type, link, title)) if link.is_empty() => {
+                        error = Some(Error::msg("There is a link that is missing a URL"));
```
Keats, line 246: Can we add the link type/title to the error message?

zbrox: I actually had it with the title in the message initially but for some reason the `title` variable there is an empty string. I didn't have more time to look into it. My initial guess was that maybe the Markdown parser is not properly parsing a link without a URL.

Keats: ah interesting, I'll see if I can reproduce it and open an issue on pulldown-cmark

Keats:
A bit late but can you add a test for that?

zbrox:
> A bit late but can you add a test for that?

Yeah, sure. I'll look into that.

components/rendering/src/markdown.rs
zbrox: @Keats did you manage to reproduce it? Is it really an issue or did I misunderstand something?

components/rendering/tests/markdown.rs
```diff
@@ -852,3 +852,24 @@ fn leaves_custom_url_scheme_untouched() {
...
+        &config,
+        "https://vincent.is/",
+        &permalinks_ctx,
+        InsertAnchor::None,
+    );
+
+    let res = render_content(content, &context);
+
+    assert!(res.is_err());
+}
```
zbrox, line 878: @Keats I was not terribly creative with the test. Should I test for something else other than erring out?

components/rendering/src/markdown.rs
Keats: I haven't tried yet, I'll check when I have time

components/rendering/tests/markdown.rs
Keats: You can check that the description contains the expected string

Keats:
Thanks!

components/rendering/src/markdown.rs
Keats: Ah nevermind, `title` is `[I'm an inline-style link with title](https://www.google.com "Google's Homepage")` Google's Homepage. It looks like you don't get the link text in the event

Issue #782, opened 2019-08-22
Try using hyper instead of actix for webserver


Keats:
To see if it helps in terms of compilation time/number of deps (on `next` branch)

See https://www.reddit.com/r/rust/comments/ctlt16/thoughts_on_rust_bloat/exnrn0i/

southerntofu:
If we're worrying about compilation time, dependencies and final binary size, how about using [feature flags](https://www.worthe-it.co.za/programming/2018/11/18/compile-time-feature-flags-in-rust.html) so users can compile only the features they need? It could look like [this](https://github.com/southerntofu/zola/tree/serve_feature_flag).

The default feature flags for the crate can include all features so users running `cargo install zola` would get all features out-of-the-box. `cargo build --release --no-default-features` opts out of optional features, passing an additional --features flag reenables individual features.

Other possible candidates for feature flags:

- zola check
- SASS support
- elasticlunr search index
- image processing

Keats:
That's not really the issue, zola is not distributed on crates.io so the speedup is only for people working on the code itself or for package managers building from scratch and in that case features would be more annoying than anything since they would all be enabled everytime.

ghost:
As mentioned in that reddit thread, is using [cobalt's serve.rs](https://github.com/cobalt-org/cobalt.rs/blob/master/src/bin/cobalt/serve.rs) an option? Not too familiar with how zola has been doing static serving but cobalt's solution seems easily extendable/modifiable.

Keats:
That's probably the first thing to try yes. That and tiny-http which might be ok too

samford:
I've been thinking about this exact topic for some time (after upgrading Zola's Actix version to 1.0) and had intended to create an issue to discuss it but I wanted to create a proof of concept first and never found the time.  It's good to see that the idea's getting some traction.

My thought was that Zola already includes Hyper because the reqwest crate is used in the `link_checker` component (which depends on Hyper), so switching from Actix seems like it would be clearly beneficial with respect to dependency count, compile time, and binary size.  The downside is that you generally have to write more lines of code with Hyper compared to Actix (e.g., static file serving in Actix is basically one line but you have to implement it yourself with Hyper, as seen in Cobalt's serve.rs).

[...]

PR #889, opened 2019-12-20
chore: Update actix to 2.0

igor-raits:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


Keats:
There is actually someone working on removing actix from Zola and using hyper directly since we already use reqwest to lower the dependency count. Thanks for that anyway!

igor-raits:
@Keats do you have some link to that work?

Keats:
https://github.com/getzola/zola/issues/782#issuecomment-565773815

PR #910, opened 2020-01-03
Further refine the slugification story

### LINKED ISSUE #768: Underscores in slugs get converted to hyphen-minus and other slugification issues

chris-morgan:
https://github.com/getzola/zola/issues/768#issuecomment-570557713 is the discussion and source of what I’ve done in this commit.

The work included in here:

- Stop stripping any characters from paths or anchors: none of the restrictions are fundamentally valid, and they prevent realistic uses.

- Split slugify_paths into slugify_paths and slugify_anchors, and change anchors to using slugify_anchors.

- Turn slugify_paths off by default, and ensure that all the documentation is consistent about this (it wasn’t).

One potentially controversial part of this is ceasing turning spaces into underscores in anchors with slugify_anchors off. I consider that a legitimate slugification scheme, and contrary to the declared intent of `slugify_anchors = false`. I think it’s best left to the future where we can specify `slugify_anchors = "Spaces_to_underscores"` or similar. (Because there are further variants of it.)

Fixes #768.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 ### Breaking
 - Remove `toc` variable in section/page context and pass it to `page.toc` and `section.toc` instead so they are
 accessible everywhere
-- [Slugification](https://en.wikipedia.org/wiki/Slug_(web_publishing)#Slug) of page paths is now optional. By default, every path will be slugified as it is happening right now.
-To keep non-ASCII characters,  set `slugify_paths = true` in your config.
+- [Slugification](https://en.wikipedia.org/wiki/Slug_(web_publishing)#Slug) of page paths is no longer performed by default.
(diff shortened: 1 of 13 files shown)


chris-morgan:
I've decided to reinstate the space → underscore translation on unslugified anchors (and done it better), and have split taxonomy terms out of slugify_paths, on realising that they’re quite different too, and *should* be enabled by default. Tests all pass this time as well.

Keats:
Closing in favour of #929

PR #926, opened 2020-01-18
Add blog.turbo.fish to EXAMPLES.md

jplatte:
Hi! Heard you were interested in more examples of Zola usage :)

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -32,3 +32,4 @@
 | [arewegameyet.rs](http://arewegameyet.rs)                          | https://github.com/rust-gamedev/arewegameyet             |
 | [klau.si](https://klau.si)                                         | https://github.com/klausi/klau.si                        |
 | [peterlyons.com](https://peterlyons.com)                           | https://github.com/focusaurus/peterlyons.com-zola        |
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

FYI, your fonts are pretty slow to load on a bad airport wifi, ~5s during which the content is not visible :/

jplatte:
Thanks for letting me know. I have a fallback to `sans-serif`, but I guess that doesn't apply on your system then? I'll investigate how to do improve fallback.

jplatte:
On my system, I also see the font change from the sans serif font I have set in Firefox's preferences, to Inter during page load. And as far as I understand <https://developer.mozilla.org/en-US/docs/Web/CSS/font-family>, `sans-serif` (as well as `serif`, `monospace`, ...) should always be present. Do you have an idea what the reason may be that it isn't working for you / how I could improve things?

Keats:
Loading fonts via `url` is blocking so it will be invisible while downloading the files afaik. Try to put a `font-display: swap` in the font-face declaration so it doesn't replace the sans-serif while loading the custom fontt

jplatte:
Done!

PR #928, opened 2020-01-20
Rewrite link_checker to use a Result internally

awonnacott:
`link_checker::LinkResult` always had exactly one member option set, so it should be an `enum`, not a `struct`. Instead of writing our own `enum` to represent sucess/failure, we can use `std::result::Result`.

### DIFF
--- a/components/link_checker/src/lib.rs
+++ b/components/link_checker/src/lib.rs
@@ -3,50 +3,33 @@ use reqwest::header::{HeaderMap, ACCEPT};
 use reqwest::{blocking::Client, StatusCode};
 
 use config::LinkChecker;
-use errors::Result;
 
 use std::collections::HashMap;
+use std::result;
 use std::sync::{Arc, RwLock};
 
-#[derive(Clone, Debug, PartialEq)]
-pub struct LinkResult {
-    pub code: Option<StatusCode>,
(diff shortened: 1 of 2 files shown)


Keats:
What's the benefit if we are not using `?`?

awonnacott:
It's not about using `Result` in particular, it's about how the data you are representing (EITHER `code` OR `error`) should be represented with an enum, whereas a struct "means" `code` AND `error`. If you use a struct to store two values and only one of them is set, you

- open up a class of errors which are impossible with an enum: using an uninitialized value or forgetting to check if error is set
- waste space, since your struct takes up `sizeof(code) + sizeof(error)` memory instead of `max(sizeof(code), sizeof(error))`.

Using types whose value sets correspond directly to the desired representable values is one of the big advantages of functional programming, and having well-designed enums is among the advantages Rust has over Go or C++. Consider how useful it is that existing Rust libraries return Results instead of Go-style pairs of value and error.

Keats:
Can you cherry-pick your commit on top of next and push again?

awonnacott:
@Keats done

awonnacott:
hmm, tests passed on my machine, will investigate

Keats:
Errors seem legit

Keats:
I'll fix the error on the next branch directly; thanks!

Issue #519, opened 2018-11-15
Asset fingerprinting and SRI


RealOrangeOne:
It'd be nice if there were a way of accessing a fingerprint or hash of static files, for a more reliable way of caching assets. 

I've given this a bit of thought, and probably need some input.

Ideally, it'd just be an additional function, say `get_static`, which returned a structure which contained the path, fingerprinted URL, hash, things like that. Because this requires the static files to exist before running, it would also serve as a link checker. However, What do we do about live reloading? When editing a static file, we'd either need to keep track of the linked resources per page, and only reload the linked pages, or just reload the whole site. (The latter is _so_ much easier!)

I see no reason why this couldnt work for any static file in the site, eg images. It wouldnt play well with images after they've been processed, but static images should work fine!

Keats:
The current timestamp is not a very good solution I agree.
I would probably just change the `cachebust` attribute of `get_url` to hash it though rather than having a new function

RealOrangeOne:
That would work for fingerprinting the URL, but wouldnt generate the required SRI data. I could change the return type to an object, but not sure I want to make a breaking change to something so fundamental to existing sites. Hence the new function.

If you're happy for me to add a breaking change, I can do that too. Or, I could just remove the `cachebust` argument to `get_url`, so people have to use `get_static` when they want their cache busted

Keats:
Do you have a link for SRI?

RealOrangeOne:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity is a pretty good one. Or the spec at https://w3c.github.io/webappsec-subresource-integrity/, which is incredibly dry!

dancek:
Is there still interest for this? I'd like hash-based cache busting. SRI is nice in theory too, but practically only useful for external resources (which zola couldn't hash during compilation).

tcurdt:
IMO this is still relevant.

Keats:
I would like a PR for that.

dancek:
I'll try to make a PR within the next couple of days.

mattly:
You know, I got all excited because I saw some activity on this ticket and punted on doing some work to my hugo site because I thought, yay, they're tackling this finally! Hugo's template system is just as insane as golang, and migrating my fledgling site would be a worthwhile yak shave.

[...]

Issue #316, opened 2018-06-09
Cannot compile with musl as target


nbigaouette:
I'm trying to build a docker image providing a statically linked musl binary of gutenberg (v0.3.3). This way the final docker image could contain only the static binary by using `FROM scratch`.

Here's the dockerfile:
```dockerfile
ARG RUSTC_VERSION=1.26.2

FROM rust:${RUSTC_VERSION}-slim-stretch as builder

ENV GUTENBERG_VERSION 0.3.3

RUN apt-get update -y && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y -q \
    wget \
    tar \
    cmake \
    build-essential \
    musl-dev \
    musl-tools \
    libpq-dev \
    libsqlite-dev \
    libssl-dev \
    pkgconf \
    xutils-dev \
    && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

RUN wget https://github.com/Keats/gutenberg/archive/v${GUTENBERG_VERSION}.tar.gz && \
    tar -zxvf v${GUTENBERG_VERSION}.tar.gz && \
    rm -f v${GUTENBERG_VERSION}.tar.gz

RUN rustup target add x86_64-unknown-linux-musl

WORKDIR gutenberg-${GUTENBERG_VERSION}

ENV RUST_BACKTRACE=1

RUN cargo build --release

FROM rust:${RUSTC_VERSION}-slim-stretch as runner
COPY --from=builder /gutenberg-${GUTENBERG_VERSION}/target/release/gutenberg .
CMD ["./gutenberg", "serve"]
```

Built and ran like this: `docker build . -t gutenberg:v0.3.3 && docker run -it --rm gutenberg:v0.3.3`

Right now it does not work. I've facing some issues.

First issue is the following:
```
error[E0518]: attribute should be applied to function
  --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/tera-0.11.6/src/parser/mod.rs:13:10
   |
13 | #[derive(Parser)]
   |          ^^^^^^ not a function
```
I'm not sure to understand this error. I was able to bypass it by running `cargo update`. Here's the resulting lock file: [Cargo.lock.txt](https://github.com/Keats/gutenberg/files/2086562/Cargo.lock.txt)
Could a release be done with updated dependencies? I guess that should fix this problem.

Secondly, I get a linking error if I try to link to musl (`cargo build --release --target=x86_64-unknown-linux-musl`):
```
[...]

PR #934, opened 2020-01-24
Add feature flags to reduce dependencies

traviscross:
We have 223 unique external direct and transitive dependencies.  Many
of these are needed only for serving files via HTTP.  Others are
needed only for making outbound HTTP/S requests.  In certain
situations, such as when building Zola as part of a continuous
integration pipeline, these features are unnecessary, and we would
prefer a faster and simpler build.

In this commit, we add `serve` and `request` feature flags to Zola.
The `serve` flag controls whether the embedded HTTP server is built
and whether the `zola serve` command is available.  The `request` flag
controls whether we support making outbound HTTP/S requests, such as
for link checking.  Both of these features are enabled by default.

With both of these features disabled, such as when building with
`--no-default-features`, we require only 160 unique external
dependencies, a 28% reduction.  Notably, we're able to drop both
`openssl` and `tokio`.  Dropping `openssl` in particular can help with
producing statically-linked builds.

With the `serve` flag, we depend on 210 unique external crates.  With
the `request` flag, we depend on 199 unique external crates.


traviscross:
In this commit, we have updated the installation documentation to reflect the available feature flags and their use.

We added these flags as part of work to support building Zola statically, so this has some relevance to issue #316.  However, the benefit of having these feature flags likely stands on its own.

traviscross:
We have ensured that all tests pass on Zola and on all components in the workspace in all feature configurations.  Further, we've tested building, serving, and checking the documentation in each configuration where the operation is relevant.

Keats:
I'm not super keen on that to be honest. I think features for Zola are only useful for some developers to avoid potential build issues, like sass-rs which is hard to build on some systems. The rest will always be built with all features enabled so there's no point splitting them.

Imo, a feature for sass (enabled by default) should be added but that's it.

traviscross:
Hi @Keats; thanks for sharing your concerns.  Understand and agree that most people building or using this on their local systems will want all features enabled (I do too!).  The use-case here is that we're building Zola to be run as part of a continuous integration system for building a site.  The embedded HTTP server and link checker don't serve a purpose in that application.  They are, in this application, actively unwanted because we don't want our CI systems accepting inbound or making outbound network requests in any case.

So even aside from reducing dependencies, that's worth it to us and probably to others using it in a similar manner.

On the dependency side, this has two main positive effects:

- Removing OpenSSL from the dependency tree makes Zola easier to build reliably and in a statically-linked manner, which is particularly important for CI.
- Removing the things that make or receive network requests and OpenSSL reduce Zola's attack surface and make security review and vulnerability tracking easier for the resulting binary, which is important for some people.

This patch is also a stepping stone to getting Zola building statically in its full configuration.  Testing and working through the problems was made easier by this.  As you mention, sass-rs is the other tricky thing to build.  (I didn't carve that one out because it seemed more central to Zola's functionality than the embedded server or link checker.)

I'm expecting to submit further PRs to sass-rs and to Zola to get Zola building statically in its full configuration.

If you start to see the value after giving some thought to the use-case and benefits I've described, great.  (I recognize the appeal may be to a niche group.)  If not, let me know, and I'll close the PR, and we'll just maintain this patch out of tree.  If you want, I'm also happy to submit a PR with a feature flag for sass, or to roll it into this PR if you think that would be better.

Keats:
Why build Zola from source everytime on CI though? It takes a good 5-10min on release mode vs grabbing it from one of the many package managers or just wget from the release page.

I think the only non-static part is reqwest right? I wanted to use the rustls feature (https://zola.discourse.group/t/remove-dependency-on-preinstalled-openssl/221) to get rid of openssl but it seems like some sites do not work with it; it might be ok for a SSG though?

traviscross:
Hi @Keats; we don't build it every time we build the site.  We build it whenever our Zola branch changes and cache the built artifact.  That cached artifact is used to build the site.  As far as why we build it ourselves rather than using a distributed binary, the immediate motivation is that we're testing in-progress improvements to Zola, such as work toward a solution to #519.  We'll be submitting those patches upstream as they become ready for your review.

Also, while our use-case is CI, and I focused a lot on that above, I think anyone running Zola on a server in any capacity would want to strip out at least the embedded HTTP server for the (at least perceived) security benefits mentioned.  Sometimes it can be useful to trust that Zola (or any binary) is providing only a strict input -> output function.

Regarding static builds, sass-rs' dependency on libstdc++ is perhaps the bigger challenge than OpenSSL.  That precludes the simple use the `musl-gcc`, which Rust uses by default when building statically with musl.  I have managed to get Zola to build statically using `musl-cross-make` and some changes to Zola/sass-rs.

To get there, I found it useful to get OpenSSL out of the way, hence the origin story of this PR.  However, I actually expect that getting Zola to build statically with OpenSSL won't be a big deal.  We'll add a feature flag to Zola that will cause `openssl-sys` to build its vendored version of OpenSSL, and that will need to be passed when building statically.  As you say, using rustls could be another option.  It may be worth adding feature flags to select which is used (there are pros and cons to each).  I can include that with the PRs I'll be submitting for the remainder of the static build work.

traviscross:
Also, it may be worth mentioning, most of the `Cargo.toml` plumbing in this PR is groundwork that would be needed anyway for plumbing openssl or rustls feature flags to where they are needed.

traviscross:
Apropos of the issue raised in the discourse group, the correct end result for Zola's own binary releases is probably to build them statically against OpenSSL, and perhaps against rustls when it's a bit more mature.  That would avoid not just the conflict that user hit, but any conflicts with system libraries.  It would also allow the Zola releases to be installed on non-gnu hosts such as Alpine Linux (which is a favorite of people using Docker... such as for CI).

Of course, that doesn't mean that the vendored OpenSSL needs to be the default for `cargo build`.  That choice can be made separately, and it may be reasonable to use the system OpenSSL by default there.

(Actually, perhaps the ideal would be if the `openssl-sys` crate were to automatically fall back to its vendored copy when `libssl-dev` isn't available, like `sass-sys` does with `libsass-dev`.)

Keats:
Thanks for the background info.

For rustls I think it's mature right now, it's just that it intentionally doesn't support every algorithms on purpose, which means it might not work with some sites. 
The main issue I have with the reqwest  feature gating is that we changes the `load_data` function. `reqwest` is only used for `zola check` (a command that you *probably* want to run on CI, with HTTP request to check external links) and that function which both becomes something very different with or without the flag on.

I was planning to add the sass-rs feature for the next Zola version since it's such a pain to build at times and prevent people from contributing.

Rather than adding a feature gate for reqwest, I am more thinking of trying to switch it to use rustls and see if people report unexpected failures; we get rid of openssl entirely and maybe can help improve rustls with issues/bugs that way. If too many people report issues, we can back down and look again at switching to the static openssl version. I don't think I want to make that choice selectable, if someone *really* needs another version of reqwest, they can fork it and it's a few chars diff.

traviscross:
Thanks @Keats for the further thoughts.  You're right about `load_data`.  I hadn't fully considered what that function does with respect to outbound requests (or else I would have at least documented the changes there).

With respect to rustls and sass, that makes sense.  I'm working up a commit series that switches to rustls, includes the sass feature flag, and rethinks the request feature a bit.

Keats:
Closing that for now as it has way too many conflicts and is better to start from scratch.

Keats:
@traviscross do you have an idea when/if you'll get to that?

PR #935, opened 2020-01-25
Correct borrow errors found running cargo install.

mysteriouspants:
Sanity check:

It didn't look like there were any open issues/PRs for this particular problem I had.

## Code changes
I ran `cargo install --path .` on zola's master branch on Ubuntu WSL and got errors. These changes correct the errors. I don't understand why `cargo build --release` worked when this didn't.

The command I ran and its output are here: https://gist.github.com/mysteriouspants/3f27d577e275ac973b0818da8c1d2f5e (baseline)  
After my changes: https://gist.github.com/mysteriouspants/62428af4f84fc138d7d1fde4afe95c50

### DIFF
--- a/components/templates/src/lib.rs
+++ b/components/templates/src/lib.rs
@@ -69,6 +69,6 @@ pub fn render_redirect_template(url: &str, tera: &Tera) -> Result<String> {
     let mut context = Context::new();
     context.insert("url", &url);
 
-    tera.render("internal/alias.html", context)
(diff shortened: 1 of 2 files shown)


mysteriouspants:
Sorry, this change failed in the CI runners. I took another look locally and it causes `cargo build --release` to fail in the same way. I'm investigating the discrepency between the output of `cargo build --release` and `cargo install --path .` as a non-Zola specific issue now.

PR #936, opened 2020-01-25
Fix warning when building site tests

traviscross:
For the site integration tests, we have a file of common code which is
used by multiple files in `tests/`.  However, not all functions in
this file are used by all files in `tests/`.

As Cargo compiles each `tests/*.rs` file as a separate crate, this
means that some of these crates end up with unused code.  Rust notices
this and prints a warning.

Let's tell Rust that we don't care about dead code in this file so
that the warning is not printed.


Keats:
Thanks; that annoyed me greatly!

Issue #938, opened 2020-01-28
balthek/zola docker image is not updated


nrjais:
# Bug Report
The image in the balthek/zola repository is 7 months old with version 0.7.0

## Expected Behavior
Updated docker image with version 0.9.0

rask:
The Docker Hub page states that the image source repository is in fact this repository, but I see no official mention anywhere regarding the Docker image?

nrjais:
Yes the source repository is this repository but there is only one image available which is 7 months old for zola version 0.7.0 which is marked as latest.
I think it is not rebuilding the image whenever there is a new release of zola.

rask:
Yeah, I see no reference to image building in the Azure Pipelines definition.

nrjais:
I think you can also enable docker automated build to automatically build and publish images as soon as a version is released on github.

Below configuration can be used to use the Github release tag as the image tag for every release of zola
<img width="1151" alt="Screenshot 2020-01-28 at 7 23 12 PM" src="https://user-images.githubusercontent.com/20153307/73269983-d723bc80-4203-11ea-89b2-b087de660586.png">

rask:
That screenshot is from Docker Hub?

nrjais:
@rask Yes from automatic build configuration

Keats:
Looks like a bug in the pipeline: 

```
Step 6/11 : RUN curl -L https://github.com/getzola/zola/releases/download/$DOCKER_TAG/zola-$DOCKER_TAG-x86_64-unknown-linux-gnu.tar.gz | tar xz
---> Running in 4e65b0453f54
[91m % Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0[0m
[91m 100 9 0 9 0 0 64 0 --:--:-- --:--:-- --:--:-- 64
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
[0m
```
It seems the DOCKER_TAG variable is not there or something?

nrjais:
@Keats The existing Dockerfile will not work as it uses hardcoded version number.
I can raise a PR with the updated Dockerfile where version is not hardcoded.

Keats:
Please do!

nrjais:
@Keats I have raised a pull with the updated dockerfile can try building again with these changes.

thomasetter:
[...]

PR #939, opened 2020-01-29
Update dockerfile to remove hardcoded version number

nrjais:
This updates the dockerfile so new images can be created when a release is made(issue #938).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,18 @@
-FROM bitnami/minideb AS builder
-RUN install_packages python-pip curl tar python-setuptools rsync binutils
-RUN pip install dockerize
+FROM rust:slim AS builder
+
+RUN apt-get update -y && \
+  apt-get install -y python-pip make g++ python-setuptools libssl-dev pkg-config rsync && \
+  pip install dockerize && \
+  rustup target add x86_64-unknown-linux-gnu
+
+WORKDIR /app
+COPY . .
+
+RUN cargo build --release --target x86_64-unknown-linux-gnu
+
+RUN mv target/x86_64-unknown-linux-gnu/release/zola /usr/bin
 RUN mkdir -p /workdir
 WORKDIR /workdir
-RUN curl -L https://github.com/getzola/zola/releases/download/$DOCKER_TAG/zola-$DOCKER_TAG-x86_64-unknown-linux-gnu.tar.gz | tar xz
-RUN mv zola /usr/bin
 RUN dockerize -n  -o /workdir  /usr/bin/zola
 
 


Keats:
The weird part is that the Dockerfile did work for 0.8 but not 0.9...
Would it make sense to do a multi-stage Dockerfile so the end image doesn't have rustup/cargo installed?

nrjais:
It is already a multi stage Dockerfile

Keats:
🤦‍♂ I should have expanded the bottom

nrjais:
> The weird part is that the Dockerfile did work for 0.8 but not 0.9...

Which Dockerfile are you talking about? I have tested the changes with master branch which is working fine.

Keats:
> Which Dockerfile are you talking about? I have tested the changes with master branch which is working fine.

On the Docker hub. The automatic build worked for 0.8.0 and failed for 0.9.0
I think the issue is that I didn't set DOCKER_TAG to be equal to the git tag in the automated build.

Screenshot 2020-01-29 at 09 21 58

So it's download the file at `latest` which doesn't exist.

nrjais:
I have mentioned the configuration in the issue #938 which can be used to automate release based on the tags.
You can also add a step to azure pipelines to build and push the image to docker hub that way there won't be any need to maintain the configuration on docker hub.
Let me know if you need anything else.

Keats:
Which approach would you recommend? The current Dockerfile fails even after fixing the docker tag value

```
 Downloading https://files.pythonhosted.org/packages/fb/40/f3adb7cf24a8012813c5edb20329eb22d5d8e2a0ecf73d21d6b85865da11/MarkupSafe-1.1.1-cp27-cp27mu-manylinux1_x86_64.whl
Building wheels for collected packages: dockerize
Running setup.py bdist_wheel for dockerize: started
Running setup.py bdist_wheel for dockerize: finished with status 'error'
Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-q1K_Rw/dockerize/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/pip-wheel-AETbdm --python-tag cp27:
usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
or: -c --help [cmd1 cmd2 ...]
or: -c --help-commands
or: -c cmd --help
error: invalid command 'bdist_wheel'
```
so I'm not attached to one or the other

nrjais:
I am able to build it when making a new release as previous releases will have the old dockerfile which only works for v0.8.0 release. After configuring the autobuild with regex on tag new releases with the this dockerfile file should work.

Keats:
Thanks!

PR #941, opened 2020-02-06
Change continue-reading to use a span with id instead of a named anchor in a paragraph

### LINKED ISSUE #940: Feature request: remove obsolete <a name=""> for continue-reading, relying on id instead.
# Feature request

## Environment

Zola version: 0.9.0

## Current Behavior

When using  in page, the target location `continue-reading` is defined using an anchor element with the name property <a name="continue-reading">, which is [marked as obsolete on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-name) as [its not in the HTML5 spec](https://www.w3.org/TR/html52/textlevel-semantics.html#the-a-element).

An empty paragraph is also created (which could be enough using `#zola-continue-reading` or renaming its id). While looking at the (firefox) accessibility inspector I'm not sure if the empty paragraph is an issue or will just be skipped by screen readers, but the <a> link also triggers 2 issues: 

- Keyboard Error: Interactive elements must be able to be activated using a keyboard. [Learn more](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#Interactive_elements_must_be_able_to_be_activated_using_a_keyboard)
- Text Labels and Names Error: Interactive elements must be labeled. [Learn more](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#Interactive_elements_must_be_labeled)

I noticed it at first because I'm decorating links using the `::after` pseudo element and had to add an exception in my css for this case.

## Expected Behavior

The default way of creating target locations in HTML5 (the id property), probably in a span element, which carries no semantic meaning.

I'm willing to create a PR to fix this if desired.

Thanks for this awesome site generator ;)

pheki:
Sanity check:

## Code changes

Closes #940, with the bonus that now it is possible to add `<!-- more -->` inside a paragraph without breaking it. In my tests the target goes to the same position as before when `<!-- more -->` is added between paragraphs.

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -17,8 +17,7 @@ use utils::vec::InsertMany;
 
 use self::cmark::{Event, LinkType, Options, Parser, Tag};
 
-const CONTINUE_READING: &str =
-    "<p id=\"zola-continue-reading\"><a name=\"continue-reading\"></a></p>\n";
+const CONTINUE_READING: &str = "<span id=\"continue-reading\"></span>";
 const ANCHOR_LINK_TEMPLATE: &str = "anchor-link.html";
 
 #[derive(Debug)]
--- a/components/rendering/tests/markdown.rs
+++ b/components/rendering/tests/markdown.rs
@@ -757,7 +757,7 @@ Bla bla
     .unwrap();
     assert_eq!(
         res.body,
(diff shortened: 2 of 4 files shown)


Keats:
The test failures are legit, there are some tests you missed (in `site` I think)

pheki:
Fixed it now, had forgotten about `--all` 🤦‍

Keats:
Thanks!

PR #945, opened 2020-02-12
Update Cargo.lock make build successfully on windows

startewho:
because the dep of  `sass-rs`. it can not build successfully on windows. And the `sass-rs` merge the patch and it update the version. and  it works now by update the deps


Keats:
Thanks for that! It is already updated on the `next` branch though.

PR #952, opened 2020-02-17
Set default user agent for external requests

### LINKED ISSUE #950: Set user agent to avoid errors in link_checker and load_data
# Bug Report
`zola check` currently reports errors for links where the server returns an error (e.g., 403, 400) when there is not a user agent in the request headers. This is expected behavior, as the current link_checker doesn't set any. Can we allow the link checker to set a user agent, and/or provide a default zola user agent?

## Environment
Ubuntu 18.04.4

Zola version:
v0.10.0

## Expected Behavior

Some servers return errors when the user agent header is missing. For example, when running the link_checker on a URL such as https://arxiv.org/abs/1906.01113, the link_checker will report a 403 and declare this as a dead link. This can be seen using an example test case:

components/link_checker/src/lib.rs
```rust
    #[test]
    fn user_agent_test() {
        let url = "https://arxiv.org/abs/1906.01113";

        let res = check_url(url, &LinkChecker::default());
        assert!(res.is_valid());
        assert!(res.code.is_some());
        assert!(res.error.is_none());
    }
```

This same test case will pass if a user agent is included, e.g.:

```rust
pub fn check_url(url: &str, config: &LinkChecker) -> LinkResult {
    {
        let guard = LINKS.read().unwrap();
        if let Some(res) = guard.get(url) {
            return res.clone();
        }
    }

    let mut headers = HeaderMap::new();
    headers.insert(ACCEPT, "text/html".parse().unwrap());
    headers.append(ACCEPT, "*/*".parse().unwrap());
    headers.append(USER_AGENT, "zola/0.10.0 link_checker".parse().unwrap());
...
```

Without a USER_AGENT, the test will fail. 

We could mitigate this issue by:
* Setting a default user agent (such as a a zola-specific user agent shown above)
* Allowing users to specify a user-agent via some configuration

For a default user-agent, we probably do not want a hard-coded string, and rather could just follow the reqwest example:

https://docs.rs/reqwest/0.10.1/reqwest/struct.ClientBuilder.html#method.user_agent
```rust
// Name your user agent after your app?
static APP_USER_AGENT: &str = concat!(
    env!("CARGO_PKG_NAME"),
    "/",
    env!("CARGO_PKG_VERSION"),
);

let client = reqwest::Client::builder()
    .user_agent(APP_USER_AGENT)
    .build()?;
```

Some other example URLs which return 400/403s without a user agent:
* https://makeymakey.com
* https://queue.acm.org/detail.cfm?id=3022184
* https://www.sushi-jiro.jp/dining-at-jiro/
* https://www.genkoreanbbq.com/

lukehsiao:
Many servers will return errors (e.g. 400/403) to requests that do not
set a User-Agent header. This results in issues in both the link_checker
and load_data components. With the link_checker these are false positive
dead links. In load_data, remote data fails to be fetched. To mitigate
this issue, this sets a default User-Agent of

    $CARGO_PKG_NAME/$CARGO_PKG_VERSION

Note that the root cause of this regression from zola v0.9.0 is that
reqwest 0.10 changed their default behavior and no longer sets a
User-Agent by default:

    https://github.com/seanmonstar/reqwest/pull/751

Fixes #950.

Sanity check:

## Code changes

This change is not visible to the end user, and doesn't have corresponding docs.

### DIFF
--- a/components/link_checker/src/lib.rs
+++ b/components/link_checker/src/lib.rs
@@ -58,7 +58,10 @@ pub fn check_url(url: &str, config: &LinkChecker) -> LinkResult {
     headers.insert(ACCEPT, "text/html".parse().unwrap());
     headers.append(ACCEPT, "*/*".parse().unwrap());
 
-    let client = Client::new();
+    let client = Client::builder()
+        .user_agent(concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION")))
+        .build()
+        .expect("reqwest client build");
 
     let check_anchor = !config.skip_anchor_prefixes.iter().any(|prefix| url.starts_with(prefix));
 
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

PR #960, opened 2020-03-04
Update configuration.md

Hustenbonbon:
Explain for newer users like me how to access the variables in code in the documentation


Keats:
Can we move that to templates/overview.md instead?

Keats:
Fixing it in next directly, thanks!

PR #969, opened 2020-03-12
Fix RSS template

raindev:
Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

---

Remove a tag imported from Atom from the otherwise valid RSS template.

PR #970, opened 2020-03-13
Fix RSS template (#969)

### LINKED ISSUE #967: Default feed template conflates rss 2.0 and Atom
# Bug Report

## Environment

Zola version: 0.10.0

## Current Behavior

Zola (with `generate_rss = true`) generates a feed with an `rss` root tag with an attribute `xmlns:atom="http://www.w3.org/2005/Atom"` and `version="2.0"`. This seems like a bug. It should either be `<rss version="2.0">` or `<feed xmlns:atom="http://www.w3.org/2005/Atom">`. This is just from my cursory look at the two formats though, maybe I'm wrong.

Also, [FeedReader](https://jangernert.github.io/FeedReader/) fails to parse the feed. I will investigate whether that is related or not soon, and split that into a new issue if it has a different cause.

raindev:
Remove a tag imported from Atom from the otherwise valid RSS template.

Fixes #967
Identical to #969 with tests updated

### DIFF
--- a/components/site/tests/site.rs
+++ b/components/site/tests/site.rs
@@ -298,7 +298,7 @@ fn can_build_site_with_taxonomies() {
     assert!(file_contains!(
         public,
         "categories/a/rss.xml",
-        "https://replace-this-with-your-url.com/categories/a/rss.xml"
+        "https://replace-this-with-your-url.com/posts/with-assets/"
     ));
     // Extending from a theme works
     assert!(file_contains!(public, "categories/a/index.html", "EXTENDED"));
--- a/components/templates/src/builtins/rss.xml
+++ b/components/templates/src/builtins/rss.xml
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
+<rss version="2.0">
     <channel>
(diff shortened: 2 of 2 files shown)


components/site/tests/site.rs
```diff
@@ -298,7 +298,7 @@ fn can_build_site_with_taxonomies() {
     assert!(file_contains!(
         public,
         "categories/a/rss.xml",
-        "https://replace-this-with-your-url.com/categories/a/rss.xml"
+        "https://replace-this-with-your-url.com/posts/with-assets/"
```
raindev, line 301: The feed no longer contains a link to itself. Assert that the category feed contains a link to one of its posts instead.

raindev:
@Keats, I just have seen your comment on #967 that this is what [Hugo](https://github.com/gohugoio/hugo/blob/5eadc4c0a8e5c51e72670591c4b7877e79c15e3c/tpl/tplimpl/embedded/templates/_default/rss.xml) does as well. This is true of [Hakyll](https://github.com/jaspervdj/hakyll/blob/b08dd317a381a0436607b44257c487b1930dcc72/data/templates/rss.xml) as well. Lets hold the PR, I'll try to figure out what's going on here.

raindev:
The template actually represents a valid RSS feed: https://github.com/getzola/zola/issues/967#issuecomment-598918979.

PR #981, opened 2020-03-31
Fail if a language is set both in config.default_languages and config…

gcaillaut:
Hi,

This PR adds a failure if a language is defined both as the default language and in the `config.languages` list.

I wrote this PR because at first I did not understand how to configure my multilingual site and ended up having a config like this:

```toml
default_language = "fr"
languages = [
    { code = "fr", rss = true },
    { code = "en", rss = true },
]
```

But I had some troubles because zola generated the two *overlapping* sections `_index.md` and `_index.fr.md`. As a consequence, the index page of my default language was empty because zola *sourced* the wrong section.

This error should not be reproducible with this PR.

### DIFF
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -215,6 +215,10 @@ impl Config {
             bail!("Highlight theme {} not available", config.highlight_theme)
         }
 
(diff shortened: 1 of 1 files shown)


Keats:
Thanks, nice catch!

PR #982, opened 2020-03-31
Add an optionnal `lang` parameter to get_url

gcaillaut:
Hi,

This PR adds an optionnal `lang` keyword to the `get_url` filter, as discussed *a long time ago* [here](https://zola.discourse.group/t/rfc-i18n/13).

If a file has the following path (relative to the content folder) `a_section/a_md_file.md`, then get_url(path="@/a_section/a_md_file.md", lang="fr") will return:

- $BASE_URL/a_section/a_md_file if the default language is fr
- $BASE_URL/fr/a_section/a_md_file if the default language is not fr and fr is in `config.languages`
- It will fail otherwise, with the *nice* error message `"'fr' is not an authorized language (check config.languages)."`

### DIFF
--- a/components/templates/src/global_fns/mod.rs
+++ b/components/templates/src/global_fns/mod.rs
@@ -53,6 +53,22 @@ impl GetUrl {
         Self { config, permalinks }
     }
 }
+
+fn make_path_with_lang(path: String, lang: &str, config: &Config) -> Result<String> {
+    if lang == &config.default_language {
+        return Ok(path);
+    }
+
+    if !config.languages.iter().any(|x| x.code == lang) {
+        return Err(format!("`{}` is not an authorized language (check config.languages).", lang).into());
+    }
+
+    let mut splitted_path: Vec<String> = path.split(".").map(String::from).collect();
+    let ilast = splitted_path.len() - 1;
+    splitted_path[ilast] = format!("{}.{}", lang, splitted_path[ilast]);
+    Ok(splitted_path.join("."))
+}
+
 impl TeraFn for GetUrl {
     fn call(&self, args: &HashMap<String, Value>) -> Result<Value> {
         let cachebust =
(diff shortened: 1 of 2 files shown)


components/templates/src/global_fns/mod.rs
```diff
@@ -516,6 +541,10 @@ mod tests {
     const TRANS_CONFIG: &str = r#"
 base_url = "https://remplace-par-ton-url.fr"
 default_language = "fr"
+languages = [
+    { code = "fr" },
```
gcaillaut, line 545: Oups, this line will prevent the tests added in #981 to pass...

Keats: indeed

Keats (commented):
Do you have a specific usecase for that feature?

gcaillaut:
> 
> 
> Do you have a specific usecase for that feature?

Yes, of course. I wouldn't bother writing code if it weren't necessary for me (or maybe I missed a feature...).

My website supports two languages (french as default langage and english). It is designed so that there is the same navbar on each page, which is what happens on most websites. When I'm on the french version, I'd like the links in the navbar to targets french content, but on the english version I'd like them to target english content, obviously.

I could easily have writtent something like this in my `layout.html` file:

```jinja
{% if lang == "" %}
<!-- French content by default -->
<a href="my-page.md">...</a>
{% else %}
<!-- English content otherwise -->
<a href="my-page.en.md">...</a>
{% endif %}
```
It would be okay, since my website is not very big and supports only two languages. But I strongly prefer the version below, since it's shorter and *should* work for any language.

```jinja
<!-- at the top of my layout template, which is extended by all the others (needs to be written once) -->
{% if lang == "" %}
  {{ set lang_trans = "fr" }}
{% else %}
  {{ set lang_trans = lang }}
{% endif %}

<!-- A multilingual link -->
<a href="{{ get_url(path='my-page.md', lang=lang_trans}}">{{ trans(key='my-key', lang=lang_trans)}}</a>
```
Note that I must define a new variable `lang_trans` because the `lang` variable is empty when the current language is the default language (why?).

Keats:
> because the lang variable is empty when the current language is the default language (why?).

It's set in some places but not all it seems, it's a leftover issue and should be fixed in the next release

PR #983, opened 2020-03-31
Preserve timestamps when copying files (#974)

### LINKED ISSUE #974: Feature Request: Preserve timestamps of files copied in.
I use rsync to deploy my site, which has some very large mp4 files.

As the timestamps are not preserved when creating the contents of the public folder for copied files and folders, rsync has to do a checksum check on each and every file to see if it really needs transferring.

magurotuna:
Hi, this PR fixes #974 

In my implementation, preserving modification timestamps is achieved by setting the source file's timestamp to the target file after copying.
And I also added a test to check this change works as expected.

Hope this meets the requirements.
Thank you.

---

Sanity check:

## Code changes

### DIFF
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -95,7 +95,10 @@ pub fn find_related_assets(path: &Path) -> Vec<PathBuf> {
 }
 
 /// Copy a file but takes into account where to start the copy as
-/// there might be folders we need to create on the way
+/// there might be folders we need to create on the way.
+/// No copy occurs if both of the following conditions are satisfied:
+/// 1. A file with the same name already exists in the dest path.
+/// 2. Its modification timestamp is identical to that of the src file.
 pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf, hard_link: bool) -> Result<()> {
     let relative_path = src.strip_prefix(base_path).unwrap();
     let target_path = dest.join(relative_path);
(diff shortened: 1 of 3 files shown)


components/utils/src/fs.rs
```diff
@@ -106,7 +107,11 @@ pub fn copy_file(src: &Path, dest: &PathBuf, base_path: &PathBuf, hard_link: boo
     if hard_link {
         std::fs::hard_link(src, target_path)?
     } else {
-        copy(src, target_path)?;
+        copy(src, &target_path)?;
+        // set original modification timestamp to the target file
+        let src_metadata = metadata(src)?;
+        let mtime = FileTime::from_last_modification_time(&src_metadata);
+        set_file_mtime(&target_path, mtime)?;
```
Keats: Can we avoid copying the file if it exists and the mtime is identical to the src?

magurotuna: Sure, that sounds a nice idea. I'll implement it.

magurotuna: I've done it :)

michael-conrad:
Shouldn't "preserving modification timestamps is achieved by setting the source file's timestamp to the target file after copying" instead read "preserving modification timestamps is achieved by setting the destination file's timestamp to the source file's timestamp after copying"?

components/utils/src/fs.rs
michael-conrad: > Can we avoid copying the file if it exists and the mtime is identical to the src?

Need to also compare SIZE in addition to the mtime.

magurotuna:
I intended our `copy_file` function to work just like Unix command `cp -p <src_path> <dest_path>`.

For example, suppose we have a file `A`, which has its modification timestamp `T_a`.
After running `cp -p A B` command, we will find the modification timestamp of `B` is just `T_a`, not the time of `cp` command executed.
I think my implementation imitates this behavior.

Please let me know if there are some cases in which this way of implementation is not suitable.

components/utils/src/fs.rs
magurotuna: You're exactly right. Like `rsync` command, the decision whether copy operation is necessary or not should be done by comparing both timestamp and filesize of two files.
cf. https://stackoverflow.com/questions/13778889/rsync-difference-between-size-only-and-ignore-times/44353773
I'll write it. Thanks.

magurotuna: Added it!

Keats:
Thanks!

PR #986, opened 2020-04-03
Various changes from Chris Morgan’s fork, for discussion

chris-morgan:
I open this for discussion; I don’t expect it to be accepted verbatim.

The first four commits (chmod -x a few files, and some documentation fixes) should be uncontroversial.

The last commit (a regular expression replacement filter) is unlikely to be desired in this repository, but I figured I’d include it for discussion. (It’s doing steadily heavier lifting on my site, with things like conditional comments to vary content for feed/HTML. You would probably scream if you saw much of my code; it’s probably for the best that I haven’t published it anywhere yet. 😛)

The RSS → feed change that I make may be controversial, especially my proposal to change the default.

Some of my HTML normalisation may be controversial.

The remaining commits are things that I think are useful, for various reasons.

The `updated` field commit is fully functional, but incomplete for landing, for the reasons noted.

To give you a taste of why I want taxonomy terms exposed in the feed template, here’s what I wrote yesterday (after having skipped it in my initial port to Zola) before deciding it’d be much better to just expose `taxonomy` and `term`:

```jinja
{# I need to be able to distinguish term-specific feeds from the global feed. This is the *crazy* way in which I do it. #}
{%- set path = feed_url | replace(from=config.base_url, to="") | replace_all(regex="^(?:/(.*))?/feed\.xml$", rep="$1") %}
{%- if path %}{# /TAXONOMY/TERM/feed.xml #}
    {%- set taxonomy = path | replace_all(regex="^(.*)/(.*)$", rep="$1") %}
    {%- set term = path | replace_all(regex="^(.*)/(.*)$", rep="$2") %}
    {%- set title = "Chris Morgan’s blog posts tagged " ~ config.extra.taxonomies[taxonomy][term].name %}
{%- else %}
    {%- set title = "Chris Morgan’s blog" %}
{%- endif -%}
…
<title>{{ title }}</title>
…
```

Just to make you cringe a little more, it occurred to me yesterday that I could put a slash in a taxonomy name, so `taxonomies."blog/tags" = […]` it is, and my URLs can go back to /blog/tags/…/ like I wanted the whole time.

This does not currently reflect what chrismorgan.info is being built with (it’s being built with https://github.com/chris-morgan/zola/tree/0.9.0.dev.294ea16d25a2a559a59472a1dbc0da4f58b395ef), but should do within a few hours.


Keats:
I like almost everything! I would just leave out the author things/replace_all for now.
As for feeds, I would probably *only* have Atom and not support RSS 2.0 if we're making a breaking change. Atom is better and is probably supported by every single tool supporting feeds so might as well focus on one and not bother with the other.

chris-morgan:
I think there’s value in providing templates for both Atom and RSS, because it makes it a very slight breaking change only—just change one line and you can continue as you were without needing to worry about whether changing from RSS to Atom may break clients, cause all of your posts to appear as though new, *&c.* I would strongly argue that the URL remain flexible even if only one feed template was provided, because URL persistence is important, and Zola sites are regularly deployed to locations that can’t do HTTP redirects, so feed URLs need to be *able* to stay the same. And once you’ve got that, providing atom.xml and rss.xml templates is trivial and has negligible cost.

(HTML has meta redirects, which `redirect_to` and `aliases` boil down to. RSS/Atom don’t have an equivalent, unless pagination could be perverted to that effect, concerning which I’m not sure. That reminds me, I plan to implement feed pagination sooner or later, because I would *like* https://chrismorgan.info/feed.xml to have all my content, but it’s 400KB and growing, currently accounting for the substantial majority of my site traffic, 3GB per day.)

I’m not sure about Atom in podcasting in particular—information about *any* of this feed stuff is extremely hard to come by, and for podcasts doubly so. I found one source from years and years ago saying iTunes supports Atom but still recommends RSS 2.0. Apple’s own site makes no mention of Atom at all, only RSS 2.0. The two validators they link to, one says everything’s hunky dory RSS even when fed an Atom document, and the other says iTunes speaks RSS 2.0 only, not older RSS or Atom. I have a dreadful premonition that I may be forced to become an Expert in these matters. (If I do, I will write about it and finally provide a canonical resource that I have wanted for at least a decade.)

I plan to add daily podcasting to my site on Bible matters in the immediate future, so I’m going to see how well Atom works to begin with—but if that doesn’t work out and I decide I care about that, I may end up feed.xml being RSS if `taxonomy.name == "podcast"`, and Atom otherwise! (This would be accompanied by a content-type switch in nginx.)

Of course the whole podcasting thing is a slight red herring for Zola itself anyway, because anyone doing that must necessarily use a custom template.

SgtIria:
I've been using Zola to produce a custom (rss.xml template) podcast feed for some time. 

(Hack to know we are dealing with podcast feed (not regular feed)
```
{% if feed_url is containing("podcast") %}
{% set podcastRSS = true %}
{% else %}
{% set podcastRSS = false %}
{% endif %}
```

Then when we are producing a podcast rss.xml we do what is needed ...

```
{% if podcastRSS %}
<?xml version="1.0" encoding="UTF-8"?>
<rss version="2.0" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:wfw="http://wellformedweb.org/CommentAPI/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:sy="http://purl.org/rss/1.0/modules/syndication/" xmlns:slash="http://purl.org/rss/1.0/modules/slash/" xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd" xmlns:rawvoice="http://www.rawvoice.com/rawvoiceRssModule/" xmlns:googleplay="http://www.google.com/schemas/play-podcasts/1.0" >

etc...
```

NOTE: Unfortunately because Zola / Terra outputs blank spaces and lines where the template code is, this will result in "invalid XML" as you cannot have spaces and blank lines preceding the actual XML tags. So I usually just strip these out of the file as part of my build process. 

`grep "\S" ~/builds/adeptenglish.com/podcast/feed/rss.xml > /opt/Caddy/www/podcast/feed/rss.xml`

We have a simple taxonomy item (podcast). The zola taxonomy section with RSS turned on (limited to 75 episodes) generated for this section.
```
taxonomies = [
    {name = "tags", rss = false, paginate_by = 6}, 
    {name = "english", rss = false, paginate_by = 6},  
    {name = "authors", rss = false, paginate_by = 2}, 
    {name = "downloads", rss = false, paginate_by = 8},
    {name = "podcast", rss = true, paginate_by = 75}, 
]
```

Inside each .md document acticle we write which is part of the podcast feed contains:

```
podcast = [ "feed" ]
```

Which results in a URL `domain.com\podcast\feed` which is out podcast feed end point.

If you are looking for documents which specify who supports what type of standard podcast xml rss feed what you won't find much. 

Apple podcasts does specify RSS 2.0 and its own feed tags, which most people (LibSyn, BluBrry) adhere to iTunes (its the main podcast distributor). (https://itunespartner.apple.com/podcasts/articles/creating-your-show_requirements)

However https://podcasters.spotify.com/terms/Spotify_Podcast_Delivery_Specification_v1.6.pdf Spotify is now in the podcast world and it's been quite specific about it's feeds. 

There are a number of feed validators (which are the main way in which feed compliance is achieved):

https://castfeedvalidator.com/
https://podba.se/validate/

Basically my podcast feed XML is RSS 2.0, with custom tags for iTunes, Google Podcasts, BluBrry.

This works for us.

Keats:
> NOTE: Unfortunately because Zola / Terra outputs blank spaces and lines where the template code is, this will result in "invalid XML" as you cannot have spaces and blank lines preceding the actual XML tags.

Have you tried whitespace control? https://tera.netlify.com/docs/#whitespace-control

SgtIria:
White space control will help trim spaces either side of a variable being output to the template. The problem is the actual tera code (loops, conditional statements etc..) also seems to leave spaces in the output. So a loop over page assets may result in 20+ blank lines in the RSS.XML. It's the same in HTML output, you get hundreds of blank lines in the output, however HTML parsers are a lot more forgiving and ignore the white space.

Keats:
Whitespace control also works on for/if etc: https://github.com/Keats/tera/blob/master/src/renderer/tests/whitespace.rs

SgtIria:
> 
> 
> Whitespace control also works on for/if etc: https://github.com/Keats/tera/blob/master/src/renderer/tests/whitespace.rs

I cannot believe I didn't spot that this was possible. Sometimes you see something you log it and never come back to it. I've just changed the rss.xml template to us `-%}` throughout and it's removed all the white space. Thanks for asking me to look again.

Thinking about it this would probably be the default usecase for most people writing templates and should maybe be the default behaviour, failing that the Tera documentation could possibly be more explicit.

Keats:
@chris-morgan can you do a PR without the author/replace_all commits?

Keats:
I've merged most of it. I don't think the auhtors make sense in Zola itself and not sure of the replace_all. I'll close this for now.

> failing that the Tera documentation could possibly be more explicit. 

I'll update the docs.

Issue #317, opened 2018-06-14
TOML is not very friendly for non-technical users


jwatt:
I've been trying to teach some older non-technical users how to use TOML front-matter for a simple recipe site that I'm putting together. This front-matter might look something like this:

    [extra]
    title = "Traditional \"Scottish\" porridge"
    description = """A traditional Scottish version of porridge made using
      water and salt.

      If you need instructions on how to make porridge, you're probably not a
      traditional Scot. ;-)  We recommend milk instead of water in that case.
      """
    servings = 1
    prep-time = "5 minutes"
    ingredients = [
      "40 g oats",
      "250 ml water",
      "pinch of salt",
      ]
    directions = [
      "Add all the ingredients to a saucepan.",
      """Stirring frequently, cook on a medium heat until the porridge thickens
        and becomes...porridgy.""",
      "Serve.",
      ]
    [extra.nutrition]
      fat = "1 g"
      carbs = "30 g"
      protein = "4 g"

I've had quite a lot of trouble getting them to feel comfortable with this, so I tried them with YAML instead, which for comparison would be:

    extra:
      title: Traditional "Scottish" porridge
      description: A traditional Scottish version of porridge made using water
        and salt.

        If you need instructions on how to make porridge, you're probably not a
        traditional Scot.  ;-) We recommend milk instead of water in that case.
      servings: 1
      prep-time: 5 minutes
      directions:
        - Add all the ingredients to a saucepan.
        - Stirring frequently, cook on a medium heat until the porridge
          thickens and becomes...porridgy.
        - Serve.
      ingredients:
        - 40 g oats
        - 250 ml water
        - pinch of salt
      nutrition:
        fat: 1 g
        carbs: 30 g
        protein: 4 g

There are a bunch of issues with the TOML (demonstrate in the TOML example above) that when taken together are making them really unhappy:

* In general there's a bunch of syntax that non-technical users aren't used to dealing with.
* As a programmer, quoting text seems normal to me, but to them it seems like an unnecessary irritation and visual clutter.
* The need to quote text also means that they would need to remember to "**escape**" quote marks using a backslash, and that probably **isn't** going to happen. :-/
* Requiring different quoting for single line text and multi-line text (triple quotes) seems to them to be gratuitous, and will likely be error prone.
[...]

PR #990, opened 2020-04-11
Yaml frontmatter

er-vin:
This is an attempt at adding YAML front matter support to Zola. I'm carrying over 16 years worth of content which moved from Jekyll to Hugo and now to Zola. Obviously this makes it easier to have YAML support in the front matter.

This feature has been discussed in #317 already and I agree that YAML should be limited only to actual content and nothing more. This is mainly useful for legacy reason and for external tooling (mainly text editors) which tend to support properly only YAML in the front matter.

### DIFF
--- a/components/front_matter/Cargo.toml
+++ b/components/front_matter/Cargo.toml
@@ -9,9 +9,14 @@ tera = "1"
 chrono = "0.4"
 serde = "1"
 serde_derive = "1"
+serde_yaml = "0.8"
 toml = "0.5"
 regex = "1"
 lazy_static = "1"
 
 errors = { path = "../errors" }
 utils = { path = "../utils" }
+
+[dev-dependencies]
+test-case = "1.0"
+
--- a/components/front_matter/src/lib.rs
+++ b/components/front_matter/src/lib.rs
@@ -3,7 +3,9 @@ use serde_derive::{Deserialize, Serialize};
 
 use errors::{bail, Error, Result};
 use regex::Regex;
+use serde_yaml;
 use std::path::Path;
+use toml;
 
 mod page;
 mod section;
@@ -12,8 +14,31 @@ pub use page::PageFrontMatter;
 pub use section::SectionFrontMatter;
 
 lazy_static! {
-    static ref PAGE_RE: Regex =
+    static ref TOML_RE: Regex =
         Regex::new(r"^[[:space:]]*\+\+\+(\r?\n(?s).*?(?-s))\+\+\+\r?\n?((?s).*(?-s))$").unwrap();
+    static ref YAML_RE: Regex =
+        Regex::new(r"^[[:space:]]*---(\r?\n(?s).*?(?-s))---\r?\n?((?s).*(?-s))$").unwrap();
+}
+
+pub enum RawFrontMatter {
+    Toml(String),
+    Yaml(String),
+}
+
+impl RawFrontMatter {
+    fn deserialize<T>(&self) -> Result<T>
+    where
+        T: serde::de::DeserializeOwned,
+    {
+        let f: T = match self {
+            RawFrontMatter::Toml(s) => toml::from_str(s)?,
+            RawFrontMatter::Yaml(s) => match serde_yaml::from_str(s) {
+                Ok(d) => d,
+                Err(e) => bail!(format!("YAML deserialize error: {:?}", e)),
+            },
+        };
+        Ok(f)
+    }
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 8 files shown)


Keats:
Is that using strict yaml?

er-vin:
Unfortunately not, to my knowledge serde_yaml currently only uses yaml-rust and not strict-yaml-rust. As soon as serde_yaml allows to use strict-yaml-rust as dependency or if a serde_strict_yaml emerges I'd expect this to be easy to migrate though.

Keats:
I would really prefer to have strict yaml only though but https://github.com/fralalonde/strict-yaml-rust doesn't seem maintained. Is there another crate?

er-vin:
Nope, didn't find any crate apart from the unmaintained one you mentioned. So went for the "next best thing" which is at least maintained.

Keats:
I didn't mean to close all the PRs by deleting the branch :(

Abdillah:
Do we wait for a maintained strict Yaml crate to appear?
Hope this merged soon..

samyakb:
[yaml-rust](https://crates.io/crates/yaml-rust) crate is fully-comaptible with YAML 1.2 specification. Can we use this library ? @er-vin

er-vin:
> [yaml-rust](https://crates.io/crates/yaml-rust) crate is fully-comaptible with YAML 1.2 specification. Can we use this library ? @er-vin

Already the case in fact, this is what serde_yaml uses under the hood.

er-vin:
Need to rebase this once more I see, will try to do it next week.

rcarmo:
this is great, I too have 18 years of content I'd like to move across :)

samyakb:
I hope this gets merged soon. Thanks @er-vin

rcarmo:
Bumping this to check if it will be merged soon.

docs/content/documentation/content/page.md
```diff
@@ -74,10 +74,10 @@ The file `content/blog/2018-10-10-hello-world.md` will yield a page at `[base_ur
 
 ## Front matter
 
-The TOML front matter is a set of metadata embedded in a file at the beginning of the file enclosed
-by triple pluses (`+++`).
+The front matter is a set of metadata embedded in a file at the beginning of the file enclosed
+by triple pluses (`+++`) if in TOML format or by triple minuses (`---`) if in YAML format.
```
Keats: Let's not document that for now, I would rather this be an undocumented feature.

docs/content/documentation/content/section.md
```diff
@@ -26,12 +26,13 @@ Just like pages sections can be drafted by setting the `draft` option in the fro
 The `_index.md` file within a directory defines the content and metadata for that section.  To set
 the metadata, add front matter to the file.
 
-The TOML front matter is a set of metadata embedded in a file at the beginning of the file enclosed by triple pluses (`+++`).
+The front matter is a set of metadata embedded in a file at the beginning of the file enclosed
+by triple pluses (`+++`) if in TOML format or by triple minuses (`---`) if in YAML format.
```
Keats: same here

components/front_matter/Cargo.toml
```diff
@@ -9,9 +9,11 @@ tera = "1"
 chrono = "0.4"
 serde = "1"
 serde_derive = "1"
+serde_yaml = "0.8"
 toml = "0.5"
 regex = "1"
 lazy_static = "1"
+test-case = "1.0"
```
Keats: this is a dev-dependency

Keats (changes requested):
thanks for the ping, I forgot about that.
The code looks ok, just a few tweaks needed

rcarmo:
I'm curious: why not document something that may bring more people to use this software?

Keats:
I don't like YAML and personally think this is not a good feature but many people want it so I'm ok adding it (like themes). But adding it to the documentation means more people will use and will want to have YAML everywhere instead of TOML (config.yaml, theme.yaml etc) and it would make sense in terms of consistency. But I really don't want to ;(

rcarmo:
Well, here’s an idea: document it as a _migration feature_ and explicitly state the limitations and that it is meant only for onboarding older content.

Otherwise nobody will discover it and you’ll lose out on people like me (who have existing content and would like to use Zola without reformatting it) but who don’t dig into the internals of their engines. Which are likely the majority of Jekyll users...

Most people don’t care about the limitations of YAML. They just care that it’s easy to write, that their editor (and GitHub preview) support it and that their content can move across.

And many don’t want to learn TOML or use it extensively, because it is just harder to write (even I am OK with writing a config file, but don’t see myself switching to TOML for content). So having YAML for front-matter might be more then enough for attracting new users without adding too much attrition to possible migrations.

Keats:
>Well, here’s an idea: document it as a _migration feature_ and explicitly state the limitations and that it is meant only for onboarding older content.

That's a good angle!

components/front_matter/Cargo.toml
er-vin: Good point, done.

docs/content/documentation/content/page.md
er-vin: So I kept the commit about documentation but reworded it with the spin @rcarmo had in mind. Hopefully that's fine now.

docs/content/documentation/content/section.md
er-vin: ditto

components/front_matter/src/lib.rs
```diff
@@ -12,8 +14,31 @@ pub use page::PageFrontMatter;
...
+    static ref TOML_RE: Regex =
         Regex::new(r"^[[:space:]]*\+\+\+(\r?\n(?s).*?(?-s))\+\+\+\r?\n?((?s).*(?-s))$").unwrap();
+    static ref YAML_RE: Regex =
+        Regex::new(r"^[[:space:]]*---(\r?\n(?s).*?(?-s))---\r?\n?((?s).*(?-s))$").unwrap();
+}
+
+pub enum RawFrontMatter {
+    Toml(String),
+    Yaml(String),
+}
```
Keats, line 26: Can we not return &'a str instead of String?

Keats (commented):
Just one small comment but lgtm otherwise

components/front_matter/src/lib.rs
er-vin: I tried that and somehow failed so that's why I stuck with String... turns out I tried again and succeeded this time. So it's done now.

Keats:
Thanks!

ttys3:
with this PR, I think I can switch from my Hugo  + YAML front matter blog to zola !

thanks for the works, guy!

PR #996, opened 2020-04-16
Theme docs: actually edit the config

elegaanz:
I hope it is just not me that misunderstood the sentence above, but I think it should be `true` and not `false`.

Sanity check:


Keats:
Nope, you're completely right! Thanks.

PR #997, opened 2020-04-20
Theme translations

southerntofu:
Discussion [on the forum](https://zola.discourse.group/t/i18n-theme-translations/388)

In investigating this problem, i uncovered a problem that using sub-tables like `[extra.foo]` would not allow to keep theme defaults when a single key was overriden in site config. A new test l.585 catches this.

I also added a test l.454 to ensure trying to override a theme's table with another type errors, so that if a theme is expecting a mapping it will not produce weirder errors.

I updated the docs.

Let me know if there's anything to change, but feel free to take/edit it in any way you please.

### DIFF
--- a/components/config/src/config.rs
+++ b/components/config/src/config.rs
@@ -306,12 +306,11 @@ impl Config {
                 continue;
             }
             if !val.is_table() {
-                // We don't want to merge this because it's not a map
+                // We don't want to merge this because it's not a map (table)
                 continue;
             }
             let child = val.as_table().unwrap();
             // We borrow mutably the site's extra subsection but only if it's a Table
-            // We can unwrap safely because we just checked the key was in there
             let site_val = self.extra.get_mut(key).unwrap();
             if !site_val.is_table() {
                 // ERROR because user rewrote a theme config's mapping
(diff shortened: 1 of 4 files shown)


southerntofu:
Hello, may i revive this PR? I'm still using `config.extra.translations` due to theme's extra being properly merged with config's extra (#1100), but it would be practical to have theme translations supported by the `trans` function directly.

Currently i use a [macro](https://tildegit.org/southerntofu/zola-water/src/branch/master/templates/widgets.html#L137) to simulate the trans function, but i would be happy to make a PR for adding a config.toml setting to zola's trans function to choose what to do in case of missing translation:

- error and abort (currently the default)
- print a warning a use a placeholder containing the name of the missing key
- print a warning and fallback to the translation in the default_language

In particular, on volunteer-run sites where translations are not always in sync, it's useful to not crash when a certain translation is missing while still producing a warning. Is a PR for that kind of stuff welcome?

southerntofu:
[I've posted in more detail on the forum to gather more feedback](https://zola.discourse.group/t/i18n-theme-translations/388/3)

PR #1002, opened 2020-04-23
Remove implicit dependency on openssl-sys (#1001)

### LINKED ISSUE #1001: Provide an entirely static build against musl
I think it would be very nice for build environments if you could provide an entirely static build that was built against musl. I run into GLIBC ABI trouble on Centos 7 as it stands with your current builds.

svenstaro:
As per #1001, this properly makes use of rustls-tls only.

NOTE: This does _not_ fix #1001 entirely. There is still a problem with libass_sys on musl. This is an isolated problem that currently exists.


Keats:
Can you rebase on next and do the PR on the next branch by any chance?

svenstaro:
Continued in #1005.

Issue #361, opened 2018-08-05
Overriding blocks


lnicola:
I'm not familiar to Jinja2 templates, so I'm probably missing something. The `hyde` theme defines a couple of blocks like `extra_head` and `sidebar_about`. Can I override them in my site? I tried dropping a file in either `templates/` or `themes/hyde/templates/`, but that didn't seem to work.

Keats:
You can override blocks, you first need to `extends` from a template and then redefining them (see https://tera.netlify.com/docs/templates/#inheritance for the docs).

In short for example in the hyde template, you would create a `index.html` template in `templates/` that  looks like:

```jinja
{% extends "themes/hyde/index.html" %}

{% block extra_head %}
// You can put what you want there
{% endblock %}
```

lnicola:
Got it, that's just like `page.html` extends `index.html`. Thanks.

lnicola:
Hmm, that doesn't work:

```
Error: Template 'index.html' is inheriting from 'themes/hyde/templates/index.html', which doesn't exist or isn't loaded.
```

The file does exist, so maybe it's not loaded yet?

Keats:
Sorry I got it wrong, it should be `{% extends "hyde/templates/index.html" %}`

lnicola:
Thanks, that works.

lnicola:
Actually, this doesn't seem great. I can make a template that extends the theme `index.html` template and override that block. However, the theme's `post.html` will the theme's own `index.html`, not mine.

Keats:
As far as I can tell it is working as expected, the page template doesn't really know about your updated `index.html`.
The renderer does those lookups:

- for rendering index page: `templates/index.html` extends `hyde/templates/index.html`
- for rendering a post page: `hyde/templates/page.html` extends `hyde/templates/index.html`

The renderer doesn't look doesn't see `templates/index.html` at all in the extend path of the page template so it assumes it's done. 
It's not ideal I agree but I don't see an easy way for the renderer to understand it needs to insert `templates/index.html` in the middle: we would need to rewrite manually the inheritance chain AFAIK

lnicola:
I realized it was working as designed, but it seems to limit the usefulness of being able to place a theme in `themes/`. Maybe the recommended approach (for now, at least) is to copy a theme and change it instead of using overrides?

JeanMertz:
The issue I found with this is _somewhat_ similar, so I figured I'd post it here, but feel free to move this if it's a separate issue:

[...]

Issue #324, opened 2018-06-25
include doesn't seem to work in a theme


sebtombs:
I am attempting to migrate a Bolt site to be static due to poor database performance by the hosting provider for the site. The reason I've picked Gutenberg rather than Hugo is because Tera is Twig-like, so it makes it easier (and also I much prefer Tera/Twig syntax).

However, I've hit a snag. The Bolt templates I am starting with use a hierarchy of partial files. Having changed the suffices and sorted out the syntactic differences between Bolt-extended Twig and Tera, I am still getting the following error:

Reason: Failed to render 'boltlike/templates/kb.html' (error happened in a parent template)
Reason: Template 'partials/_topbar.html' not found

Now, partials/_topbar.html is included into partials/_master.html, which is itself extended by kb.html. If I move the include to kb.html, or put the included file in the theme templates directory rather than partials subdirectory, I get the same issue.

So, "extends" seems to work, but "include" doesn't. In the three sample themes (book, even, and hyde) that I've looked at, "include" isn't used.

However, if I put _topbar.html in the actual templates directory, rather than the templates directory in the theme, all works as expected. So the problem is only present when the include destination is in a theme.

This is no big deal for me at the moment, as I can just set things up in the template, but may catch others out who want to use include in a theme.

Keats:
It is mentioned in https://www.getgutenberg.io/documentation/themes/creating-a-theme/#caveat
The main reason is that include can be anywhere in the AST and I didn't want to update the code that rewrites import to be recursive because of laziness. I'd happily take a PR for it though

Keats:
If anyone is interested in fixing it, it should be in `components/utils/templates.rs:rewrite_theme_paths` but it will need to walk the AST recursively. Not very hard and I might do it myself after the 0.4 release but that would be a nice first issue for someone interested in contributing

sebtombs:
Apologies for not reading the documentation properly, and for my tardy response. I hope to work on this issue some time in the next few weeks - not quite there yet with my site conversion which needs to come first.

Keats:
No worries, it can be worked on any time

southerntofu:
After some testing, i believe [this](https://github.com/southerntofu/zola/commit/5b618b3ff5b7bb86312d32c2530c04cbe764b118) fixes the issue. Could people confirm or point out problems with this approach? (tests pass, i use it on some sites)

Keats:
[...]

PR #1004, opened 2020-04-23
Make themes more flexible

southerntofu:
* [x] The change is fully backwards-compatible 

Fix for both #324 and #361. I open this PR to bring more attention so everyone can test and see if it makes them happy.

Following remarks [here](https://github.com/getzola/zola/issues/324#issuecomment-618257210), i made sure this patch was fully backwards-compatible by copying the template's theme (inside tera) to  "THEME/templates/index.html" in all cases, where THEME is the theme name.

In addition, I propose to copy all theme templates to "theme/templates/" where "theme" is a literal. This would allow site templates to extend the theme's template regardless of the theme name, so that  you can swap themes seemlessly as long as they implement the same blocks in the base template. Would this be ok?

Waiting for feedback before i edit the docs


southerntofu:
done. the section is formatted properly and tests pass

Keats:
It looks like it works while being much simple?! Can you add tests for shortcodes and an actual `include`?

Keats:
ping @southerntofu

Keats:
I didn't mean to close all the PRs by deleting the branch :(

southerntofu:
> Can you add tests for shortcodes and an actual include?

done. i also updated the docs

> In addition, I propose to copy all theme templates to "theme/templates/" where "theme" is a literal. This would allow site templates to extend the theme's template regardless of the theme name, so that you can swap themes seemlessly as long as they implement the same blocks in the base template. Would this be ok?

^ Would this change be welcome?

Keats:
> This would allow site templates to extend the theme's template regardless of the theme name, so that you can swap themes seemlessly as long as they implement the same blocks in the base template.

I don't think it's common enough. And right now you can easily tests multiple themes and pick the one you prefer, which was the original intent.

Keats:
It looks good to me; anything else you want to add to the PR?

PR #1008, opened 2020-04-25
Access heading level in anchor-link.html

southerntofu:
This is a super trivial patch to access heading level as ~~lvl~~ `level` in anchor-link.html. The usecase is i really enjoy markdown-style headings on my blog. However, without this i have to use CSS tricks which means i can't make the # clickable. Here's what i want to do:

```
<a class="zola-anchor" href="#{{ id }}" aria-label="anchor link for {{ id }}">{% for i in range(end=level) %}#{% endfor %}</a> 
```

### DIFF
--- a/docs/content/documentation/content/linking.md
+++ b/docs/content/documentation/content/linking.md
@@ -39,7 +39,12 @@ This option is set at the section level: the `insert_anchor_links` variable on t
 
 The default template is very basic and will need CSS tweaks in your project to look decent.
 If you want to change the anchor template, it can be easily overwritten by
-creating an `anchor-link.html` file in the `templates` directory, which gets an `id` variable.
+creating an `anchor-link.html` file in the `templates` directory.
+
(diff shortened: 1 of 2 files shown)


docs/content/documentation/content/linking.md
```diff
@@ -39,7 +39,12 @@ This option is set at the section level: the `insert_anchor_links` variable on t
 
 The default template is very basic and will need CSS tweaks in your project to look decent.
 If you want to change the anchor template, it can be easily overwritten by
-creating an `anchor-link.html` file in the `templates` directory, which gets an `id` variable.
+creating an `anchor-link.html` file in the `templates` directory.
+
+The anchor link template has the following variables:
+
+- `id`: the heading's id after applying the rules defined by `slugify.anchors`
+- `lvl`: the heading level (between 1 and 6)
```
Keats: can you name it level?

Keats:
Thanks!

PR #1011, opened 2020-04-26
Fix path of generated page

ad-si:
I tried it with zola 0.10.1, and this is how it seems to work. Is this also how it's supposed to work? 🤔


Keats:
yes it will, thanks

PR #1023, opened 2020-05-12
Make path to config.toml independent from root dir

### LINKED ISSUE #1022: Path to config file should not be relative to root dir
# Bug Report
Command line description `-c, --config <config>    Path to a config file other than config.toml [default: config.toml]` doesn't suggest, that the path to a config file will be treated relatively to the root directory of a project.
https://github.com/getzola/zola/blob/79b2338bd3a0cda334e4f642696553a91e02b364/components/config/src/lib.rs#L8-L19

The misunderstanding occurs only when one wants to specify both the  `root` and  `config` parameters with values other than the default ones. Personally I think, that in this situation it's probably most likely that the user wants to use a custom config file from other location than the project directory.

Note that the second command line parameter taking a path as a value, `output_dir`, is not treated relatively to the project directory.

## Environment
Zola version: zola 0.10.1

## Expected Behavior
The path should be treated relatively to the current work directory with the current CLI description.

## Step to reproduce
Simply create a new site and then try to build it being outside of it:
```sh
zola init test
zola -r test/ -c test/config.toml build
```
what results in:
```
Building site...
Failed to load test/config.toml
Error: No `"config.toml"` file found. Are you in the right directory?
```

## Notes
I would change the description to something like:  
`Path to a config file other than config.toml in the project directory`,  
and then treat the path relatively to the cwd.  
I can implement this solution if it's ok.

satakuma:
fixes #1022

## Code changes

~~If the change is a new feature or adding to/changing an existing one:~~

* [ ] ~~Have you created/updated the relevant documentation page(s)?~~

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -62,9 +62,10 @@ impl<'a> SerializedTaxonomyItem<'a> {
 impl Site {
     /// Parse a site at the given path. Defaults to the current dir
     /// Passing in a path is used in tests and when --root argument is passed
-    pub fn new<P: AsRef<Path>>(path: P, config_file: &str) -> Result<Site> {
+    pub fn new<P: AsRef<Path>, P2: AsRef<Path>>(path: P, config_file: P2) -> Result<Site> {
         let path = path.as_ref();
-        let mut config = get_config(path, config_file);
+        let config_file = config_file.as_ref();
+        let mut config = get_config(config_file);
         config.load_extra_syntaxes(path)?;
 
         let tpl_glob =
--- a/completions/_zola.ps1
+++ b/completions/_zola.ps1
@@ -20,8 +20,8 @@ Register-ArgumentCompleter -Native -CommandName 'zola' -ScriptBlock {
 
     $completions = @(switch ($command) {
         'zola' {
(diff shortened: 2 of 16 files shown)


components/site/src/lib.rs
```diff
@@ -62,9 +62,10 @@ impl<'a> SerializedTaxonomyItem<'a> {
 impl Site {
     /// Parse a site at the given path. Defaults to the current dir
     /// Passing in a path is used in tests and when --root argument is passed
-    pub fn new<P: AsRef<Path>>(path: P, config_file: &str) -> Result<Site> {
+    pub fn new<P: AsRef<Path>, P2: AsRef<Path>>(path: P, config_file: P2) -> Result<Site> {
```
Keats, line 65: I think you can reuse `P` no?

Keats (commented):
Looks ok to me, I just need to try it locally to make sure I understand well the behaviour.

components/site/src/lib.rs
satakuma: It would force `path` and `config_file` to have the same type, what isn't necessarily required here. One may want to use for example &PathBuf and &str

But yes, currently this function is always called with `&PathBuf` so it doesn't matter that much + makes this line quite long.

Keats:
Thanks!

PR #1029, opened 2020-05-16
Fix demo links for docsascode theme

parlough:
The upper link did not work and the lower link used http prefix despite https being supported.

Sanity check:


Keats:
That needs to be fixed on the repo itself. These docs are generated from https://github.com/getzola/themes
If it has been updated in the repo, I just need to regenerate those pages.

PR #1032, opened 2020-05-18
Change get_url(cachebust=true) to use a hash

### LINKED ISSUE #519: Asset fingerprinting and SRI

dancek:
Cache-busting was previously done with a compile-time timestamp. Change
to the SHA-256 hash of the file to avoid refreshing unchanged files.

The implementation could be used to add a new global fn (say,
get_file_hash) for subresource integrity use, but that's for another
commit.

Fixes #519.

Sanity check:

## Code changes


dancek:
I'm not very seasoned with Rust, so I'll appreciate any review comments. Especially the `TestContext` part seems a bit questionable to me, but this way I *think* the directory should get cleaned up even when a test panics.

dancek:
```
error: failed to run custom build command for `onig_sys v69.2.0`
```
I don't think this is caused by my changes. 🤔

Keats:
Don't worry about that, it's being fixed upstream.

Keats:
Thanks!

PR #1033, opened 2020-05-22
Update lockfile to new format

XAMPPRocky:
This PR updates the lock file to use the new Cargo.lock format that was introduced in 1.41. See https://github.com/rust-lang/cargo/pull/7070 for more details on the format.


Keats:
Can you do it on the `next` branch if it's not already the case? Thanks

PR #1041, opened 2020-05-26
Error messaging in tera_fn

stanistan:
_This is a follow up to https://github.com/getzola/zola/pull/1036_

I wanted it to be a bit easier to identify that this error had come
from a specific function, and instead of adding that information to the
load_data function itself, wanted to investigate how errors were
reported in the first place.

```
Error: Failed to render page '/Users/stanistan/dev/stanistan.com/content/invoice/01.md'
Reason: Failed to render 'invoice.html'
>>> Reason: Failed in tera function call to `load_data` <<< NEW
Reason: Could not get response status for url: http://localhost:3000/invoice/01
```

---

This PR adds information to _every_ `global_fn` registered in tera by the zola.

The approach I've taken is:

1. create a `NamedTeraFn` trait which then means the `tera::Function` can have
   a statically associated name, the `site` no longer has to know the
   name of the function it is registering.
2. Wrap a `tera::Function` in another `tera::Function` which adds this
   name to any error that might have been produced, this is the
   `WrappedTeraFn`.
3. The functions are registered with a `register!` macro, which DRYs up
   those calls a bit.

LMK what you think-- I was also considering adding this functionality to `tera` itself, so that its transparent when calling the `register` methods and not here, but thought this was a better place to open this up for discussion given the context.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -9,7 +9,7 @@ use glob::glob;
 use rayon::prelude::*;
 use sass_rs::{compile_file, Options as SassOptions, OutputStyle};
 use serde_derive::Serialize;
-use tera::{Context, Tera};
+use tera::{Context, Tera, Function as TeraFn};
 
 use config::{get_config, Config, Taxonomy as TaxonomyConfig};
 use errors::{bail, Error, ErrorKind, Result};
@@ -19,6 +19,7 @@ use library::{
     TaxonomyItem,
 };
 use templates::{global_fns, render_redirect_template, ZOLA_TERA};
+use templates::global_fns::{NamedTeraFn, WrappedTeraFn};
 use utils::fs::{copy_directory, create_directory, create_file, ensure_directory_exists};
 use utils::net::get_available_port;
 use utils::templates::{render_template, rewrite_theme_paths};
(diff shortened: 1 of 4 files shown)


Keats:
I don't think this belongs in Tera and I don't know if it makes sense there. I would prefer making sure that that there are no expect/unwrap in the code than adding a trait.

stanistan:
For tera, it wouldn’t have to be a trait- something that wraps the current fn and reports an error message (like the struct in the PR). 

Agree on getting to all the experts/unwraps to avoid panics. I think the error messaging addition would also be helpful otherwise all error messages will have to know the function name from which they came. Which is kind of true now. 

---

Edit: 

More specifically, `WrappedTeraFn` can end up having another field, `name`.

```rust
struct <F> WrappedTeraFn {
    pub name: String,
    pub f: F
}
```

And the `Function` implementation for `call` uses `self.name` instead of `F::NAME`.

stanistan:
Looking again, we already have all of that context here in `tera`.

https://github.com/Keats/tera/blob/6c804557d619d1bcb63991757742aed7025211d0/src/renderer/processor.rs#L471-L483
Going to try something there.

PR #1044, opened 2020-06-01
Implement get_file_hash

dancek:
# ~WARNING: contains a bug.~

As mentioned in #519, both the `cachebust` commit I previously posted and this one look for static files directly in `content_path` when they're really in `static/`. I'll fix that tonight if I have time.

EDIT: it's fixed now, in the first commit.

---

Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/global_fns/mod.rs
+++ b/components/templates/src/global_fns/mod.rs
@@ -49,11 +49,11 @@ impl TeraFn for Trans {
 pub struct GetUrl {
     config: Config,
     permalinks: HashMap<String, String>,
-    content_path: PathBuf,
+    search_paths: Vec<PathBuf>,
 }
 impl GetUrl {
-    pub fn new(config: Config, permalinks: HashMap<String, String>, content_path: PathBuf) -> Self {
-        Self { config, permalinks, content_path }
+    pub fn new(config: Config, permalinks: HashMap<String, String>, search_paths: Vec<PathBuf>) -> Self {
+        Self { config, permalinks, search_paths }
     }
 }
 
@@ -74,11 +74,34 @@ fn make_path_with_lang(path: String, lang: &str, config: &Config) -> Result<Stri
     Ok(splitted_path.join("."))
 }
 
-fn compute_file_sha256(path: &PathBuf) -> result::Result<String, io::Error> {
-    let mut file = fs::File::open(path)?;
+fn open_file(search_paths: &Vec<PathBuf>, url: &String) -> result::Result<fs::File, String> {
+    let cleaned_url = url.trim_start_matches("@/").trim_start_matches("/");
+    for base_path in search_paths {
+        match fs::File::open(base_path.join(cleaned_url)) {
+            Ok(f) => return Ok(f),
+            _ => continue
+        };
+    }
+    Err(format!("file {} not found; searched in {}", url,
+        search_paths.iter().fold(String::new(), |acc, arg| acc + " " + arg.to_str().unwrap())))
+}
+
+fn compute_file_sha256(mut file: fs::File) -> result::Result<String, String> {
     let mut hasher = Sha256::new();
-    io::copy(&mut file, &mut hasher)?;
-    Ok(format!("{:x}", hasher.result()))
+    io::copy(&mut file, &mut hasher)
+        .and_then(|_| Ok(format!("{:x}", hasher.result())))
+        .map_err(|e| format!("{}", e))
+}
+
+fn get_cachebust_hash(search_paths: &Vec<PathBuf>, url: &String) -> Option<String> {
+    match open_file(search_paths, url).and_then(compute_file_sha256) {
+        Ok(hash) => Some(format!("?h={}", hash)),
+        Err(e) => {
+            // NOTE: there should be a better way to print warnings
+            eprintln!("WARN: `get_url`/`cachebust`: {}", e);
+            None
+        }
+    }
 }
 
 impl TeraFn for GetUrl {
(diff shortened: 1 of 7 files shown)


dancek:
This is now two commits, with the first one fixing my previous blunders with `cachebust`. I erroneously read files from `content_path` when static files actually are in `static_path`.

The implementation has different functions `compute_file_sha256`, `compute_file_sha384` and `compute_file_sha512` for different hash types, which looks a bit repetitive and dumb.

Another option is to make the calls like `compute_file_hash<Sha256>`, but that requires a complicated trait bound to the function:
```rust
fn compute_file_hash<T>(path: &PathBuf) -> result::Result<String, io::Error>
where
    T: Digest + io::Write,
    <T as Digest>::OutputSize: ops::Add,
    <<T as Digest>::OutputSize as ops::Add>::Output: generic_array::ArrayLength<u8>
{
    let mut file = fs::File::open(path)?;
    let mut hasher = T::new();
    io::copy(&mut file, &mut hasher)?;
    Ok(format!("{:x}", hasher.result()))
}
```

That would also add a direct dependency on `generic_array` (though sha2 depends on it anyway). I'll switch to that version if you prefer it. I just thought it's better to err on the side of readability in a FLOSS project.

Also, let me know if you'd rather have a separate PR and possibly a bug report for the first commit.

dancek:
Oh boy, the cache-busting is still broken on generated assets (compiled scss).

dancek:
Ok, the final try. Both the cachebusting implementation and `get_file_hash` will search in `static_path`, `output_path` and `content_path`. If none of them works, cachebusting reverts to timestamp and logs a warning to stderr. I realized it's annoying to crash just because cachebusting couldn't be done in the best possible way.

Hope this is good.

dancek:
Oh, `build_timestamp` was removed after I branched from `next`. Let's error out if the file is not found, then.

dancek:
The windows build seems to be failing because the files in `test_site` have different SHA sums.
```
      <link href="https://replace-this-with-your-url.com/site.css?h=8661df0fd2f603ff954e7e04b66d014c90c72305855e5c360bff4813186d122b" rel="stylesheet">
...
thread 'can_cachebust_static_files' panicked at 'assertion failed: file_contains!(public, "index.html",
               "<link href=\"https://replace-this-with-your-url.com/site.css?h=83bd983e8899946ee33d0fde18e82b04d7bca1881d10846c769b486640da3de9\" rel=\"stylesheet\">")', components\site\tests\site.rs:692:5
```

This seems to be because git converts line endings on Windows.
```
$ unix2dos test_site/static/site.css && sha256sum test_site/static/site.css
unix2dos: converting file test_site/static/site.css to DOS format...
8661df0fd2f603ff954e7e04b66d014c90c72305855e5c360bff4813186d122b  test_site/static/site.css
``

I'll see if this can be fixed with `.gitattributes` later when I have time.

File: components/templates/src/global_fns/mod.rs [hunk 1]
```diff
@@ -74,11 +74,35 @@ fn make_path_with_lang(path: String, lang: &str, config: &Config) -> Result<Stri
...
+        .and_then(|_| Ok(format!("{:x}", hasher.result())))
+        .map_err(|e| format!("{}", e))
+}
+fn compute_file_sha384(mut file: fs::File) -> result::Result<String, String> {
+    let mut hasher = Sha384::new();
+    io::copy(&mut file, &mut hasher)
+        .and_then(|_| Ok(format!("{:x}", hasher.result())))
+        .map_err(|e| format!("{}", e))
+}
+fn compute_file_sha512(mut file: fs::File) -> result::Result<String, String> {
```
Keats: Can we not return an Error rather than a string so we can use `?`?

docs/content/documentation/templates/overview.md
```diff
@@ -146,6 +146,21 @@ In the case of non-internal links, you can also add a cachebust of the format `?
...
+
+```jinja2
+{{/* get_file_hash(path="js/app.js", sha_type=256) */}}
+```
+
+This can be used to implement subresource integrity. Do note that subresource integrity is typically used when using external scripts, which `get_file_hash` does not support.
+
+```jinja2
+<script src="{{/* get_url(path="js/app.js") */}}"
+        integrity="sha384-{{/* get_file_hash(path="js/app.js", sha_type=256) */}}"></script>
```
Keats: sha256 at the beginning no? or remove the sha_type argument

docs/content/documentation/templates/overview.md
```diff
@@ -146,6 +146,21 @@ In the case of non-internal links, you can also add a cachebust of the format `?
 by passing `cachebust=true` to the `get_url` function.
 
 
+### 'get_file_hash`
```
Keats, line 149: can that doc mention something about the search paths?

Keats (changes requested):

dancek:
Thanks for reviewing! Good points, I'll fix these.

docs/content/documentation/templates/overview.md
dancek: Yes, this was supposed to be 384 (as that's recommended with SRI in general).

docs/content/documentation/templates/overview.md
dancek: Good point. I wrote a bit on this.

components/templates/src/global_fns/mod.rs
dancek: I'll see what I can do. I had some difficulties getting everything to work the way I want, but that's just because I'm a Rust newbie. :)

dancek: Yeah, this is better. I split the error message creation into its own function.

dancek:
I tried to fix the Windows tests by forcing `eol=lf` in `.gitattributes` for the relevant files and fixed the issues from review comments.

Keats:
Thanks!

PR #1051, opened 2020-06-05
Don't panic on bad date strings

### LINKED ISSUE #993: Zola panics on bad date strings
# Environment

Zola version: 0.10.1

## Expected Behavior

There should be an error message that the date string is badly formatted, instead of a cryptic error message.

## Current Behavior

```
$ env RUST_BACKTRACE=1 zola serve                                                                                                                                                (base) 
Building site...
-> Creating 15 pages (0 orphan), 1 sections, and processing 0 images
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /private/tmp/rust-20200227-78610-djkaw6/rustc-1.41.1-src/src/libcore/macros/mod.rs:15:40
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::rust_panic_with_hook
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::panicking::panic
   8: library::sorting::sort_actual_pages_by_date
   9: rayon::slice::quicksort::shift_tail
  10: rayon::slice::quicksort::recurse
  11: site::Site::render_rss_feed
  12: site::Site::build
  13: zola::cmd::serve::create_new_site
  14: zola::cmd::serve::serve
  15: zola::main
  16: std::rt::lang_start::{{closure}}
  17: main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

## Step to reproduce

Make a page with a badly formatted date, like so (it should have a `+00:00` at the end).

```md
+++
title = "Building 1:1 Scale Minecraft Replicas with Google Maps"
template = "post.html"
date = 2020-04-12T19:52:46
+++
```

## Side note

How are you supposed to generate the type of date string that zola wants? On MacOS, the `date` command doesn't have an ISO8601 mode, and Python `datetime.now().isoformat()` outputs something that Zola doesn't understand (and makes it panic).

dancek:
Instead, show a helpful error message explaining only RFC3339 is
supported.

Fixes #993.

Sanity check:

## Code changes

### DIFF
--- a/components/front_matter/src/page.rs
+++ b/components/front_matter/src/page.rs
@@ -57,6 +57,18 @@ pub struct PageFrontMatter {
     pub extra: Map<String, Value>,
 }
 
+/// Parse a TOML datetime from a string.
+/// There are three alternatives: an offset datetime (plain RFC3339), a local datetime
+/// (RFC3339 with timezone omitted) and a local date (YYYY-MM-DD). This tries each in
+/// order.
+fn parse_datetime(d: &String) -> Option<NaiveDateTime> {
+    DateTime::parse_from_rfc3339(d)
+        .or(DateTime::parse_from_rfc3339(format!("{}Z", d).as_ref()))
+        .map(|s| s.naive_local())
+        .or(NaiveDate::parse_from_str(d, "%Y-%m-%d").map(|s| s.and_hms(0, 0, 0)))
+        .ok()
+}
+
 impl PageFrontMatter {
     pub fn parse(toml: &str) -> Result<PageFrontMatter> {
         let mut f: PageFrontMatter = match toml::from_str(toml) {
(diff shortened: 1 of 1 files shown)


dancek:
I realize it would be better to support full TOML dates, but that seems nontrivial as `chrono` has quite rigid parsing and `toml::value::Datetime` has private structs for its data.

This will get rid of the panics, at least.

dancek:
More specifically, the panics are caused when code somewhere checks `date.is_some()` or such, but actually depends on `datetime.is_some()`. That happens at least when rendering feeds.

dancek:
That was easier than I thought. Or rather, `DateTime::parse_from_rfc3339` is smarter than I gathered from its documentation. This should parse the full range of the TOML spec. I wouldn't trust this to be exactly the same as the parser in the `toml` crate, though, so checking that both `self.date` and `self.datetime` are `Some` is still useful IMHO.

Keats:
Thanks!

PR #1057, opened 2020-06-09
add .dockerignore file to reduce intermediate docker image size and build time

thomasetter:
This reduced the intermediate container size by a few GBs (mainly because `target/` is that large) for me when building locally and made the build faster because of that.

When working on the `Dockerfile`, this should greatly speed up the development because the changes on the file itself do not invalidate the build results.

This also helps to keep the build more reproducible as it avoids e.g. the contents of `target/` being used for the docker build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+.dockerignore
+.git*
+Dockerfile
+
+# From .gitignore
+target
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

Issue #905, opened 2019-12-27
Support for non-HTML templates


rotty:
# Feature Request

## Environment

Zola version: 0.9.0

## Feature Proposal

As I've initially described this in #840, it is currently impossible (AFAICT) to manually emulate the `generate-rss` feature by using an appropriate template, as Zola assumes all pages to render to HTML output. There are two aspects to this:

- Zola does not handle templates with an extension different to `.html`. For an RSS feed template, it would make sense to have a non-HTML extension for the template, such as `.xml` or `.rss`. As such non-HTML templates are currently not supported, it should be possible to enable special behavior when adding support for such templates; see below.
- For non-HTML output, the "beautiful URL" scheme employed, i.e., generating "foo/index.html" instead of "foo.html", doesn't make sense, as webservers typically just serve `index.html` when a request that maps to a directory name comes in.

Thus, I'd like to discuss the following potential change to Zola:

> When a template is specified with a path that does not end in `.html`, use the extension from the template to form the resulting path, and don't do the `index.html` transformation. For example, a `foo.md` page specifying `template = "feed.xml"` in its frontmatter will be rendered to `section/foo.xml` instead of `section/foo/index.html`. When a `path` is specified for such a page, the given path would be used verbatim.

With this change, I think one would get pretty close to being able to implement the `generate-rss` feature in "user-space", so to speak, which would be nice from an elegance and versatility perspective, besides allowing for my specific use-case.

## Drawbacks

Implementing the proposed change will still treat templates with the `.html` extension specially, as it does the `index.html` transformation. However, if that were to be made configurable, as proposed in #840, `.html` and other templates would be treated exactly the same, which mitigates that concern, I think.

Keats:
I like that change. I don't really want to support ugly urls if possible so I don't mind not having a configuration option for now.

I'm wondering if it conflicts with https://zola.discourse.group/t/proposal-custom-output-formats/68 or not at all, I'm still unclear on that.

rotty:
Quoting from [your comment on #840](https://github.com/getzola/zola/issues/840#issuecomment-569284995):

> Hmm so for that specific usecase I think it would make sense for section to be able to have a `rss = true` setting that automatically create a RSS feed for that section which could use its own issue.

[...]

PR #1059, opened 2020-06-11
[RFC] non-HTML output

dancek:
I'm posting this for discussion. Feedback is welcome and I'm willing to work on this.

I'll switch my own gemini site to use this probably this weekend.

---
Support non-HTML templates

- allow templates with any file extension
- use the template extension for output filenames
- add plain_content that has been rendered with Tera, but not
  markdown->HTML
- add simple example/test: test_site_gemini
- write basic documentation

See discussion in issue #905.
---

Sanity check:

## Code changes

### DIFF
--- a/components/library/src/content/ser.rs
+++ b/components/library/src/content/ser.rs
@@ -57,6 +57,7 @@ impl<'a> TranslatedContent<'a> {
 #[derive(Clone, Debug, PartialEq, Serialize)]
 pub struct SerializingPage<'a> {
     relative_path: &'a str,
+    plain_content: &'a str,
     content: &'a str,
     permalink: &'a str,
     slug: &'a str,
@@ -121,6 +122,7 @@ impl<'a> SerializingPage<'a> {
         SerializingPage {
             relative_path: &page.file.relative,
             ancestors,
+            plain_content: &page.plain_content,
             content: &page.content,
             permalink: &page.permalink,
             slug: &page.slug,
@@ -178,6 +180,7 @@ impl<'a> SerializingPage<'a> {
         SerializingPage {
             relative_path: &page.file.relative,
             ancestors,
+            plain_content: &page.plain_content,
             content: &page.content,
             permalink: &page.permalink,
             slug: &page.slug,
@@ -211,6 +214,7 @@ impl<'a> SerializingPage<'a> {
 #[derive(Clone, Debug, PartialEq, Serialize)]
 pub struct SerializingSection<'a> {
     relative_path: &'a str,
+    plain_content: &'a str,
     content: &'a str,
     permalink: &'a str,
     ancestors: Vec<String>,
@@ -252,6 +256,7 @@ impl<'a> SerializingSection<'a> {
         SerializingSection {
             relative_path: &section.file.relative,
             ancestors,
+            plain_content: &section.plain_content,
             content: &section.content,
             permalink: &section.permalink,
             title: &section.meta.title,
@@ -289,6 +294,7 @@ impl<'a> SerializingSection<'a> {
         SerializingSection {
             relative_path: &section.file.relative,
             ancestors,
+            plain_content: &section.plain_content,
             content: &section.content,
             permalink: &section.permalink,
             title: &section.meta.title,
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -69,7 +69,7 @@ impl Site {
         config.load_extra_syntaxes(path)?;
 
         let tpl_glob =
-            format!("{}/{}", path.to_string_lossy().replace("\\", "/"), "templates/**/*.*ml");
+            format!("{}/{}", path.to_string_lossy().replace("\\", "/"), "templates/**/*");
         // Only parsing as we might be extending templates from themes and that would error
         // as we haven't loaded them yet
         let mut tera =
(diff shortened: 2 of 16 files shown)


components/library/src/content/ser.rs
```diff
@@ -178,6 +180,7 @@ impl<'a> SerializingPage<'a> {
         SerializingPage {
             relative_path: &page.file.relative,
             ancestors,
+            plain_content: &page.plain_content,
```
Keats, line 183: That approach is a no-go, it will make Zola roughly twice slower since it has to serialize the content twice in the HTML case.

components/site/src/lib.rs
```diff
@@ -69,7 +69,7 @@ impl Site {
         config.load_extra_syntaxes(path)?;
 
         let tpl_glob =
-            format!("{}/{}", path.to_string_lossy().replace("\\", "/"), "templates/**/*.*ml");
+            format!("{}/{}", path.to_string_lossy().replace("\\", "/"), "templates/**/*");
```
Keats, line 72: That's going to mess up with various editor swap files..

docs/content/documentation/templates/non-html.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Whitespace
+
+For text-based formats, the [whitespace control](https://tera.netlify.app/docs/#whitespace-control) is quite useful. An useful pattern is removing the whitespace at the end whenever there's a Tera delimiter at the end a line (that is, use `-}}` or `-%}`).
+
+## Markdown rendering
+
+The main focus of Zola is in rendering markdown to HTML. That's what the `markdown` filter does and `page.content` is the resulting HTML.
+
+For non-HTML formats it's usually better to use `page.plain_content` which has Tera expressions and statements evaluated, but no HTML conversion.
```
Keats, line 28: it's only shortcodes that are rendered no?

Keats (changes requested):
I think the 2-steps rendering approach is not great perf wise.
I'm also wondering how this features intersects with https://zola.discourse.group/t/proposal-custom-output-formats/68 which is a generalisation of that, and which allows multiple output formats.
It probably needs some more thoughts I think

components/library/src/content/ser.rs
dancek: Alright, I'll have to either find a way to only serialize one of them or to reuse `page.content`. Didn't realize serializing is the bottleneck.

components/site/src/lib.rs
dancek: Oh, right. I have vim configured with `directory=~/.local/share/nvim/swap//` so 
I didn't notice that. If this is a bad problem, the target file extension needs to be set in front matter. Or alternatively there needs to be another way of selecting files, eg. multiple globwalk patterns or a regex, but that would require changes to Tera.

docs/content/documentation/templates/non-html.md
dancek: I thought that control structures such as `for` in Tera are different from shortcodes.

dancek: Oh, I see your point. Only shortcodes in content are rendered, indeed. In templates, the full range of Tera is rendered.

dancek:
> I'm also wondering how this features intersects with https://zola.discourse.group/t/proposal-custom-output-formats/68 which is a generalisation of that, and which allows multiple output formats.

I've thought of that, and for different kinds of wishlist items there are different considerations. I think all are doable with this basic approach; it's more a matter of which features are worth adding and maintaining.

## Page listings (feeds, search indices, sitemaps, ...)

It's possible to create a page listing in a custom format for a single section already with this commit. Caveats include

* The url is always `something/index.*`. Supporting ugly urls would help this, and shouldn't be very hard.
* The normal section `_index.md` is usually used to generate normal content, so you need to create something like `atom.md` rendered with a specific template to create a feed for the section. Multiple templates per page would help.
* To create a site-wide feed it would be very useful to have a function like `get_all_pages`. It might be doable by crawling subsections form the root index.

## Multi-format output per page

This is not directly possible with this commit. It would require implementing new array-based front matter entries for `templates` and `page_templates`. Then you could use multiple templates and get multiple output formats in one go. This is not necessary for my personal use, but I do understand the appeal.

The main caveats here pertain to content formatting:

* Markdown is the only possible input format, and only rendering markdown to HTML is supported. Other formats can include page content as either raw markdown or HTML. This will make other formats second class citizens. IMHO this is acceptable.
* Shortcodes need some way to output different things based on format. Setting the file extension as a page variable is enough to support this.

## Summa summarum

These two larger features (and a bunch of smaller changes) would be useful additions.

1. **Ugly urls**. The question is, should the decision be per page, per template or per filetype.
2. **Multiple templates**

I'm happy to work on these if it seems useful. Of course if this seems like out of scope for Zola or if another approach is better, that's OK too. What do you think, @Keats?

Keats:
I'm still not 100% sold on the custom output format idea since I haven't seen many examples using it, do you know any?

As for ugly urls, it's probably not worth the effort, very few people asked for it.

dancek:
Well, I think it would be good if RSS/Atom feeds didn't need any special functionality. I can come up with reasons to generate robots.txt, iCalendar feeds or some generic data formats (json, csv), but who knows if it would happen in practice. Certainly many CMS-based websites have APIs like that, and many of those could be statically generated.

Of course, being everything to everyone can turn out to be a maintenance nightmare. So maybe scoping Zola to only Markdown -> only HTML is a good decision. In that case, I think #905 should be closed as wontfix. As for my own needs, I can work with my own fork or possibly build another tiny SSG for Gemini.

dancek:
I think a plaintext-only SSG would be a better fit for gopher, gemini and such. Closing.

skyfaller:
@dancek Wait, where are you going instead? Take me with you...

dancek:
@skyfaller I have some ideas of a plaintext-output SSG (for Gemini, Gopher etc), but nothing concrete. It should be much simpler than targeting the web (and as a side effect would not be suitable for HTML output). If I write one I'll announce it on the Gemini mailing list and my gemlog.

PR #1063, opened 2020-06-15
Add support for SVG files to `get_image_metadata` (Fixes #769)

### LINKED ISSUE #769: Support get_image_metadata for SVG

mre:
This is the second attempt to fix #769. 🤞 
Tested it on my blog and it was working nicely.

Tried a couple of variants to write the size calculation idiomatically, but if anyone has a better idea, I'm open for suggestions.

#### Important caveat
If a user works with SVGs on their site **and** an SVG neither provides width/height nor viewport information **and** the user is calling `get_image_metadata()`, the build will fail with an error message, because zola could not determine the image proportions. I think that's an acceptable tradeoff, though as the user was explicitly asking for the metadata in a template. I'd even go so far to say that it's expected behavior to stop compilation instead of silently rendering with missing image proportions, but it's certainly something to be aware of.

### DIFF
--- a/components/templates/src/global_fns/mod.rs
+++ b/components/templates/src/global_fns/mod.rs
@@ -2,9 +2,11 @@ use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, RwLock};
 use std::{fs, io, result};
+use std::ffi::OsStr; 
 
 use sha2::{Digest, Sha256, Sha384, Sha512};
 use tera::{from_value, to_value, Error, Function as TeraFn, Result, Value};
+use svg_metadata as svg;
 
 use config::Config;
 use image;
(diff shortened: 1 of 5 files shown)


components/templates/src/global_fns/mod.rs
```diff
@@ -279,15 +281,31 @@ impl TeraFn for GetImageMeta {
...
-        map.insert(String::from("height"), Value::Number(tera::Number::from(img.height())));
-        map.insert(String::from("width"), Value::Number(tera::Number::from(img.width())));
+        map.insert(String::from("height"), Value::Number(tera::Number::from(height)));
+        map.insert(String::from("width"), Value::Number(tera::Number::from(width)));
         Ok(Value::Object(map))
     }
 }
 
+// Try to read the image dimensions for a given image
+fn image_dimensions(path: &PathBuf) -> Result<(u32, u32)> {
```
Keats, line 293: that's a place where anonymous structs would be very helpful, no need to member which u32 is height and width.

Keats (changes requested):
Looks good thanks!
Can you update the docs to mention SVG support?

components/templates/src/global_fns/mod.rs
mre: True. We could add a `Dimensions` struct instead, but I think that would unnecessarily increase the complexity.

mre:
Updated the Markdown documentation now. Added the supported image formats. Hope that's what you meant.

mre:
Also I noticed that we don't support Webp yet, which is another format we could add in the future. (Use it on my blog, too since Firefox started supporting it.)

Keats:
Thanks!
For webp, is it supported by `image` or would it be another crate?

mre:
Just checked and webp should be supported by `image`. Now I wonder why it doesn't work out of the box if I try it on my blog. 🤔

Jieiku:
I seem to be getting backwards info: https://github.com/getzola/zola/issues/1877

height reports width, and width report height

PR #1065, opened 2020-06-16
Add --force/-f flag to init, for creation in non-empty dir

perlun:
Sorry for not discussing this first. I needed the feature for my own good, since I was about to create my first Zola web site (:tada:) tonight and I had the previous (Jekyll) version of the site in a submodule already. It was easier/more fun to try to add this missing flag to Zola rather than trying to workaround it... :see_no_evil: 

Note that `--force` is not as evil as one might think in this case. It merely bypasses the "folder exists" check. We could do a real `--force` flag that overwrites any existing content & configuration, but that's definitely something much more dangerous than what this PR aims to achieve.

----

Sanity check:

## Code changes


Keats:
Thanks!

PR #1073, opened 2020-07-04
Katex

technicalguy:
[Discourse thread](https://zola.discourse.group/t/use-mathematical-formula-with-zola/308)

Sanity check:

## Code changes

* The change is a new feature.


technicalguy:
I don't know why the windows build isn't working, but perhaps it's related to this: https://github.com/xu-cheng/katex-rs/issues/1#issuecomment-633731592

technicalguy:
@Keats requesting a review :)

Keats:
I would prefer to review it once all the CI pass, this Windows target should work

technicalguy:
I don't actually know what I need to do to fix the windows target.

liushuyu:
> I don't actually know what I need to do to fix the windows target.

I found that by upgrading `katex` dependency will fix the build. Just `cargo update` and it should do.

technicalguy:
> I found that by upgrading `katex` dependency will fix the build. Just `cargo update` and it should do.

I've tried this and I don't think it's working :/

l0calh05t:
What is the blocker right now? The checks list failures, but when I try to look at the failure details, I only get a "Build not found." message. In any case, I would love to see this feature and I couldn't replicate any build or test failures on my Windows machine.

components/rendering/src/lib.rs
```diff
@@ -2,21 +2,28 @@ mod context;
...
-    if content.contains("{{") || content.contains("{%") {
-        let rendered = render_shortcodes(content, context)?;
-        let html = markdown_to_html(&rendered, context)?;
-        return Ok(html);
+    if preprocessed.contains("{{") || preprocessed.contains("{%") {
+        preprocessed = render_shortcodes(&preprocessed, context)?;
     }
 
-    markdown_to_html(&content, context)
+    preprocessed = render_katex(&preprocessed);
```
Keats, line 24: this needs to be gated on some config option, I can add it myself later. Any preference on the name? `katex = true`?

Keats (commented):
Can you rebase and resolve the conflict? I can't re-run the pipeline it seems.

components/rendering/src/lib.rs
technicalguy: yeah sure, or `render_latex_math = true`?

technicalguy:
Just did a rebase.

@l0calh05t, It works fine for me on Linux, and I don't have a windows machine to test it on. I think it's something to do with `libquickjs`. If you can figure it out that would be great!

l0calh05t:
> 
> 
> Just did a rebase.
> 
> @l0calh05t, It works fine for me on Linux, and I don't have a windows machine to test it on. I think it's something to do with `libquickjs`. If you can figure it out that would be great!

I'll gladly check things on Windows. It appears I built the wrong branch yesterday, because I definitely can't build on Windows now. It also fails on the `patch`-command like in the pipeline, except that I get a file not found error (I have no `patch` in my `PATH`). In any case, it appears QuickJS doesn't support Windows with MSVC. As evidenced by:

- https://github.com/theduke/quickjs-rs/issues/72
- https://github.com/theduke/quickjs-rs/blob/master/libquickjs-sys/build.rs which directly uses `gcc`-Flags to build QuickJS
- https://github.com/mengmo/QuickJS-Windows-Build which also uses MinGW to build on Windows

Is there any possible alternative to QuickJS?

l0calh05t:
There is also https://github.com/c-smile/quickjspp which adds MSVC-support to QuickJS (but requires premake to build). And https://github.com/boa-dev/boa which is a JS interpreter written in pure Rust (but still "experimental")

components/rendering/src/lib.rs
greizgh: Since katex and mathjax have slightly different feature sets, maybe make it clear that rendering rely on katex?
`katex = true` seems a good choice.

l0calh05t: I'd also prefer `katex = true` (Does MathJax require processing at render time?)

technicalguy:
> There is also https://github.com/c-smile/quickjspp which adds MSVC-support to QuickJS (but requires premake to build). And https://github.com/boa-dev/boa which is a JS interpreter written in pure Rust (but still "experimental")

The katex-rs crate depends on quickjs, but I don't see why one shouldn't be able to replace it with something else.

technicalguy:
AFAIK MathJax only works as a client-side renderer (scanning the page for
LaTeX to convert).

l0calh05t:
> 
> 
> > There is also https://github.com/c-smile/quickjspp which adds MSVC-support to QuickJS (but requires premake to build). And https://github.com/boa-dev/boa which is a JS interpreter written in pure Rust (but still "experimental")
> 
> The katex-rs crate depends on quickjs, but I don't see why one shouldn't be able to replace it with something else.

Quick update on this: I managed to get quickjs-rs to *compile* on MSVC (https://github.com/theduke/quickjs-rs/issues/72, https://github.com/l0calh05t/quickjs-rs) using quickjspp. However, it crashes immediately. I believe it is an issue at the FFI, as the output of the C code looks sane, but the input after going through Rust does not.

I also tried to run a Katex using Boa, and that doesn't work either (fails with a `for in loops not supported`-error).

technicalguy:
I intend to rebase this branch and make a new PR at some point, but as of yet I still don't know what's wrong with libquickjs and Windows, unfortunately :(

Keats:
(my bad I didn't mean to close it)

edmellum:
Is [Rex](https://github.com/ReTeX/ReX) an alternative to katex-rs here that could side-step all the compilation issues being pure Rust? 😄

l0calh05t:
For one:

 

> First note that ReX is currently in heavy development and is not intended to be used in any way other than for testing and debugging.

 

Furthermore, it renders to SVG or Canvas and accessibility is a lot worse than KaTeX (which produces MathML for semantics and HTML for visualization)

l0calh05t:
> Is [Rex](https://github.com/ReTeX/ReX) an alternative to katex-rs here that could side-step all the compilation issues being pure Rust? 😄

For one:

>  First note that ReX is currently in heavy development and is not intended to be used in any way other than for testing and debugging.


Furthermore, it renders to SVG or Canvas and accessibility is a lot worse than KaTeX (which produces MathML for semantics and HTML for visualization)

edmellum:
Huge difference in accessibility! I had no idea, thanks for taking the time to expound 😄

Keats:
I'm closing this because it's stale now but if someone manages to make Katex work on all OSes, please open a new PR!

micahcantor:
So I took a look at this because I think katex rendering would be useful and possible. I believe the original PR used the [pcre2 crate](https://github.com/BurntSushi/rust-pcre2) for regex which is by its own admission not fully featured or well supported. This might have been why the Windows builds were failing (though I know other have said there were issues with the katex crate dependencies, that might have to be looked at as well).

I suggest switching to the [regex](https://docs.rs/regex/1.5.4/regex/) crate supported by Rust language. Much of the code can be used, but because the regex crate does not support look around, which the original PR uses, the regex patterns to match the $...$ of inline math and $$ ... $$ of display math will have to be altered.

Unfortunately, I have no clue how to write regex but if there were someone else willing to take a look at updating those capture patterns, I could create a new PR with the changes I have so far, which add a markdown option for katex and make sleight adjustments with the switch to the regex crate. 

I don't have a windows machine to test on but the changes I made compiled, but failed at run time due to the unsupported refer features.

[Here's a link to the changes in the fork](https://github.com/micahcantor/zola/pull/1/files).

Keats:
Would [oniguruma](https://github.com/rust-onig/rust-onig) work as a regex engine?

technicalguy:
> I believe the original PR used the [pcre2 crate](https://github.com/BurntSushi/rust-pcre2) for regex which is by its own admission not fully featured or well supported.

I used lookbehind and lookahead for the regular expressions because it's the easiest way to deal with escaping dollar symbols using backslash. e.g. `\$1.33,\$2.55,\$ 3.96\$1.01` should not use KaTeX rendering mode at any point. It also makes it easier to distinguish between single and double dollar. I don't know how hard it would be to avoid using lookaround (or if it is possible to do it without).

My inclination would be to try oniguruma instead.

l0calh05t:
> So I took a look at this because I think katex rendering would be useful and possible. I believe the original PR used the [pcre2 crate](https://github.com/BurntSushi/rust-pcre2) for regex which is by its own admission not fully featured or well supported. This might have been why the Windows builds were failing (though I know other have said there were issues with the katex crate dependencies, that might have to be looked at as well).
> 

From what I remember `quickjs` (used by `katex-rs`) not supporting Windows was the main issue, see also https://github.com/xu-cheng/katex-rs/issues/6

technicalguy:
> From what I remember `quickjs` (used by `katex-rs`) not supporting Windows was the main issue, see also [xu-cheng/katex-rs#6](https://github.com/xu-cheng/katex-rs/issues/6)

Yes, that is what I think. I don't use Windows nor do I have any interest in adding windows support for `quickjs`.

uwni:
The katex-rs has added two new js engine in addition. Maybe this problem could be solved

technicalguy:
> The katex-rs has added two new js engine in addition. Maybe this problem could be solved

Indeed! I've added it to my to-do list to look into this, but no idea how long it will take me to get around to it.

PR #1084, opened 2020-07-11
Ignore sass directories starting with _

iirelu:
Prior to this change, only sass files starting with _ were ignored by Zola's sass compiler. This made using sass frameworks incredibly inconvenient, as Zola attempted to compile every single sass file in the framework, and inevitably errored due to them not being standalone. For example, to use the Bulma framework, you had to manually add an underscore to the beginning of *every* sass file in it so Zola would stop trying to compile them as standalone css files. Now you can change the directory name to _bulma and have the same result.


Keats:
Thanks!

PR #1085, opened 2020-07-11
Allow emitting newlines and whitespace in shortcodes and introduce markdown shortcodes

eaon:
Preface: I personally wouldn't call this a new feature but a remedy (though I'm hesitant to call it a fix) for a hack that is currently in place.

I attempted to use a shortcode in a theme I'm working on that would also be used to render preformatted text with markdown. To my surprise multiple newlines and whitespaces following newlines were scrapped. My investigation led me here. It isn't the most beautiful solution but it doesn't seem to break any scenarios that I came up with so far.

Happy to make changes when given feedback. Thanks!


Keats:
I'm away for a week but it looks good. I'll check it when I'm back.

Keats:
I was hoping that https://zola.discourse.group/t/problems-generating-complex-markdown-in-shortcodes/477 would work but that's not the case :(
The root issue is the same at the one you encountered I think but this patch does not fix it as well. It would be nice to have any shortcode rendering markdown to work :/

eaon:
I've not used TOCs myself but I'll have a look at how to better integrate shortcode output it if you'd like.

eaon:
Just got some time to have a look and *now* I get it. Using shortcodes like this never occured to me before, but I guess that makes sense. I do think that an easy way to handle this would be to have markdown shortcodes (with an .md suffix). In a way that would be nicer (IMO :shrug:) by making it explicit that that is something you can do with shortcodes. If you think that's a reasonable approach, I'll happily work on that feature.

Keats:
The issue is not really ToC but more the markdown rendering where you need to preserve new lines.
Maybe keeping the newlines as is (eg removing the code that removes them like you do in that PR) but using `trim_start` on every lines of the rendered shortcodes could work? This way we can't trigger the code block issue and markdown content should be rendered as expected.
Ideally we wouldn't need specific .md shortcode rendering, it should just work

eaon:
I think the conflict there is `pre` tags still (which was my original concern). Pre-formatted text would break again if only `trim_start` is used, and even if it is only used on output *other* than the inside of `pre` tags, the whitespace that may be rendered inside them could subsequently trip up the markdown parser again, no? Checking for pre-formatted text and applying my `<!--\n-->` workaround only for that and `trim_start` for the rest may work though. Not sure I love it, but I can try.

Keats:
Sorry what's the issue with `pre` tags? I meant to use `trim_start` only on the output of the shortcode renderer, ie there: https://github.com/getzola/zola/pull/1085/files#diff-87e6423cab04c90e477134980c7d11fdR129

It would be annoying for people with shortcodes where they rely on the indentation somehow but I don't think it's problematic?

eaon:
To quote from above:

> I attempted to use a shortcode in a theme I'm working on that would also be used to render preformatted text with markdown.

It was my whole point of even looking at this in the first place ;)

I'm not sure if you mean to imply that but just applying `trim_start` there in my pull request wouldn't do much because all newlines are replaced with HTML comments, breaking all "markdown shortcodes" because their newlines are lost before the shortcode output is handed over to the markdown parser along with the rest of the page.

As far as I can tell, if you'd like to go with a `trim_start` based solution, this pull request is void.

Keats:
> It was my whole point of even looking at this in the first place ;)

Ah sorry, let's not do just trim_start then :( Do you have the shortcode you are writing somewhere? I'm curious what it is

I ran this branch on the repo from discourse above and it worked okayish, except the first title that contained the whole entry for some reason.

eaon:
```
<figure>
<blockquote{% if uri %} cite="{{ uri | safe }}"{% endif %}>
{{ body | markdown | safe }}
</blockquote>
{% set atu = author or title or uri -%}
{% if atu or note -%}
<figcaption>{% if author %}— {{ author }}{% endif %}{% if author and title or author and uri %} in {% endif %}{% if title or uri %}<cite>{% if uri %}<a href="{{ uri | safe }}">{% endif %}{{ title | default(value=uri | safe) }}{% if uri %}</a>{% endif %}</cite>{% endif %}{% if note %}{% if atu and not note is starting_with(",") %}. {% endif %}{{ note | markdown(inline=true) | safe }}{% endif %}</figcaption>
{% endif -%}
</figure>
```

And then the quote might be from a tech focused mailing list with inline code ([e.g.](https://marc.info/?l=openbsd-tech&m=159524634423806&w=2))

> I ran this branch on the repo from discourse above and it worked okayish, except the first title that contained the whole entry for some reason.

Yes, makes sense, I think. I assume since all the newlines will be comments when it's passed to the rest of the document, the markdown parser thinks everything is part of a really big `<h2>` heading :smile:

So what do you think is the better route to take, distinguishing between `.md` and `.html` shortcodes (btw, I think I got that for the most part, but for some reason my glob pattern never seems to load `.md` files?), or `trim_start` combined with special treatment for the inside of `<pre>` tags?

Edit: thought of a third option that might work: instead of replacing newlines with comments, do that to the white space at the beginning of lines instead. (Although that would break white space based pre-formatted sections…)

Edit 2: I'm not sure what I did wrong yesterday but the glob pattern worked after all, I might've run `cargo build` in the wrong directory. https://github.com/eaon/zola/tree/shortcode-markdown has the ".md template / shortcode" version. No tests yet because I'm not sure you're interested in merging this, but I tested the discourse entry and it works.

Keats:
> Yes, makes sense, I think. I assume since all the newlines will be comments when it's passed to the rest of the document, the markdown parser thinks everything is part of a really big `<h2>` heading 😄

Yes, but the next entries are completely fine 🤔 

Semantically, having .md shortcodes makes sense but I'm a bit wary about it. I'm wondering if using inline templating in the markdown file itself would be better and whether it would solve the problem more elegantly. 

cc @killercup do you know whether inline templating is a big thing in Jekyll?

killercup:
> cc @killercup do you know whether inline templating is a big thing in Jekyll?

Not sure it's a "big" thing, but I've sure found it very convenient. And, [well](https://github.com/killercup/scribbles/blob/597fbc3833a983ea067675d3153e89da40ad12f8/_posts/2020-04-16-several-months-of-reading-material.md), I might have used it to build a mini CMS in a blog post as well, so I'm very biased :smile:

Keats:
Bah let's go with markdown shortcodes, it's more reusable across pages/sections if needed.

eaon:
Cool cool, if you I want I can edit this pull request and throw the markdown shortcode bits in. I might be able to get to writing some tests for that later tonight.

Keats:
Go for it

eaon:
That's it I think. Happy to hear feedback, squash etc :upside_down_face:

Keats:
@eaon looks good! If you have time can you add some docs about markdown shortcodes in the docs? Otherwise I'll merge and do it myself

eaon:
I'll try to come up with something, I haven't been able to find the right phrasing that would go along with the already existing one.

Keats:
No hurry, and it doesn't need to be perfect either.

eaon:
How 'bout that? Couldn't think of a simpler way to get it across.

southerntofu:
Really nice feature @eaon and i find the docs about it are very clear. One question though why is `html.body.replace("<!--\\n-->", "\n")` taking place in render_content and not at the end of markdown_to_html directly?

EDIT: For reference in case somebody wants to work on interop, there's at least two other syntaxes for Markdown shortcodes.

```
:::shortcodename
shortcode body
:::
{{< shortcodename key="value" >}}
[[ shortcodename key="value" ]]
```

First one is used by [CodiMD](https://github.com/codimd/server), second one by [hugo](https://gohugo.io/content-management/shortcodes/) and the third one i've never met but is referenced [here](https://github.com/djm/remark-shortcodes).

eaon:
> Really nice feature @eaon and i find the docs about it are very clear.

That's good to hear, thanks!

> One question though why is `html.body.replace("<!--\\n-->", "\n")` taking place in render_content and not at the end of markdown_to_html directly?

While it's an unlikely thing to have `<!--\n-->` appear anywhere else, and even if, what's the harm of having it replaced with a newline (OK, hmm, I can think of *some* harm given how browsers deal with whitespace …), but it seemed like the cleaner way to only try and replace it if there's reason to. `render_shortcodes` potentially emits it with the intention of having it replaced later, so I that happens only after that actually ran.

Keats (approved):

Keats:
Very nice, thanks @eaon !

Keats:
@eaon in the end it was a bit too naive as an approach, this will fail if the newline is in a HTML tag since HTML comments are not allowed there (see https://github.com/getzola/zola/issues/1172 for an example). I'm looking into it but open if you have ideas.

PR #1089, opened 2020-07-15
Zola is now available in the official Arch repos

svenstaro:
I just put zola in the official Arch repos and the docs should reflect this change.

I hope it's correct to make this PR for `master` as opposed to `next` as it's a docs change.


Keats:
Very nice, thanks!

PR #1094, opened 2020-07-19
Fixed a compiler warning

imor:
When compiling with rust 1.45.0, I get the following warning:

use of deprecated item 'std::error::Error::description': use the Display impl or to_string()

This PR fixes the above warning.


Keats:
It looks like you need to rebase?

imor:
@Keats yeah sorry, I initially opened the PR against master and then realized that it has to be against the next branch. Messed it up, will start over again.

PR #1095, opened 2020-07-21
Add warning to not put taxonomies key in [extra] section

uggla:
Sanity check:

Explanation:
Just a little change on the documentation to warn user to not put taxonomies key in the [extra] section of config.toml.
I did this mistake, because naturally you tend to put the taxonomies key at the end of the config file.


Keats:
Thanks!

PR #1097, opened 2020-07-22
Fix pagination section memory issue

justinturpin:
When paginate_by is zero, set number_pagers to 1 so at least 1 sitemap section is pushed. I don't know if this breaks sitemap generation and will need a bit of help testing.

Sanity check:

## Code changes

* [n/a] Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/components/library/src/content/section.rs
+++ b/components/library/src/content/section.rs
@@ -252,6 +252,16 @@ impl Section {
     pub fn to_serialized_basic<'a>(&'a self, library: &'a Library) -> SerializingSection<'a> {
         SerializingSection::from_section_basic(self, Some(library))
     }
+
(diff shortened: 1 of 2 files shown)


Keats:
I think a better fix would be to have a method on the section like

```
fn paginate_by(&self) -> Option<usize> {
  // returns None if self.meta.paginate_by is None or  0
  // returns the number if >0
}
```
so the issue can't happen in multiple places - I can't remember if we calculate those pagers anywhere else.

justinturpin:
Okay made the update, is that kind of what you were thinking?

Keats:
Yep, exactly. Thanks!

PR #1100, opened 2020-07-26
Tables in config.extra can be merged with those in theme.extra

southerntofu:
This is a bugfix ported from #997 which allows to use tables (subsections) in config.extra/theme.extra and have them merged without losing information.

Example config.toml:

```
[extra]
[extra.foo]
bar = "something"
```

Example theme.toml:

```
[extra]
[extra.foo]
bar = "default"
baz = "some more things"
```

Without this patch, accessing `extra.foo.baz` in a template will fail because the `add_theme_extra` function will entirely replace the theme's extra.foo with the config's extra.foo without merging child keys (bar, and baz in this example).

The behavior introduced here is to crash when the theme is expecting a table which is overridden by some data of another type in the config. Not sure if that's the best approach.

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -212,27 +212,8 @@ impl Config {
                 self.extra.insert(key.to_string(), val.clone());
                 continue;
             }
-            if !val.is_table() {
-				// The key is overriden in site config, and is not a mapping (table)
-				// so we discard it (don't merge)
-                continue;
-            }
-			// The key is a table, so we want to merge its child entries with config
-			// Safe unwraps because we have checks before
-			let theme_table = val.as_table().unwrap();
-			let config_table = match self.extra.get_mut(key).unwrap().as_table_mut() {
-                Some(t) => t,
-                None => return Err(Error::msg(&format!("config.toml and theme.toml have incompatible types for entry theme.extra.{}", key)))
-            };
-
-			for (child_key, child_val) in theme_table {
-				if !config_table.contains_key(child_key) {
-					// Theme setting is not overriden, insert it
-					config_table.insert(child_key.to_string(), child_val.clone());
-				}
-			}
-		}
-
+            merge(self.extra.get_mut(key).unwrap(), val)?;
+        }
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


components/config/src/config/mod.rs
```diff
@@ -585,4 +604,25 @@ languages = [
...
+        "#;
+        let mut config = Config::parse(config_str).unwrap();
+        let theme_str = r#"
+[extra]
+[extra.foo]
+bar = "baz"
+        "#;
+        let theme = Theme::parse(theme_str).unwrap();
+        // Should panic!
+        assert!(config.add_theme_extra(&theme).is_ok());
```
Keats: can we just assert it's `assert_eq(false, ...)` instead of panicking?

components/config/src/config/mod.rs
```diff
@@ -206,18 +206,30 @@ impl Config {
...
+			let config_table = self.extra.get_mut(key).unwrap()
+				.as_table_mut().expect(&format!("config.toml and theme.toml have incompatible types for entry theme.extra.{}", key));
+
+			for (child_key, child_val) in theme_table {
+				if !config_table.contains_key(child_key) {
+					// Theme setting is not overriden, insert it
+					config_table.insert(child_key.to_string(), child_val.clone());
+				}
+			}
+		}
```
Keats: It would need to be recursive to be fully working no?

southerntofu: Yes, without recursion we only do one more level deep merging, so a subsubtable `[config.extra.foo.bar]` would not be merged properly. Should i make it correct and merge recursively?

Keats: If it's not too much work, sure!

southerntofu:
Recursion is here. Any further remarks?

BertalanD:
I am sorry if it is not relevant, but I think the `translations` for the page's language should also be exposed in the `extra` object, to make transparent localization of templates easier.

Keats:
`translations` is going to be dumped in favour of fluent (https://github.com/getzola/zola/pull/1040 don't mind that it is closed) so I wouldn't worry too much about it

BertalanD:
I am currently working on that, will soon add a comment with my plans.

I think `translations` should be kept, because Fluent is too clunky for basic things like using the correct localized URL for a link. Many themes also rely on setting `extra` variables for things like page title, author name and copyright text, and these would all need to be rewritten to get these things from `.ftl` files, not to mention all of the user's config.

In my opinion, Fluent is best used only for creating internationalized templates/themes.

Will elaborate on that PR, but I am a noob.

Keats:
@southerntofu do you want to become a member/admin of the Zola org? You're doing great work

Keats:
@BertalanD if we keep `translations` then we would have two ways of handling translations in the same app which is clearly not ideal.

southerntofu:
> do you want to become a member/admin of the Zola org? You're doing great work

Thanks, i accepted the invitation (i'm honored). I'm not sure what that entails but i intend to continue contributing however i can. Is there a team meeting planned sometime soon so i can get up to speed? Or a IRC/XMPP chatroom i can join?

>  if we keep translations then we would have two ways of handling translations in the same app which is clearly not ideal.

I understand @BertalanD 's request for a simpler system for very straightforward cases. Maybe we could use `lang` keys in config/theme extra for this? Like `extra.{en,fr}.dateFormat`. If that sounds reasonable to you both, we could even automatically load the translations table into the extra table for backwards-compatibility.

Keats:
>  I'm not sure what that entails but i intend to continue contributing however i can. Is there a team meeting planned sometime soon so i can get up to speed? Or a IRC/XMPP chatroom i can join?

Mostly just planning the next versions via GitHub project and fixing bugs/docs, reviewing PRs. We can do an async meeting on https://zola.discourse.group/ so more people can collaborate.
The focus for 0.12 is the updated `zola serve` + bug fixes. I don't think it makes sense to get fluent in there since it needs a lot of testing/fine-tuning probably.
I don't think there are other big features coming though.

PR #1106, opened 2020-07-28
Support relative links for root directory (fix #1028)

### LINKED ISSUE #1028: Zola can't deal with custom templates and root_dir argument
# Bug Report

Zola fails to find custom templates while building with custom `root_dir` parameter other than the current working directory. 

## Environment

Zola version: 0.10.1

## Step to reproduce

 - In `testsite/content/_index.md` put:
```
+++
template = "custom.html"
+++
```
 - Create the corresponding template file, for example `touch testsite/templates/custom.md`
 - Run `zola -r testsite build`
It outputs:
```
Building site...
Failed to build the site
Error: Failed to render section 'testsite/content/_index.md'
Reason: Tried to render `custom.html` but the template wasn't found
```
It works perfectly fine when you call zola from `testsite` directory

southerntofu:
This patch should fix the issues mentioned in #1028 by:

- canonicalizing the path inherited from the CLI (i.e. expanding relative path to an absolute one)
- prefixing the serve's watchers paths with the root directory

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -233,45 +233,36 @@ pub fn serve(
     )?;
     console::report_elapsed_time(start);
 
+    // An array of (path, bool) where the path should be watched for changes, and the boolean value
+    // indicates whether this file/folder must exist for zola serve to operate
+    let watch_this = vec!(
+        ("config.toml", true),
+        ("content", true),
+        ("sass", false),
+        ("static", false),
+        ("templates", false),
+        ("themes", false)
+    );
+
     // Setup watchers
-    let mut watching_static = false;
-    let mut watching_templates = false;
-    let mut watching_themes = false;
     let (tx, rx) = channel();
     let mut watcher = watcher(tx, Duration::from_secs(1)).unwrap();
-    watcher
-        .watch(root_dir.join("content/"), RecursiveMode::Recursive)
-        .map_err(|e| ZolaError::chain("Can't watch the `content` folder. Does it exist?", e))?;
-    watcher
-        .watch(root_dir.join(config_file), RecursiveMode::Recursive)
-        .map_err(|e| ZolaError::chain("Can't watch the `config` file. Does it exist?", e))?;
 
-    let mut dir = root_dir.join("static");
-    if dir.exists() {
-        watching_static = true;
-        watcher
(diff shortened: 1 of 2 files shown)


src/cmd/serve.rs
```diff
@@ -241,35 +240,38 @@ pub fn serve(
...
         watching_templates = true;
         watcher
-            .watch("templates/", RecursiveMode::Recursive)
+            .watch(dir, RecursiveMode::Recursive)
             .map_err(|e| ZolaError::chain("Can't watch the `templates` folder.", e))?;
     }
 
-    if Path::new("themes").exists() {
+    dir = root_dir.join("themes");
+    if dir.exists() {
```
Keats: can we name them properly instead of reusing `dir`? It's easy to miss an assignment

Keats (changes requested):
A very small nit but looks good otherwise

southerntofu:
I think it should be harder to miss an assignment this way

Quick notes:

- config.toml was watched with RecursiveMode::Recursive, i don't know if that's supposed to be correct but i kept this behavior here (though it's trivial to change)
- other paths would fail in case they existed but couldn't be watched, but not sass. now the behavior for sass has been aligned with others
- the serve command has no idea whether sass is enabled or not, so the folder will be watched if it exists, and changes in there will be applied to the site (sass will be compiled, or an error will be displayed)

Keats:
> config.toml was watched with RecursiveMode::Recursive, i don't know if that's supposed to be correct but i kept this behavior here (though it's trivial to change)

It's probably a copy/paste, fine to change.

> the serve command has no idea whether sass is enabled or not

Isn't that

```
    if site.config.compile_sass {
        watchers.push("sass");
    }
```

southerntofu:
> It's probably a copy/paste, fine to change.

According to the [docs](https://docs.rs/notify/5.0.0-pre.3/notify/trait.Watcher.html): « If the path is a file, recursive_mode will be ignored and events will be delivered only for the file. »

No need to change anything apparently

> watchers.push("sass")

watchers is used exactly once, for printing "Listening for changes in..." so this condition will not disable SASS support in the serve command, although the message won't mention that the sass folder is being watched. The latest commit should provide the expected behavior

Keats:
Thanks!

Issue #877, opened 2019-12-12
Link generated for relative image does not reflect intuition


timotheos-firestone:
# Bug Report

Directory structure:

- `_index.md`
- `page.md`
- `sanctuary-map.jpg`

`page.md` includes the following Markdown:

`![Map of Sanctuary](sanctuary-map.jpg)`

On using `zola serve`, the HTML output for this becomes:

```
<p><img alt="Map of Sanctuary" src="http://127.0.0.1:1111/page/sanctuary-map.jpg"></p>
```

Expected HTML output:

```
<p><img alt="Map of Sanctuary" src="http://127.0.0.1:1111/sanctuary-map.jpg"></p>
```

Surely I would manually specify `![Map of Sanctuary](page/sanctuary-map.jpg)` if I wanted that behaviour?

I took a look on the documentation and I couldn't find this current behaviour documented in the content overview.

It also doesn't generate any warnings that the image couldn't be found.

There's no documentation on how to reference an image that's within the same folder as the `page.md`, that I could see?

## Environment

Zola version: 0.9.0

Keats:
That's because the renderer assumes it's a colocated asset, therefore nested in the folder and prefix the page path. This is not really supported but the link should not be invalid.

itf:
It is confusing that valid links in the source do not map to valid links on the output because of this issue. It seems that forcing the colocation asset assumption causes this issue.  I really like #840 as a solution option for people not manually creating /folder/index.md; because with "ugly urls", both the output file and the relative links inside the file work as expected.  In particular it is easier to have 2 files in the same folder sharing the same asset, rather than either having assets being fully "public" in the statics folder, or associated with a single file

Keats:
It's actually a bug imo, we should not copy such a file in the public directory. You cannot share the assets, you will have to copy it once per .md file in that folder

tad-lispy:
Hi! Sorry for re-heating this old issue. I'd like to share an idea. Please let me know if it's a bad one.

Currently only pages from `some-dir/index.md` have the `collocated_path` property. We have a notion of colocated and un-colocated pages. Would it make sense for **every page and section** to have a `colocated_path`? Basically it would be **the path that will receive the assets** from a directory of the `.md` file.  When rendering images and links, this ever-present `colocated_path` would be used as a prefix. 

For illustration:

| source                             | destination                         | colocated_path     |
|------------------------------------|-------------------------------------|--------------------|
[...]

Issue #687, opened 2019-05-17
Templates need a variable that could be “section” or “page”


chris-morgan:
I find that I need this in templates, e.g. in the base template:

```jinja
{% set section_or_page = section | default(value = page) -%}
```

That way I can do things like active page highlighting in top nav:

```jinja
<a href="/help/" {% if section_or_page.path is starting_with('help/') %}class="is-selected"{% endif %}">Help</a>
```

Something to think about. As you see, there is a workaround, but it’s not exactly *elegant*. At the least, it should be documented.

Keats:
For that specific usecase, all templates get `current_path` so you don't need to access the section/page variable for that.
Do you have an example of a common variable needed apart from the path?

chris-morgan:
OK, that was indeed a bad example; I’d forgotten about `current_path`.

`description` and `extra.*` for various things like meta tags in the base template, and other properties in certain other templates: basically I find that in various places I don’t want most of the distinction between sections and pages, and so want to be able to reuse template portions for both sections and pages.

Still, I think the path is the most probable case where people in general will want to do this, so it’d probably be acceptable to just document the workaround for when you wish to treat section and page alike in a template.

Walther:
Cross-referencing a related discussion thread here: https://zola.discourse.group/t/idea-better-page-metadata-for-published-pages/111

Keats:
Any good name that could represent both?

chris-morgan:
I’ve been using `resource`. My base templates are always starting with this now:

```jinja
{%- if page %}
	{%- set resource = page %}
{%- elif section %}
	{%- set resource = section %}
{%- endif %}
```

Also I become steadily more and more opposed to the page/section distinction, extremely strongly wishing for them to be unified into one type. It’s been causing me *so much* trouble on a couple of sites I’ve been making. I’d say that was a different matter from this issue, but it’s all bound up in the way that they’re regularly treated equivalently, and that I see no reason why they shouldn’t *always* be treated equivalently—each “section” could then opt into a different template if it wanted it. It’s not enough to make me abandon Zola, but it’s close on one site I’m redoing from Drupal right now.

Keats:
I don't see how having them be the same thing makes sense. Can you have a RSS feed for a page, paginate a single page? What is a transparent page? What's a page that has `render = false`? They are 2 different things and merging them would make it too confusing imo.

I'm not too keen on `resource` as a name as we already use `assets` for files and those are synonyms.

[...]

PR #1107, opened 2020-07-28
WIP: Pages "inherit" assets from a section they colocate with (#877)

southerntofu:
The problem described in #877 is solved by this patch though in a very unoptimal way because all assets shared between pages in a section will be copied over to all pages output folders. This is because declaring the files as page assets will trigger their copy when rendering the page. I don't see a more elegant way to solve it (yet). Any ideas?


Keats:
How about having a flag on a page whether it's a page with colocated assets or not. If there is, do like right now but if there isn't, in the markdown renderer (L125 of markdown.rs) make a link from the parent instead.

southerntofu:
> How about having a flag on a page whether it's a page with colocated assets or not

Markdown links would work but then `page.assets` would be empty for those pages which is not expected behavior (eg. if a template lists a page's related assets). I'll try to dig deeper into how asset management works.

Keats:
I think that's fine, page.assets is only for colocated assets and if the asset is in the directory it isn't colocated with anything.

southerntofu:
The latest commit only copies assets once, but has them declared in the associated section/pages as expected. Relative links from Markdown are fixed on the go, and assets still returns paths from the content folder, which means we can't use page.assets in templates as relative URLs, but have to concatenate with the base_url (as was already the case).

This commit was tested with the following `page.html`:

```
{% for i in page.assets %}
{% if i is ending_with(".png") %}
<img src="{{ resize_image(path=i, width=400, op="fit_width") }}">
<img src="{{ config.base_url }}/{{ i }}">
{% endif %}
{% endfor %}
```

Waiting for feedback before i add some tests.

Not very interesting but worth mentioning for #687: for a "page in page" where the "parent" page has a different path/slug, the link produced in the "child" page will be incorrect.

Keats:
I still don't know whether to support that or not...

PR #1108, opened 2020-07-29
WIP: Don't fail to rebuild the site when a page is edited (fix #1014)

### LINKED ISSUE #1014: Rebuild is broken by index.md file
# Bug Report

Rebuild keep getting this error: `Error: Change on "/content/somepage.md" detected but only files named index.md with an optional language code are allowed` 

## Environment

Zola version: 0.10.1

## Expected Behavior

Rebuild should just works, and the error message should be more helpful.

## Step to reproduce

Put an `index.md` file in `content/`, create any other valid pages under `content/`, then edit one of these pages should trigger the error.

Delete `index.md` then everything is ok.

southerntofu:
Removing the block which bailed seems to match expected behavior in #1014. I don't believe this change will break anything.

I'm not entirely sure this function does what it's supposed to do. I'm not entirely sure what it's supposed to do. I would expect something like this:

- 1. Entry is a file => What is it?
  - 1. Entry is markdown => Does it still exist?
    - 1. Yes => handle_{page,section}_editing
    - 2. No => delete_element
  - 2. Entry is something else => Does it still exist?
    - 1. Yes => copy it as a colocated asset, rebuild associated pages
    - 2. No => remove from colocated assets, rebuild associated pages
- 2. Entry is a folder => Does it still exist?
  - 1. Yes => find pages/sections inside (add pages/sections inside)
  - 2. No => remove pages/sections matching this basepath

Instead i see delete_element is called only if `!index.exists() && !path.exists()`. why? shouldn't delete_element be called if we have content/{_index,test}.md and have deletion of test.md?

### DIFF
--- a/components/rebuild/src/lib.rs
+++ b/components/rebuild/src/lib.rs
@@ -342,11 +342,11 @@ pub fn after_content_change(site: &mut Site, path: &Path) -> Result<()> {
     let is_md = path.extension().unwrap() == "md";
     let index = path.parent().unwrap().join("index.md");
 
-    let mut potential_indices = vec![path.parent().unwrap().join("index.md")];
+    /*let mut potential_indices = vec![path.parent().unwrap().join("index.md")];
     for language in &site.config.languages {
         potential_indices.push(path.parent().unwrap().join(format!("index.{}.md", language.code)));
     }
-    let colocated_index = potential_indices.contains(&path.to_path_buf());
+    let colocated_index = potential_indices.contains(&path.to_path_buf());*/
 
     // A few situations can happen:
     // 1. Change on .md files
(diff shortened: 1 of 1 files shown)


Keats:
I'm actually planning to scrap completely the rebuild component and have zola serve rebuild the whole thing on every change. 
To make it fast, all the HTML content will be help in a hashmap that the webserver will check before looking on disk. Assets will still be copied to avoid some big memory usage but that's the gist of it: https://zola.discourse.group/t/virtual-file-system/415/3

I made a poc locally the other day where it __seemed__ to work well.

southerntofu:
> zola serve rebuild the whole thing on every change

Nice, that seems a lot easier than debugging all the weird serve issues. And if somebody's not happy with rebuilding the whole site on every change, it will still be possible for them to reimplement differential logic from scratch (without accumulated cruft).

> To make it fast, all the HTML content will be help in a hashmap that the webserver will check before looking on disk.

Your proposal is certainly an improvement over the current situation. Do you think it would suit memory-constrained systems such as Raspberry Pi or low-end VMs? Whatever the approach, it should be possible to free the memory from site building (which is mostly the library) before serving the pages.

Personally, i appreciate the folder output of `zola serve` because it will let me explore the website output with my file browser, and when i'm done i can just ^C. I'm aware this is not what the serve command is for, so do not consider this a strong argument to keep the current behavior. It's just a trick i use rather often when investigating output paths because it's considerably shorter than typing `rm -r public/` after a build.

> I made a poc locally the other day where it seemed to work well.

Can't wait to try this. Should we close this PR in the meantime (because your PoC will make it obsolete) or should it still be merged so we can close the associated issue sooner?

Keats:
> Do you think it would suit memory-constrained systems such as Raspberry Pi or low-end VMs?

Only the HTML will be loaded in memory so hopefully that won't be huge.

> Personally, i appreciate the folder output of zola serve because it will let me explore the website output with my file browser, and when i'm done i can just ^C

That would be the same as zola build then no?

> Can't wait to try this. Should we close this PR in the meantime (because your PoC will make it obsolete) or should it still be merged so we can close the associated issue sooner?

The poc was extremely minimal and didn't have any rebuild whatsoever. I'll take quite some time to get it fully working but I'll try to do it asap. That's the main feature I want in for 0.12 since half of the issues on that repo are with zola rebuild.

PR #1114, opened 2020-08-01
Change zola serve to load HTML from memory instead of disk

Keats:
TODO:

- [x] Copy page/section assets using copy_file to avoid needless copying
- [x] Do not re-render markdown when templates change unless it is in shortcodes
- [x] Find ways to avoid re-rendering Markdown as much as possible
- [ ] Try it more with taxonomies/pagination etc


Keats:
cc @southerntofu so you have an idea of it.

Still a few things to do but it should be better than the current state, with a nice loc reduction as well.

Keats:
The obvious downside with that approach is that the reload time on serve is magnitude bigger BUT avoids a lot of bugs.
Eg if the initial build is 300ms, the rebuild with that PR would be 200ms instead of 10ms with the current approach.
I think this is a worthwhile goal though, I'll add a `--fast` flag to only render the section/page that was changed, which should improve the speed when iterating on something particular.

Any feedback?

Keats:
I've added a `zola serve --fast` that only reloads the section/page that changed. This brings us back to the current speed but it explicitly does not try to be smart and will not reload everything you might expect in some cases (such as making a section transparent). You will have to restart the server or use plain `zola serve` for it.

Any comments on it? It's missing some documentation and some more testing but it should be way more resilient than the current system.

src/cli.rs
```diff
@@ -90,6 +90,11 @@ pub fn build_cli() -> App<'static, 'static> {
                         .long("open")
                         .takes_value(false)
                         .help("Open site in the default browser"),
+                    Arg::with_name("fast")
+                        .short("f")
+                        .long("fast")
+                        .takes_value(false)
+                        .help("Only rebuild the minimum on change - useful when working on specific a page"),
```
uggla: sentence error: useful when working on specific **a** page  --> useful when working on **a** specific page

Keats: Fixed, thanks!

Hugo-Trentesaux:
Maybe we should keep an option to copy html to disk.
I'm working on a setup to automatically rebuild a static site when files synchronized with Nextcloud change. Having the files on the disk is a feature I miss in zola 0.12.
Plus the "output dir" option is now useless since it creates an empty folder and does not put content in it.

Keats:
It will put all content in it.
I don't think I would add it back since it was very buggy and didn't rebuild everything that was needed. Depending on your sites/templates, it's very likely that some could be out of date if you were relying on zola serve filesystem changes.

Hugo-Trentesaux:
> It will put all content in it.
With 0.12.1 it just creates an empty folder for me.
Then an option for a complete on-disk html re-export on build could be useful in the case I describe (and will announce on the forum once I get a satisfying configuration).

Issue #1102, opened 2020-07-27
Issues with building search index contains Chinese contents


liushuyu:
# Bug Report

## Environment

Zola version: 0.11.0 (from `crates.io`)

## Expected Behavior
The search index builds correctly with no error since the underlying crate (`elasticlunr-rs`) supports Chinese through the de-facto `Jieba` segmentation library.

## Current Behavior
It does not work when the website has multiple Chinese languages (Simplified and Traditional) and fails with 

```
Tried to build search index for language zh-cn which is not supported
```

The root cause is that [upstream implementation](https://github.com/mattico/elasticlunr-rs/blob/beedaf86038839b07df007b0f7cf8417e4414956/src/lang/mod.rs#L133) recognizes the Chinese as a whole (and their implementation works for common variants of Chinese) and the only language code it accepts is `zh`.

## Step to reproduce

Put the following into `config.toml`:

```toml
# The URL the site will be built for
base_url = "https://example.com"

# Whether to automatically compile all Sass files in the sass directory
compile_sass = true

# Whether to do syntax highlighting
# Theme can be customized by setting the `highlight_theme` variable to a theme supported by Zola
highlight_code = false

# Whether to build a search index to be used later on by a JavaScript library
build_search_index = true

theme = "book"

languages = [
  {code = "zh-cn", search = true},
  {code = "zh-tw", search = true},
]

[extra]
# Put all your custom variables here
```

## Extra notes

I didn't open a PR since this may need some discussion as I can see Zola thrives for cleaner implementations for everything.

My suggested solution would be having extra handling in https://github.com/getzola/zola/blob/97e772868d8892874cfb825c1acd789d9ad725f3/components/search/src/lib.rs#L32-L48.

The extra handling could be just simply stripping out the variant suffix like `zh-cn => zh`.

If you think the extra string manipulation would hurt the performance, then it could be gated behind a feature switch; and if you think this should be better resolved on the upstream, then I can open an issue on the upstream repository as well.

Keats:
Sadly, I actually had to remove Chinese & Japanese support in the `next` branch for the search index generation as it was causing the binary size to inflate **a lot** (or not build at all due to asking too much RAM).
I'm happy to turn them back on if we find a way to not end up with a 80mb+ binary size though.

liushuyu:
> Sadly, I actually had to remove Chinese & Japanese support in the `next` branch for the search index generation as it was causing the binary size to inflate **a lot** (or not build at all due to asking too much RAM).
[...]

PR #1115, opened 2020-08-01
Making Chinese and Japanese search indexing optional

liushuyu:
Previous discussion: #1102.

This Pull Request still requires some discussions as the current implementation is less than ideal due to the following reasons:

- The languages included by default are hardcoded, if the upstream decides to add a new language, then that ~~laundry~~ list needs to be updated.
- The Chinese segmentation library is significantly smaller (~5 MB) than the Japanese one (~60 MB), is it a good idea to just includes the Chinese segmentation library by default?
- Many languages have several different variants, if someone wants to, say include Simplified Chinese and Traditional Chinese on the same website, then there will be issues with the search indexing (as the language codes will not be ISO 639-1 compliant). So do we need to implement a system to convert the language codes to ISO 639-1 code so that `elasticlunr-rs` will work?

--------------------------------------------------------------------------------------------------

Sanity check:

## Code changes

---
This change is [Reviewable](https://reviewable.io/reviews/getzola/zola/1115)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2817,6 +2817,7 @@ dependencies = [
  "notify",
  "open",
  "rebuild",
+ "search",
  "site",
  "termcolor",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ errors = { path = "components/errors" }
 front_matter = { path = "components/front_matter" }
 utils = { path = "components/utils" }
 rebuild = { path = "components/rebuild" }
+search = { path = "components/search" }
 
 [workspace]
 members = [
--- a/components/search/Cargo.toml
+++ b/components/search/Cargo.toml
@@ -12,3 +12,8 @@ lazy_static = "1"
 errors = { path = "../errors" }
 library = { path = "../library" }
 config = { path = "../config" }
+
+[features]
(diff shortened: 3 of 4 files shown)


Keats:
> So do we need to implement a system to convert the language codes to ISO 639-1 code so that elasticlunr-rs will work?

That should be done as part of https://github.com/getzola/zola/pull/1040#issuecomment-666504613

> The Chinese segmentation library is significantly smaller (~5 MB) than the Japanese one (~60 MB), is it a good idea to just includes the Chinese segmentation library by default?

That is like a 30% increase in binary size, I would still disable it by default.

liushuyu:
> > So do we need to implement a system to convert the language codes to ISO 639-1 code so that elasticlunr-rs will work?
> 
> That should be done as part of [#1040 (comment)](https://github.com/getzola/zola/pull/1040#issuecomment-666504613)
> 
> > The Chinese segmentation library is significantly smaller (~5 MB) than the Japanese one (~60 MB), is it a good idea to just includes the Chinese segmentation library by default?
> 
> That is like a 30% increase in binary size, I would still disable it by default.

Understood. Then this Pull Request is ready for review now.

Keats:
Thanks!

PR #1118, opened 2020-08-06
Support anchors without quotes in link checker

siiptuo:
Sanity check:

## Code changes

### DIFF
--- a/components/link_checker/src/lib.rs
+++ b/components/link_checker/src/lib.rs
@@ -101,11 +101,15 @@ fn has_anchor(url: &str) -> bool {
 fn check_page_for_anchor(url: &str, body: String) -> errors::Result<()> {
     let index = url.find('#').unwrap();
     let anchor = url.get(index + 1..).unwrap();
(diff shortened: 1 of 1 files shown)


Keats:
Is that even valid HTML? :o

siiptuo:
I'm pretty sure it's valid because [HTML Living Standard](https://html.spec.whatwg.org/) itself doesn't use quotes.

Keats:
> I'm pretty sure it's valid because HTML Living Standard itself doesn't use quotes.

It does? I see all the ids between double quotes in the console

siiptuo:
Here's the relevant part of the specification: https://html.spec.whatwg.org/#unquoted

> > I'm pretty sure it's valid because HTML Living Standard itself doesn't use quotes.
> 
> It does? I see all the ids between double quotes in the console

I mean quotes are not used in the actual HTML source of the specification. Browser developer tools usually display double quotes when inspecting the DOM.

Keats:
Thanks!

PR #1120, opened 2020-08-07
Update filetime crate to 0.2.12

protoCall7:
This PR updates the `filetime` dependency in `components/utils` to version 0.2.12. This version contains a bug fix for atime/mtime setting, which fixes the new file copying routines on OpenBSD.


Keats:
Thanks!

PR #1128, opened 2020-08-13
Per section/subsection feeds

Yaroslav-95:
* Added `generate_feed` variable to section front matter.
* Generate atom/rss feeds for sections/subsections that have the
`generate_feed` variable set to true (false by default); this works
independent of the `generate_feed` variable in the root `config.toml`
file, however, the name (and template) of the feed file for each section
is the same as `feed_filename` in `config.toml`, just located in the
root of each section.
* Slightly edited `atom.xml` and `rss.xml` so that they include the
section title (if any), and the url of a section, if it's a section
feed.

I haven't updated the docs, since I first wanted to see what the thoughts of the maintainers are. A thread on this topic was already started in the forums: <https://zola.discourse.group/t/separate-feeds-for-each-section/461/3/>, it didn't seem to move forward much, though.

Sanity check:

## Code changes


Keats:
Thanks! Can you add some tests?

Yaroslav-95:
Sure! I've added/edited the tests and updated the relevant documentation.

components/site/src/lib.rs
```diff
@@ -901,6 +901,31 @@ impl Site {
...
+                format!("{}/", &section.lang)
+            };
+            let pages = section
+                .pages
+                .iter()
+                .map(|k| library.get_page_by_key(*k))
+                .collect();
+            self.render_feed(
+                pages,
+                Some(&PathBuf::from(section.file.components.iter().fold(
```
Keats: you can use `section.path` here I think

Keats (commented):
Just a small comment, if you have time to fix great otherwise I'll fix it after merging.

components/site/src/lib.rs
Yaroslav-95: You're right. It should be fixed now.

components/site/src/lib.rs
```diff
@@ -901,6 +901,24 @@ impl Site {
...
+        if section.meta.generate_feed {
+            let library = &self.library.read().unwrap();
+            let pages = section
+                .pages
+                .iter()
+                .map(|k| library.get_page_by_key(*k))
+                .collect();
+            self.render_feed(
+                pages,
+                Some(&PathBuf::from(&section.path[1..])),
```
Keats, line 913: why the [1..]?

Yaroslav-95: Because section.path start with a "/" which means it will try to render the feed starting at the root of the filesystem instead of the root of the site directory.

Keats:
Thanks!

PR #1142, opened 2020-08-18
A Fix for the permalinks in #1136

aslowwriter:
## Code changes
When taking another look at my PR #1136 I noticed that it didn't actually solve the problem because it didn't merge the permalinks properly. apologies for that. I checked that Also, my git fu is not amazing but I've tried to keep the commit history as clean as possible, I hope it's okay like this. 

## Checklist


Keats:
The test failures seem legit?

aslowwriter:
they were, I had forgotten to run the entire test suite rather than just the ones for the taxonomy, that's my bad. The point is that the taxonomy names now get saved slightly differently to detect the collisions where necessary. The only change necessary was to add `| replace(from="-", to=" ") |  title ` to the relevant templates where `tag.name` was used, it should be good now

components/templates/src/builtins/atom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <feed xmlns="http://www.w3.org/2005/Atom" xml:lang="{{ lang }}">
 	<title>{{ config.title }}
-		{%- if term %} - {{ term.name }}
+		{%- if term %} - {{ term.name | replace(from="-", to=" ") |  title}}
```
Keats: sadly that doesn't work if you have unicode :/
The deduplication should happen without needing any changes to the templates

Keats (changes requested):

components/templates/src/builtins/atom.xml
aslowwriter: okay I'll have a look and see what I can do

Keats:
Can you do a separate PR with some failing tests for the permalinks? I can have a look as well.

aslowwriter:
I think I found a solution that works and doesn't require moddifications to the templates to pass all tests. If you're happy about the changes I'll try and clean up the history before you merge the request

components/library/src/taxonomies/mod.rs
```diff
@@ -70,22 +71,38 @@ impl TaxonomyItem {
...
 
-        TaxonomyItem { name: name.to_string(), permalink, slug, pages }
+        TaxonomyItem { name: name.to_string(), permalink, slug: item_slug, pages }
     }
 
     pub fn serialize<'a>(&'a self, library: &'a Library) -> SerializedTaxonomyItem<'a> {
         SerializedTaxonomyItem::from_item(self, library)
     }
+
+    pub fn merge(&mut self, other: &Self) {
```
Keats: if you merge you can pass `Self` I think

components/library/src/taxonomies/mod.rs
```diff
@@ -121,8 +138,23 @@ impl Taxonomy {
         for (name, pages) in items {
             sorted_items.push(TaxonomyItem::new(&name, &kind, config, pages, library));
         }
-        sorted_items.sort_by(|a, b| a.name.cmp(&b.name));
+        // make sure we have stabel sorting when slugs are equal
```
Keats: stable

components/library/src/taxonomies/mod.rs
```diff
@@ -70,22 +71,38 @@ impl TaxonomyItem {
...
+            config.make_permalink(&format!("/{}/{}/{}", taxonomy.lang, taxo_slug, item_slug))
         } else {
-            config.make_permalink(&format!("/{}/{}", taxonomy.name, slug))
+            config.make_permalink(&format!("/{}/{}", taxo_slug, item_slug))
         };
+        //let permalink = if taxonomy.lang != config.default_language {
+        //config.make_permalink(&format!("/{}/{}/{}", taxonomy.lang, taxonomy.name, slug))
+        //} else {
+        //config.make_permalink(&format!("/{}/{}", taxonomy.name, slug))
+        //};
```
Keats: can be removed?

Keats (changes requested):
Looks ok to me. Do you think there are enough tests covering that or should we add more?

components/library/src/taxonomies/mod.rs
aslowwriter: noted will apply this shortly

components/library/src/taxonomies/mod.rs
aslowwriter: I assume you mean that you prefer the original bit since that is already stable, which is fine, I'll revert it when I make the change

aslowwriter: Edit, sorry I just realised sorting by slug first is actually necessary since the deduping otherwise fails. Tests fail with the original sorting

components/library/src/taxonomies/mod.rs
aslowwriter: I thought this would cause ownership conflicts but after trying it out you are actually correct. Changed this in the next commit

aslowwriter:
> Looks ok to me. Do you think there are enough tests covering that or should we add more?

Thanks for the feedback. I added tests for all cases that I could think of that would result in the same permalink (added whitespace, inconsistent capitalisation and substituting spaces for hyphens) so unless you can think of something on top of that I think it should be okay? 

The only thing that is perhaps left untested is how this all interacts with Unicode, but I thought that valid URLs are limited to ASCII, so I didn't add anything for that. I don't actually know enough to make a definitive comment on that so I'll defer to your judgement on that

Keats:
> The only thing that is perhaps left untested is how this all interacts with Unicode, but I thought that valid URLs are limited to ASCII, so I didn't add anything for that.

They get escaped but you can get URL outside of ASCII: https://fr.wikipedia.org/wiki/%C3%89cole (eg École).
If someone is not turning paths into slugs, the 2 taxonomies ecole and école should NOT be merged. If you are slugifying them on the other hand, they should end up merged.

aslowwriter:
I've added two tests to cover those two cases

Keats (approved):

Keats:
Thanks!

Issue #1056, opened 2020-06-08
Link checker sometimes flaky 


ralfbiedert:
# Bug Report

## Environment

Zola version: `0.11.0`

## Current Behavior

Hi, when running the link checker on my site zola reports spurious network errors, although opening the same site in Firefox works fine. As a subsequent error, if I try to re-run the checker I then sometimes get blocked by other hosts.

For example, first run:

```
D:\Development\Source\cheats.rs>bash ./deploy.sh --live
Checking site...
Checking 0 internal link(s) with an anchor.
Checking 256 external link(s).
> Checked 256 external link(s): 3 error(s) found.
Failed to check the site
Error: Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://rustup.rs/: error sending request for url (https://rustup.rs/): error trying to connect: dns error: This is usually a temporary error during hostname resolution and means that the local server did not receive a response from an authoritative server. (os error 11002)
Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://www.jetbrains.com/idea/: error sending request for url (https://www.jetbrains.com/idea/): error trying to connect: dns error: This is usually a temporary error during hostname resolution and means that the local server did not receive a response from an authoritative server. (os error 11002)
Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://www.jetbrains.com/clion/: error sending request for url (https://www.jetbrains.com/clion/): error trying to connect: dns error: This is usually a temporary error during hostname resolution and means that the local server did not receive a response from an authoritative server. (os error 11002)
error with 'zola'
```

Running again:

```
D:\Development\Source\cheats.rs>bash ./deploy.sh --live
Checking site...
Checking 0 internal link(s) with an anchor.
Checking 256 external link(s).
> Checked 256 external link(s): 3 error(s) found.
Failed to check the site
Error: Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://github.com/bheisler/criterion.rs: Client error status code (429 Too Many Requests) received
Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://github.com/async-rs/async-std: Client error status code (429 Too Many Requests) received
Dead link in D:/Development\Source\cheats.rs\content\_index.md to https://github.com/rust-lang/rust/blob/master/RELEASES.md: Client error status code (429 Too Many Requests) received
error with 'zola'
```

In contrast to them being labeled "temporary", once I encounter `os error 11002` messages I can
[...]

PR #1145, opened 2020-08-20
Add documentation build and links checking in CI.

uggla:
Sanity check:

## Code changes

Explanations:

@Keats, you told me in a comment that it was something you wanted to add to the CI. I did it as I could handle it and let you manage more complex stuffs.
* Add `zola build` and `zola check` to ensure documentation can be built and does not contain invalid links.
* Fix links in the current documentation that were broken.
Maybe I would have to open this PR on the master branch ? Let me know if you would rather this PR on the master branch.


uggla:
Hum, seems like this is suffering from #1056. I think I'm gonna add a couple of retries to try to avoid link checking issue. Maybe not the best solution but better than nothing.

azure-pipelines.yml
```diff
@@ -41,6 +41,24 @@ stages:
...
+            displayName: Check documentation build windows
+
+          - script: |
+              cd docs && ../target/debug/zola check
+            condition: ne( variables['Agent.OS'], 'Windows_NT' )
+            displayName: Check documentation links
+          - script: |
+              cd docs && ..\target\debug\zola check
+            condition: eq( variables['Agent.OS'], 'Windows_NT' )
+            displayName: Check documentation links windows
```
Keats: That's no good, it would block the release of Zola if there is a dead link. I would be tempted to move `zola check` to GH actions actually.

docs/config.toml
```diff
@@ -8,5 +8,16 @@ insert_anchor_links = true
...
+# Skip link checking for external URLs that start with these prefixes
+skip_prefixes = [
+    "http://127.0.0.1",
+]
+
+# Skip anchor checking for external URLs that start with these prefixes
+# skip_anchor_prefixes = [
+#     "https://caniuse.com/",
+# ]
+
```
Keats, line 21: what's the reason for that change?

Keats (changes requested):

azure-pipelines.yml
uggla: Hello @Keats, in fact that was the intention. :smile: 
To not pass the CI process if for any reason the documentation was broken or of boken links was introduced (Like the ones I corrected). Also it would check this on a regular basis and avoid documentation to deviate and keep it completely updated.
I like to put all checks in the CI, but if you think it annoying I can remove them from this PR.

docs/config.toml
uggla: In fact I did a configuration documentation copy / paste.
```
[link_checker]
# Skip link checking for external URLs that start with these prefixes
skip_prefixes = [
    "http://127.0.0.1",
]

# Skip anchor checking for external URLs that start with these prefixes
# skip_anchor_prefixes = [
#     "https://caniuse.com/",
# ]
```
I wanted to avoid broken links errors with `http://127.0.0.1` (there are a several errors in the documentation), but so far I don't have to use the "anchors skip part". So I commented it out.

azure-pipelines.yml
Keats: It's a good instinct but the link checker is not robust yet so that could cause me to comment out these lines in anger if I want to do a release and it's failing for whatever reason

uggla: So I comment out the link check part.
I introduced nb_threads_per_cpu to configure the link_checker number of threads. As reducing the number of threads looks more stable at least on AzureDevOps. Maybe because there is some kind of limitation to traffic and especially number of simultaneous connections outgoing from AzureDevOps to avoid DDOS.

azure-pipelines.yml
```diff
@@ -41,6 +41,26 @@ stages:
...
             displayName: Cargo test
 
+          - script: |
+              cd docs && ../target/debug/zola build
+            condition: ne( variables['Agent.OS'], 'Windows_NT' )
+            displayName: Check documentation build
+          - script: |
+              cd docs && ..\target\debug\zola build
+            condition: eq( variables['Agent.OS'], 'Windows_NT' )
+            displayName: Check documentation build windows
```
Keats, line 51: I like that!

components/config/src/config/link_checker.rs
```diff
@@ -7,10 +7,15 @@ pub struct LinkChecker {
     pub skip_prefixes: Vec<String>,
     /// Skip anchor checking for these URL prefixes
     pub skip_anchor_prefixes: Vec<String>,
+    pub nb_threads_per_cpu: usize,
```
Keats: I don't think this should be configurable. Let's keep it at 4 for now and we'll see for configuration later on.

components/site/src/link_checking.rs
```diff
@@ -119,9 +119,14 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
-    // (almost) all pages simultaneously
-    let threads = std::cmp::min(all_links.len(), 32);
+    // create thread pool with threads depending on the number of cpus (8 threads per cpu)
+    // so we can fetch (almost) all pages simultaneously on large config
+    let threads = std::cmp::min(all_links.len(), num_cpus::get() * nb_threads_per_cpu);
+    println!(
+        "Number of cpus: {}, using {} threads",
+        num_cpus::get(),
+        num_cpus::get() * nb_threads_per_cpu
+    );
```
Keats: we don't want to print that

docs/content/themes/codinfox-zola/index.md
```diff
@@ -21,11 +21,11 @@ homepage = "https://svavs.github.io/"
 
 # Codinfox-Zola
 
-This is a [Zola](https://www.getzola.com) theme inspired to [Codinfox-Lanyon](https://codinfox.github.com/), a Lanyon based theme for [Jekyll](http://jekyllrb.com). See a live demo [here](https://codinfox-zola.vercel.app/).
+This is a [Zola](https://www.getzola.org) theme inspired to [Codinfox-Lanyon](https://codinfox.github.com/), a Lanyon based theme for [Jekyll](http://jekyllrb.com). See a live demo [here](https://codinfox-zola.vercel.app/).
 
 This theme places content first by tucking away navigation in a hidden drawer.
 
-* Built for [Zola](https://www.getzola.com)
+* Built for [Zola](https://www.getzola.org)
```
Keats, line 28: Those are generated by the https://github.com/getzola/themes repo, fixing it here will not prevent more bad permalinks on the next update.

Keats (commented):
Overall I like the build on CI but not sure about the rest. We can definitely lower the number of threads for the link checker though, in a more static way for now.

components/config/src/config/link_checker.rs
uggla: So I come back to the static version

components/site/src/link_checking.rs
uggla: I have removed it.

PR #1146, opened 2020-08-24
Add minify support

### LINKED ISSUE #542: Built in Minify Support
#### Problem

As the title says, currently Zola doesn't have built-in html minifier support. And as for comparison, Hugo has one by invoking:

```bash
$ hugo --minify
```

It will generate production-ready minified HTML (Inline Styles + Inline Scripts) pages.

#### Idea

By borrowing the idea from Hugo we can copy like:

```bash
$ zola build --minify
```

I'm not familiar with Rust ecosystem. So maybe using https://docs.rs/html-minifier/1.1.3/html_minifier/ is possible.

areille:

Sanity check:

## Code changes

Closes #542

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -75,7 +75,7 @@ pub struct Config {
     /// Whether to compile the `sass` directory and output the css files into the static folder
     pub compile_sass: bool,
     /// Whether to minify the html output
-    pub minify: bool,
+    pub minify_html: bool,
     /// Whether to build the search index for the content
     pub build_search_index: bool,
     /// A list of file glob patterns to ignore when processing the content folder. Defaults to none.
@@ -322,7 +322,7 @@ impl Default for Config {
             hard_link_static: false,
             taxonomies: Vec::new(),
             compile_sass: false,
-            minify: false,
+            minify_html: false,
             mode: Mode::Build,
             build_search_index: false,
             ignored_content: Vec::new(),
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -74,6 +74,8 @@ pub struct Config {
 
     /// Whether to compile the `sass` directory and output the css files into the static folder
     pub compile_sass: bool,
+    /// Whether to minify the html output
+    pub minify: bool,
     /// Whether to build the search index for the content
     pub build_search_index: bool,
     /// A list of file glob patterns to ignore when processing the content folder. Defaults to none.
@@ -320,6 +322,7 @@ impl Default for Config {
             hard_link_static: false,
             taxonomies: Vec::new(),
             compile_sass: false,
+            minify: false,
             mode: Mode::Build,
             build_search_index: false,
             ignored_content: Vec::new(),
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -567,12 +567,11 @@ impl Site {
     pub fn render_page(&self, page: &Page) -> Result<()> {
         let output = page.render_html(&self.tera, &self.config, &self.library.read().unwrap())?;
         let content = self.inject_livereload(output);
-        let minified_content = self.minify(content).expect("Couldn't minify html file");
         let components: Vec<&str> = page.path.split('/').collect();
         let current_path = self.write_content(
             &components,
             "index.html",
-            minified_content,
+            content,
             !page.assets.is_empty(),
         )?;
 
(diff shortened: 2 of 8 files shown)


components/config/src/config/mod.rs
```diff
@@ -74,6 +74,8 @@ pub struct Config {
 
     /// Whether to compile the `sass` directory and output the css files into the static folder
     pub compile_sass: bool,
+    /// Whether to minify the html output
+    pub minify: bool,
```
Keats: this needs to be `minify_html`

components/site/src/lib.rs
```diff
@@ -540,7 +541,24 @@ impl Site {
...
+                Ok(_len) => {
+                    output = std::str::from_utf8(&mut output_bytes).expect("Unable to truncate html file.").to_string();
+                },
+                Err(MinifyError{error_type, position})=> {
+                    eprintln!("Failed to truncate html at character {}:", position);
+                    eprintln!("{}", error_type.message())
+                }
+            }
+
+        }
```
Keats: you need to have the same code everywhere we are rendering a template: section and taxonomies at least

components/site/src/lib.rs
```diff
@@ -540,7 +541,24 @@ impl Site {
...
+                minify_js: false,
+            };
+            let mut output_bytes = output.as_bytes().to_vec();
+            match truncate(&mut output_bytes, cfg) {
+                Ok(_len) => {
+                    output = std::str::from_utf8(&mut output_bytes).expect("Unable to truncate html file.").to_string();
+                },
+                Err(MinifyError{error_type, position})=> {
+                    eprintln!("Failed to truncate html at character {}:", position);
+                    eprintln!("{}", error_type.message())
```
Keats: we should bail rather printing to stderr

Keats (changes requested):

areille:
Gone to WIP status because of a problem found with `minify-html`.

It appears to turn :
`<link href="http:&#x2F;&#x2F;areille.com&#x2F;style.css" rel="stylesheet">`
into :
`<link href=http:%%areille.com%style.css rel=stylesheet>`

areille:
@Keats this is ready for review now.

components/site/src/lib.rs
```diff
@@ -445,6 +446,26 @@ impl Site {
         html
     }
 
+    /// Minifies html content
+    fn minify(&self, mut html: String) -> Result<String> {
+        if self.config.minify_html {
```
Keats: can you reverse the condition so it's not indented? eg `if not self.config.minify_html { return Ok(html) } ..`

components/site/src/lib.rs
```diff
@@ -542,9 +563,14 @@ impl Site {
     pub fn render_page(&self, page: &Page) -> Result<()> {
         let output = page.render_html(&self.tera, &self.config, &self.library.read().unwrap())?;
         let content = self.inject_livereload(output);
+        let minified_content = self.minify(content).expect("Couldn't minify html file");
```
Keats: we need to remove the expect and bubble up the error

components/site/src/lib.rs
```diff
@@ -707,7 +733,8 @@ impl Site {
         context.insert("config", &self.config);
         let output = render_template("404.html", &self.tera, context, &self.config.theme)?;
         let content = self.inject_livereload(output);
-        self.write_content(&[], "404.html", content, false)?;
+        let minified_content = self.minify(content).expect("Couldn't minify html file");
```
Keats: same here, need to remove the expect

components/site/src/lib.rs
```diff
@@ -747,7 +774,8 @@ impl Site {
         let list_output =
             taxonomy.render_all_terms(&self.tera, &self.config, &self.library.read().unwrap())?;
         let content = self.inject_livereload(list_output);
-        self.write_content(&components, "index.html", content, false)?;
+        let minified_content = self.minify(content).expect("Couldn't minify html file");
```
Keats: same here and everywhere.

Keats (changes requested):
I think we should move the minification to `write_content` so we can't forget it and no need to repeat ourselves everywhere

areille:
I moved the `minify()` call inside `write_content()`. Hope that it is ok now !

Keats:
Seems good, can you add a couple of tests for the minified output?

areille:
I added a test for `index.html`. Please let me know if this is not enough and I will add some tests for other files.

EDIT : it appears to break other tests, will correct that

Keats (approved):

Keats:
Thanks! I'll bump the pinned version myself for the build to pass

Issue #776, opened 2019-08-13
Reverse pagination


spl:
_I'm not sure if this is a bug report, documentation request, or feature request._

Is it possible to do pagination with a reversed `sort_by`?

If I do this:

```
{% for page in section.pages | reverse % }
...
{% endfor %}
```

I get the pages in the reverse order of whatever the `sort_by` field is set to in this section's `_index.md`.

However, if I add `paginate_by = 3` to that `_index.md`, the pages are first sorted (according to `sort_by`) and then paginated. So, if I do this:

```
{% for page in paginator.pages | reverse % }
...
{% endfor %}
```

I get the pages _within_ a chunk (as grouped by pagination) sorted in reverse. That is, I seem to only get pages:

1. sorted,
2. grouped by pagination, and
3. reversed within a group.

But I would like pages:

1. sorted,
2. reversed, and then
3. grouped by pagination.

My particular use case is:
* I'm using weights to sort pages.
* I want newer pages to come before older pages.
* Since weights are non-negative integers, it makes sense to start the oldest page with a weight of 0 and increment by 1 for each newer page.
* But weights are sorted from least to greatest.

I can work around this by starting with an arbitrarily large weight and decrementing by 1 for each page. However, in the long run, I think it would be better to improve Zola's pagination to support the above (if it doesn't already).

Keats:
> Is it possible to do pagination with a reversed sort_by?

No, pagination is done in Zola itself. Templates cannot generate new files.

A while ago, there was a 3rd pagination method `order` which was the opposite of `weight` but it was making the whole pagination super confusing.
A possible way to fix your issue would be to add a `reverse_pagination = true` config option in a section front-matter but I'm going to wait to have more feedback on that, as 99% of the pagination usecase is with newer-to-older dates.

spl:
> A possible way to fix your issue would be to add a `reverse_pagination = true` config option in a section front-matter

That would solve my particular use case, but I'm not sure it's the best general solution.

What about supporting compositionality in `sorted_by`? I'm not sure if this is the best/right syntax, but I could imagine the following:

```
sort_by = "weight | reverse"
sort_by = "date | reverse"
```

At the moment, I can't think of other things you'd use instead of `reverse` here, but I haven't given it much thought.

I tend to think it would be better to use a modified `sort_by` option or to define a config option related to sorting instead of to pagination.

rochacbruno:
It is also common to have `asc` and `-desc` using the `-` and implicit `+` signs example:

[...]

PR #1147, opened 2020-08-27
Reverse pagination

aslowwriter:
Sanity check:

# Description
This PR is in response to issue #776. I've wanted the reverse pagination myself so I've implemented it. I know the wat this should be exposed to the user hasn't been agreed upon yet, but I wanted to make the PR anyway for the following reasons:

1. Since pagination happens before the template get's access to the paginator it was clear to me that this would have to be set in the section front_matter so it could be handled in time. 
2. the functionality is the same regardless of how it's handled in the front matter
3. changing the names of the sort_by options is quite easy. 

Given these observations, I think the option I went with (e.g. `weight_asc` and `weight_desc` instead of `weight`) is the most organic/consistent. However, if people prefer backwards-compatibility, which would also be fair, we can revert the names of the `sort_by` and add a `paginate_reversed` option, this would be a very easy change (that was actually how I did it originally). 
If people want a different naming altogether that is also possible. I wanted to make the PR anyway so we have something to use during the discussions. I'm open to change requests or other feedback 

P.s. I also removed the `order` field in the front matter of the page since this should now be obsolete and wasn't used anywhere else.


Keats:
> Given these observations, I think the option I went with (e.g. weight_asc and weight_desc instead of weight) is the most organic/consistent. However, if people prefer backwards-compatibility, which would also be fair, we can revert the names of the sort_by and add a paginate_reversed option, this would be a very easy change (that was actually how I did it originally).

I think it's better to go the `paginate_reversed` way, the default order should be fine for most people , I'll review the PR once it's switched

aslowwriter:
@Keats Done

components/front_matter/src/lib.rs
```diff
@@ -19,11 +19,8 @@ lazy_static! {
 #[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
 pub enum SortBy {
-    /// Most recent to oldest
     Date,
-    /// Lower weight comes first
     Weight,
-    /// No sorting
     None,
```
Keats: why remove those comments?

components/library/src/content/ser.rs
```diff
@@ -150,6 +150,10 @@ impl<'a> SerializingPage<'a> {
         }
     }
 
+    // just for testing the paginator
```
Keats: can that be removed?

components/library/src/pagination/mod.rs
```diff
@@ -262,15 +273,22 @@ mod tests {
...
-        library.insert_page(Page::default());
-        library.insert_page(Page::default());
-        library.insert_page(Page::default());
+    fn create_library(
+        is_index: bool,
+        num_pages: usize,
+        paginate_reversed: bool,
+    ) -> (Section, Library) {
+        let mut library = Library::new(num_pages, 0, false);
+        for i in 1..num_pages + 1 {
```
Keats: isn't `...` equivalent to `end + 1`?

components/site/src/lib.rs
```diff
@@ -901,6 +901,24 @@ impl Site {
...
+            self.render_feed(
+                pages,
+                Some(&PathBuf::from(&section.path[1..])),
+                &section.lang,
+                |mut context: Context| {
+                    context.insert("section", &section.to_serialized(library));
+                    context
+                },
+            )?;
+        }
```
Keats: looks like there are unrelated changes in the commits

components/library/src/pagination/mod.rs
```diff
@@ -45,11 +45,13 @@ impl<'a> Pager<'a> {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Paginator<'a> {
     /// All pages in the section/taxonomy
-    all_pages: &'a [DefaultKey],
+    all_pages: Vec<DefaultKey>,
```
Keats: Would `Cow` work here to avoid the performance penalty when cloning the keys if we don't need to?

Keats (changes requested):
Can you clean the commits? I think a rebase one `next` should work, or manually removing the unrelated commits from the history

components/front_matter/src/lib.rs
aslowwriter: A mistake on my part when reverting from the `weight_asc` config. I'll revert it

components/library/src/content/ser.rs
aslowwriter: sure

components/library/src/pagination/mod.rs
aslowwriter: I looked it up `..=` is an inclusive range. I'll make that change if you prefer,

components/site/src/lib.rs
aslowwriter: probably my mistake while pulling `next` so the PR wouldn't have merge conflicts, I'll remove it

components/library/src/pagination/mod.rs
aslowwriter: good suggestion, I added it in the next commit

aslowwriter:
My git-fu isn't too great so I'm struggling a little to get the commit history correct and cleaned up but I'm working on it

aslowwriter:
Okay, I think I got rid of all unrelated changes now. I tried to clean up the commit history as well as I could too, I hope it's okay like this

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ binary size
 - Add Markdown-outputting shortcodes
 - Add `minify_html` option to `config.toml` to minify HTML automatically
 
+
```
Keats: you can add this change to that file ;)

docs/content/documentation/content/section.md
```diff
@@ -164,6 +165,13 @@ to newest (at the bottom).
 
 `reverse` has no effect on `page.later`/`page.earlier` or `page.heavier`/`page.lighter`.
 
+However this will not work correctly if the pages are paginated, because it
+will only apply to the pagers and not to all pages. For example if you have
+pages `[1,2,3,4]` and a `paginate_by=2` they will be paginated as
+`[[2,1],[4,3]]` instead of `[[4,3],[2,1]]` as you might expect. To achive this
```
Keats: achieve

docs/content/documentation/content/section.md
```diff
@@ -156,6 +156,7 @@ This will be sort all pages by their `weight` field, from lightest weight
 page gets `page.lighter` and `page.heavier` variables that contain the
 pages with lighter and heavier weights, respectively.
 
+### reversed sorting
```
Keats: Reversed sorting

docs/content/documentation/getting-started/configuration.md
```diff
@@ -88,6 +88,10 @@ compile_sass = false
 # When set to "true", the HTML output files will be minified
 minify_html = false
 
+# When set to "true", a search index is built from the pages and section
+# content for `default_language`.
+build_search_index = false
```
Keats: Isn't it already in there?

components/library/src/content/ser.rs
```diff
@@ -150,6 +150,10 @@ impl<'a> SerializingPage<'a> {
         }
     }
 
+    pub fn get_title(&'a self) -> &'a Option<String> {
```
Keats, line 154: Can you add back the docstring saying it's only used in tests?

Keats (changes requested):
Looks like we need to add `.swp` to gitignore

docs/content/documentation/content/section.md
aslowwriter: does that mean you want this part removed?

Keats: no, just the typo

Keats (approved):

Keats:
Thanks!

PR #1148, opened 2020-08-28
WIP: Refactor internationalization system

BertalanD:
This pull request implements the redesigned localization system as described in my [forum post].

In summary, the goal is to be transparent towards templates, but customizable and explicit towards users. A [Project Fluent]-based string translation system will eventually be introduced that's similar to #1040 by @XAMPPRocky.

See the commit messages for an in-depth changelog.

# To-do
- [X] Refactor existing code
- [ ] Merge `extra` from page/section header?? + language in front matter
- [x] Add proper linking in doc comments
- [ ] Add more tests
- [x] Add fluent function
- [ ] Load locales from both site and theme
- [ ] Generate Fluent stuff on init
- [ ] Reduce the amount of `clone()`s (kind of done)
- [ ] User testing
- [X] Add user documentation
- [ ] Add template API documentation
- [ ] Create a demo theme or internationalize an existing theme

[forum post]: https://zola.discourse.group/t/rfc-internationalization-system-rework/546
[Project Fluent]: https://www.projectfluent.org/

### DIFF
--- a/components/config/src/config/languages.rs
+++ b/components/config/src/config/languages.rs
@@ -1,16 +1,95 @@
 use std::collections::HashMap;
 
 use serde_derive::{Deserialize, Serialize};
+use toml::Value as Toml;
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+/// Contains site settings that can be set differently for each language
+///
+/// When rendering pages, this will be merged with the options for the default language, the
+/// details of which are documented for each field separately.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 #[serde(default)]
-pub struct Language {
-    /// The language code
-    pub code: String,
-    /// Whether to generate a feed for that language, defaults to `false`
-    pub feed: bool,
-    /// Whether to generate search index for that language, defaults to `false`
-    pub search: bool,
+pub struct LocaleOptions {
+    /// How to refer to the language in URLs
+    ///
+    /// The localization functions require a valid language identifier to work correctly. Some
+    /// multilingual sites might already be using language names that do not conform to the
+    /// canonical [BCP 47] syntax, and this lets us keep links to them working currently.
+    ///
+    /// If empty or not set, the `language_code` will be used.
+    ///
+    /// [BCP 47]: https://tools.ietf.org/html/bcp47
+    pub language_alias: String,
+
+    /// Title of the site. Defaults to none
+    ///
+    /// This will not fall back to how it's set for the default language.
+    pub title: Option<String>,
+    /// Description of the site. Defaults to None.
+    ///
+    /// This will not fall back to how it's set for the default language.
+    pub description: Option<String>,
+
+    /// Taxonomies available for a language
+    ///
+    /// This will not fall back to how it's set for the default language.
+    pub taxonomies: Vec<super::taxonomies::Taxonomy>,
+
+    /// Whether to generate a feed for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// Its default value for the primary language is `false`.
+    pub generate_feed: Option<bool>,
+    /// The number of articles to include in the feed for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// Its default value for the primary language is all.
+    ///
+    /// All has been represented as None since the beginning, hence the awkward Option<Option<_>>.
+    pub feed_limit: Option<Option<usize>>,
+
+    /// Whether to generate search index for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// Its default value for the default language is `false`.
+    pub build_search_index: Option<bool>,
+    /// The search config, telling what to include in the search index for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// If unset for the default language, it will be set to [Search::default].
+    pub search: Option<super::search::Search>,
+
+    /// All user parameters set in `[extra]`
+    ///
+    /// This supersedes `[translations]` from previous versions. Serde ensures that this is a Toml
+    /// table by attempting to convert it into a `HashMap<String, toml::Value>`. This would,
+    /// however, make us do `.get().unwrap().get().unwrap()` for retrieveing a nested value; but we
+    /// treat this as opaque data, and do not attempt to modify it.
+    ///
+    /// For localizations, this will be recursively merged with that of the default language.
+    pub extra: HashMap<String, Toml>,
 }
 
-pub type TranslateTerm = HashMap<String, String>;
+impl LocaleOptions {
+    /// Sets the default values when set for the default language.
+    ///
+    /// This had to be done here, because serde does not yet support #[serde(default)] for newtype
+    /// structs.
+    pub fn default_for_default_lang() -> Self {
+        LocaleOptions {
+            language_alias: String::new(),
+            title: None,
+            description: None,
+            taxonomies: Vec::new(),
+            generate_feed: Some(false),
+            feed_limit: Some(None),
+            build_search_index: Some(false),
+            search: Some(super::search::Search::default()),
+            extra: HashMap::new(),
+        }
+    }
+
+    pub fn default_for_additional_lang() -> Self {
+        LocaleOptions::default()
+    }
+}
--- a/components/config/src/config/languages.rs
+++ b/components/config/src/config/languages.rs
@@ -16,7 +16,7 @@ pub struct LocaleOptions {
     /// multilingual sites might already be using language names that do not conform to the
     /// canonical [BCP 47] syntax, and this lets us keep links to them working currently.
     ///
-    /// If empty or not set, the `language_code` will be used.
+    /// If empty or not set, the language code will be used.
     ///
     /// [BCP 47]: https://tools.ietf.org/html/bcp47
     pub language_alias: String,
@@ -73,7 +73,7 @@ pub struct LocaleOptions {
 impl LocaleOptions {
     /// Sets the default values when set for the default language.
     ///
-    /// This had to be done here, because serde does not yet support #[serde(default)] for newtype
+    /// This had to be done here, because serde does not yet support `#[serde(default)]` for newtype
     /// structs.
     pub fn default_for_default_lang() -> Self {
         LocaleOptions {
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -4,7 +4,7 @@ pub mod search;
 pub mod slugify;
 pub mod taxonomies;
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 
 use globset::{Glob, GlobSet, GlobSetBuilder};
@@ -156,27 +156,27 @@ impl Config {
             bail!("Highlight theme {} not available", config.highlight_theme)
         }
 
+
         config.lang = config.default_language.clone();
-
-        // Find duplicate language aliases/names by first extracting them to a vector while also
-        // setting aliases.
-        let mut language_names: Vec<String> = Vec::new();
-
         if config.default_language_options.language_alias.is_empty() {
             config.default_language_options.language_alias = config.default_language.to_string();
         }
-        language_names.push(config.default_language_options.language_alias.clone());
         for (identifier, options) in &mut config.languages {
-            if options.language_alias == "" {
+            if options.language_alias.is_empty() {
                 options.language_alias = identifier.to_string();
             }
-            language_names.push(options.language_alias.clone());
         }
-        // https://stackoverflow.com/a/46766782
-        // TODO: actually print which identifier/alias is causing a problem
-        if (1..language_names.len()).any(|i| language_names[i..].contains(&language_names[i - 1])) {
-            bail!("A language code or alias should not appear twice in config.toml");
+
+
+        let mut language_names = HashSet::with_capacity(config.languages.len() + 1);
+        let _ = language_names.insert(&config.default_language_options.language_alias);
+        for options in config.languages.values() {
+            let duplicate = !language_names.insert(&options.language_alias);
+            if duplicate {
+                bail!("Multiple languages cannot have the same name: {}", options.language_alias)
+            }
         }
+
 
         if !config.ignored_content.is_empty() {
             // Convert the file glob strings into a compiled glob set matcher. We want to do this once,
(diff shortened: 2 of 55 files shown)


Keats:
@BertalanD I will have a look at that soon, I just want to get a 0.12 release first so this PR can be the main focus of 0.13

alerque:
I'll definitely have my eyes pealed for 0.13 or when this PR hits master.

BertalanD:
I've just been put in quarantine, so I'll start working on it tomorrow.

BertalanD:
I'm finally done with the core functionality and have written up some user docs.

Updating the docs for the template API is a big task, so before that, I would like to ask for help in a couple of places:
- LocalizedConfig and the copied standalone Tera contexts for each language seem awkward and slow for me. Does anybody have an idea how they might be replaced?
- Are there any config values you would like to be overridable for each language, but aren't yet?
- Doing `{{ config.extra.whatever }}` in templates seems awkward to me (especially since `extra` values come from translations, themes, and hopefully in the future, from the front matter, too). Any better ideas?

Lastly, I'd like @Keats' help about what tests to implement and finding beta testers.

Requesting a review.

components/config/src/config/languages.rs
```diff
@@ -1,16 +1,95 @@
...
+    /// Taxonomies available for a language
+    ///
+    /// This will not fall back to how it's set for the default language.
+    pub taxonomies: Vec<super::taxonomies::Taxonomy>,
+
+    /// Whether to generate a feed for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// Its default value for the primary language is `false`.
+    pub generate_feed: Option<bool>,
```
Keats, line 42: I would rather deprecate the top `generate_feed` in favour of `feed` than having `generate_feed` everywhere.

components/config/src/config/languages.rs
```diff
@@ -1,16 +1,95 @@
...
+    /// Its default value for the primary language is all.
+    ///
+    /// All has been represented as None since the beginning, hence the awkward Option<Option<_>>.
+    pub feed_limit: Option<Option<usize>>,
+
+    /// Whether to generate search index for a language
+    ///
+    /// If unset in a translation, it will fall back to how it's set for the default language.
+    /// Its default value for the default language is `false`.
+    pub build_search_index: Option<bool>,
```
Keats, line 55: Note: we should show a message if the language cannot be built into a search index, only a few languages are supported right now.

components/config/src/config/mod.rs
```diff
@@ -100,16 +137,10 @@ pub struct Config {
 
     /// The setup for which slugification strategies to use for paths, taxonomies and anchors
     pub slugify: slugify::Slugify,
-
-    /// The search config, telling what to include in the search index
-    pub search: search::Search,
-
-    /// All user params set in [extra] in the config
-    pub extra: HashMap<String, Toml>,
 }
```
Keats, line 140: I don't know about that..
Things like `compile_sass`, `minify_html` etc are not language specific. Maybe pass something like `language_config` to the template instead?

components/config/src/config/mod.rs
```diff
@@ -125,8 +156,26 @@ impl Config {
...
+        for (identifier, options) in &mut config.languages {
+            if options.language_alias == "" {
+                options.language_alias = identifier.to_string();
+            }
+            language_names.push(options.language_alias.clone());
+        }
+        // https://stackoverflow.com/a/46766782
+        // TODO: actually print which identifier/alias is causing a problem
+        if (1..language_names.len()).any(|i| language_names[i..].contains(&language_names[i - 1])) {
+            bail!("A language code or alias should not appear twice in config.toml");
```
Keats: just use a set instead of a vec?

components/config/src/config/mod.rs
```diff
@@ -262,26 +393,39 @@ impl Config {
...
+            Ok(LocalizedConfig(self.clone()))
+        } else if self.languages.contains_key(lang) {
+            let mut languages = self.languages.clone();
+            let (_, default_language_options) = languages.remove_entry(&lang).unwrap();
+            languages.insert(self.default_language.clone(), self.default_language_options.clone());
+            Ok(LocalizedConfig(Config {
+                lang: lang.clone(),
+                default_language_options,
+                languages,
+                ..self.clone()
```
Keats, line 407: I guess this means we clone the Config every time we render a template?

components/config/src/config/mod.rs
```diff
@@ -312,35 +454,63 @@ impl Default for Config {
...
-            title: None,
-            description: None,
             theme: None,
             highlight_code: false,
             highlight_theme: "base16-ocean-dark".to_string(),
-            default_language: "en".to_string(),
-            languages: Vec::new(),
-            generate_feed: false,
-            feed_limit: None,
+            lang: langid!("en"),
```
Keats, line 460: do we need that macro? Is the macro-less approach harder to use?

components/config/src/config/taxonomies.rs
```diff
@@ -11,9 +13,12 @@ pub struct Taxonomy {
     pub paginate_path: Option<String>,
     /// Whether to generate a feed only for each taxonomy term, defaults to false
     pub feed: bool,
-    /// The language for that taxonomy, only used in multilingual sites.
-    /// Defaults to the config `default_language` if not set
-    pub lang: String,
+
+    // Only for convenience/backwards compatibility
+    #[serde(skip_deserializing)]
+    pub lang: LanguageIdentifier,
```
Keats, line 19: that's still going to break if someone used an invalid language identifier right?

components/config/src/theme.rs
```diff
@@ -2,40 +2,55 @@ use std::collections::HashMap;
...
-            if let Some(ex) = theme_table.get("extra") {
-                if ex.is_table() {
-                    extra = ex.clone().try_into().unwrap();
-                }
-            }
-        } else {
-            bail!("Expected the `theme.toml` to be a TOML table")
-        }
-
-        Ok(Theme { extra })
```
Keats, line 38: how come we don't need that anymore?

components/library/src/content/file_info.rs
```diff
@@ -26,15 +34,48 @@ pub fn find_content_components<P: AsRef<Path>>(path: P) -> Vec<String> {
...
+        return (
+            caps.name("name").unwrap().as_str().to_string(),
+            Some(caps.name("datetime").unwrap().as_str().to_string()),
+        );
+    }
+    (name.as_ref().to_string(), None)
+}
+
+/// Strip `{date}{_,-}` prefix and `.{lang}` suffix from name
+pub fn clear_filename<S: AsRef<str>>(name: S) -> String {
```
Keats: I don't like the name but I can't think of anything better gith now.

components/library/src/content/file_info.rs
```diff
@@ -49,26 +90,39 @@ pub struct FileInfo {
     /// This is `parent` + `name`, used to find content referring to the same content but in
     /// various languages.
     pub canonical: PathBuf,
+    /// The `lang` part of the {name}.{lang}.md naming scheme. There is no guarantee that this
+    /// language is enabled for the site.
+    pub maybe_lang: Option<String>,
```
Keats: let's call it lang rather than maybe_lang.

components/library/src/content/page.rs
```diff
@@ -110,27 +104,47 @@ impl Page {
...
+                // TODO: add test
+                bail!(
+                    "Page `{}` has a language set, but `languages` in config.toml is empty. Note that file names must not contain dots.",
+                    file_path.display()
+                );
+            } else if l.is_empty() {
+                // TODO: add test
+                bail!("Page `{}` has an empty language. Did you mistakenly put 2 dots before the extension?", file_path.display());
+            } else if l == &config.default_language_options.language_alias {
+                bail!("Pages in the default language should not set it in their title (caused by `{}`).", file_path.display());
```
Keats, line 118: does that cause any issues?

components/library/src/content/page.rs
```diff
@@ -110,27 +104,47 @@ impl Page {
...
-            }
-        } else {
-            page.file.name.replace(".md", "")
-        };
-        if let Some(ref caps) = RFC3339_DATE.captures(&file_path) {
-            slug_from_dated_filename = Some(caps.name("slug").unwrap().as_str().to_string());
+        if let Some(ref date) = page.file.date {
             if page.meta.date.is_none() {
-                page.meta.date = Some(caps.name("datetime").unwrap().as_str().to_string());
+                // TODO: do not fail silently on invalid formats
```
Keats, line 146: do we have a test for that?

components/library/src/content/section.rs
```diff
@@ -86,21 +90,54 @@ impl Section {
...
+            }
+        } else {
+            section.lang = config.default_language.clone();
+            section.language_alias = config.default_language_options.language_alias.clone();
+        }
+
+        // Catch programming errors
+        // `lang` and `language_alias` must be non-empty
+        assert_ne!(section.lang, LanguageIdentifier::default());
+        assert!(!section.language_alias.is_empty());
```
Keats, line 124: looks like an almost copy/paste than the page.rs file

components/search/src/lib.rs
```diff
@@ -26,42 +26,45 @@ lazy_static! {
     };
 }
 
-fn build_fields(config: &Config) -> Vec<String> {
+fn build_fields(config: &LocalizedConfig) -> Vec<String> {
     let mut fields = vec![];
-    if config.search.include_title {
+    let search = config.0.default_language_options.search.clone().expect("search set");
```
Keats, line 31: can we ever have None?

components/search/src/lib.rs
```diff
@@ -80,26 +83,28 @@ fn fill_index(
...
-pub fn build_index(lang: &str, library: &Library, config: &Config) -> Result<String> {
-    let language = match Language::from_code(lang) {
-        Some(l) => l,
-        None => {
-            bail!("Tried to build search index for language {} which is not supported", lang);
-        }
-    };
+pub fn build_index(library: &Library, config: &LocalizedConfig) -> Result<String> {
+    let language = Language::from_code(config.0.lang.language.as_str()).ok_or_else(|| {
+        format!("Tried to build search index for language {} which is not supported", config.0.lang)
```
Keats, line 88: we can move that error at start time

Keats (changes requested):
Ok I started reviewing it but the PR is way too big to be properly reviewed on its own.
Can we split it in smaller PRs?

BertalanD:
I'm not sure how to split this up. The new `LocaleOptions` + `LanguageIdentifier` stuff takes up most of the change anyway, and dividing that into parts if not really an option, because almost all lines referencing `Config` would still need to be modified.
I'm open to leaving the Fluent/Docs part for later.

Sorry for the code spam.

components/config/src/config/mod.rs
BertalanD: We could have it be parsed from a str. Parsing and using an `unwrap()` in a constructor seems a bit dodgy to me.

components/config/src/theme.rs
BertalanD: The code previously parsed the toml into  a Toml object and just copied the `extra` key from it into the struct. This did not use Serde at all.

The new code just uses Serde to directly parse that into the `Theme` struct. This actually saves a `clone()` and having to do manual error handling. Also this code won't need to be modified for an eventual YAML or JSON (ew) config option.

I should probably look up if an attribute needs to be set to ignore unknown keys.

components/library/src/content/file_info.rs
BertalanD: `strip` as in my comment?

components/library/src/content/page.rs
BertalanD: Don't think so. The code currently does fail though if the default language is set in a file name.

```rust
if !config.languages_codes().contains(&parts[1].as_ref()) {
            bail!("File {:?} has a language code of {} which isn't present in the config.toml `languages`", self.path, parts[1]);
        }
```

components/config/src/config/mod.rs
BertalanD: Yep. I mainly want to have language-specific search because CJK search indices can get huge; and tight bandwidth might exclude things like feeds

components/config/src/config/mod.rs
BertalanD: Unfortunately, yes. I'm thinking about having a struct with just references like `SerializingTaxonomy` and co. I'm way too inexperienced to not resort to cloning something with that many borrows.

components/config/src/config/taxonomies.rs
BertalanD: This field [won't be read](https://serde.rs/field-attrs.html#skip_deserializing) from the config, just passed to templates. We set it from the containing `LocaleOptions`' key in `config.languages`. Any invalid language will error out when parsing the config.

I might not get what you're saying though.

components/rendering/tests/codeblock_hl_lines.rs
```diff
@@ -40,20 +43,28 @@ fn hl_lines_simple() {
...
-    "#, &context).unwrap();
-    assert_eq!(res.body, colored_html!(
-        @no "foo",
-        @hl "bar",
-        @no "bar\nbaz",
-    ));
+    "#,
+        &context,
+    )
+    .unwrap();
```
BertalanD, line 59: Rustfmt really messed up this part. Sorry about that. I probably should revert changes to this file.

components/config/src/config/languages.rs
BertalanD: It would feel a bit ridiculous alongside `build_search_index`. How would you feel about also renaming `build_search_index` to `searchable`?

Keats:
I put a comment on https://zola.discourse.group/t/rfc-internationalization-system-rework/546/7, let's think about the "what are we trying to achieve" first so you can do more incremental changes.

Keats:
sorry >_>

Keats:
Closing in favour of https://github.com/getzola/zola/pull/1391 as it addresses some existing issues and this PR would be too much work to rebase cleanly.

Thanks a lot @BertalanD !

alerque:
This PR (and #1040) should be reconsidered. I saw in notifications that #1391 got merged and came back to check out the project and see if it would work yet for any of my sites. The answer seems to be no, the i18n work actually done there is very rudimentary. I don't mean this to be rude @Keats, but please consider that while you may not have a use case for a true multilingual site this is a pretty big deal — for many a make or break issue. Please consider trusting the people that are trying to contribute something more here. Having a Fluent based static site generator with multi language templating capabilities would be a big deal. Seeing multiple people try to contribute towards this and having their issues shut down is immensely discouraging. This is still up for grabs because no other static site generator does this right yet either, but as long as you keep shutting down PRs that attempt to address this I don't see how we're going to get anywhere.

PR #1129, opened 2020-08-14
Strip unc if it exists, fix #1110

uggla:
Sanity check:

* [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
* [X] Are you doing the PR on the `next` branch?

Explanation:
Remove unc on windows platform see #1110. Unfortunately I have no windows system at the moment to check that this patch is working as expected.
I'm opening this PR to check if the CI is ok.

PR #1151, opened 2020-08-31
Strip unc if it exists, fix #1110 #1129

### LINKED ISSUE #1110: Your site was created in ...
# Bug Report

`zola init`
`Done! Your site was created in "\\\\?\\C:\\Users\\Mus\\Code\\testwebsite"`

Would be nice to remove the UNC path prefix since it's unecessary.

uggla:
Sanity check:

## Code changes

Explanation:
Remove unc on windows platform see #1110 .
Sorry I mistakenly closed the previous PR on that topic #1129 .
 
So here is the new one I did 3 things:
* Bump to rust 1.45 version which is the minimal one required.
* Put a comment as requested on the LOCAL_UNC const.
* Add a new test that will fail as soon as the canonicalize function will be fixed by the rust team. So this workaround will have to be removed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ stages:
               rustup_toolchain: stable
             linux-pinned:
               imageName: 'ubuntu-16.04'
-              rustup_toolchain: 1.43.0
+              rustup_toolchain: 1.45.2
         pool:
           vmImage: $(imageName)
         steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ stages:
               rustup_toolchain: stable
             linux-pinned:
               imageName: 'ubuntu-16.04'
-              rustup_toolchain: 1.45.2
+              rustup_toolchain: 1.45
         pool:
           vmImage: $(imageName)
         steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ stages:
               rustup_toolchain: stable
             linux-pinned:
               imageName: 'ubuntu-16.04'
-              rustup_toolchain: 1.45
+              rustup_toolchain: 1.45.0
         pool:
           vmImage: $(imageName)
         steps:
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

azure-pipelines.yml
```diff
@@ -21,7 +21,7 @@ stages:
               rustup_toolchain: stable
             linux-pinned:
               imageName: 'ubuntu-16.04'
-              rustup_toolchain: 1.43.0
+              rustup_toolchain: 1.45.0
```
Keats, line 24: why do we need the bump btw?

uggla: @Keats, 
```
error[E0658]: use of unstable library feature 'str_strip': newly added
  --> src/cmd/init.rs:65:26
   |
65 |     match path_to_refine.strip_prefix(LOCAL_UNC) {
   |                          ^^^^^^^^^^^^
   |
   = note: see issue #67302 <https://github.com/rust-lang/rust/issues/67302> for more information

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: could not compile `zola`.
```
This was stabilized in 1.45, (https://github.com/rust-lang/rust/blob/master/RELEASES.md#stabilized-apis-1), I manage to confirm using rustc as well. So minimum version to bump is 1.45 to avoid this error. I'm gonna change that accordingly.

Keats: Wouldn't `trim_start_matches` do the same?

uggla: In that case, I think yes. To be honest I used strip_prefix because I looked at the documentation and it was what I needed and I did not look further at it. (shame on me)
As I used a version > 1.45, I have not realize that this method was not stable before that release.
Anyway, would you like me to use trim_start_matches instead of strip_prefix and come back to 1.43 for the pinned version ?

Keats: If you have time and don't mind sure

OvermindDL1:
It looks like this is only doing a trivial string prefix removal instead of actually testing that it's valid.  UNC paths can represent paths on windows that are otherwise unrepresentable, so removing it can make a path invalid.  It should only be removed (if at all) if the resultant path is still valid.  This is also why the current rust method returns what it does, because it is the kernel itself performing the canonicalization.

PR #1154, opened 2020-09-02
Better minify errors

### LINKED ISSUE #1152: minfying has unhelpful error messaging and breaks on html chars in inner text
# Bug Report

## Environment

Zola version: 0.12.0 on `next` branch

## Expected Behavior
I tried to build my own website with minification turned on. The website should get built and be minified. 

## Current Behavior
when I build my website with minification turned on I simply get this message: 
```
Error: Failed to truncate html at character: 15769
```
with no further information, making it impossible to tell what the problem is for the user. As I have my own version installed I did some extra digging and found the issue causing the crash for which I will open a separate issue, but this should be fixed in some way. At least I think the file currently being minified should be displayed, but ideally, a window around the error in the string currently being minified would also be shown. Displaying the entire string helps but is still very cumbersome as the files can be very large so displaying all of it is unhelpful. 

As for what is causing the exit, after doing some digging in my own installation I found out that it is because one of my pages containing the text `<<` (for nav buttons) so it seems that the inner HTML text is not properly escaped/encoded before it is minified. 

## Step to reproduce
```bash
mkdir bug && cd bug
zola init
sed -i  's/\[extra\]/minify_html = true \n [extra]/' config.toml # enable minification
echo '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd"> <html lang="en"> <head> <meta http-equiv="content-type" content="text/html; charset=utf-8"> <title>Title Goes Here</title> </head> <body> <p>This is my web page</p>
        <p> <<<<<<  </p><script src="/livereload.js?port=1024&amp;mindelay=10"></script></body> </html> 
' >> templates/index.html
zola serve
```

areille:

Sanity check:

## Code changes

## Description

Closes #1152.

### DIFF
--- a/components/site/Cargo.toml
+++ b/components/site/Cargo.toml
@@ -8,7 +8,7 @@ include = ["src/**/*"]
 [dependencies]
 tera = "1"
 glob = "0.3"
-minify-html = "0.3.6"
+minify-html = "0.3.8"
 rayon = "1"
 serde = "1"
 serde_derive = "1"
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -11,7 +11,7 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use glob::glob;
 use lazy_static::lazy_static;
-use minify_html::{truncate, Cfg};
+use minify_html::{with_friendly_error, Cfg};
 use rayon::prelude::*;
 use tera::{Context, Tera};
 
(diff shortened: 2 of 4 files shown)


areille:
@Keats do you want me to revert back the previous pr (#1146) in this one ?

Keats:
It's already merged in released in 0.12, it just needs to be enabled + add back the  docs

Keats:
Thanks!

areille:
@Keats should I open a new PR to re-enable minify-html ? I saw that it is forced set to false in config's `mod.rs`

Keats:
Yes please! Hopefully the big issues are fixed now

PR #1155, opened 2020-09-02
fix `get_taxonomy_url` to handle merged taxonomies properly

aslowwriter:
This is a small bug fix. The `get_taxonomy_url` wasn't handling the newly merged taxonomies properly, this PR fixes that, and adds a test for it. Prior to this fix, Zola will crash on taxonomies that would be merged if `get_taxonomy_url` is used.


Keats (approved):

Keats:
Thanks!

PR #1159, opened 2020-09-05
Add Void Linux installation instructions

sburris0:

Sanity check:

It appears that Void Linux installation instructions were removed from the docs when the project was renamed, presumably because Void hadn't update its package to reflect the change at the time.


Keats:
Thanks!

PR #1171, opened 2020-09-18
Make gitlab CI example script faster

raffomania:
By using alpine, the time to pull and start the container decreases.
By using the official alpine zola package, the script gets simpler.
Also, this updates the zola version to the latest release.

I can understand if you don't want to depend on a lesser known distro for this example, but I thought you might also like faster builds :)

Sanity check:


Keats:
Thanks!

PR #1210, opened 2020-10-16
Fix RSS block example in documentation

kiyanmair:
Copying the commit message here:

> This commit fixes a bug in the documentation for the feeds page which
> shows how to use the RSS block to enable feed autodiscovery.
> 
> The bug used an em space character instead of a space in one part of the
> code snippet. If a user were to copy the code snippet into their code as
> directed, the Zola build would fail with a parse error. The em space
> appears identical to a regular space in monospaced fonts, making the
> error seem mysterious or incorrect.
> 
> I believe the em space was used in order to prevent the shortcode from
> rendering, as the code snippet is just meant to show what the shortcode
> looks like. However, it is possible to escape the shortcode so that it
> renders correctly without causing confusion for the user who expects to
> be able to copy and paste it.
> 
> This commit replaces the em space in both code snippets with regular
> spaces and escapes the shortcodes.

I actually looked through the parser code and compared older Zola versions before I realised the problem was just a rogue space haha. Hopefully this will save others the confusion.

--

Sanity check:

### DIFF
--- a/docs/content/documentation/templates/feeds/index.md
+++ b/docs/content/documentation/templates/feeds/index.md
@@ -48,7 +48,7 @@ As an example this is how it looks like using [Firefox](https://en.wikipedia.org
 You can enable posts autodiscovery modifying your blog `base.html` template adding the following code in between the `<head>` tags.
 ```html
 {% block rss %}
-  <link rel="alternate" type="application/rss+xml" title="RSS" href="{{ get_url(path="rss.xml", trailing_slash=false) }}">
+  <link rel="alternate" type="application/rss+xml" title="RSS" href="{{/* get_url(path="rss.xml", trailing_slash=false) */}}">
 {% endblock %}
 ```
 You can as well use an Atom feed using `type="application/atom+xml"` and `path="atom.xml"`.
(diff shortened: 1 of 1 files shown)


kiyanmair:
Sorry, I realise now that the em spaces were probably intentional to avoid Zola attempting to expand the shortcode.

However, I feel that using an em space is confusing because it's not *obviously* incorrect. Would you consider changing the em space to a more clearly incorrect character that could then be replaced with a space? At the very least, I think it should be clear in the documentation that the code needs to be changed before using.

Or perhaps there's another method that would allow copying and pasting without shortcode expansion? I imagine that other SSGs have had to deal with this issue too. I can create a separate issue if that's preferable to discuss the em space situation more thoroughly, I had just thought it was a mistake.

Either way I'm happy to contribute towards clarifying this situation. Please let me know if I'm misinterpreting anything. Thank you!

kiyanmair:
Sorry, I didn't realise there was an existing way to escape shortcodes that just wasn't followed for this code snippet. I'm still new to Zola. I didn't want to clutter the commit graph with a second commit for such a minor detail, so I just fixed up my existing commit.

Keats:
Thanks!

Issue #927, opened 2020-01-19
Sections can’t be drafted


manu-cyber:
Sections don’t have the `draft` variable, so setting `draft = true` still builds them and includes them in the output.
Would it be a desirable feature to be able to save sections as drafts?

Keats:
What would a draft section mean?

manu-cyber:
One that wouldn’t be rendered/built with the `build` command. Would that make sense?

Keats:
I get that, but I meant more what's a usecase for a draft section? Would that draft status apply to all pages/sections below it?

manu-cyber:
I guess it would have to, for it to make sense.
I use sections to organise my notes in a hierarchical, tree-like manner. Doing so, I can add sub-sections everywhere I like, or just add "regular" content just as I would for pages.
It could also make sense if one was to extend a website by a whole section, e.g. "projects", but would want to wait until every page in that section, i.e. every project, has a description one is happy with.

Keats:
Makes sense. It's a bit tricky to implement though since section processing is done without any specific order so we might end up adding a child section of a draft section before seeing that it shouldn't have been added.
If someone has a good idea, feel free to work on it!

jasikpark:
In my opinion, simply making the drafts variable a single page flag and adding a warning to the output letting you know that your draft section has non-draft pages beneath it, explaining why that’s not a good idea, seems best. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jasikpark:
or defining the drafts variable a recursive search upwards until you reach the root for *every* page + seems like an inefficient but simple and correct way to do it?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

florianb:
We built single-page sites and missed something like a section-based `render`-flag.

I don't know the mechanisms behind the scenes yet, but wouldn't it be possible to simply derive these flags?

Keats:
Zola doesn't want to assume too many things, and assuming that a section is a draft because all pages are a draft is not something wanted: the section could have a "Coming soon" page for example.

aslowwriter:
I have a working implementation for this I just need some time to clean up the PR

PR #1218, opened 2020-10-29
Make sections draftable

aslowwriter:
This PR is in response to #927. With the current implementation the content directory is processed in a recursive infix order (basically like a classic tree walk). This did require a bit of an ugly loop, but has the added benefit that it skips processing drafted sections entirely which is a nice free performance save. It also has no impact on the default cases so almost all the tests could remain unchanged. 

Addition to the docs about this is quite short, but not sure how much else would be needed. Open to feedback.  I implemented this because I use an `unpublished` folder in my content for WIPs that I wanted to make a draft to make sure none of those would ever end up in the published website. Could add this as an example if that seems useful to people. 

Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2205,10 +2205,12 @@ dependencies = [
  "search",
  "serde",
  "serde_derive",
+ "slotmap",
  "tempfile",
  "templates",
  "tera",
  "utils",
+ "walkdir",
 ]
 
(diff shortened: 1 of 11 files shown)


Keats (approved):
This looks good to me. Hopefully we can get some Windows users to double check the new page/section loading still work for them.

Issue #672, opened 2019-04-27
Zola needs a default theme that always runs unless otherwise specified.


jrmiller82:
Even if it's CSS free, it needs to have a default theme that slaps up pages quickly and serves as a base for theming and templating.

Keats:
What would the default templates show? Just the rendered content?

jrmiller82:
Yes.

trevordmiller:
This would be nice. I'd love to defer to a default simple theme.

PR #1220, opened 2020-10-29
init now builds a default barebones theme

aslowwriter:
This PR is in response to #672 I wasn't sure what people wanted from a default theme, so I went for the most bare bones thing possible: No CSS, an `index.html` that just displays the title and any pages and their content, or some filter text if those are not available, a 404 page that just says "not found" and page.html that just lists the title and content of a page. Since they were small enough I embedded them in the `cmd::init` file for now just like the `Config.toml`, but this will likely not work for a more fleshed out basic theme. If no theme is supplied, it will look for this one. Since any other theme or template will override this it should not have an impact on most users.

I have not updated docs yet because I wanted to get some feedback on the theme beforehand. If people want adjustments to the theme or the way it is generated, I'm open to suggestions.

## Checklist:


aslowwriter:
A question for @Keats with the idea of backwards compatibility. If the default theme is not found, should we just rebuild it, or just fail?

src/cmd/init.rs
```diff
@@ -131,6 +205,21 @@ fn populate(path: &Path, compile_sass: bool, config: &str) -> Result<()> {
         create_dir(path.join("sass"))?;
     }
 
+    setup_default_theme(&path.join("themes"))?;
+    Ok(())
+}
+
+fn setup_default_theme(path: &Path) -> Result<()> {
```
Keats, line 212: I think we don't want a theme, we just want some default templates. We do have some that get automatically used in `serve` and I would prefer expanding those. Once they are in a good shape, we can write them to disk in the init.

Keats (changes requested):

src/cmd/init.rs
aslowwriter: I've tried to look for them but for the life of me I can't find where those get defined. Mind pointing me in the right direction?

Keats: https://github.com/getzola/zola/blob/master/components/utils/src/templates.rs extremely basic as you can see, https://github.com/getzola/zola/blob/master/components/utils/src/default_tpl.html for the actual template.

PR #1226, opened 2020-11-05
cmd/serve: stop removing output folder

### LINKED ISSUE #1173: `zola serve` deletes /public/
# Bug Report

## Environment

Zola version: 0.12.0

## Expected Behavior
I would expect Zola to not delete the build directory after I run `zola serve`. (imo `zola serve` should be conceptually equivalent to running `zola build` and then `cd public && python3 -m http.server`, and if it doesn't behave like that then it should be clearly documented)

## Current Behavior
Once `zola serve` exits, the /public/ directory no longer exists.

## Step to reproduce
- run `zola build`
- notice that /public/ exists
- run `zola serve`
- ctrl-c
- notice that /public/ is gone

a8nJ:
Serve currently removes output folder on exit. This is not desirable, as some
users will simply have to remake the folder with `zola build` every single time
they exit the server.

Fixes #1173

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -21,7 +21,7 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-use std::fs::{read_dir, remove_dir_all};
+use std::fs::read_dir;
 use std::path::{Path, PathBuf};
 use std::sync::mpsc::channel;
 use std::thread;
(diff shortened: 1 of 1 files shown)


Keats:
This is not doing what you think it is doing. zola serve doesn't write HTML files to disk so you would be missing most of your site.

> This is not desirable, as some
users will simply have to remake the folder with zola build every single time
they exit the server.

This is desirable, the HTML output of zola serve is not suitable to be the live site (base url change, livereload injection etc).

a8nJ:
`zola serve` should not be removing anything that could be the result of `zola build`!

If `zola build` outputs to `public`, `zola serve` should not use or touch that folder!

If the issue is that `zola serve` also uses `public` by default, that should be changed so that, either `zola serve` runs from memory, or `zola serve` runs from some temporary folder. How can it possibly make sense that a user:

1. run `zola build`, which output to `public`
2. run `zola serve` to test the result
3. have to run `zola build` again, as `zola serve` removed everything from step 1

is this really what you are suggesting?

PR #1236, opened 2020-11-11
Add shortcode 'invocation' variable to allow a shortcode to track how…

NathanaelLane:
… many times it has been invoked in a given Markdown file., as per [this Discourse topic](https://zola.discourse.group/t/can-shortlinks-get-an-invocation-count-variable/671).

Sanity check:

## Code changes

### DIFF
--- a/components/rendering/src/shortcode.rs
+++ b/components/rendering/src/shortcode.rs
@@ -114,7 +114,7 @@ fn render_shortcode(
         // Trimming right to avoid most shortcodes with bodies ending up with a HTML new line
         tera_context.insert("body", b.trim_end());
     }
-    tera_context.insert("invocation", &invocation_count);
+    tera_context.insert("nth", &invocation_count);
     tera_context.extend(context.tera_context.clone());
 
     let mut template_name = format!("shortcodes/{}.md", name);
--- a/components/rendering/src/shortcode.rs
+++ b/components/rendering/src/shortcode.rs
@@ -140,22 +140,14 @@ fn render_shortcode(
     }
 }
 
-struct InvocationTracker(HashMap<String, u32>);
-
-impl InvocationTracker {
-    fn new() -> Self {
-        InvocationTracker(HashMap::new())
-    }
-    fn get(&mut self, name: &str) -> u32 {
-        let invocation_number = self.0.entry(String::from(name)).or_insert(0);
-        *invocation_number += 1;
-        *invocation_number
-    }
-}
-
 pub fn render_shortcodes(content: &str, context: &RenderContext) -> Result<String> {
     let mut res = String::with_capacity(content.len());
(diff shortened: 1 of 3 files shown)


components/rendering/src/shortcode.rs
```diff
@@ -137,8 +140,22 @@ fn render_shortcode(
...
+    fn get(&mut self, name: &str) -> u32 {
+        let invocation_number = self.0.entry(String::from(name)).or_insert(0);
+        *invocation_number += 1;
+        *invocation_number
+    }
+}
+
 pub fn render_shortcodes(content: &str, context: &RenderContext) -> Result<String> {
     let mut res = String::with_capacity(content.len());
+    let mut invocation_counts = InvocationTracker::new();
```
Keats: why a struct and not simply a hashmap?

Keats (changes requested):

components/rendering/src/shortcode.rs
NathanaelLane: The intent was to simplify code at the call site by wrapping the lazy-initialization and increment logic in the `get()` method. Would it be better to do something like:
```rust
    let mut invocation_map: HashMap<String, u32> = HashMap::new();
    let mut get_invocation_count = |name: &str| {
        let invocation_number = invocation_map.entry(String::from(name)).or_insert(0);
        *invocation_number += 1;
        *invocation_number
    };
```
instead?

Keats: Yep

NathanaelLane:
Apologies for all the commit noise; not sure what happened with the merge

docs/content/documentation/content/shortcodes.md
```diff
@@ -134,6 +134,24 @@ If you want to have some content that looks like a shortcode but not have Zola t
 you will need to escape it by using `{%/*` and `*/%}` instead of `{%` and `%}`. You won't need to escape
 anything else until the closing tag.
 
+### Invocation Count
+
+Every shortcode context is passed in a variable named `invocation` that tracks how many times a particular shortcode has
```
Keats: Ok the code looks good but I'm not sold on the name `invocation`. `nth`? `nth_invocation`? `invocation_count`?

NathanaelLane: I agree that the name leaves something to be desired. I think `nth` or `invocation_count` are probably the best options; `invocation_count` is more descriptive, but in my subjective opinion it feels too verbose relative to the rest of Zola's features/API. How much concern is there around breaking users' existing shortcodes if we were to go with `nth`?

Keats: > How much concern is there around breaking users' existing shortcodes if we were to go with nth?

None I think?

Keats:
Thanks!

PR #913, opened 2020-01-07
Syntax highlighting with CSS classes

uwearzt:
Todo

- [x]  Wait for syntect release (incl. fix for rust-onig)
- [x] Documentation -> @uwearzt 
- [x] Example -> @uwearzt (My Website)
- [ ] Check interface (especially config.toml) @Keats

PR #1242, opened 2020-11-24
Classed syntax highlighting + Line numbers

evan-brass:
CSS classed syntax highlighting

This extends #913 
`ClassedHtmlGenerator` uses `tokens_to_classed_spans` internally which creates a new scope stack for each line causing the panic when a restore occurs on a different line from a clear.

* [x] Documentation
  * [x] Explain how to style classed highlighting (`.code` class, samples for `<mark>`, etc.)
* [x] Implement hl_lines (`<mark>` tags?)
* [x] Implement line numbers
  * Table
  * `linenostart` fence option
* [x] Tests

* Is any more documentation needed?
* Is it ok for `highlighting_css_themes` to not be under the new markdown section?


components/rendering/src/markdown/codeblock.rs
```diff
@@ -1,92 +1,206 @@
...
-                false => None,
-            },
-            background,
-            theme,
-
-            highlight_lines: fence_info.highlight_lines,
-            num_lines: 0,
+fn get_include_background(theme: &Theme) -> IncludeBackground {
+    IncludeBackground::IfDifferent(
+        theme.settings.background.unwrap_or(::syntect::highlighting::Color::WHITE),
```
evan-brass, line 48:
```suggestion
        theme.settings.background.unwrap_or(Color::WHITE),
```

components/rendering/src/markdown/codeblock.rs
```diff
@@ -1,92 +1,206 @@
...
+                // This essentially does the same thing as ClassedHtmlGenerator, except:
+                // 1. It outputs each line one at a time
+                //   * This will be helpful for line numbers
+                // 2. It shares a scope_stack across lines which solves the JSON syntax crash
+                // TODO: Support highlighting lines
+                // TODO: Handle if text doesn't end in a newline?
+                let repo =
+                    SCOPE_REPO.lock().expect("A thread must have poisened the scope repo mutex.");
+                let mut html = String::new();
+                // TODO: Use something other than split_inclusive because it requires a nightly feature:
```
evan-brass, line 144:
```suggestion
```

evan-brass:
Hello @Keats.  Is this something you're interested in?  Before I continue, I wanted to see if this is going in a right direction.

Keats:
Sorry I thought I replied to it already! This is something we want but I'm wondering if we need to override the HTML syntect outputs so we can support line numbers and highlight in the best possible way (https://zola.discourse.group/t/highlighting-improvements/19/11?u=keats)

evan-brass:
I don't think it's too hard to override the HTML at a line granularity which enables hl_lines and line numbers, but overwriting at a character granularity (for character level highlighting) might be hard.  I'll try that next.
So:
* [ ] Character highlighting
  * Would a `language, hl_lines=1:6-3:20` where 6 and 20 are columns be ok?  Or is a shortcode better?
* [ ] Insufficient contrast for highlight-background
  * Could be solved with a highlight background override in config (as was mentioned on #1131), or by allowing custom themes in a similar way to custom syntaxes
  * Using `<mark>` tags for hl_lines in both classed and inline means we can reuse the code between them.
* [ ] Theme CSS output
  * [ ] I don't think that Syntect's theme->css output `<mark>`  styles to match the theme's `line_highlight` setting.
* [ ] Line numbers
  * [ ] Extend fence info to accept an initial line number like: `language, linenos=5`
  * [ ] Table output

I was thinking of working on line numbers after classed highlighting, but they seem to be more connected than I thought.

Currently, classed highlighting for my site looks like this:
Dark mode:
image
Light mode:
image

I set the mark tags to `display: block` in my CSS which highlights empty lines and extends to the end of the line, like you mentioned here: https://github.com/getzola/zola/pull/1131#pullrequestreview-476968411
Display block wouldn't work for character highlighting, so if we add that, then we'd probably need to add a CSS class ('full-line', 'partial-line' perhaps?) to distinguish those.  And for inline styling - if we switch to `<mark>` tags - an inline style to display: block for full line highlights, maybe.  Or maybe that's too far into the weeds.

Keats:
1. Character highlighting: I don't think we want to do that, just lines are fine.
2. Higher contrast: if we rewrite the HTML output, we can use `mark` as you say for highlighted text and let people overwrite the style via CSS. I'm not sure about using `mark` for syntax instead of `span` though. Why do you say it will reduce duplication?
3. Line numbers: I believe to have it nicely working (copy/paste without the line numbers etc), we need to rewrite the HTML from syntect completely.

I think if we go the rewrite html route, adding the classes is trivial since we just put it instead of inline css?

components/rendering/Cargo.toml
```diff
@@ -16,6 +16,7 @@ pest_derive = "2"
 regex = "1"
 lazy_static = "1"
 gh-emoji = "1.0"
+html-escape = "0.2.6"
```
evan-brass, line 19: TODO: Use Tera's escape instead.

Keats:
I think if we go the rewrite HTML route, the first thing to do is to write by hand the HTML we would like to end up so we can experiment easily

uwearzt:
@evan-brass Thanks for pushing my PR forward. This year i have time to work on Zola (and my website). If you need any help, just drop me a note.

evan-brass:
Sorry for the delay.

I'm still not sure what line number style to implement so I implemented the few discussed already.  There's some bugs and no tests yet.

I don't like how much code I wrote in the codeblock module - it feels like there's some simple design that I'm not seeing.  I tried to break the line numbers, highlighted lines, and syntax highlighting into multiple "passes", but it still doesn't seem right.

Keats:
I did it again sorry >_>

Keats:
@evan-brass is it ready or is it still a WIP?

Keats:
Another ping @evan-brass 
I'll pick it up later this month otherwise

kzurawel:
I've been using this branch for a while and so far I'm quite happy with it, but I had one suggestion/request: when using the "css" highlight theme, it would be nice to include the `language-*` classname from the `<code>` block on the `<pre>` tag as well. This would allow for doing things like setting different background colors for different types of code block (in my use case, allowing "plain text" fenced code blocks to blend in more with the page by having no background color, while allowing specific language code blocks to be more visually defined).

Keats:
It looks __really__ good! Merge conflict is going to be a bit annoying (hidden lines, how it interacts with line numbers etc) but I'll try to merge things through the `syntax-highlight-++` branch this week or so

Keats:
Erhm the merge is annoying, do you mind if I do it manually in a new commit and mark you as the commit author @evan-brass ? I will probably tweak some bits but most of it will be this PR.

evan-brass:
Sure.  Sounds good.

Keats:
@kzurawel should the lang be moved entirely to `pre` instead of `code`? It's a bit silly to repeat the information 3-4 times

kzurawel:
I would be fine with that; I can’t think of a situation where having the inner lang class would matter

Keats:
It's if you want to use something like highlightjs. I think I will just the class on `pre` and `code` actually so it covers everything.

Keats:
Not finished yet but https://github.com/getzola/zola/pull/1531 is a simplified implementation as it doesn't try to unify various spans, which sounds like it would something better done in syntect directly.
However the class based highlighting panics when ran on the Zola site despite the code being 99% copy/pasted from syntect itself so I might copy the code from this PR instead if I can't figure it out.

Keats:
https://github.com/trishume/syntect/issues/336 and I made a PR to fix it so it should be ok

Keats:
@kzurawel can you try https://github.com/getzola/zola/pull/1531 ?

PR #1246, opened 2020-11-27
update to tokio 1.0

### LINKED ISSUE #1240: Update zola serve to use latest tokio on next branch
I don't really know Tokio or the best way to upgrade it, if someone has time to do it it shouldn't take too long.

ibraheemdev:
I updated the tokio version to `0.3.4` and updated to the new `runtime::Builder` api. The updated runtime is equivalent to the current single-threaded runtime.

Resolves #1240

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -786,7 +786,7 @@ dependencies = [
  "http",
  "indexmap",
  "slab",
- "tokio",
+ "tokio 0.2.22",
  "tokio-util",
  "tracing",
 ]
@@ -896,7 +896,7 @@ dependencies = [
  "itoa",
  "pin-project",
  "socket2",
- "tokio",
+ "tokio 0.2.22",
  "tower-service",
  "tracing",
  "want",
@@ -913,7 +913,7 @@ dependencies = [
  "hyper",
  "log",
  "rustls",
- "tokio",
+ "tokio 0.2.22",
  "tokio-rustls",
  "webpki",
 ]
@@ -930,7 +930,7 @@ dependencies = [
  "hyper",
  "mime_guess",
  "percent-encoding",
- "tokio",
+ "tokio 0.2.22",
  "url",
  "winapi 0.3.9",
 ]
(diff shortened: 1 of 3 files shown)


ibraheemdev:
Unrelated question: Is there any reason that you do not use tokio's multi-threaded runtime?

Keats:
I don't think this works, it's going to miss the reactor. Running it locally I get:

```
thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of Tokio runtime', /Users/vincentprouillet/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.22/src/io/driver/mod.rs:202:14
```

ibraheemdev:
Tokio 0.2 and 0.3 are not backward compatible. Hyper still depends on tokio 0.2. They completed the 0.3 upgrade on their dev branch but haven't released a new version yet. Because we depend on Hyper, we have to wait till they release a new version, or use the [`tokio-compat-02`](https://tokio.rs/blog/2020-10-tokio-0-3#compatibility-between-02-and-03) compatibility layer.

Keats:
Ah I thought it had been updated. Let's wait until hyper gets released then

ibraheemdev:
I updated tokio to 1.0.1 and hyper to 0.14.1. The issue now is that hyper-staticfile seems to be unmaintained and is still using hyper 0.13 which is causing compatibility issues. 

@Keats I could also write the file server directly with `hyper` and `tokio::fs` is you want.

You might also consider using [warp](https://github.com/seanmonstar/warp) instead of hyper directly, which would simplify things a lot

Keats:
Which one would be the simplest without adding many new dependencies?

ibraheemdev:
The simplest would probably be to just write the file server manually. It would only require enabling the `fs` feature of Tokio.

Keats:
Let's go with that then if it's not too much work

ibraheemdev:
I wrote a simple custom file server and successfully tested it locally. This should conclude the Tokio 1.0 migration.

The `hyper-staticfile` dep has been removed.

Keats:
Thanks it works well! I'll merge it when reqwest is updated to avoid having multiple tokio versions

Keats:
Thanks!

Keats:
(I messed up an didn't notice the PR was on `master` instead of `next` so I've reverted it while I push the commit on the right branch)

PR #1251, opened 2020-12-06
Gitlab Pages CI : Get zola from alpine community repository

fredrik-lindseth:
Zola is not available in Alpine's `testing` repository, and we should be getting it from `community` instead

Sanity check:

* [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?


Keats (approved):

Keats:
Thanks!

Issue #1260, opened 2020-12-13
elasticlunr is outdated


ghost:
# Bug Report

Hi,
While navigating the sources, I noticed that `elasticlunr.js`(https://github.com/getzola/zola/blob/master/components/search/src/elasticlunr.min.js) is outdated: `0.9.5` instead of `0.9.6` https://github.com/weixsong/elasticlunr.js/tags).

It may be nice to upgrade.

I can make a PR if you want.

Keats:
Go for it!

ghost:
closed in #1261

PR #1261, opened 2020-12-13
search: Update elasticlunr.min.js to 0.9.6 (#1260)

ghost:
## Code changes

After lengthy investigation it appeared that  there was a problem in `elasticlunr.min.js` build system and the embedded library was indeed  `elasticlunr.min.js` `v0.9.6` but tagged as `v0.9.5`.

We can see it be reading this `0.9.6` build: https://cdnjs.cloudflare.com/ajax/libs/elasticlunr/0.9.6/elasticlunr.js

And searching for code changes (with `ctrl+f`) made after the `0.9.5` release like this one https://github.com/weixsong/elasticlunr.js/commit/c16fd0792f29f3631f71f6989f90e1e867736ad0#diff-b885d55a0b1142614118e64583ac39314f5ab077670ed1a7bc78636edd3abea7R913

So I just updated the version in the header, to avoid further confusion.

https://github.com/getzola/zola/issues/1260

### DIFF
--- a/components/search/src/elasticlunr.min.js
+++ b/components/search/src/elasticlunr.min.js
@@ -1,6 +1,6 @@
 /**
  * elasticlunr - http://weixsong.github.io
- * Lightweight full-text search engine in Javascript for browser search and offline search. - 0.9.5
+ * Lightweight full-text search engine in Javascript for browser search and offline search. - 0.9.6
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

Issue #1268, opened 2020-12-24
Supported language Clojure is not highlighted properly


vise890:
**NOTE**: I opened a PR that seems to fix the issue #1269 

# Bug Report

## Environment

Kernel: Linux 5.9.15-1-MANJARO
Zola version: zola 0.12.2

## Expected Behavior

I would like to see my Clojure code blocks highlighted properly, like in the picture below:

![image](https://user-images.githubusercontent.com/1628760/103083859-c7605400-45dd-11eb-8cb9-1cfeeff45b87.png)

## Current Behavior

Everything after the first comment character (`;`), is highlighted as a comment:
![image](https://user-images.githubusercontent.com/1628760/103083889-da732400-45dd-11eb-953f-a2f8bdcf1b77.png)

## Step to reproduce

1. in config.toml:
   ```toml
   highlight_code = true
   ```
1. In a page `.md` source, add a clojure code block:
   ![image](https://user-images.githubusercontent.com/1628760/103083911-e828a980-45dd-11eb-8792-d9997a8fc00e.png)
3. Run `zola server` and navigate to that page, you will see (roughly):
   ![image](https://user-images.githubusercontent.com/1628760/103083897-de9f4180-45dd-11eb-8fd5-8bc18a795ada.png)

## Workaround

1. in the project root
   ```bash
   mkidr syntaxes
   cd syntaxes
   git submodule add https://github.com/tonsky/sublime-clojure
   ```
1. in `config.toml`
   ```toml
   extra_syntaxes = ["syntaxes/sublime-clojure"]
   ```
1. **Edit `sublime-clojure/ClojureC.sublime-syntax`** so that 
`file-extensions` look like this:
   ```sublime-syntaxt
   file_extensions: [bloop, clj, cljs, cljc, cljx, boot]
   ```
1. in the page source:
   ![image](https://user-images.githubusercontent.com/1628760/103084238-d1368700-45de-11eb-895b-ac4da3ff4510.png)

## Notes
- I couldn't find a clojure `.sublime-syntax in [`/zola/sublime/syntaxes/`](https://github.com/getzola/zola/tree/master/sublime/syntaxes)
- **It doesn't seem to be possible to override built in short-names** [listed in the docs](https://www.getzola.org/documentation/content/syntax-highlighting/), maybe it should?

Keats:
[...]

PR #1269, opened 2020-12-24
fix clojure syntax highlighting

vise890:
Fix clojure syntax highlighting as per #1268,

 also [sublime-clojure provides better highlighting in some corner-cases](https://github.com/tonsky/sublime-clojure)

I tried to follow the procedure in the contrib guidelines, but feel free to close or ask for edits if something is not right! :smile:


Keats:
I'm not sure it will work well, syntect will load the whole folder but I don't know if the order will be deterministic

vise890:
I see, it's probably worth closing this until we find a proper way to deal with it then? Is it worth opening an issue with syntect then since it is likely to be affected too?

Keats:
Not worth a syntect issue, we just need to have a better way to override the syntaxes in zola. I'll check next week, it shouldn't be too hard.

vise890:
Ok great, thank you for looking into it! :smile:

Issue #1112, opened 2020-07-31
Architecture for less iterations over the library


southerntofu:
This is not exactly a bug report, but when i was investigating site::load() i realized the pages/sections collections get iterated over **many** times. For a simple build, it iterates over:

- sections (loop l. 174)
- pages (loop l. 182)
- sections & pages (library.check_for_path_collisions() l.197)
- pages (library.populate_taxonomies() l. 205)
- sections and pages (self.populate_sections() l. 207, 5 times for sections, 2 for pages because of other calls inside this function)
- sections and pages (self.render_markdown l. 208, 2 times for pages)
- sections and pages (check_internal_links_with_anchors() l. 212)
- sections and pages (check_external_links() l.215, only in check mode)

Without check mode, that's a total of 9 iterations over sections and 8 iterations over pages. This is only for iterations after the library is built. There may be some compiler magic behind the scenes to "reunite" some iterations in there, but my intuition is there's a lot of potential performance improvements to be investigated here.

I think some of the useless iterations are due to the absence of library structure due to the glob pattern (reuniting children pages with related sections, and building content hierarchy). Maybe a recursive descent into the content folder could help here?

Keats:
We can probably trim that number down but I don't think it will make a significant change in the speed.
The main speed bottleneck is serialising the content to JSON for Tera and that is a much bigger overhead than iterating. The other speed bottleneck is syntax highlighting because it's essentially doing many many regexes and is roughly doubling the build time when enabled.

Of course if we can improve that without making the code harder to understand, I would be very interested

southerntofu:
> The main speed bottleneck is serialising the content to JSON for Tera

What's the best way to have some good data on this? I tried running the benchmarks but they don't really help me understand the bottlenecks. So i ran [flamegraph](https://github.com/flamegraph-rs/flamegraph) on building medium-blog and this was the result. I've never done profiling before so i'm not entirely sure what's happening here. But rayon appears to take quite some space on there.

![zola_flame](https://user-images.githubusercontent.com/52931252/89068361-964a6300-d360-11ea-9d11-a36a5fc26735.jpg)

> Of course if we can improve that without making the code harder to understand, I would be very interested

It's entirely possible, but it will take some time. Let's not hurry too much ;). In the meantime, are benchmarks run periodically? It would be nice to have benchmarks on every commit/PR. Is it easy to achieve with the current pipelines?

Keats:
[...]

PR #1272, opened 2020-12-25
move translation population and path collision checking to insert stage

aslowwriter:
PR in response to #1112. 

## Code changes
I moved the path collision checking and the translation population to the insertion into the library stage. This is to save a few iterations across pages and sections. It did necessitate extra indices for the library, but for aliases and translations, I think that is acceptable since it should be relatively slim in most cases. I have some comments about other possible improvements but I'll add those to the issue discussion itself

### DIFF
--- a/components/library/src/library.rs
+++ b/components/library/src/library.rs
@@ -41,6 +41,12 @@ pub struct Library {
     pub paths_to_sections: HashMap<PathBuf, DefaultKey>,
     /// Whether we need to look for translations
     is_multilingual: bool,
+
+    // aliases -> files,
+    // so we can easily check for conflicts
+    pub reverse_aliases: HashMap<String, HashSet<String>>,
+
+    pub translations: HashMap<PathBuf, HashSet<DefaultKey>>,
 }
 
 impl Library {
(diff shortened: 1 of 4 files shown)


components/library/src/library.rs
```diff
@@ -41,6 +41,12 @@ pub struct Library {
     pub paths_to_sections: HashMap<PathBuf, DefaultKey>,
     /// Whether we need to look for translations
     is_multilingual: bool,
+
+    // alieas -> files,
```
Keats: aliases

components/library/src/library.rs
```diff
@@ -51,22 +57,78 @@ impl Library {
...
+        for alias in &section.meta.aliases {
+            self.reverse_aliases
+                .entry(alias.to_string())
+                .and_modify(|s| {
+                    s.insert(file_rel_path.clone());
+                })
+                .or_insert_with(|| {
+                    let mut s = HashSet::new();
+                    s.insert(file_rel_path.clone());
+                    s
```
Keats, line 83: seems like copy/paste where it's not actually inserting the alias? It would be better to insert the aliases in the same pass as the actual file path for the section (ie l 71-80)

components/library/src/library.rs
```diff
@@ -51,22 +57,78 @@ impl Library {
...
+                });
+        }
```
Keats, line 112: same as for sections

Keats (changes requested):

components/library/src/library.rs
aslowwriter: Not quite sure how that would work, as l 71-80 isn't actually a loop, it just creates the "identity alias". We do need the loop though because the aliases are the keys with the same values. So I'm not sure how you would do that.

Keats: Yeah, I was wrong. I think something like (pseudocode)
```
let entries = vec![rel_path] + section.meta.aliases;
for entry in entries {
        self.reverse_aliases
            .entry(entry)
            .and_modify(|s| {
                s.insert(file_rel_path.clone());
            })
            .or_insert_with(|| {
                let mut s = HashSet::new();
                s.insert(file_rel_path.clone());
                s
            });
}
```
would be neater though

aslowwriter: Done

Keats:
Thanks!

Keats:
FYI I just did a commit to DRY it a bit: https://github.com/getzola/zola/pull/1189/commits/66220780383bb0f1bc07d8db3466dd937d146a6f

PR #1290, opened 2021-01-07
Include "markdown" in list of config.toml sections

flother:
This is a teeny-tiny documentation change to include `[markdown]` in the list of sections available in `config.toml`. The section was added recently and the example configuration code block includes it, but the list that precedes it wasn't updated.


Keats:
Thanks!

PR #1293, opened 2021-01-09
Next version

Keats:


meithecatte:
The code blocks on the documentation preview look wrong — I don't think they're supposed to be red: https://deploy-preview-1293--getzola.netlify.app/documentation/content/syntax-highlighting/

Keats:
Thanks for catching that, I guess the theme doesn't work with markdown.

PR #1305, opened 2021-01-13
Ensure minified HTML is truncated before converting to String

### LINKED ISSUE #1304: HTML minification fails with CJK text
# Bug Report

Zola sometimes fails to build or serve content with `minify_html = true` when source files contain CJK characters, depending on the number of characters around them.

## Environment

Zola version: 0.13.0

## Expected Behavior

Zola should compile and minify all files without issue.

## Current Behavior

Site fails to build sometimes, and in other situations fail to be served, if there are CJK characters. It's interesting that `zola serve` and `zola build` behave differently (I haven't run into a situation where both failed to work on the same content), but I guess this probably has to do with the injected reload script/different `base_url`.

```
Building site...
Failed to build the site
Error: Failed to convert bytes to string : invalid utf-8 sequence of 1 bytes from index 18
```
(the third line doesn't show up when running `zola serve`)

## Steps to reproduce

As it's not a consistent issue, I've prepared a small example site: [minify-cjk-bug.zip](https://github.com/getzola/zola/files/5805531/minify-cjk-bug.zip)

Zola successfully builds this example site if I ensure three bytes (ASCII characters, newlines), or multiples of three bytes (including 0 bytes), are before the CJK character.

I haven't been able to create a minimal site to reproduce the failure during `zola serve`, but [`my website`](https://github.com/lae/web/tree/97569df5d9850656f8428709d9a2fad034733346) currently fails to build on serve.

lae:
Closes #1304

------------------
Sanity check:

### DIFF
--- a/components/utils/src/minify.rs
+++ b/components/utils/src/minify.rs
@@ -6,8 +6,8 @@ pub fn html(html: String) -> Result<String> {
     let mut input_bytes = html.as_bytes().to_vec();
 
     match with_friendly_error(&mut input_bytes, cfg) {
-        Ok(len) => match std::str::from_utf8(&input_bytes) {
-            Ok(result) => Ok(result[..len].to_string()),
+        Ok(len) => match std::str::from_utf8(&input_bytes[..len]) {
+            Ok(result) => Ok(result.to_string()),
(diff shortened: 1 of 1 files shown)


Keats:
Woops, thanks!

PR #1306, opened 2021-01-14
load_data() template function takes a `required` boolean flag (closes #1209)

### LINKED ISSUE #1209: Deal with failure from load_data
There is currently no way to deal with failure from `load_data`, whether due to a 404 for remote data, or nonexistent files. Right now, the build just fails.

I can work on this issue. The best way I can think of is to add an optional `required` (default true to prevent breaking changes) parameter to `load_data`. If required is false and the function fails, it simply returns null (does tera have nulls?), and the user can deal with it, knowing that it failed.

We could also add an optional `default` fallback value: `load_data(path="data.json", default={...}, required=false)`

Would a PR implementing the above behaviour be accepted?

southerntofu:
load_data() now takes an optional `required` boolean flag (defaults to true). Docs have been udpated and tests also.

I'm not too happy with the implementation because it's not too expressive. It uses a Result<Option<DataSource>> where Err are hard error and None is only used for missing file. I'd be happier with typed Errors from which we could discard specific variants depending on context but that would be a lot more work. Maybe error handling could use some love in the coming versions, but this is not the concern here?

Also i don't know of a way to test in a cross-platform manner that a permissions problem still produces error when `required` is false. I would read from `/etc/passwd` but it's outside of basedir, and also would not work on Windows. So i didn't write a test case for that.

### DIFF
--- a/components/templates/src/global_fns/load_data.rs
+++ b/components/templates/src/global_fns/load_data.rs
@@ -415,6 +415,17 @@ mod tests {
     }
 
     #[test]
+    fn doesnt_fail_when_missing_file_is_not_required() {
+        let static_fn = LoadData::new(PathBuf::from("../utils"));
+        let mut args = HashMap::new();
+        args.insert("path".to_string(), to_value("../../../READMEE.md").unwrap());
+        args.insert("required".to_string(), to_value(false).unwrap());
+        let result = static_fn.call(&args);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), tera::Value::Null);
+    }
+
+    #[test]
     fn cant_load_outside_content_dir() {
         let static_fn = LoadData::new(PathBuf::from(PathBuf::from("../utils")));
         let mut args = HashMap::new();
@@ -511,6 +522,26 @@ mod tests {
             format!("Failed to request {}: 404 Not Found", url)
         );
     }
+
+    #[test]
+    fn doesnt_fail_when_request_404s_is_not_required() {
+        let _m = mock("GET", "/aazeow0kog")
+            .with_status(404)
+            .with_header("content-type", "text/plain")
+            .with_body("Not Found")
+            .create();
+
+        let url = format!("{}{}", mockito::server_url(), "/aazeow0kog");
+        let static_fn = LoadData::new(PathBuf::new());
+        let mut args = HashMap::new();
+        args.insert("url".to_string(), to_value(&url).unwrap());
+        args.insert("format".to_string(), to_value("json").unwrap());
+        args.insert("required".to_string(), to_value(false).unwrap());
+        let result = static_fn.call(&args);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), tera::Value::Null);
+    }
+
 
     #[test]
     fn set_default_user_agent() {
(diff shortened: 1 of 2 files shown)


Keats:
> Maybe error handling could use some love in the coming versions, but this is not the concern here?

It probably needs to at some point.

components/templates/src/global_fns/load_data.rs
```diff
@@ -72,26 +72,32 @@ impl OutputFormat {
...
+                return Ok(None);
             }
-            return Ok(DataSource::Path(full_path));
+            return Ok(Some(DataSource::Path(full_path)));
         }
 
         if let Some(url) = url_arg {
             return Url::parse(&url)
                 .map(DataSource::Url)
+                .map(|x| { Some(x) })
```
Keats: I don't think the braces are needed?

docs/content/documentation/templates/overview.md
```diff
@@ -212,10 +212,20 @@ As a security precaution, if this file is outside the main site directory, your
 {% set data = load_data(path="content/blog/story/data.toml") %}
 ```
 
+The optional `required` boolean argument can be set to false so that missing data does not produce an error, but returns a null value instead. For URLs, all HTTP errors are discarded. For local paths, missing file errors are discarded, but permissions errors are kept. In both cases, invalid data that could not be parsed to the requested data format will still produce an error.
+
+The snippet below outputs the HTML from a Wikipedia page, or "No data found" if the page was not reachable, or did not return a successful HTTP code:
+
+```jinja2
+{% set data = load_data(path="https://en.wikipedia.org/wiki/Commune_of_Paris", required=false %}
```
Keats: missing closing )

docs/content/documentation/templates/overview.md
```diff
@@ -212,10 +212,20 @@ As a security precaution, if this file is outside the main site directory, your
...
+{% if data %}{{ data | safe }}{% else %}No data found{% endif %}
```
Keats, line 221: I believe you need to use `{%/*` and `*/%}` in that snippet to avoid zola calling it.

docs/content/documentation/templates/overview.md
```diff
@@ -212,10 +212,20 @@ As a security precaution, if this file is outside the main site directory, your
 {% set data = load_data(path="content/blog/story/data.toml") %}
 ```
 
+The optional `required` boolean argument can be set to false so that missing data does not produce an error, but returns a null value instead. For URLs, all HTTP errors are discarded. For local paths, missing file errors are discarded, but permissions errors are kept. In both cases, invalid data that could not be parsed to the requested data format will still produce an error.
```
Keats: It's not clear to me what you mean by discarded. 

How about:

The optional `required` boolean argument can be set to `false` so that missing data (HTTP error or local file not found) does not produce an error, but returns a null value instead. However, permission issues with a local file and  invalid data that could not be parsed to the requested data format will still produce an error even with `request=false`.

?

Keats (changes requested):
It looks ok to me, we can iterate on it later.

docs/content/documentation/templates/overview.md
southerntofu: This does not seem to be the case (anymore?). i rendered the docs and it appeared as expected.

For example markdown filter is already described like this:

````
```jinja2
{{ some_text | markdown(inline=true) }}
```
````

Keats: Hu weird

PR #1307, opened 2021-01-14
Don't panic for missing highlight theme (close #1299)

### LINKED ISSUE #1299: crash when using a theme that doesn't exist
# Bug Report

Zola panics when I make a mistake for the theme in config.toml. Here is the error message:

```
thread 'main' panicked at 'no entry found for key', components/rendering/src/markdown.rs:230:38
```

## Environment

Running `zola serve` locally.

Zola version: 0.13.0

## Expected Behavior

Don't crash, show error that theme doesn't exist.

## Current Behavior

Crash.

## Step to reproduce

Enter a non-existing theme in config.toml and run `zola serve`

southerntofu:
So on load config was only checking if config.highlight_theme did exist, but it's been moved to config.markdown.highlight_theme so i added that possibility to the check. I did manual testing (it works) but didn't rerun the tests because it's such a trivial change.

Also missing theme error is different from missing highlight theme error. The former is red and does not specify the missing file. The latter is not colored but is more informative:

```
Failed to load /tmp/site/config.toml
Error: Highlight theme foo not available
```

```
Error: No `theme.toml` file found. Is the `theme` defined in your `config.toml` present in the `themes` directory and does it have a `theme.toml` inside?
Reason: No such file or directory (os error 2)
```

I can add a commit to this PR for this, but i would like feedback on how errors should look like (color wise) first.

### DIFF
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -48,7 +48,9 @@ pub fn create_directory(path: &Path) -> Result<()> {
 /// Return the content of a file, with error handling added
 pub fn read_file(path: &Path) -> Result<String> {
     let mut content = String::new();
-    File::open(path)?.read_to_string(&mut content)?;
+    File::open(path)
+        .map_err(|e| Error::chain(format!("Failed to open '{}'", path.display()), e))?
+        .read_to_string(&mut content)?;
 
     // Remove utf-8 BOM if any.
     if content.starts_with("\u{feff}") {
--- a/components/utils/src/fs.rs
+++ b/components/utils/src/fs.rs
@@ -48,9 +48,7 @@ pub fn create_directory(path: &Path) -> Result<()> {
 /// Return the content of a file, with error handling added
 pub fn read_file(path: &Path) -> Result<String> {
     let mut content = String::new();
-    File::open(path)
-        .map_err(|e| Error::chain(format!("Failed to open '{:?}'", path.display()), e))?
-        .read_to_string(&mut content)?;
+    File::open(path)?.read_to_string(&mut content)?;
 
     // Remove utf-8 BOM if any.
     if content.starts_with("\u{feff}") {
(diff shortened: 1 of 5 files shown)


Keats:
I knew I forgot more of those, hopefully we can get rid of those duplicated fields in the next version entirely if it looks like a 0.13.1 is unlikely.
If possible the error should be red and I thought I fixed all the IO errors where the name wasn't indicated 🤔

southerntofu:
```
Building site...
Error: Failed to load config from file /tmp/site/config.toml
Reason: Highlight theme foo not available
```

```
Building site...
Error: Failed to load theme foo from file /tmp/site/themes/foo/theme.toml
Reason: No such file or directory (os error 2)
```

Is that better?

components/utils/src/fs.rs
```diff
@@ -48,9 +48,7 @@ pub fn create_directory(path: &Path) -> Result<()> {
 /// Return the content of a file, with error handling added
 pub fn read_file(path: &Path) -> Result<String> {
     let mut content = String::new();
-    File::open(path)
-        .map_err(|e| Error::chain(format!("Failed to open '{:?}'", path.display()), e))?
```
Keats, line 52: We don't want to remove that, otherwise we don't know what failed in the various calls across the project. It's ok if the file is mentioned twice but at least we're sure we get the path.

southerntofu: Very good point. Maybe read_file() can also take a name argument so it can produce all the errors itself?

- Failed to load config from /foo/config.toml
- Failed to load page from /foo/content/bar.md
- Failed to load theme foobar from /foo/themes/foobar/theme.toml
- etc..

Keats: I'm not sure it's worth passing another argument

Keats: Can you revert that? It's ok to merge otherwise I think

southerntofu: OK i just removed the extra quotes printed due to debug output.

southerntofu:
OK i think errors are ok now, and printed in red. Highlight theme error:

```
Error: Highlight theme foobar defined in config does not exist.
```

Theme error:

```
Error: Failed to load theme bar
Reason: Failed to open '/tmp/foo/themes/bar/theme.toml'
Reason: No such file or directory (os error 2)
```

Keats:
Thanks!

PR #1312, opened 2021-01-15
Fix fallback to syntax highlighting theme (closes #1309)

### LINKED ISSUE #1309: Extra syntax conflict
# Bug Report
When defining `extra_syntaxes` in `config.toml`, standard highlighters no longer work.

## Environment
Zola version: 0.13, also tested master and next branches (2021-01-14).

This was working with the next branch some days ago (not sure about the exact date though..)

## Step to reproduce
Test article. `py3` is a standard language. `shell-session` is a custom language, defined in `syntaxes/shell-session.sublime-syntax`

    +++
    title = "highlight"
    slug = "highlight"
    
    date = 2099-01-01
    
    draft = true
    
    [taxonomies]
    tags = ["test"]
    +++
    
    ```py3
    python = "python"
    ```
    
    ```shell-session
    $ echo shell
    ```

1st test case: here I set `extra_syntaxes` in cargo.toml:

    [markdown]
    extra_syntaxes = ["syntaxes"]
    highlight_code = true

Result:
image

Note that the python code is node highlighted, that's the issue

2nd test case: `extra_syntaxes` is commented:

    [markdown]
    #extra_syntaxes = ["syntaxes"]
    highlight_code = true

image

The python code is now properly highlighted.

If you want to reproduce, here is the `shell-session.sublime-syntax` (disclaimer: it's just a quick and dirty hack...)

    %YAML 1.2
    ---
    file_extensions:
      - shell-session
    scope: source.test
    
    contexts:
      main:
        - match: "^((?:\\(.*?\\) )?[^ ]*(?:\\$|#|>))( .*)?"
        #           ^^^^^^^^^^^^^^ python venv
        #                          ^^^^ user@hostname
        #                               ^^^^^^^^^^^^ last prompt character
        #                                           ^^^^^ command
          captures:
            1: constant.prompt
            2: keyword.command

southerntofu:
As suggested in the discussion there, i also added a warning for an unrecognized language so issues are easier to diagnose. Maybe some people won't like that? It would be useful to have configuration for something like this to be either an error, a warning, or ignored entirely, don't you think?

### DIFF
--- a/components/rendering/src/markdown/codeblock.rs
+++ b/components/rendering/src/markdown/codeblock.rs
@@ -1,4 +1,4 @@
-use config::highlighting::{get_highlighter, SYNTAX_SET, THEME_SET};
+use config::highlighting::{get_highlighter, HighlightSource, SYNTAX_SET, THEME_SET};
 use config::Config;
 use std::cmp::min;
 use std::collections::HashSet;
@@ -22,16 +22,27 @@ pub struct CodeBlock<'config> {
 }
 
 impl<'config> CodeBlock<'config> {
-    pub fn new(fence_info: &str, config: &'config Config, background: IncludeBackground) -> Self {
+    pub fn new(fence_info: &str, config: &'config Config, background: IncludeBackground, path: Option<&'config str>) -> Self {
         let fence_info = FenceSettings::new(fence_info);
         let theme = &THEME_SET.themes[config.highlight_theme()];
-        let (highlighter, in_extra) = get_highlighter(fence_info.language, config);
+        let (highlighter, highlight_source) = get_highlighter(fence_info.language, config);
+        let extra_syntax_set = match highlight_source {
+            HighlightSource::Extra => config.markdown.extra_syntax_set.as_ref(),
+            HighlightSource::NotFound => {
+                // Language was not found, so it exists (safe unwrap)
+                let lang = fence_info.language.unwrap();
+                if let Some(path) = path {
+                    eprintln!("Warning: Highlight language {} not found in {}", lang, path);
+                } else {
+                    eprintln!("Warning: Highlight language {} not found", lang);
+                }
+                None
+            },
+            _ => None,
+        };
         Self {
             highlighter,
-            extra_syntax_set: match in_extra {
-                true => config.markdown.extra_syntax_set.as_ref(),
-                false => None,
-            },
+            extra_syntax_set,
             background,
             theme,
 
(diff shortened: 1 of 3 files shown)


tatref:
I think it would be better to also have the name of the page triggering the issue (or the path).

I think it will have to be done outside of this function, but it will probably require to change the signature of the function.

southerntofu:
@tatref What do you think like this?

```
Building site...
Warning: Highlight language foo not found in bar.md
```

I don't like it too much because the code is more convoluted and we're passing path information further down. I'd rather have rendering provide more feedback on potential errors, which the caller may turn into either warnings, errors, or discard. But that would require more reflection on how to architect things so it may not be appropriate for this pull request.

Also i added a second warning message in case the page path was not found (currently extracting from tera context's page or section -> relative_link) which i assume is going to be the case when running codeblocks through the markdown filter (but i'm not sure) and maybe other cases ? maybe it's just unneeded

@Keats any feedback?

components/rendering/src/markdown/codeblock.rs
```diff
@@ -22,16 +22,27 @@ pub struct CodeBlock<'config> {
...
+        let extra_syntax_set = match highlight_source {
+            HighlightSource::Extra => config.markdown.extra_syntax_set.as_ref(),
+            HighlightSource::NotFound => {
+                // Language was not found, so it exists (safe unwrap)
+                let lang = fence_info.language.unwrap();
+                if let Some(path) = path {
+                    eprintln!("Warning: Highlight language {} not found in {}", lang, path);
+                } else {
+                    eprintln!("Warning: Highlight language {} not found", lang);
+                }
```
Keats, line 38: Should we error instead? There's no point putting a lang that we don't support so it's either a missing highlighting or a typo, both of those could be considered as errors.

southerntofu: There's usecase for language that is not highlighted by Zola but is still styled by CSS using the data-lang attribute of the code tag. Can be used to differentiate lyrics, poetry, quotations, etc.. So error is not a universal answer though it is the good answer for a programming blog. Adding error here would be breaking change, especially for people who use markdown shared with other programs than zola who may produce/want such information.

Maybe we can open a discussion about errors/warnings on the forums to have feedback for the next releases, not just about this specific case. In the meantime, what to do? It's not hard to add a config setting like `markdown.missing_codeblock_lang` which could be "error", "warn" or "off". But do you think it's a good idea? Otherwise, i'd be tempted to just remove the warning until we have more feedback.

Keats: Good point, let's keep it as warning for now. We can add config to disable those warnings later if needed.

Keats (approved):

PR #1315, opened 2021-01-18
Add sort_by title

xpe:
Add sorting by title.

Discussed at https://zola.discourse.group/t/allowing-sorting-by-title/369/3

Sanity check:

Code changes:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,6 +52,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "any_ascii"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70033777eb8b5124a81a1889416543dddef2de240019b674c81285a2635a7e1e"
+
+[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1161,6 +1167,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lexical-sort"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c09e4591611e231daf4d4c685a66cb0410cc1e502027a20ae55f2bb9e997207a"
+dependencies = [
+ "any_ascii",
+]
+
+[[package]]
 name = "libc"
 version = "0.2.82"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1176,6 +1191,7 @@ dependencies = [
  "front_matter",
  "globset",
  "lazy_static",
+ "lexical-sort",
  "rayon",
  "regex",
  "rendering",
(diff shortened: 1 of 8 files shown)


southerntofu:
That looks good @xpe, thanks for the patch. Do you think the sorting works well across casing and for non-english locales? I'm under the impression you're currently sorting the strings as byte vectors, so following unicode points.

Maybe [lexical-sort](https://docs.rs/lexical-sort/0.3.1/lexical_sort/) would help for case-insensitive lexicographic sorting? I haven't tested it but according to docs but it looks like what we're looking for. Let me know if you need help to come up with test cases for that, i'm a french speaker and we've got quite a few characters unicode-insensitive systems mess up with ;)

xpe:
@southerntofu Excellent point. Actually, I was just finishing up adding [lexical-sort] to the PR when I saw your comment. :)

[lexical-sort]: https://docs.rs/lexical-sort/0.3.1/lexical_sort/

xpe:
@southerntofu I included `"métro"` as part of a test, along with other words relating to subway / mass-transit systems.

southerntofu (approved):

southerntofu:
Looks good! :)

Keats (commented):
It looks good but I'm not a fan of title_prev/title_next but I can't come up with something better either.

southerntofu:
> It looks good but I'm not a fan of title_prev/title_next but I can't come up with something better either.

Same here. Though, why are we using different variables for that in the first place? Couldn't we populate some `prev|next` variables with whatever sorting mechanism is enabled, and maybe add a `sorted_by` variable to the templates. This way no need to change the templates for different ordering methods (you can just change config) and the templates can figure out what to display to end-users, even fetching translations for the sorted_by and accounting for right-to-left languages.

I skimmed through #338 and i think providing ordering type context along with unified "next/prev" variables to the templates should be enough to address concerns exposed there while being simpler for theme authors to deal with. Am i missing something?

xpe:
Yes, I was also wondering if it was necessary or useful to have multiple forward/backward links.

Keats:
> Though, why are we using different variables for that in the first place? 

I thought it was nicer to use in the template than `previous` and `next` since you don't really change from weight to date or vice versa in practice. I'm happy to remove them and just have previous/next though if it makes more sense to other people.

xpe:
So this PR is good to go? (for the next release?)

Keats:
Yep, just thinking a bit more for the names. I think `previous`/`next` would work for a-z order, or `before`/`after`.
Or we could just have universal previous/next. What do you think @southerntofu ?

southerntofu:
> we could just have universal previous/next

That's a lot preferable, because it requires less work to make a theme.

> So this PR is good to go?

For me, yes. But if @Keats is okay and you would like to take care of it, you can add a commit here to unify the variables. Otherwise we can merge this PR and we'll do that later :)

Keats:
> Otherwise we can merge this PR and we'll do that later :)

Let's do that.

xpe:
Thanks for the eyeballs and quick feedback (and Zola!)

Issue #1316, opened 2021-01-19
Zola unable to markdown relative links that are not in page.content


crzysdrs:
# Bug Report

## Environment

Zola version: 0.13.0

## Expected Behavior
Zola should be able to `markdown` internal links that come from other fields besides the `page.content`.

## Current Behavior

If an internal link is in field that is converted via `markdown` (that is not `page.content`), zola is unable to find the corresponding internal link.
```
$ zola serve
Building site...
-> Creating 2 pages (0 orphan), 0 sections, and processing 0 images
Error: Failed to render page '/home/crzysdrs/proj/zola_test/content/1.md'
Reason: Failed to render 'page.html'
Reason: Filter call 'markdown' failed
Reason: Failed to render markdown filter: Error { kind: Msg("Relative link @/2.md not found."), source: None }
```

## Step to reproduce

```
======= ./config.toml =======
title= "Test"
base_url = "/"

======= ./templates/page.html =======
{{ page.content | safe }}

{{ page.extra.md | markdown | safe }} {# FAILS #}

======= ./content/2.md =======
---
title: 2
extra:
    md: ""
---

======= ./content/1.md =======
---
title: "1"
extra:
    md : "[Page 2](@/2.md)"
---

[Page 2](@/2.md)
```

`zola build`

southerntofu:
Hi @crzysdrs am i correct that this bug is a duplicate of #1296? If so i just proposed a patch at #1318 feel free to take a look there see if it's good for you :)

crzysdrs:
I agree, this is a duplicate. Surprised I didn't find your issue when I searched.

Issue #1296, opened 2021-01-10
[Feature Request] Support Internal Links in Shortcodes


lukehsiao:
# Bug Report

## Environment

Zola version: v0.13.0

## Expected Behavior
Previously, I was using a shortcode that did something like {{ body | markdown | safe }}, and it would correctly parse Internal links like `[Link](@/other_page.md)`. 

## Current Behavior
With v0.13.0, this behavior is broken, and the links that used to work now show
```
Reason: Filter call 'markdown' failed
Reason: Failed to render markdown filter: Error { kind: Msg("Relative link @/other_page.md not found."), source: None }
```
even though `contents/other_page.md` does in fact exist.

## Step to reproduce
Please provide the steps to reproduce the issue.
If the issue is hard to reproduce, please provide a sample repository or sample
that triggers the bug.

Keats:
I'm a bit surprised it worked... Before 0.13 it was just running pulldown_cmark directly so it shouldn't have known about internal links at all. I'll need to bisect to figure out what happened.

lukehsiao:
@Keats Upon further inspection, I'm mistaken. On zola v0.12.2, the internal link is incorrect as well, it just didn't throw an error like it does in v0.13.0. With v0.12.2, the internal link is turned into literally `http://127.0.0.1:1111/example_post/@/other_page.md`, which, as expected, will 404.

So, rather than a bug report, this should be a feature request to support the internal link syntax in this context. I assume this is something to do with the `permalinks` passed to `resolve_internal_links()` when used in this context? 

I've updated the title to frame this as a feature request, but will need your help updating the label.

As a side note, have absolutely loved the new features in v0.13.0! Thank you to you and all the zola contributors!

southerntofu:
Hi @lukehsiao, have you considered using a Markdown shortcode as explained [in the docs](https://www.getzola.org/documentation/content/shortcodes/) instead of a HTML shortcode? I just tried and it works for a simple case. If that doesn't work for you, could you post your templates so we can try to see what's wrong, and maybe find a workaround?

In any case, if i understand well we would like the markdown filter in the templates to support internal links, which requires inheriting the rendering context (for the `permalinks`) instead of rebuilding it from config like MarkdownFilter::filter() does? However the markdown filter is registered only once for the site so i have no idea how to do this. Any ideas @Keats ?

lukehsiao:
Hmmm, I could explore that. I'm using an HTML shortcode because I'm using it as a caption. The template is this:

```
[...]

PR #1318, opened 2021-01-19
Internal links are supported in markdown filter

southerntofu:
I went the easy way not bothering with lifetimes and simply cloning (as is done with config, though permalinks is potentially a lot bigger) is that ok?

docs are updated and the change is documented in CHANGELOG.md, and there's a corresponding test

#1296 #1316

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## unreleased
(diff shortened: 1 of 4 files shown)


Keats:
Thanks! We can make all the Tera functions more efficient later on if needed.

PR #1321, opened 2021-01-21
Update to 2021

ejiektpobehuk:
Just a year bump in the docs template =]

### DIFF
--- a/docs/templates/index.html
+++ b/docs/templates/index.html
@@ -99,7 +99,7 @@
             {% endblock content %}
         </div>
         <footer>
-            ©2017-2019 — <a class="white" href="http://vincentprouillet.com">Vincent Prouillet</a> and <a class="white" href="https://github.com/getzola/zola/graphs/contributors">contributors</a>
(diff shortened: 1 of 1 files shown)


docs/templates/index.html
```diff
@@ -99,7 +99,7 @@ <h2>Augmented Markdown</h2>
             {% endblock content %}
         </div>
         <footer>
-            ©2017-2020 — <a class="white" href="http://vincentprouillet.com">Vincent Prouillet</a> and <a class="white" href="https://github.com/getzola/zola/graphs/contributors">contributors</a>
+            ©2017-2019 — <a class="white" href="http://vincentprouillet.com">Vincent Prouillet</a> and <a class="white" href="https://github.com/getzola/zola/graphs/contributors">contributors</a>
```
Keats: sounds like we went back in time?

ejiektpobehuk: Wow, I do sometimes forget how Ctrl+A in Vim works when there is `-` in play =\
Thanks for a quick catch!

Keats:
Thanks!

PR #1323, opened 2021-01-24
Feature: allow spaces between dashes in filename for page

digitalresistor:
Apologies for creating a PR without asking/discussing it first. I needed this for a site I am migrating, as I didn't want to rename all the files and lose easy access to history in git. I hope this small change will be accepted.

This change allows files named:

```
2021-01-23 - hello new world.md
```

to be treated the same as files named

```
2021-01-23-hello-new-world.md
```

And have the date stripped from the slug that is generated.

## Code changes

### DIFF
--- a/components/library/src/content/page.rs
+++ b/components/library/src/content/page.rs
@@ -25,7 +25,7 @@ lazy_static! {
     // Based on https://regex101.com/r/H2n38Z/1/tests
     // A regex parsing RFC3339 date followed by {_,-}, some characters and ended by .md
     static ref RFC3339_DATE: Regex = Regex::new(
-        r"^(?P<datetime>(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)([01][0-9]|2[0-3]):([0-5][0-9])))?)(_|-)(?P<slug>.+$)"
+        r"^(?P<datetime>(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)([01][0-9]|2[0-3]):([0-5][0-9])))?)\s?(_|-)(?P<slug>.+$)"
     ).unwrap();
 }
 
@@ -701,6 +701,23 @@ Hello world
 
         assert_eq!(page.meta.date, Some("2018-10-08".to_string()));
         assert_eq!(page.slug, " こんにちは");
+    }
+
+    #[test]
+    fn can_get_date_from_filename_with_spaces() {
+        let config = Config::default();
+        let content = r#"
++++
++++
+Hello world
+<!-- more -->"#
+            .to_string();
+        let res = Page::parse(Path::new("2018-10-08 - hello.md"), &content, &config, &PathBuf::new());
+        assert!(res.is_ok());
+        let page = res.unwrap();
+
+        assert_eq!(page.meta.date, Some("2018-10-08".to_string()));
+        assert_eq!(page.slug, "hello");
     }
 
     #[test]
--- a/components/library/src/content/page.rs
+++ b/components/library/src/content/page.rs
@@ -721,6 +721,24 @@ Hello world
     }
 
     #[test]
+    fn can_get_date_from_filename_with_spaces_respects_slugification() {
+        let mut config = Config::default();
+        config.slugify.paths = SlugifyStrategy::Off;
+        let content = r#"
++++
++++
+Hello world
+<!-- more -->"#
+            .to_string();
+        let res = Page::parse(Path::new("2018-10-08 - hello.md"), &content, &config, &PathBuf::new());
+        assert!(res.is_ok());
+        let page = res.unwrap();
+
+        assert_eq!(page.meta.date, Some("2018-10-08".to_string()));
+        assert_eq!(page.slug, " hello");
+    }
+
+    #[test]
     fn can_get_date_from_full_rfc3339_date_in_filename() {
         let config = Config::default();
         let content = r#"
--- a/docs/content/documentation/content/page.md
+++ b/docs/content/documentation/content/page.md
@@ -65,12 +65,12 @@ When the article's output path is not specified in the frontmatter, it is extrac
 - if the filename is `index.md`, its parent folder name (`bar`) is used as output path
 - otherwise, the output path is extracted from `thing` (the filename without the `.md` extension)
 
-If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
+If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by optional whitespace and then an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
 
 The output path extracted from the file path is then slugified or not, depending on the `slugify.paths` config, as explained previously.
 
 **Example:**
-The file `content/blog/2018-10-10-hello-world.md` will yield a page at `[base_url]/blog/hello-world`.
+The file `content/blog/2018-10-10-hello-world.md` will yield a page at `[base_url]/blog/hello-world`. With optional whitespace, the file `content/blog/2021-01-23 -hello new world.md` will yield a page at `[base_url]/blog/hello-new-world`
 
 ## Front matter
 


digitalresistor:
@Keats will this be considered at all, what are the next steps you'd like me to take to move this forward? I've rebased this on the last commit of the `next` branch again.

Keats:
Sorry I missed that. 
I'm not super convinced about that feature, so far everything in Zola is WYSIWYG and spaces can be allowed in a page/section path. This change would break that assumption.

digitalresistor:
@Keats how would it break that assumption? The slug that is generated already replaces all spaces with `-`, and strips leading whitespace that does exist.

`2021-02-14- this is a page.md` currently turns into: `this-is-a-page`, with the page getting the date `2021-02-14`.

Is it expected that: `2021-02-14- this is a page.md` gets slugified as: `-this-is-a-page`?

This change allows optional whitespace before the dash and after the dash.

This way `2021-02-14 - this is a page` gets slugified as `this-is-a-page` instead of as `2021-02-14-this-is-a-page` and the page not getting any date (and thus getting ignored):

```
1 page(s) ignored (missing date or weight in a sorted section):
- ~/sample/content/2021/2021-02-14 - this is a page.md
```

I think the examples I gave above were bad ones, let me try again, this change treats the following:

```
2021-02-14 - this is a page.md
```

the same as:

```
2021-02-14-this is a page.md
```

or

```
2021-02-14- this is a page.md
```

or

```
2021-02-14-this-is-a-page.md
```

All would get a date of `2021-02-14` and all would get slugified to `this-is-a-page`.

Keats:
> Is it expected that: 2021-02-14- this is a page.md gets slugified as: -this-is-a-page?

Hmm I didn't realise this could happen right now so that's probably a bug?

In theory they should be different things depending on the slugification approach chosen: if you have selected no slugification they would refer to 2 different pages.
It looks like we need to ensure extracting the slug from the filename conforms to the strategy.

As for the PR, if the bug above is fixed then `2021-02-14 - this is a page.md` and `2021-02-14 -this is a page.md` would be two different pages in some cases.

digitalresistor:
> > Is it expected that: 2021-02-14- this is a page.md gets slugified as: -this-is-a-page?
> 
> Hmm I didn't realise this could happen right now so that's probably a bug?

No, it doesn't happen right now. At the moment spaces are stripped, and the slug for `2021-02-14- this is a page.md` turns into `this-is-a-page`.

The regex components before this PR are:

- datetime: `2021-02-14`
- slug: ` this is a page`

Thee regex components after this PR (by allowing optional whitespace around the `_` or `-`):

- datetime: `2021-02-14`
- slug: `this is a page`

Both of these get turned into:

slug: `this-is-a-page` by `slug::slugify`

Looking at the docs from: https://docs.rs/slug/0.1.4/slug/fn.slugify.html this matches the documented behavior in either case.

> 
> In theory they should be different things depending on the slugification approach chosen: if you have selected no slugification they would refer to 2 different pages.

That is true, if I changed the regex to accept a whitespace before the dash instead, that would then change no further behavior about the rest of the path.

> It looks like we need to ensure extracting the slug from the filename conforms to the strategy.
> 
> As for the PR, if the bug above is fixed then `2021-02-14 - this is a page.md` and `2021-02-14 -this is a page.md` would be two different pages in some cases.

If slugification is turned off yes, that is indeed true.

Would you accept a PR that would allow an optional space between the `date` component and the `(_|-)`, but not add optional whitespace between the `(_|-)` and the `slug` part?

This would match the original intent, and not affect page generation with existing whitespace in the slug part (when slugification is turned off), and would allow `<date>\s?-<slug>`, and since I do use slugification, my files will be named as appropriate.

I've got a couple hundred posts all named `date - name.md` and I'd prefer not to rename them all as it will break my existing blogging setup which requires the space between the date and the `-`, while I try to get things moved to Zola.

components/library/src/content/page.rs
```diff
@@ -25,7 +25,7 @@ lazy_static! {
     // Based on https://regex101.com/r/H2n38Z/1/tests
     // A regex parsing RFC3339 date followed by {_,-}, some characters and ended by .md
     static ref RFC3339_DATE: Regex = Regex::new(
-        r"^(?P<datetime>(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)([01][0-9]|2[0-3]):([0-5][0-9])))?)(_|-)(?P<slug>.+$)"
+        r"^(?P<datetime>(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)([01][0-9]|2[0-3]):([0-5][0-9])))?)\s?(_|-)\s?(?P<slug>.+$)"
```
digitalresistor:
```suggestion
        r"^(?P<datetime>(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])(T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)([01][0-9]|2[0-3]):([0-5][0-9])))?)\s?(_|-)(?P<slug>.+$)"
```

@Keats this is the change I would make, it would drop the optional whitespace after the `-` to make sure that existing processing continues to work, and while still allowing an optional whitespace after the date.

docs/content/documentation/content/page.md
```diff
@@ -65,12 +65,12 @@ When the article's output path is not specified in the frontmatter, it is extrac
 - if the filename is `index.md`, its parent folder name (`bar`) is used as output path
 - otherwise, the output path is extracted from `thing` (the filename without the `.md` extension)
 
-If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
+If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by optional whitesppace and then an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
```
digitalresistor:
```suggestion
If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by optional whitespace and then an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
```

Keats:
> No, it doesn't happen right now. At the moment spaces are stripped, and the slug for 2021-02-14- this is a page.md turns into this-is-a-page.

I've just added tests to make sure it was working in https://github.com/getzola/zola/pull/1293/commits/7fc7ef4720cf85866a7734dca83a152a9d2b2be9 and it does.
You do need to set your slugification strategy in the `config.toml` though.

> Would you accept a PR that would allow an optional space between the date component and the (_|-), but not add optional whitespace between the (_|-) and the slug part?

I think it's acceptable

digitalresistor:
@Keats I've updated this PR to match the proposed allowance of a space after the dash.

components/library/src/content/page.rs
```diff
@@ -683,6 +683,23 @@ Hello world
...
+    #[test]
+    fn can_get_date_from_filename_with_spaces() {
+        let config = Config::default();
+        let content = r#"
++++
++++
+Hello world
+<!-- more -->"#
+            .to_string();
+        let res = Page::parse(Path::new("2018-10-08 - hello.md"), &content, &config, &PathBuf::new());
```
Keats, line 715: Didn't we say the space after the `-` was not meant to be there? Eg the slug should be ` hello` depending on the slugification method chosen. You can override the slug strategy of the config to keep everything so we actually check it's working

docs/content/documentation/content/page.md
```diff
@@ -65,12 +65,12 @@ When the article's output path is not specified in the frontmatter, it is extrac
...
+If the path found starts with a datetime string (`YYYY-mm-dd` or [a RFC3339 datetime](https://www.ietf.org/rfc/rfc3339.txt)) followed by optional whitespace and then an underscore (`_`) or a dash (`-`), this date is removed from the output path and will be used as the page date (unless already set in the front-matter). Note that the full RFC3339 datetime contains colons, which is not a valid character in a filename on Windows.
 
 The output path extracted from the file path is then slugified or not, depending on the `slugify.paths` config, as explained previously.
 
 **Example:**
-The file `content/blog/2018-10-10-hello-world.md` will yield a page at `[base_url]/blog/hello-world`.
+The file `content/blog/2018-10-10-hello-world.md` will yield a page at `[base_url]/blog/hello-world`. With optional whitespace, the file `content/blog/2021-01-23 - hello new world.md` will yield a page at `[base_url]/blog/hello-new-world`
```
Keats: For the example it would be better to not have a space after the `-`

Keats (changes requested):

components/library/src/content/page.rs
digitalresistor: I've added another test that validates that the space is not lost when using a slugification strategy of off. So yes, this depends on the slugification strategy.

docs/content/documentation/content/page.md
digitalresistor: I've updated the documentation to remove the extra space after the `-`.

digitalresistor:
Thank you @Keats!

PR #1333, opened 2021-02-02
Bugfix: index can have translations, too (fix #1332)

### LINKED ISSUE #1332: Regression? index section has no translations since zola 0.13
Works on zola v0.12.2. Starting on v0.13, up to recent from two weeks ago, the index page no longer has translations. `section.translations` is empty. Other pages/sections have translations just fine.

So far i've tested on two different machines with two different sites and can reproduce. I'll try with the latest next branch, investigate what's up and report back.

southerntofu:
A regression in zola v0.13 disabled translations for the index page. See #1332

It's due to index section being specifically skipped (for optimization) before translations are mapped. This conditional skip was already here, but an optimization for translations was added in 369fb40 after it.

Simply reordering the blocks fixes the problem and does not break any test.

### DIFF
--- a/components/site/tests/common.rs
+++ b/components/site/tests/common.rs
@@ -182,9 +182,14 @@ impl Translations {
             panic!("No such page or section: {}", path);
         };
 
-        let translations = library.translations
-            .get(&unified_path)
-            .expect(&format!("Could not find page {} in translations", unified_path.display()))
+        let translations = library.translations.get(&unified_path);
+        if translations.is_none() {
+            println!("Page canonical path {} is not in library translations", unified_path.display());
+            panic!("Library error");
+        }
+
+        let translations = translations
+            .unwrap()
             .iter().map(|key| {
                 // Are we looking for a section? (no file extension here)
                 if unified_path.ends_with("_index") {
--- a/components/site/tests/common.rs
+++ b/components/site/tests/common.rs
@@ -1,6 +1,7 @@
 #![allow(dead_code)]
 use std::env;
-use std::path::PathBuf;
+use std::path::{PathBuf, Path};
+use std::collections::HashMap;
 
 use site::Site;
 use tempfile::{tempdir, TempDir};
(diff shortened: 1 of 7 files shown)


Keats:
Is it hard to make a test for that? It was easy to miss :/

southerntofu:
I should be able to add a test.

Slightly off-topic: don't you think adding so many small unit tests for corner cases is exhausting and misses many issues? I'm thinking maybe we could have more smaller test_sites with high-level integration tests (eg. all pages who have markdown translation produce HTML with equivalent number of translations), while still retaining the big test_site to ensure all features work well together. Does that make sense to you?

Keats:
I would prefer integration tests yes since otherwise unit tests make it too easy to miss things like that.

southerntofu:
So that's my first attempt in that direction. There's a lot that can be improved. With this patch, running the tests with the original regression would have triggered:

```
---- correct_translations_on_all_pages stdout ----
Page canonical path /tmp/zola/test_site_i18n/content/_index is not in library translations
thread 'correct_translations_on_all_pages' panicked at 'Library error', components/site/tests/common.rs:188:13

failures:
    correct_translations_on_all_pages
```

It was not really hard, but required some time and concentration to write these tests, because all paths are somewhat different:

- library.translations uses absolute canonical paths (extension and language code stripped)
- library.get_page|section() use absolute paths
- library.permalinks has paths relative to the content folder
- library.permalink has permalinks ending with / (not with /index.html)
- looking for files in output folder needs /index.html

PS: Hmmm apparently fails on windows because of \ being used as separator for some (all?) paths. I don't have a windows machine to know where this comes from.. Does anyone here?

components/site/tests/common.rs
```diff
@@ -67,3 +68,247 @@ where
...
+        return Some((no_ext, None));
+    }
+}
+
+/// Recursively process a folder to find translations, returning a list of every language
+/// translated for every page found. Translations for the default language are stored as "DEFAULT"
+/// TODO: This implementation does not support files with a dot inside (foo.bar.md where bar is
+/// not a language), because it requires to know what languages are enabled from config, and it's
+/// unclear how to distinguish (and what to do) between disabled language or "legit" dots
+pub fn add_translations_from(dir: &Path, strip: &str, default: &str) -> HashMap<String, Vec<String>> {
```
Keats, line 114: what are `strip` and `default` in this context?

components/site/tests/common.rs
```diff
@@ -67,3 +68,247 @@ where
...
+
+pub struct Translations {
+    trans: Vec<Translation>,
+}
+
+impl Translations {
+    pub fn for_path(site: &Site, path: &str) -> Translations {
+        let library = site.library.clone();
+        let library = library.read().unwrap();
+        // WORKAROUND because site.content_path is private
```
Keats, line 176: you can make it `pub`

components/site/tests/common.rs
```diff
@@ -67,3 +68,247 @@ where
     site.build().expect("Couldn't build the site");
     (site, tmp_dir, public.clone())
 }
+
+/// Finds the unified path (eg. _index.fr.md -> _index.md) and
+/// potential language (if not default) associated with a path
+/// When the path is not a markdown file (.md), None is returned
+/// Strips base_dir from the start of path
+fn find_lang_for(entry: &Path, base_dir: &str) -> Option<(String, Option<String>)> {
```
Keats, line 76: Can you not build that from the `FileInfo` struct?

Keats:
Ping @southerntofu 
Sadly the azure build disappeared so I can't re-run it to see the errors. Can you check my comments and push something to re-trigger them?

mtolk:
Hi @Keats, @southerntofu 

I think I managed to fix the windows specific issues.

There paths that were translated to strings and then path specific operations were done on those strings.
This caused problems related to the handling of file seperators.
When finally still having to write paths to strings I used the crate path-slash to make sure this happens in a cross-platform way.

Since I dont know if i can edit someone else's PR i just appended my changed files.
Hope this helps.


[fix.zip](https://github.com/getzola/zola/files/6121771/fix.zip)

Keats:
I can't seem to push either because it doesn't work with 2 factor auth apparently? @southerntofu can you apply the path or I'll close this PR and start another.

Keats:
Done in https://github.com/getzola/zola/pull/1417

PR #1343, opened 2021-02-08
Remove extra config parameter

ChrisChinchilla:
There was an extra `[extra.social]` not needed

### DIFF
--- a/docs/content/themes/DeepThought/index.md
+++ b/docs/content/themes/DeepThought/index.md
@@ -56,7 +56,6 @@ avatar = "/images/avatar.png"
 
 # Social link setup
 [extra.social]
-[extra.social]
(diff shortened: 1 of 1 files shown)


Keats:
Hey,

The theme section of the docs is automatically generated from https://github.com/getzola/themes so this change would be reverted on the next update. You can do the PR on the actual repo of the theme though.

ChrisChinchilla:
OK, thanks @Keats however, I did notice that there are quite a few changes to the config example in the repo for the theme, which mean that the example in the Zola docs as is doesn't work, so I am not sure if the update is manual or automatic, but might be good to update sooonish.

Keats:
I need to update the submodules once in a while but I keep forgetting to do it without some reminders :( I'll do it now

PR #1346, opened 2021-02-10
Add AVIF support to resize_image()

BezPowell:

Sanity check:

## Code changes

---
This pr adds support for AVIF images to the `resize_image()` function. As `image` does not currently allow setting the speed and quality parameters via the `write_to()` method this manually creates an encoder and writes the result out to a file.

There are 2 caveats to this implementation:
1. The `speed` parameter is hardcoded. This currently uses the same default as the `cavif-rs` crate (which `image` uses for avif encoding). This encoding speed can be very slow, so it might be nice to allow changing this. I did not add it as another argument to `resize_image()` as I felt it would start to get unwieldy. Perhaps eventually this could be overriden in `config.toml`?
2. AVIFs quality levels, while accepting the same range as jpeg are very different. The default value of 75 actually results in a file slight bigger than the equivalent jpeg. I do not know how AVIF SSIM values compare to jpeg for different quality settings, but it might be nice to choose a different default for AVIF that has a similar SSIM to jpegs 75 quality level.


BezPowell:
Just checking this over and the moment I clicked to create the PR it pulled in a load of other commits from other people. Shall I delete and try again?

Keats:
Yep, if you can just do a PR with just your commits on top of `next` that would be great

PR #1351, opened 2021-02-13
Fix "serve" subcommand for URL encoded static files

wdv4758h:
There is a bug in `zola serve` for static files. If the filename contains something that will be changed after URL encoding (e.g. CJK), then I can't access it with the builtin dev server.

For example, if I put a file like `static/測試.png` in the project.
Before:
```
http://localhost:1111/測試.png -> 404 Not Found
```
After:
```
http://localhost:1111/測試.png -> get PNG image
```

Not sure if there is more concern for this PR, but this change solves my issue.

----

Sanity check:

## Code changes


Keats:
Thanks!

Issue #1350, opened 2021-02-12
Can't find shortcode when parsing string with markdown filter


hannobraun:
# Bug Report

## Environment

Zola version: 0.13.0

## Expected Behavior

In a template, I'm trying to parse a string that contains a shortcode as markdown and insert it into the page:
```
{{ section.extra.header | markdown | safe }}
```

I expect this work like it would with normal section/page content.

## Current Behavior

I get an error that the shortcode isn't found:
```
Failed to build the site
Error: Failed to render section '/home/hanno/Projects/braun-embedded/braun-embedded.com/content/_index.md'
Reason: Failed to render 'index.html'
Reason: Filter call 'markdown' failed
Reason: Failed to render markdown filter: Error { kind: Msg("Failed to render testimonial shortcode"), source: Some(Error { kind: Msg("Tried to render `shortcodes/testimonial.html` but the template wasn\'t found"), source: None }) }
```

I did some digging and I think this line is the culprit: https://github.com/getzola/zola/blob/def95886fb086261a7218e9d5b9f92d71d27f68c/components/templates/src/filters.rs#L22

This call creates a `RenderContext` with a default `Tera` instance, which doesn't know anything about Zola's templates.

For comparison, here's how `Site` creates its `Tera` instance: https://github.com/getzola/zola/blob/def95886fb086261a7218e9d5b9f92d71d27f68c/components/site/src/lib.rs#L83

This requires a path though, and that isn't available in `MarkdownFilter`. I'll do some more digging, maybe I can find a way to fix this.

## Step to reproduce

1. Add a string containing Markdown with a reference to a shortcode to a section's frontmatter, under `[extra]`. Presumably it works the same with a page, but I haven't verified this.
2. In the section's template, try to parse that string using the `markdown` filter.
3. You should see the error when building the page.

hannobraun:
Okay, I got a fix. Preview here: https://github.com/hannobraun/zola/tree/filter-shortcode

Unfortunately I made my changes against `master`, not `next`, and my changes conflict with the changes there. I'll figure that out next week and will open a pull request then.

hannobraun:
@Keats I think this issue needs a `done` label, according to `CONTRIBUTING.md`.

pkgw:
I have a follow-up question about this change. I'm passing some text into the `markdown` filter and invoking a shortcode named `ulink`, which invokes the function `get_url`. It's causing this error:

```
Reason: Filter call 'markdown' failed
[...]

PR #1358, opened 2021-02-15
Make shortcodes work in `markdown` filter

hannobraun:
See #1350. Other details in the commits messages.

### DIFF
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -64,7 +64,8 @@ Zola adds a few filters in addition to [those](https://tera.netlify.com/docs/#fi
 in Tera.
 
 ### markdown
-Converts the given variable to HTML using Markdown. Shortcodes won't work within this filter.
+Converts the given variable to HTML using Markdown. Please note that shortcodes evaluated by this filter won't be able to access any context variables.
+
 By default, the filter will wrap all text in a paragraph. To disable this behaviour, you can
 pass `true` to the inline argument:
 
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -64,7 +64,7 @@ Zola adds a few filters in addition to [those](https://tera.netlify.com/docs/#fi
 in Tera.
 
 ### markdown
-Converts the given variable to HTML using Markdown. Please note that shortcodes evaluated by this filter won't be able to access any context variables.
+Converts the given variable to HTML using Markdown. Please note that shortcodes evaluated by this filter cannot access the current rendering context. `config` will be available, but accessing `section` or `page` (among others) from a shortcode called within the `markdown` filter will prevent your site from building. See [this discussion](https://github.com/getzola/zola/pull/1358).
 
 By default, the filter will wrap all text in a paragraph. To disable this behaviour, you can
 pass `true` to the inline argument:
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -295,7 +295,7 @@ impl Site {
         // taxonomy Tera fns are loaded in `register_early_global_fns`
         // so we do need to populate it first.
         self.populate_taxonomies()?;
-        tpls::register_early_global_fns(self);
+        tpls::register_early_global_fns(self)?;
         self.populate_sections();
         self.render_markdown()?;
         tpls::register_tera_global_fns(self);
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -20,7 +20,7 @@ use front_matter::InsertAnchor;
 use library::{find_taxonomies, Library, Page, Paginator, Section, Taxonomy};
 use relative_path::RelativePathBuf;
 use std::time::Instant;
-use templates::render_redirect_template;
+use templates::{load_tera, render_redirect_template};
 use utils::fs::{
     copy_directory, copy_file_if_needed, create_directory, create_file, ensure_directory_exists,
 };
@@ -80,7 +80,7 @@ impl Site {
             config.merge_with_theme(&path.join("themes").join(&theme).join("theme.toml"), &theme)?;
         }
 
-        let tera = tpls::load_tera(path, &config)?;
+        let tera = load_tera(path, &config)?;
 
         let content_path = path.join("content");
         let static_path = path.join("static");
--- a/components/site/src/tpls.rs
+++ b/components/site/src/tpls.rs
@@ -1,15 +1,16 @@
 use crate::Site;
 use templates::{filters, global_fns};
+use tera::Result as TeraResult;
 
 /// Adds global fns that are to be available to shortcodes while rendering markdown
-pub fn register_early_global_fns(site: &mut Site) {
+pub fn register_early_global_fns(site: &mut Site) -> TeraResult<()> {
     site.tera.register_filter(
         "markdown",
         filters::MarkdownFilter::new(
             site.base_path.clone(),
             site.config.clone(),
             site.permalinks.clone(),
-        ),
+        )?,
     );
 
     site.tera.register_function(
@@ -44,6 +45,8 @@ pub fn register_early_global_fns(site: &mut Site) {
             site.content_path.clone(),
         ]),
     );
+
+    Ok(())
 }
 
 /// Functions filled once we have parsed all the pages/sections only, so not available in shortcodes
--- a/components/site/src/tpls.rs
+++ b/components/site/src/tpls.rs
@@ -5,7 +5,11 @@ use templates::{filters, global_fns};
 pub fn register_early_global_fns(site: &mut Site) {
     site.tera.register_filter(
         "markdown",
-        filters::MarkdownFilter::new(site.config.clone(), site.permalinks.clone()),
+        filters::MarkdownFilter::new(
+            site.base_path.clone(),
+            site.config.clone(),
+            site.permalinks.clone(),
+        ),
     );
 
     site.tera.register_function(
--- a/components/site/src/tpls.rs
+++ b/components/site/src/tpls.rs
@@ -1,52 +1,5 @@
-use std::path::Path;
-
-use tera::Tera;
-
 use crate::Site;
-use config::Config;
-use errors::{bail, Error, Result};
-use templates::{filters, global_fns, ZOLA_TERA};
-use utils::templates::rewrite_theme_paths;
-
-pub fn load_tera(path: &Path, config: &Config) -> Result<Tera> {
-    let tpl_glob =
-        format!("{}/{}", path.to_string_lossy().replace("\\", "/"), "templates/**/*.{*ml,md}");
-
-    // Only parsing as we might be extending templates from themes and that would error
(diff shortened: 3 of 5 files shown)


southerntofu:
Hey thanks for the patch! That's was quick :)

I have two questions/critiques about it:

- do you think we can find a workaround to not reload all templates every time the MarkdownFilter is run (`filter()`)? that sounds very expensive, though i haven't done any benchmarking
- shortcodes usually [inherit](https://github.com/getzola/zola/blob/1ef8c85f53b4988fdafc0e6271cce590515d55aa/components/library/src/content/section.rs#L180) the page or section's rendering context, with your patch i believe shortcodes using such context will just crash (failed to access a missing variable), though i have not tested ; is that also your understanding?

hannobraun:
Thanks for the review, @southerntofu! This is my first time diving into the innards of Zola, so your feedback is very helpful.

> do you think we can find a workaround to not reload all templates every time the MarkdownFilter is run (`filter()`)? that sounds very expensive, though i haven't done any benchmarking

Good point, I didn't think of that. I didn't notice any difference in my site, for what it's worth, but that isn't worth much :-)

> shortcodes usually [inherit](https://github.com/getzola/zola/blob/1ef8c85f53b4988fdafc0e6271cce590515d55aa/components/library/src/content/section.rs#L180) the page or section's rendering context, with your patch i believe shortcodes using such context will just crash (failed to access a missing variable), though i have not tested ; is that also your understanding?

I wasn't aware of that, but it certainly makes sense the way you explain it. I didn't encounter any errors building my site, but I don't think any of my shortcodes access such context.

I will look into both of these points, and will report back once I know more (probably within the next few days).

hannobraun:
@southerntofu:
I've pushed two new commits. One extends the test site to demonstrate the error that you suspected would happen (and which is now confirmed). The other addresses your concern over reloading templates too often, which wasn't hard.

I'm still looking into approaches to address the error.

hannobraun:
I've been thinking about how to solve the issue. The problem is that we need to somehow pass the current context into the markdown filter before it is run. I see multiple ways to do this, and none of them are any good:

1. Pass the context into `MarkdownFilter::new`. Right now, if I understand correctly, filters are [constructed on initialization](https://github.com/getzola/zola/blob/d734b1723c4097fe2bf5da1fe6b0f5c0060c6c2d/components/site/src/tpls.rs#L53-L56). We'd have to change that and construct a new `MarkdownFilter` every time we render a section or page. Sounds workable, but I'm not sure what the performance implications are, or if this would be undesirable for other reasons.
2. Pass the context into `tera::Filter::filter`. Since this isn't currently supported, we'd need to add an additional argument to the trait method, an associated type to the trait to specify the type of that argument, and some kind of mechanism in Tera to pass that argument to the filter. I don't think that this could be done in a backwards-compatible way (that would require associated type defaults, at least), and I'm not sure if such extensive changes to Tera would be welcome anyway.
3. Add an additional method (`filter_with_context`) to `tera::Filter` to pass dynamic context to. It might be possible to pass the context dynamically, as a `Box<Any>` and downcast it to the type that `MarkdownFilter` needs. I think this could be done in a backwards-compatible way (the new filter method could have a default implementation that just calls the old one), but would still require adding custom parameter infrastructure to Tera. The `Box<Any>` thing also feels like a hack.
4. Add a method to `MarkdownFilter` that accepts a context and call that whenever rendering a page. The problem here is that the filters live as `Arc<dyn Filter>` in `Tera`, so we'd at least need to add an `Any` bound to `Filter` to do anything with it. This would be an incompatible change to Tera and would limit which types could be used as filters (although I don't know if that would matter in practice).

Please let me know if I'm missing anything!

A much more practical solution would be to just merge this PR as-is, more of less. Yes, shortcodes in Markdown rendered by the `markdown` filter can't access page or section context, but given that without this PR the `markdown` filter doesn't render shortcodes at all, this is a strict improvement. Given that I've already spent more time on this than I wanted, and don't run into this limitation in my site, this is my preferred solution, to be honest.

Keats:
> A much more practical solution would be to just merge this PR as-is, more of less.

I'm tempted to agree on that. Hopefully people using the `markdown` filter are not expecting a page or section argument since it can be used anywhere. As long as it's documented, I think it's fine.
The failing tests need to be fixed though

hannobraun:
@Keats Great, I'll get the PR ready tomorrow, hopefully (next week, maybe).

The tests are failing because I demonstrated the shortcode context error in the test site, so I just have to drop that one commit to get them running again.

southerntofu:
Hey, thanks for the in-depth analysis @hannobraun. I believe solution 2 would be viable given minor modifications. I wrote a proof-of-concept [here](https://github.com/southerntofu/zola/commit/929048375767388511e372ae3c3ccb95b6a7b072) which passes the test. i ended up passing from the template a list of (String, Value) (context elements)  but flattened as `Vec<Value>` directly to the markdown filter, using a dedicatd `context` argument. Some notes:

- tera currently has no way (that i could see) to designate the current context ; `__tera_context` is a pretty-printed string, so it's not suitable to be passed around
- tera has no Map literals and weird (broken?) List<List>> support
- it's just a proof of concept and i wouldn't recommend to adopt this list of context key/values to insert ; Map literals in tera (or another trick) would make it nicer
- i did not find a convenient way to write a unit test for the filter
- context has to be passed explicitly to the markdown filter ; i don't exactly like that (it's counter-intuitive from a template designer's perspective), but i appreciate the value of declaring what should be passed precisely and not cloning around a lot of data by default all the time

Any ideas?

Side note: there is a `json_encode` but i could not see a corresponding `json_decode` filter. The closest was `load_data` function but i don't think it supports a string source (that's not a path or a URL). I was thinking maybe there could be a way to turn `__tera_context` back into a Map after all, but well...

I'm happy with the PR as it is which is already a neat improvement @hannobraun, i'm merely continuing the conversation (which may be moved to a new issue).

About tera:

- there is already a ticket there about [support for Maps](https://github.com/Keats/tera/issues/420), however @Keats said it represented a considerable amount of work
- having tera filters aware of the context (your solution 3 or 4) is being discussed on [tera bugtracker](https://github.com/Keats/tera/issues/543#issuecomment-774698933) ; i believe it makes sense to have separate types of filters for functional filters (no context) and contextual filters

Keats:
> there is already a ticket there about support for Maps, however @Keats said it represented a considerable amount of work

I'm holding off any work on Tera parser waiting for pest3 but it doesn't look like this is happening any time soon. I'll have to investigate other libraries but rewriting the parser is a lot of work.

> i believe it makes sense to have separate types of filters for functional filters (no context) and contextual filters

I think the context will be a function since there can be multiple contexts while rendering (eg a forloop might shadow some values from the context *inside* the loop). I don't think we want to distinguish between no context and with context though: the user can just ignore the context argument if they don't use it.

Tera v2 will basically be rewritten parser + context aware filters/functions. And if someone figures out how to avoid JSON serialisation it would probably be its own major version.

southerntofu:
> waiting for pest3 but it doesn't look like this is happening any time soon

The maintainer [has said](https://github.com/pest-parser/pest/issues/489#issuecomment-775114684) they are too busy at the moment, but are happy to guide contributions ("I have a clear path forward if anyone would be interested to contribute"). So maybe passing around that call to contribution can help? We could have a new "Contributing" section in the README highlighting good-first-issues, issues we need help with, but also upstream issues and calls for maintainers. What do you think?

> I don't think we want to distinguish between no context and with context though

Well if you have clues how to make it zero-copy, sure. But as a template designer, i appreciate that how complex my build is depends on explicit trade-offs i make in the functions i call (eg. when i load information from other pages onto a page, i'm expecting a performance hit). If context passing is not zero-copy, i'm afraid trivial "functional templating" patterns (small macros everywhere) will be less attractive with tera.. but maybe that's a discussion for tera bugtracker?

About my poc context passing to the markdown filter, do you think:

- any other existing filters need to support that?
- if no Map literals are on the short-term TODO list, would a `newmap` function (or other map manipulation functions) be welcome either in zola or tera? ([this neater approach for example](https://github.com/southerntofu/zola/commit/9d5d93a02df7bf99e26bef04405657aab4a3126e) also passes the tests) this could include Map support in concat, or even an insert filter which works on both Lists and Maps (`{% set context = context | insert(key="foo", value="bar") %}`)
- current context could be exposed in a non-stringy way? (like go templates do with `.`) ; perhaps pretty printing structures could be a dedicated filter (or is that already what json_encode does?) so there would be no need to compute pretty-printed `__tera_context` on every render

hannobraun:
I've update this pull request:
- Dropped the commit that demonstrated the shortcode context problem in the test site (tests pass again).
- Updated the respective commit message to mention the shortcode context limitation.
- Updated the commit that changes the documentation to mention the limitation.

Should be good to go now!

docs/content/documentation/templates/overview.md
```diff
@@ -64,7 +64,8 @@ Zola adds a few filters in addition to [those](https://tera.netlify.com/docs/#fi
 in Tera.
 
 ### markdown
-Converts the given variable to HTML using Markdown. Shortcodes won't work within this filter.
+Converts the given variable to HTML using Markdown. Please note that shortcodes evaluated by this filter won't be able to access any context variables.
```
southerntofu: I believe config context (and maybe some other things) will be available inside the markdown. Maybe something like:
```
Please note that shortcodes evaluated by this filter cannot currently access the current rendering context. `config` will be available, but accessing `section` or `page` (among others) from a shortcode called within the `markdown` filter will prevent your site from building. See [this discussion](https://github.com/getzola/zola/pull/1358).
```

hannobraun: Done!

(Minus one word; "currently access the current" read a bit awkward, so I dropped "currently".)

Keats:
> The maintainer has said they are too busy at the moment, but are happy to guide contributions ("I have a clear path forward if anyone would be interested to contribute"). So maybe passing around that call to contribution can help? We could have a new "Contributing" section in the README highlighting good-first-issues, issues we need help with, but also upstream issues and calls for maintainers. What do you think?

I've offered to help with the testing part since he said he had a plan for pest3 but I don't really know what that plan is and it isn't documented afaik. I don't know if the overlap Zola <-> Pest is big enough to be useful for the contribution bits. There's been discussion about pest3 since 2018 and at this point I would rather go with something more maintained I think.

> Well if you have clues how to make it zero-copy, sure. But as a template designer, i appreciate that how complex my build is depends on explicit trade-offs i make in the functions i call (eg. when i load information from other pages onto a page, i'm expecting a performance hit). If context passing is not zero-copy, i'm afraid trivial "functional templating" patterns (small macros everywhere) will be less attractive with tera.. but maybe that's a discussion for tera bugtracker?

It would have to be zero copy definitely. I have an idea but I don't know if it will actually work. If it does it would be trivial though but it would be a breaking change so for Tera v2.

I think your poc would be a bandaid and map literals/context should be fixed directly in Tera. Intermediate helpers like a new `insert` filter would probably cause too much churn imo.

Keats:
Thanks @hannobraun !

PR #1376, opened 2021-02-22
Document conditional TOC rendering

yhoiseth:
It took me some tinkering to figure out the simplest way to conditionally render the table of contents. So I think this might be useful to others as well.

Sanity check:

### DIFF
--- a/docs/content/documentation/content/table-of-contents.md
+++ b/docs/content/documentation/content/table-of-contents.md
@@ -34,3 +34,20 @@ While headers are neatly ordered in this example, it will work just as well with
 
 Note that all existing HTML tags from the title will NOT be present in the table of contents to
 avoid various issues.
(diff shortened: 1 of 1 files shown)


docs/content/documentation/content/table-of-contents.md
```diff
@@ -34,3 +34,20 @@ While headers are neatly ordered in this example, it will work just as well with
 
 Note that all existing HTML tags from the title will NOT be present in the table of contents to
 avoid various issues.
+
+## Conditional rendering
```
Keats: I think it would be clearer if you just wrapped the example above in `{% if page.toc %}`

Keats (changes requested):

Keats:
Thanks!

PR #1377, opened 2021-02-22
Document cache invalidation opportunity

yhoiseth:
It took me a while to discover `cachebust`, so it might be useful to others to have it in the Sass docs.

Sanity check:

### DIFF
--- a/docs/content/documentation/content/sass.md
+++ b/docs/content/documentation/content/sass.md
@@ -40,3 +40,11 @@ Files with the `scss` extension use "Sassy CSS" syntax,
 while files with the `sass` extension use the "indented" syntax: <https://sass-lang.com/documentation/syntax>.
 Zola will return an error if `scss` and `sass` files with the same
 base name exist in the same folder to avoid confusion -- see the example above.
+
+### Cache invalidation
+
+Often, browsers will cache files such as stylesheets. In order to make sure that the visitors to your site always get the latest version, you can use the global `get_url` function’s `cachebust` option. For example:
+
+```jinja2
+<link href="{{ get_url(path='style.css', cachebust=true) }}" rel="stylesheet">
+```


docs/content/documentation/content/sass.md
```diff
@@ -40,3 +40,11 @@ Files with the `scss` extension use "Sassy CSS" syntax,
 while files with the `sass` extension use the "indented" syntax: <https://sass-lang.com/documentation/syntax>.
 Zola will return an error if `scss` and `sass` files with the same
 base name exist in the same folder to avoid confusion -- see the example above.
+
```
Keats, line 43: This part of the docs is about the content itself, not the templates. None of the other pages in that section mention template specific things like functions.

I agree it would be useful to show this kind of section _somewhere_ but I don't know where. @southerntofu any ideas?

southerntofu: It should probably go [here](https://www.getzola.org/documentation/templates/overview/#data-caching) with the `get_url` docs no? (by the way having built-ins in theme overview has always confused me i'd rather have a "FIlters/Functions" pages)

Also maybe explain what `cachebust` does (checksumming) by providing an example output HTML?

Keats: > by the way having built-ins in theme overview has always confused me i'd rather have a "FIlters/Functions" pages

Yeah... this was made when there was one function, it should be split.

yhoiseth: We now have 3 filters and 10 functions. 

Maybe it would make sense to have two separate pages — one for filters and one for functions? I see a few advantages to having separate pages:

1. Users may more quickly understand the difference between filters and functions.
2. Shorter pages means less scrolling to get to the thing you are looking for.
3. If we use one page now and add more content, we might want to split the pages later. In that case, we may break links to the documentation. Using two pages right away is more future-proof.

On the other hand, it would mean a bigger menu. I see that as less of a concern than the ones outlined above.

Keats: I would put them all in the same page though, at least for now

Issue #1250, opened 2020-12-05
Can't generate language-specific feed url with get_url


bemyak:
# Bug Report

## Environment

Zola version:`0.12.2`, `next`

## Expected Behavior

Calling `get_url(path=config.feed_filename, lang=lang)` for non-default language should produce `/<lang>/atom.xml`

## Current Behavior
`get_url` ignores `lang` argument and always outputs `/atom.xml`

## Step to reproduce
1. Create a new zola site
1. Setup two languages: default and non-default
1. Try calling `get_url` function for non-default language

A very simple test project can be found [here](https://gitlab.com/bemyak/zola-test). Click `Switch to "non-default" language` and verify that `Feed url` did not change.

This looks relatively easy to fix, but I'm afraid to break backward compatibility.

Keats:
Well that's a bug so fixing it is fine. Hopefully no one relies on a bug

PR #1381, opened 2021-02-28
Fix language specific non-internal urls for get_url.

snake66:
Here's a take on fixing issue #1250. I've only added tests for the feed filename, but I think that should cover most cases. Let me know if you want me to add more tests cases, or if you want me to make other changes to the patch.

Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/global_fns/mod.rs
+++ b/components/templates/src/global_fns/mod.rs
@@ -149,20 +149,15 @@ impl TeraFn for GetUrl {
             }
         } else {
             // anything else
-            let mut pathbuf = PathBuf::new();
+            let mut segments = vec![];
 
             if lang != self.config.default_language {
-                pathbuf.push(lang);
+                segments.push(lang);
             };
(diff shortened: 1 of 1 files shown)


components/templates/src/global_fns/mod.rs
```diff
@@ -149,21 +149,37 @@ impl TeraFn for GetUrl {
             }
         } else {
             // anything else
-            let mut permalink = self.config.make_permalink(&path);
+            let mut pathbuf = PathBuf::new();
```
Keats: Using PathBuf for urls won't work in Windows.

Keats (changes requested):

components/templates/src/global_fns/mod.rs
snake66: Ack, that makes sense. I'll find another way.

Keats:
Thanks!

PR #1382, opened 2021-03-01
Add CORS header

lf-:
I was having an issue with `zola serve` where the browser was not
fetching fonts on account of it not sending a CORS header with the
fonts

I've added a CORS header specifying the host that is printed out at the
command line.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -128,7 +128,11 @@ async fn handle_request(req: Request<Body>, mut root: PathBuf) -> Result<Respons
 
     Ok(Response::builder()
         .status(StatusCode::OK)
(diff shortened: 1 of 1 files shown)


lf-:
This leaves a potential footgun where the user goes to `localhost:1111` and doesn't have their fonts and suchlike load. Not sure if we should make Access-Control-Allow-Origin `*`.

Keats:
> Not sure if we should make Access-Control-Allow-Origin *.

Keats:
> Not sure if we should make Access-Control-Allow-Origin *.

I think it would be sensible to do that

Keats:
> Not sure if we should make Access-Control-Allow-Origin *.

I think it would be sensible to do that

lf-:
alright, this should be ok now.

Keats:
Thanks!

Issue #294, opened 2018-04-24
Can't serve using a custom config file


raffomania:
Starting gutenberg like so:

    gutenberg --config config.dev.toml serve

I get the following output:
```
Building site...
-> Creating 0 pages (0 orphan) and 0 sections
Done in 14ms.

Error: Can't watch the `config.toml` file. Does it exist?
Reason: entity not found
Reason: No such file or directory (os error 2)
```

Keats:
Which version of Gutenberg are you using? It should be working in the latest version (0.3.3)

raffomania:
I'm using version 0.3.3, and still get the error. You should be able to easily reproduce this:

    gutenberg init test
    cd test
    mv config.toml config.dev.toml
    gutenberg --config config.dev.toml serve

Keats:
Thanks, that's the watcher that was expecting a `config.toml` file to be present, I should have read the error message in the first post more carefully >_>
It's fixed in the patch above that will be in 0.4.0

raffomania:
Thanks :)

PR #1385, opened 2021-03-03
Fix serve command when used with config option

kristoforsalmin:
Hi there!

I noticed that the `serve` command isn't working as expected when used together with the `--config` option.

When I run `zola -c ./config.prod.toml serve`, I expect `config.prod.toml` (assuming `config.prod.toml` is the only config file I have) to be watched and a website to be built, but I was getting this:

```sh
❯ zola -c ./config.prod.toml serve
Building site...
Done in 44ms.

Error: Can't watch `config.toml` for changes in folder `...`. Does it exist, and do you have correct permissions?
Reason: entity not found
```

The quick fix is to create an empty `config.toml` file. That no longer yields an error, but `config.prod.toml` still isn't watched.

Similar issue was reported and fixed a couple of years ago (see #294).

I have no prior Rust experience, so code adjustments and comments are very much appreciated.

---

Sanity check:

## Code changes

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -689,36 +689,48 @@ mod tests {
                 (ChangeKind::Templates, PathBuf::from("/templates/hello.html")),
                 Path::new("/home/vincent/site"),
                 Path::new("/home/vincent/site/templates/hello.html"),
+                "config.toml",
             ),
             (
                 (ChangeKind::Themes, PathBuf::from("/themes/hello.html")),
                 Path::new("/home/vincent/site"),
                 Path::new("/home/vincent/site/themes/hello.html"),
+                "config.toml",
             ),
             (
                 (ChangeKind::StaticFiles, PathBuf::from("/static/site.css")),
                 Path::new("/home/vincent/site"),
                 Path::new("/home/vincent/site/static/site.css"),
+                "config.toml",
             ),
             (
                 (ChangeKind::Content, PathBuf::from("/content/posts/hello.md")),
                 Path::new("/home/vincent/site"),
                 Path::new("/home/vincent/site/content/posts/hello.md"),
+                "config.toml",
             ),
             (
                 (ChangeKind::Sass, PathBuf::from("/sass/print.scss")),
                 Path::new("/home/vincent/site"),
                 Path::new("/home/vincent/site/sass/print.scss"),
+                "config.toml",
             ),
             (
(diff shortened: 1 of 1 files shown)


src/cmd/serve.rs
```diff
@@ -631,7 +637,7 @@ fn detect_change_kind(pwd: &Path, path: &Path) -> (ChangeKind, PathBuf) {
         ChangeKind::StaticFiles
     } else if partial_path.starts_with("/sass") {
         ChangeKind::Sass
-    } else if partial_path == Path::new("/config.toml") {
+    } else if partial_path.extension().unwrap() == "toml" {
```
kristoforsalmin: I'm not entirely sure about this one. Basically any TOML file is treated as Zola's `config.toml`. Should we somehow enforce both "config" and "toml" to be present?

kristoforsalmin:
Also, while I was testing I noticed that `zola serve` picks up changes in `config.toml` only once:

1. `zola serve` — build ✔
2. Change in `config.toml` — change detected, rebuild ✔
3. Change in `config.toml` — nothing happens ❌

This is happening in the `next` branch (even without my changes) and I can't seem to understand why. Seems like watcher doesn't trigger change events after the first rebuild.

src/cmd/serve.rs
Keats: You can pass the config filename to the function so it watches the correct one

Keats:
The watcher issue is odd, can you check if we get an error when setting it up?

src/cmd/serve.rs
kristoforsalmin: @Keats good point :+1: Fixed!

kristoforsalmin:
> The watcher issue is odd, can you check if we get an error when setting it up?

Yes, it's odd and actually I don't think it's an issue anymore. I'm sorry for the confusion.

I use Windows and WSL as my main development environment, but I had problems building `sass-rs`, so I quickly booted Ubuntu from an USB stick (live version) and that's where I encountered the issue in question. Today, unsatisfied with performance over USB, I installed regular distribution of Ubuntu and the issue is gone. I suppose live version might have some filesystem limitations.

src/cmd/serve.rs
kristoforsalmin: Fixing tests...

Keats:
Thanks!

kristoforsalmin:
@Keats and thanks to you for the continuous effort on building Zola — it's awesome!

PR #1389, opened 2021-03-05
Official zola.386 theme's demo page corrected

thedamian:
Official zola.386 theme's demo page corrected as per their own documentation at https://github.com/lopes/zola.386

Sanity check:

* [*] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?


Keats:
Thanks for the PR but these docs are actually automatically generated from https://github.com/getzola/themes so I think I just need to update the submodules there.

PR #1390, opened 2021-03-06
Documentation: Clarify where exactly to define the theme in the .toml file

medizinmensch:
It was not immediately clear for me (as a first time .toml user) that the user has to place the theme variable in the top level of the config file. I had an issue with that and would like to save some time for fellow zola users. 
[Others](https://zola.discourse.group/t/i-am-missing-something-to-start-using-the-product/423/12) have encountered this as well.

The only thing I would like you to think about is: Is this a general rule or theme dependent? If the latter, then I have to rephrase oc. 

Sanity check:


Keats:
> It was not immediately clear for me (as a first time .toml user)

You found the worst part of TOML imo, which makes it kind of annoying sometimes...
Thanks for the PR!

hmzbg:
Hey

Issue #1287, opened 2021-01-02
Extra character in front-matter led to 30 minute hunt for the extra <ul>


digitalresistor:
# Enhancement

## Environment

Zola version: 0.12.2

## Expected Behavior

It'd be great if Zola warned if the front-matter fences were malformed.

## Current Behavior

Zola strips off the front-matter, and passes the file to the markdown renderer

## Step to reproduce

Create a piece of content that looks like this:

```
+++
title = "Some random title"
date = 2021-01-02
++++
```

Then look at the output of the generated content:

```
<ul><li></li></ul>
```

I will admit that I am tired, and thus it was my fault, however instead of warning me that the fence around the front-matter was malformed, zola passed the extra `+` to the markdown renderer. Leading me on a weird goose chase.

It'd be great if it were a requirement that no other text were allowed to exist on the closing line fo the front matter fence!

xpe:
Here are the relevant pieces of code from https://github.com/getzola/zola/blob/master/components/front_matter/src/lib.rs

```rs
lazy_static! {
    static ref PAGE_RE: Regex =
        Regex::new(r"^[[:space:]]*\+\+\+(\r?\n(?s).*?(?-s))\+\+\+\r?\n?((?s).*(?-s))$").unwrap();
}
```

```rs
/// Split a file between the front matter and its content
/// Will return an error if the front matter wasn't found
fn split_content<'c>(file_path: &Path, content: &'c str) -> Result<(&'c str, &'c str)> {
    if !PAGE_RE.is_match(content) {
        bail!(
            "Couldn't find front matter in `{}`. Did you forget to add `+++`?",
            file_path.to_string_lossy()
        );
    }

    // 2. extract the front matter and the content
    let caps = PAGE_RE.captures(content).unwrap();
    // caps[0] is the full match
    // caps[1] => front matter
    // caps[2] => content
    Ok((caps.get(1).unwrap().as_str(), caps.get(2).unwrap().as_str()))
}
```

xpe:
It seems reasonable to me require `\+\+\+\r?\n` to close the frontmatter. Would this break expectations? Some specification? Important use cases?

Keats:
I think it would be fine

lf-:
I got surprised by this behaviour and just patched it out on account of not having looked at the git history. The trouble that the current impl causes is that if there is not a trailing newline in an intentionally empty file, `_index.md`, for example. But I get this and I see it probably stops a bigger footgun. So it's probably fine with me.

[...]

PR #1408, opened 2021-03-15
Feature: allow optional newline after frontmater

digitalresistor:
This fixes an issue whereby the user may want to have a stand-alone `_index.md` file that contains nothing but a front matter. Without content there is no requirement for there to be a newline following the front-matter.

This PR is thanks to https://github.com/getzola/zola/issues/1287#issuecomment-787970320 (an issue I also ran into myself). This updates the regular expression to look for either the end of the file OR it requires a newline before the content.

This satisfies the original bug report (new tests added to validated that as well), as well as some new test cases to validate that no newline is required after the front matter.


Keats:
Thanks!

PR #1421, opened 2021-03-30
link_checking: prevent rate-limiting

angristan:
Fix for https://github.com/getzola/zola/issues/1056.

- assign all links for a domain to the same thread
- reduce number of threads from 32 to 8
- add sleep between HTTP calls

Demo

```
devenv :: ~/dawn-zola ‹main› » time ~/zola/target/debug/zola check
Checking site...
Checking 0 internal link(s) with an anchor.
Checking 502 external link(s).
Thread for domain: soundiiz.com
Thread for domain: grafana.com
Thread for domain: web.archive.org
[...]
Thread for domain: docs.docker.com
Domain: docs.docker.com, url: "https://docs.docker.com/docker-hub/github/"
Domain: docs.docker.com, url: "https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin"
Thread for domain: wiki.diasporafoundation.org
Domain: wiki.diasporafoundation.org, url: "https://wiki.diasporafoundation.org/Installation/Debian/Jessie"
Thread for domain: core.telegram.org
Domain: core.telegram.org, url: "https://core.telegram.org/bots"
Domain: grafana.com, url: "https://grafana.com/plugins?type=datasource"
Domain: twitter.com, url: "https://twitter.com/fuolpit"
Domain: docs.docker.com, url: "https://docs.docker.com/storage/storagedriver/zfs-driver/"
Thread for domain: caniuse.com
Domain: caniuse.com, url: "https://caniuse.com/#feat=tls1-3"
Domain: twitter.com, url: "https://twitter.com/torproject?ref_src=twsrc%5Etfw"
[...]
Domain: github.com, url: "https://github.com/hyperic/sigar/issues/74"
Domain: github.com, url: "https://github.com/Angristan/dockerfiles/tree/master/diaspora"
Domain: github.com, url: "https://github.com/angristan/dockerfiles/tree/master/diaspora"
Domain: github.com, url: "https://github.com/caddyserver/caddy/issues/2080"
Domain: github.com, url: "https://github.com/caddyserver/caddy/releases/tag/v0.11.5"
Domain: github.com, url: "https://github.com/sivel/speedtest-cli"
Domain: github.com, url: "https://github.com/sivel/speedtest-cli#installation"
Domain: github.com, url: "https://github.com/sivel/speedtest-cli#usage"
Domain: github.com, url: "https://github.com/Microsoft/vscode/issues/51132"
Domain: github.com, url: "https://github.com/alexanderyakusik"
Domain: github.com, url: "https://github.com/Microsoft/vscode/issues/51132#issuecomment-424132330"
> Checked 502 external link(s): 1 error(s) found.
Failed to check the site
Error: Dead link in /root/dawn-zola/content/2018-03-diaspora-in-docker/index.md to https://gitlab.koehn.com/docker/diaspora: error sending request for url (https://gitlab.koehn.com/docker/diaspora): operation timed out
~/zola/target/debug/zola check  14.13s user 1.44s system 8% cpu 3:05.32 total
```

Considering there is 500 links to check, it takes about 2.7s per link. Out of these 500 links there are 150 `github.com` links, that's why we see all the github links at the end, since this is the last thread running. I didn't get any `429 Too Many Requests` using this patch.

---

This is my first time using Rust 🦀, so I'm not familiar with the idiomatic way of doing things. I had a hard times with iterators, but it looks like I ended up with something that works! I look forward to some review so I can see what can be improved. 🙏  I added a few comments on things that I'm not sure how to handle.

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -156,14 +156,11 @@ pub fn check_external_links(site: &Site) -> Result<()> {
     let errors = pool.install(|| {
         links_by_domain
             .par_iter()
-            .map(|(domain, links)| {
-                println!("Thread for domain: {}", domain);
-
+            .map(|(_domain, links)| {
                 let mut links_to_process = links.len();
                 links
                     .into_iter()
                     .filter_map(move |(page_path, link)| {
-                        println!("Domain: {}, url: {:?}", domain, link);
                         links_to_process -= 1;
 
                         if site
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -2,7 +2,8 @@ use rayon::prelude::*;
 
 use crate::Site;
 use core::time;
-use errors::{Error, ErrorKind, Result};
+use errors::{bail, Result};
+use errors::{Error, ErrorKind};
 use std::{collections::HashMap, path::PathBuf, thread};
 use url::Url;
 
@@ -97,43 +98,35 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
     Err(Error { kind: ErrorKind::Msg(msg), source: None })
 }
 
+fn get_link_domain(link: &str) -> Result<String> {
+    return match Url::parse(&link) {
+        Ok(url) => match url.host_str().map(String::from) {
+            Some(domain_str) => Ok(domain_str),
+            None => bail!("could not parse domain `{}` from link", link),
+        },
+        Err(err) => bail!("could not parse domain `{}` from link: `{}`", link, err),
+    };
+}
+
 pub fn check_external_links(site: &Site) -> Result<()> {
     let library = site.library.write().expect("Get lock for check_external_links");
-    let page_links = library
-        .pages()
-        .values()
-        .map(|p| {
-            let path = &p.file.path;
-            p.external_links.iter().map(move |l| {
-                let domain = match Url::parse(l) {
-                    Ok(url) => match url.host_str().map(String::from) {
-                        Some(domain_str) => domain_str,
-                        None => "".to_string(),
-                    },
-                    Err(why) => panic!("{:?}", why),
-                };
-                (path.clone(), l.to_string(), domain)
-            })
-        })
-        .flatten();
-    let section_links = library
-        .sections()
-        .values()
-        .map(|p| {
-            let path = &p.file.path;
-            p.external_links.iter().map(move |l| {
-                let domain = match Url::parse(l) {
-                    Ok(url) => match url.host_str().map(String::from) {
-                        Some(domain_str) => domain_str,
-                        None => "".to_string(),
-                    },
-                    Err(why) => panic!("{:?}", why),
-                };
-                (path.clone(), l.to_string(), domain)
-            })
-        })
-        .flatten();
-    let all_links = page_links.chain(section_links).collect::<Vec<_>>();
+
+    let mut all_links: Vec<(PathBuf, String, String)> = vec![];
+
+    for p in library.pages_values().into_iter() {
+        for external_link in p.clone().external_links.into_iter() {
+            let domain = get_link_domain(&external_link)?;
+            all_links.push((p.file.path.clone(), external_link, domain));
+        }
+    }
+
+    for s in library.sections_values().into_iter() {
+        for external_link in s.clone().external_links.into_iter() {
+            let domain = get_link_domain(&external_link)?;
+            all_links.push((s.file.path.clone(), external_link, domain));
+        }
+    }
+
     println!("Checking {} external link(s).", all_links.len());
 
     let mut links_by_domain: HashMap<String, Vec<(PathBuf, String)>> = HashMap::new();
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -159,12 +159,12 @@ pub fn check_external_links(site: &Site) -> Result<()> {
             .map(|(domain, links)| {
                 println!("Thread for domain: {}", domain);
 
-                let mut links_processed = 0;
+                let mut links_to_process = links.len();
                 links
                     .into_iter()
                     .filter_map(move |(page_path, link)| {
                         println!("Domain: {}, url: {:?}", domain, link);
-                        links_processed += 1;
+                        links_to_process -= 1;
 
                         if site
                             .config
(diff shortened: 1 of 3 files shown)


components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
         .map(|p| {
             let path = &p.file.path;
-            p.external_links.iter().map(move |l| (path.clone(), l))
+            p.external_links.iter().map(move |l| {
+                let domain = match Url::parse(l) {
+                    Ok(url) => match url.host_str().map(String::from) {
+                        Some(domain_str) => domain_str,
+                        None => "".to_string(),
+                    },
+                    Err(why) => panic!("{:?}", why),
```
angristan: Not sure what to do here. I'm pretty sure we would never get an error since the links are already parsed by the markdown renderer?

components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
+                };
+                (path.clone(), l.to_string(), domain)
+            })
```
angristan: This is identical to the `page_links` snippet above. Is that a problem? Should we put these lines a function?

components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
     if all_links.is_empty() {
         return Ok(());
     }
 
     // create thread pool with lots of threads so we can fetch
-    // (almost) all pages simultaneously
-    let threads = std::cmp::min(all_links.len(), 32);
+    // (almost) all pages simultaneously, limiting all links for a single
+    // domain to one thread to avoid rate-limiting
+    let threads = std::cmp::min(links_by_domain.len(), 8);
```
angristan, line 150: 8 threads seems to be a reasonable default for me, but this is kind of an arbitrary change 😅

components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
-                } else {
-                    Some((page_path, link, res))
-                }
+            .map(|(domain, links)| {
+                println!("Thread for domain: {}", domain);
+
+                links
+                    .into_iter()
+                    .filter_map(move |(page_path, link)| {
+                        println!("Domain: {}, url: {:?}", domain, link);
```
angristan: I'll remove this `println` and the one above, this was just for debugging/demo purposes

components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
+                            .link_checker
+                            .skip_prefixes
+                            .iter()
+                            .any(|prefix| link.starts_with(prefix))
+                        {
+                            return None;
+                        }
+                        let res = link_checker::check_url(&link, &site.config.link_checker);
+                        // Prevent rate-limiting
+                        thread::sleep(time::Duration::from_millis(500));
```
angristan: - I chose 500ms. What do you think?
- Ideally, I would like to sleep only if there is a "next" element in the iterator (if that makes sense)
  -  i.e. if we're at the last link for a domain, no need to sleep since the next one will be for another domain
  - this would overall save a few seconds of run time for `check`
  - is this achievable easily in rust? 🙂

components/site/src/link_checking.rs
```diff
@@ -101,53 +104,94 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
+                        if site
+                            .config
+                            .link_checker
+                            .skip_prefixes
+                            .iter()
+                            .any(|prefix| link.starts_with(prefix))
+                        {
+                            return None;
+                        }
+                        let res = link_checker::check_url(&link, &site.config.link_checker);
```
hugomd, line 176: I'm not familiar with Zola or Rust, but I wonder if you could also check the [`Retry-After` header](https://tools.ietf.org/html/rfc7231#section-7.1.3) to determine whether you've been rate limited, and determine the amount of time to wait before retrying.

This is probably out of the scope of this PR though 🤷🏻‍♂️

components/site/src/link_checking.rs
Keats: They are not parsed afaik so they could be invalid.
I think it's a case where it would probably be easier to follow the code if we add errors by using a for loop rather than iterators.

components/site/src/link_checking.rs
Keats: So was the original 32 :p

components/site/src/link_checking.rs
Keats: That would be a nice follow up addition but definitely not needed for this PR.

components/site/src/link_checking.rs
Keats: Depending on the way you write it, you can use https://doc.rust-lang.org/std/iter/struct.Peekable.html

components/site/src/link_checking.rs
Keats: Yep, I want to add some tests for it too

Keats (changes requested):

components/site/src/link_checking.rs
angristan: Done in https://github.com/getzola/zola/pull/1421/commits/72f21d7b578422abfa84ed6b1e912bb03fc2f430, it looks simpler to me. What do you think? Also, is the error handling ok?

components/site/src/link_checking.rs
angristan: Moved domain parsing in `get_link_domain()` in https://github.com/getzola/zola/pull/1421/commits/72f21d7b578422abfa84ed6b1e912bb03fc2f430.

- Is the name ok?
- Not sure where to put it 😄 
- And also as said above, not sure about the handling of `Option` and `Result`.

components/site/src/link_checking.rs
angristan: That's what I was thinking about! I'll take a look

angristan: I wasn't sure that `Peekable` was the way to go, so I went for a simpler and more classic solution: https://github.com/getzola/zola/pull/1421/commits/a7f862c3bbdc161f4c181f2f2ed87113b8fb4098

angristan: Better: https://github.com/getzola/zola/pull/1421/commits/cd2d021b15fbec2abb41405567d72a9e95c156de

components/site/src/link_checking.rs
Keats: Can you remove them?

Keats (approved):
LGTM, just need to remove the println

components/site/src/link_checking.rs
angristan: oops, I forgot about them, done!

Keats:
Thanks!

PR #1424, opened 2021-03-30
rendering: keep track of all internal links

lucab:
This updates rendered markdown structures in order to keep track
of all internal links, not anymore limiting to only those targeting
an explicit anchor fragment.
The goal of this rework is to allow building other features, such
as backlinks, on top of the existing collection of internal links.

---

Sanity check:
  * Not applicable

### DIFF
--- a/components/library/src/content/page.rs
+++ b/components/library/src/content/page.rs
@@ -82,12 +82,11 @@ pub struct Page {
     pub lang: String,
     /// Contains all the translated version of that page
     pub translations: Vec<DefaultKey>,
-    /// Contains the internal links that have an anchor: we can only check the anchor
-    /// after all pages have been built and their ToC compiled. The page itself should exist otherwise
-    /// it would have errored before getting there
-    /// (path to markdown, anchor value)
-    pub internal_links_with_anchors: Vec<(String, String)>,
-    /// Contains the external links that need to be checked
+    /// The list of all internal links (as path to markdown file), with optional anchor fragments.
+    /// We can only check the anchor after all pages have been built and their ToC compiled.
+    /// The page itself should exist otherwise it would have errored before getting there.
+    pub internal_links: Vec<(String, Option<String>)>,
+    /// The list of all links to external webpages. They can be validated by the `link_checker`.
     pub external_links: Vec<String>,
 }
 
@@ -262,7 +261,7 @@ impl Page {
         self.content = res.body;
         self.toc = res.toc;
         self.external_links = res.external_links;
-        self.internal_links_with_anchors = res.internal_links_with_anchors;
+        self.internal_links = res.internal_links;
 
         Ok(())
     }
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -3,16 +3,21 @@ use rayon::prelude::*;
 use crate::Site;
 use errors::{Error, ErrorKind, Result};
 
-/// Very similar to check_external_links but can't be merged as far as I can see since we always
-/// want to check the internal links but only the external in zola check :/
+/// Check whether all internal links pointing to explicit anchor fragments are valid.
+///
+/// This is very similar to `check_external_links`, although internal links checking
+/// is always performed (while external ones only conditionally in `zola check`).
 pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
+    println!("Checking all internal links with anchors.");
     let library = site.library.write().expect("Get lock for check_internal_links_with_anchors");
+
+    // Chain all internal links, from both sections and pages.
     let page_links = library
         .pages()
         .values()
         .map(|p| {
             let path = &p.file.path;
-            p.internal_links_with_anchors.iter().map(move |l| (path.clone(), l))
+            p.internal_links.iter().map(move |l| (path.clone(), l))
         })
         .flatten();
     let section_links = library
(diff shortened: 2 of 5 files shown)


components/library/src/content/page.rs
```diff
@@ -82,13 +82,12 @@ pub struct Page {
...
-    /// (path to markdown, anchor value)
-    pub internal_links_with_anchors: Vec<(String, String)>,
-    /// Contains the external links that need to be checked
-    pub external_links: Vec<String>,
+    /// The list of all internal links (as path to markdown file), with optional anchor fragments.
+    /// We can only check the anchor after all pages have been built and their ToC compiled.
+    /// The page itself should exist otherwise it would have errored before getting there.
+    pub internal_links: Vec<(String, Option<String>)>,
+    /// The list of all links to external webpages. They can be validated by the `link_checker`.
+    pub external_web_links: Vec<String>,
```
Keats: why did you change that name?

components/site/src/link_checking.rs
```diff
@@ -3,84 +3,68 @@ use rayon::prelude::*;
...
+        .inspect(|_| anchors_total = anchors_total.saturating_add(1));
+
+    // Check for targets existence (including anchors), then keep only the faulty
+    // entries for error reporting purposes.
+    let missing_targets = links_with_anchors.filter(|(_, md_path, anchor)| {
+        // There are a few `expect` here since the presence of the .md file will
+        // already have been checked in the markdown rendering
+        let mut full_path = site.base_path.clone();
+        full_path.push("content");
+        for part in md_path.split('/') {
```
Keats, line 50: is that causing issues on Windows?

components/utils/src/site.rs
```diff
@@ -14,12 +14,15 @@ pub fn get_reading_analytics(content: &str) -> (usize, usize) {
...
 
+/// Result of a successful resolution of an internal link.
 #[derive(Debug, PartialEq, Clone)]
 pub struct ResolvedInternalLink {
+    /// Resolved link target, as absolute URL address.
     pub permalink: String,
-    // The 2 fields below are only set when there is an anchor
-    // as we will need that to check if it exists after the markdown rendering is done
-    pub md_path: Option<String>,
+    /// Internal path to the .md file, without the leading `@/`.
```
Keats, line 22: is that path normalised or not?

Keats (changes requested):

components/library/src/content/page.rs
lucab: While going through the existing logic, I realized there is a filtering step applied when building this collection, so that it effectively only holds HTTP(S) links.
I was mildly surprised that this field is only a subset of "all the external links in the page", thus I aligned the name and docstrings to mention web/webpages (as noted in the commit message).

This rename is split in its own commit, so it can be quickly dropped if you prefer so.

components/utils/src/site.rs
lucab: I don't know what is the definition of "normalised" here, so I'm not sure. The logic to compute this value is the same as before (minus the `Option` wrapper), I only added a docstring to the best of my understanding.
If you have a specific suggestion here to make it clearer, I'll be happy to update.

components/site/src/link_checking.rs
lucab: Likely not? This is not newly added logic (see L45 of existing codebase), so if it didn't get any bug reports before I assume it does work on Windows too (but I don't have a machine to test it manually).

To the best of my understanding, Zola does not require internal links to be written differently according to the platform (i.e. `@/foo/bar.md` on Mac, and `@/foo\bar.md` on Windows). Thus I guess `md_path.split('/')` would work everywhere.

bemyak: Could [Path iter](https://doc.rust-lang.org/std/path/struct.Path.html#method.iter) be used here?

components/library/src/content/page.rs
Keats: Let's drop that change then, I don't think anyone expects phone numbers etc to be part of that list.

components/utils/src/site.rs
Keats: That's fine I got confused whether it was the on disk path or the in md path!

components/site/src/link_checking.rs
lucab: As far as I can tell this isn't a filesystem path (i.e. the separator does not change according to the runtime OS), so `Path` and its components splitting/iterating logic won't be the correct one to use here.

lucab:
@Keats thanks for the feedback! I've amended the PR according to comment, it's ready for another round of reviews.

components/site/src/link_checking.rs
```diff
@@ -3,84 +3,68 @@ use rayon::prelude::*;
...
-            "> Checked {} internal link(s) with an anchor: {} error(s) found.",
-            all_links.len(),
-            errors.len()
-        );
-    }
-
-    if errors.is_empty() {
-        return Ok(());
-    }
+        .inspect(|_| anchors_total = anchors_total.saturating_add(1));
```
Keats, line 41: TIL about `inspect`. Do we expect to have more than u64::MAX_SIZE anchors?
Is it not as simple as doing `links_with_anchors.len()`?

lucab: > Do we expect to have more than u64::MAX_SIZE anchors?

Quite unlikely. The `saturating_add` only makes the arithmetic explicit. If it feels un-ergonomic in this codebase, it can be changed to a `+= 1`, whose behavior depends on build-time configuration (however neither panic-ing nor overflowing is really useful here IMHO).

> Is it not as simple as doing links_with_anchors.len()?

Not really, mostly because `links_with_anchors` does not exist as a collection. If you look at the rest of the changes to this function, you'll notice that it used to iterate multiple times over the input and collect into two intermediate vectors.
This instead builds a single iterator and performs a single collection at the end, however it has to count elements on the fly.

Keats:
Thanks!

Keats:
@lucab are you planning to work on adding the backlinks feature or should I handle it?

lucab:
@Keats feel free to tackle it, unfortunately I think I won't have time to dig into that in the very near future. Thanks for your ongoing work!

Keats:
No worries, did you have something specific in mind? I was thinking of each page/section having a `backlinks` property that is `Vec<Backlink {title, permalink}>`. Anything on top?

lucab:
@Keats I didn't keep notes when I started looking at this, but I seem to remember these (confused) details:
 * having to look at shortcodes to retrieve the backlinks (possibly because the full site library needs to be populated first?)
 * trying to have the user-facing API exposing something like `Array<Page>`
 * whether it was possible to query backlinks for any other arbitrary pages (I think so)
 * being able to pick apart "backlinks from pages" and "backlinks from sections" in the user-facing API (similar to `get_taxonomy` maybe?)
 * not finding a solid answer about how backlinks and translations could mix. I thought it would be nice to let the user decide whether to split or not split backlinks according to languages. I didn't dig too much into this topic as there was a full i18n rework ongoing at the time.

Issue #1359, opened 2021-02-16
`zola serve` fails to serve directories under /static/


jli:
# Bug Report

## Environment

Zola version: 0.13.0

## Expected Behavior

I have a subdirectory in the `static/` directory in Zola:
```
static
└── artifact1
    └── index.html
```

When I run `zola serve` and then request `http://localhost:1111/artifact1/`, I expect Zola to serve `static/artifact1/index.html`.

## Current Behavior

When I request `http://localhost:1111/artifact1`, Chrome gives `ERR_EMPTY_RESPONSE` and I see the following error from `zola serve`:

```
thread '<unnamed>' panicked at 'Is a directory (os error 21)', src/cmd/serve.rs:123:18
stack backtrace:
   0: _rust_begin_unwind
   1: std::panicking::begin_panic_fmt
   2: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   3: <hyper::server::conn::upgrades::UpgradeableConnection<I,S,E> as core::future::future::Future>::poll
   4: <hyper::server::conn::spawn_all::NewSvcTask<I,N,S,E,W> as core::future::future::Future>::poll
   5: tokio::runtime::task::raw::poll
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I'm guessing the issue is here:  https://github.com/getzola/zola/blob/v0.13.0/src/cmd/serve.rs#L109

```rust
    // Remove the trailing slash from the request path
    // otherwise `PathBuf` will interpret it as an absolute path
    root.push(&req.uri().path()[1..]);
```

## Step to reproduce

Set up the static directory:
```
$ zola init .
$ mkdir static/artifact1
$ echo 'hello' > static/artifact1/index.html
$ zola serve
```

Then, request `http://localhost:1111/artifact1/`

kubukoz:
Was this solved in #1426?

Keats:
It should be yes

PR #1426, opened 2021-04-04
Issue/1359

bneijt:
If it's a directory that `zola serve` tries to read, make sure we try `/index.html` instead of crashing.

This will fix issue #1359 

I have not yet written a lot of Rust so the solution is probably a bit ugly, feel free to edit and update.


Keats (approved):

Keats:
Thanks!

PR #1432, opened 2021-04-16
doc(installation): add Debian package

barnumbirr:
Hello,

just adding [a Debian package I've created](https://github.com/barnumbirr/zola-debian) to the documentation.

Cheers.

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -30,6 +30,15 @@ Zola is available in the official Alpine Linux repository, only on the `edge` ve
 $ apk add zola --repository http://dl-cdn.alpinelinux.org/alpine/edge/community/
 ```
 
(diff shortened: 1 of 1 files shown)


Keats:
Is there something like AUR for debian? Or is it relying on individuals repos?

barnumbirr:
There unfortunately isn't, although the idea has been brought up before ([here for example](https://lists.debian.org/debian-devel/2013/05/msg00131.html)).
There's a few solutions:
- Use the provided package as is, relying on a SPOF (me 🙈) to match Zola's release cycle
- Get an official Zola repository through services like [packagecloud.io](https://packagecloud.io/)
- Try and get the provided package into the official Debian repositories. Would require a Debian Developer's sponsorship.

Keats:
Ok let's go with the github repo for now! Thanks

PR #1452, opened 2021-05-09
Add new sorting: UpdateDate that sorts by `meta.updated`.

### LINKED ISSUE #1384: Feature Request: sort section by "updated"
Hello,

https://www.getzola.org/documentation/content/section/#sorting describes how to sort by "date" within a section. pages also have an "[updated](https://www.getzola.org/documentation/content/page/#front-matter)" variable in the front matter that I'd like to sort my section by. Would it be difficult to add that as an option?

WaffleLapkin:
This resolves #1384.

I don't know where to add tests/update documentation (it's my first time contributing to zola) so I didn't.

-----

Sanity check:

## Code changes

### DIFF
--- a/components/front_matter/src/lib.rs
+++ b/components/front_matter/src/lib.rs
@@ -48,6 +48,8 @@ impl RawFrontMatter<'_> {
 pub enum SortBy {
     /// Most recent to oldest
     Date,
+    /// Most recent to oldest
+    UpdateDate,
     /// Sort by title
     Title,
     /// Lower weight comes first
--- a/components/front_matter/src/page.rs
+++ b/components/front_matter/src/page.rs
@@ -20,6 +20,12 @@ pub struct PageFrontMatter {
     /// Updated date
     #[serde(default, deserialize_with = "from_toml_datetime")]
     pub updated: Option<String>,
+    /// Chrono converted update datatime
+    #[serde(default, skip_deserializing)]
+    pub updated_datetime: Option<NaiveDateTime>,
+    /// The converted update datetime into a (year, month, day) tuple
+    #[serde(default, skip_deserializing)]
+    pub updated_datetime_tuple: Option<(i32, u32, u32)>,
     /// Date if we want to order pages (ie blog post)
     #[serde(default, deserialize_with = "from_toml_datetime")]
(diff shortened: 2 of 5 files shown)


WaffleLapkin:
Also, the original issue proposes using `max(created, updated)` date, while this PR only uses `created`. Which behaviour is more desired?

Keats:
> Also, the original issue proposes using max(created, updated) date, while this PR only uses created. Which behaviour is more desired?

This would mean that pages that were not updated would be shown at the end in your code right? `max(created, updated)` makes much more sense to me.

WaffleLapkin:
> This would mean that pages that were not updated would be shown at the end in your code right?

I think they won't be shown at all. Anyway, yeah, `max(created, updated)` is much more reasonable. I'll fix this.

Keats:
Looks good, can you add some tests for it?
Also what are your thoughts on https://zola.discourse.group/t/remove-lighter-heavier-and-earlier-later/767 ?

WaffleLapkin:
> Looks good, can you add some tests for it?

Sure, can you point me to where I should add them? I'm kinda lost 😅

> Also what are your thoughts on https://zola.discourse.group/t/remove-lighter-heavier-and-earlier-later/767 ?

Hm, we only set one pair of `lighter/heavier`, `earlier/later`, … at a time, right? If so, then having only `prev/next` (or other neutral names, it's not the point) makes perfect sense to me.

Keats:
I thought `library.rs` had tests for that but it doesn't seem that way, don't worry about it then I'll write some myself later.

Keats:
> Hm, we only set one pair of lighter/heavier, earlier/later, … at a time, right? If so, then having only prev/next (or other neutral names, it's not the point) makes perfect sense to me.

Yep. We lose a bit of expliciteness but we only keep one name around rather than `n` for each sorting methods (and it's becoming hard to come up with them)

WaffleLapkin:
The only question is: how to move to `prev/next` without breaking changes?

Keats:
That would have to be a breaking change.

WaffleLapkin:
I'm too used to people that are scared almost to death about breaking changes :D

PR #1453, opened 2021-05-10
Hide lines in code block

mockersf:
https://zola.discourse.group/t/hide-lines-from-code-block/898

Adds an annotation to code block `hide_lines` that accept ranges of lines to hide. Most of code is copied from `hl_lines`.

Sanity check:

## Code changes

### DIFF
--- a/components/rendering/tests/codeblock_hide_lines.rs
+++ b/components/rendering/tests/codeblock_hide_lines.rs
@@ -7,14 +7,8 @@ use front_matter::InsertAnchor;
 use rendering::{render_content, RenderContext};
 
 macro_rules! colored_html_line {
-    ( @no $s:expr ) => {{
+    ( $s:expr ) => {{
         let mut result = "<span style=\"color:#c0c5ce;\">".to_string();
-        result.push_str($s);
-        result.push_str("\n</span>");
-        result
-    }};
-    ( @hl $s:expr ) => {{
-        let mut result = "<span style=\"background-color:#65737e30;color:#c0c5ce;\">".to_string();
         result.push_str($s);
         result.push_str("\n</span>");
         result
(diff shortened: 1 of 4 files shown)


components/rendering/tests/codeblock_hide_lines.rs
```diff
@@ -0,0 +1,67 @@
...
+        "",
+        &permalinks_ctx,
+        InsertAnchor::None,
+    );
+    let res = render_content(
+        r#"
+```hide_lines=2
+foo
+bar
+bar
```
Keats: It would be better as a test to not have `bar` twice since then you can't be sure which one was removed.

components/rendering/tests/codeblock_hide_lines.rs
```diff
@@ -0,0 +1,67 @@
+use std::collections::HashMap;
+
+use tera::Tera;
+
+use config::Config;
+use front_matter::InsertAnchor;
+use rendering::{render_content, RenderContext};
+
+macro_rules! colored_html_line {
+    ( @no $s:expr ) => {{
```
Keats: what does `no` means here?

Keats (changes requested):

components/rendering/tests/codeblock_hide_lines.rs
mockersf: it means I copied the test from hl_lines and didn't read enough what it did

mockersf:
CI on windows failed to install Rust...

I changed how to handle `hide_lines` to share more code with `hl_lines`

Keats:
> CI on windows failed to install Rust...

Yeah it started to happen super often after upgrading to some more recent Rust version :(
I need to move to Github Actions but the libsass build system is a real pain.

components/rendering/tests/codeblock_hide_lines.rs
Keats: Ah sorry, I completely forgot that code

Keats:
Thanks!

PR #1456, opened 2021-05-13
Fix next/previous page links when using transparent sections

scouten:
## Code changes

There is a race condition in the generation of next/previous page links when combined with sections that use `transparent = true`. I use next/previous generically to all manner of sorting (date, title, weight).

The next/previous links are generated per-section, but the sections themselves are iterated in an apparently arbitrary sequence. A parent section may be iterated before or after its child. For each page, the last section which _might_ contain it wins, which means the next/previous links _may or may not_ include page from sibling sections.

This change disables the generation of next/previous page references for sections with `transparent = true`, meaning that larger list of pages available to the (grand)parent section will be the only (sorted) list used to generate next/previous references.

I believe this is in keeping with the intent of the transparent flag, as quoted from documentation here:

```
# If set to "true", the section will pass its pages on to the parent section. Defaults to `false`.
# Useful when the section shouldn't split up the parent section, like
# sections for each year under a posts section.
transparent = false
```

* [ ] Have you created/updated the relevant documentation page(s)? _n/a - no new API created_

### DIFF
--- a/components/library/src/library.rs
+++ b/components/library/src/library.rs
@@ -319,28 +319,38 @@ impl Library {
         }
 
         for (key, (sorted, cannot_be_sorted, sort_by)) in updates {
-            // Find sibling between sorted pages first
-            let with_siblings = find_siblings(&sorted);
+            let section_is_transparent = if let Some(section) = self.sections.get(key) {
+                section.meta.transparent
+            } else {
+                false
+            };
 
-            for (k2, val1, val2) in with_siblings {
-                if let Some(page) = self.pages.get_mut(k2) {
-                    match sort_by {
-                        SortBy::Date => {
-                            page.earlier = val2;
-                            page.later = val1;
+            if !section_is_transparent {
+                // Find sibling between sorted pages first
+                let with_siblings = find_siblings(&sorted);
+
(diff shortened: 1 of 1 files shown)


Keats:
Yep that makes sense. Considering that it is a race condition, I guess it's hard to write a test for it?

scouten:
@Keats do you have an example of a similar test that I could follow?

scouten:
Given that I _think_ I'm fixing a race condition, it should in theory be feasible to write a stable test.

Keats:
Don't worry about it unless it takes 5 minutes

scouten:
I'm fast, but maybe not _that_ fast … 😆

Keats:
ahah I'll leave it open for a few days and then I'll merge it. Thanks again!

WaffleLapkin:
This pr is having conflicts because of #1452, so now it has to be slightly adjusted :^)

scouten:
No worries. I’ll take a look later today or tomorrow.

scouten:
@WaffleLapkin conflict with #1452 resolved (easy merge conflict). It passes my local (manual) test.

Keats:
Thanks!

PR #1460, opened 2021-05-16
Add `num_format` filter

### LINKED ISSUE #1459: [Feature Request] Add built-in filter for formatting numbers
This may be a somewhat niche use case, but it would be nice to be able to format numbers that are generated dynamically. Some background context [here](https://maxdeviant.com/shards/2021/how-hard-is-it-to-format-a-number/).

The way I envision it would be using something like [`num_format`](https://docs.rs/num-format/0.4.0/num_format/) to power a filter that would take an (optional) locale and format the input accordingly:

```mustache
{{ 1000000 | num_format }} <!-- 1,000,000 -->

{{ 1000000 | num_format(locale="fr") }} <!-- 1 000 000 -->
```

Would be happy to open a PR if this is something we want to add.

maxdeviant:
This PR adds a `num_format` filter to Zola for formatting numbers using [`num_format`](https://docs.rs/num-format/0.4.0/num_format/).

This allows for numbers generated dynamically to be displayed using proper formatting:

```mustache
{% set n = 1000 * 1000 %}
{{ n | num_format }}
<!-- 1,000,000 -->
```

`num_format` accepts an optional `locale` argument that can be used to format a number using a specific locale;

```mustache
{% set n = 1000 * 1000 %}
{{ n | num_format(locale="fr") }}
<!-- 1 000 000 -->
```

Resolves #1459.

---

Sanity check:

## Code changes


components/templates/src/filters.rs
```diff
@@ -72,13 +75,32 @@ pub fn base64_decode<S: BuildHasher>(
...
 
+pub fn num_format(value: &Value, args: &HashMap<String, Value>) -> TeraResult<Value> {
+    use num_format::{Locale, ToFormattedString};
+
+    let num = try_get_value!("num_format", "value", i64, value);
+
+    let locale = match args.get("locale") {
+        Some(locale) => try_get_value!("num_format", "locale", String, locale),
+        None => "en".to_string(),
+    };
```
Keats: I think making this field required would be nicer, or pass the config default_language to the filter. No reasons to default to `en` if we don't know anything about the current locale.

components/templates/src/filters.rs
```diff
@@ -201,4 +223,44 @@ mod tests {
...
+
+    #[test]
+    fn num_format_filter_with_locale() {
+        let tests = vec![
+            ("en", 1_000_000, "1,000,000"),
+            ("en-IN", 1_000_000, "10,00,000"),
+            // Note:
+            // U+202F is the "NARROW NO-BREAK SPACE" code point.
+            // When displayed to the screen, it looks like a space.
+            ("fr", 1_000_000, "1\u{202f}000\u{202f}000"),
```
Keats, line 265: Why not just use a space?

Keats (changes requested):
I'll need to figure out how it works with the other i18n locale functions so I'll keep it un-merged for now and will merge when I have a clearer idea.

components/templates/src/filters.rs
maxdeviant: @Keats This is coming from `num_format`. I think it's so you don't have numbers wrapping across line boundaries when they're split.

crepererum: Apart from "it doesn't line break" it's also narrower than a normal inter-word space. These are typographical details which in sum will give the published blog / website a more professional look. And since the Unicode character is standardized, all user agent can be expected to deal with it (even things like TTS).

components/templates/src/filters.rs
maxdeviant: I opted to make it default to the `default_language` in the config, as I think it feels more ergonomic than having to pass a locale every time.

Keats:
So that looks ok to me minus the small conflict, I can resolve it myself if you prefer.

maxdeviant:
> So that looks ok to me minus the small conflict, I can resolve it myself if you prefer.

Feel free to resolve, if it's not too much trouble for you 👍🏻

Keats (approved):

Keats:
Ah I thought I could just use the GH UI >_> that's going to be for another day then

maxdeviant:
@Keats I fixed the merge conflict. Not sure if I need to regenerate `Cargo.lock` as well?

Keats:
I'll do that, I mess up the tests a bit yesterday evening. Thanks!

PR #1464, opened 2021-05-20
Specify default output directory

rootkea:


src/cli.rs
```diff
@@ -46,6 +46,7 @@ pub fn build_cli() -> App<'static, 'static> {
                         .short("o")
                         .long("output-dir")
                         .takes_value(true)
+                        .default_value("$PROJECT_ROOT/public")
```
rootkea: Or may be it should be `project_root/public` as `$PROJECT_ROOT` may lead to a confusion of `PROJECT_ROOT` being an acceptable variable by zola?

src/cli.rs
```diff
@@ -46,6 +46,7 @@ pub fn build_cli() -> App<'static, 'static> {
...
                         .help("Outputs the generated site in the given path"),
```
rootkea: Or may be a change like this:
```
.help("Outputs the generated site in the given path (by default directory named public in project root)"),
```

rootkea:
Also a similar change for `serve -o`

src/cli.rs
Keats: I would prefer that change

rootkea: I made the suggested change further shortening the string.

src/cli.rs
```diff
@@ -69,7 +69,7 @@ pub fn build_cli() -> App<'static, 'static> {
                         .short("o")
                         .long("output-dir")
                         .takes_value(true)
-                        .help("Outputs the generated site in the given path"),
+                        .help("Outputs the generated site in the given path (by default 'public' dir in project root)"),
```
Keats, line 72: Thinking about that, `zola serve` doesn't actually outputs the generated site, only the assets. The HTML is kept in memory. I'll tweak the message later.

Keats (approved):
Thanks!

src/cli.rs
rootkea: Just leaving a note here: when you'll tweak the message for `--output-dir` in `zola serve`, also mention the same on https://www.getzola.org/documentation/getting-started/cli-usage/#serve

Currently it says nothing about output_dir only containing the assets and HTML being in memory.

Issue #515, opened 2018-11-05
Allow shortcodes in shortcode bodies


ghost:
If a short-code with a body is used in the body of a short-code it's `{% end %}` is matched with the outer short-code.  This prevents the use of nested short-codes with bodies.

### Use Case

I'm trying to write [`reveal.js`] presentations in Gutenberg currently and I have defined simple shortcodes for vertical slides, fragments, and notes:

```html
<aside class="notes">
{{ body | safe | trim | markdown }}
</aside>
```

```html
<section{%- if id %} id="{{id}}" {%- endif -%}
        {%- if attributes -%}
          {%- for attr in attributes %}{#
            #} {{attr.attribute}}="{{attr.value}}"
          {%- endfor -%}
        {%- endif -%}>
{{ body | safe | markdown(inline=true) }}
</section>
```

```html
<span class="fragment">
    {{ body | safe | trim | markdown }}
</span>
```

This makes it easier to write slides like so:

```md
+++
title = "Installation instructions"
+++

{% vertical_slide() %}
Main content

{% notes() %}
- notes on main content
{% end %}
{% end %}


{% vertical_slide() %}
Basement slide with extra content

{% notes() %}
- notes on basement content
{% end %}
{% end %}
```

### Workaround

Of course I can just throw in html to work around this, but I find short-codes are more pleasant to work with, especially when I have more complicated things going on like figures with complex markup that I've created with short-codes showing up in vertical slides.

For example:

```md
{% vertical_slide() %}
**Linux** (Ubuntu/Debian)

`sudo apt-get install git`

<figure class="toggle-figure">
    <span class="toggle-figure__button"></span>
    <img class="toggle-figure__figure" alt="`sudo apt-get install git`" src="img/gif/linux-install-git.gif"/>
</figure>

<aside class="notes">
- Linux typically comes with `git` installed
- `git` was created specifically to manage the Linux project
</aside>
{% end %}
```

However, this isn't ideal since the list inside the aside _is not rendered as markdown_.

[`reveal.js`]: https://revealjs.com/#/

Keats:
Having shortcodes inside shortcodes is very unlikely to happen imo, it doesn't feel great and would probably be a bit messy since there are already issues with the shortcodes <> markdown interactions in some cases (https://github.com/Keats/gutenberg/issues/479)

If you have time you could write the code and I will have a look ('m not promising to merge it just to be clear).

Keats:
[...]

PR #1475, opened 2021-05-22
Allow recursively nested shortcodes

meithecatte:
This PR modifies how the parser will parse the following markdown:

```
{% first() %}

  {% second() %}

  {% end %}

{% end %}
```

Previously, this would parse as the shortcode `first` with `{% second() %}` as the body, all followed by the text `{% end %}`. Now it nests properly.

Together with the already-merged PR #1358, this fixes issue #515.


Keats:
Thanks! Can you add some tests?

meithecatte:
I'm not sure what's the best way to test this. I tried the following:
```rust
    #[test]
    fn recursive_shortcodes() {
        let mut tera = Tera::default();
        let md = MarkdownFilter::new(PathBuf::new(), Config::default(), HashMap::new()).unwrap();
        tera.register_filter("markdown", md);
        tera.add_raw_template("shortcodes/outer.html", "This is the outer shortcode. {{ body | markdown | safe }}").unwrap();
        tera.add_raw_template("shortcodes/inner.html", "We're inside! {{ body | markdown | safe }}").unwrap();
        let res = render_shortcodes("{% outer() %} One! {% inner() %} Two! {% end %} Three! {% end %} Fin.", &tera);
        assert_eq!(res, "TODO");
    }
```

This fails for the same reason #1358 was necessary - the markdown filter as constructed here cannot see the `inner` filter. I don't see how to mock the filesystem like this so it affects `MarkdownFilter` :/

Keats:
You can probably skip the `markdown` filter in the test and have only the shortcodes for the test

meithecatte:
Okay, I've added a test.

meithecatte:
Did you have a chance to look at this?

Keats:
Not yet sorry, I will have a look this week.

components/rendering/src/shortcode.rs
```diff
@@ -479,4 +479,12 @@ Some body {{ hello() }}{%/* end */%}"#,
         let res = render_shortcodes("{{ youtube() }}", &tera);
         assert_eq!(res, "* 1\n* 2\n* 3");
     }
+
+    #[test]
+    fn recursive_shortcodes() {
+        let mut tera = Tera::default();
+        tera.add_raw_template("shortcodes/outer.html", "This is the outer shortcode {{ body }} outer ends here").unwrap();
+        let res = render_shortcodes("{% outer() %} One! {% inner() %} Two! {% end %} Three! {% end %} Fin.", &tera);
+        assert_eq!(res, "<pre data-shortcode>This is the outer shortcode One! {% inner() %} Two! {% end %} Three! outer ends here</pre> Fin.");
```
Keats: I would have expected that to error since `inner` is not defined and we only render shortcodes once so `render_shortcode` should handle the recursion

Keats (changes requested):

meithecatte:
The inner shortcode will be rendered if the user passes the body through
the `markdown` filter.

Keats:
Ok had another look.
I think the test should create the temp dir and the shortcode files so the markdown filter is aware of it so we can test the actual behaviour.
To create temp files, see https://github.com/getzola/zola/blob/next/components/templates/src/global_fns/images.rs#L141-L150 and just pass the `dir` as base path to the MarkdownFilter, which will load the Tera templates.

Sorry for the confusion

Keats:
@NieDzejkob ping

meithecatte:
Thanks for the ping! I edited the test, but it still fails — the inner call to `markdown` doesn't see the filter. I don't exactly see why that is. Could you take a look?

Keats:
The issue is:
```
        let tera = load_tera(&path, &config).map_err(tera::Error::msg)?;
```
 in the `MarkdownFilter::new`. This means it essentially creates a new instance of Tera which will have the shortcodes but this instance will not have the markdown filter since it hasn't been registered.

I don't know what to do about that tbh.

meithecatte:
Perhaps the markdown filter shouldn't create its own instance of Tera at all? The cleanest solution would be to make `tera::Filter::filter` take a reference to the `Tera` instance, though if that's unacceptable, we could make the `Tera` variable in `struct Site` an `Arc<Tera>`, and keep a `Weak<Tera>` in the filter.

Keats:
> The cleanest solution would be to make `tera::Filter::filter` take a reference to the `Tera` instance

That would have to be for Tera v2 so not anytime soon.

We can also keep Arc<Tera> in the filter no?

meithecatte:
> We can also keep Arc in the filter no?

AFAICS, keeping a full `Arc` in the filter (as opposed to `Weak`) would create a cycle and leak memory. I'm not sure whether a new instance of Tera gets created for each rebuild in a `zola serve` session, but if not, this cycle would indeed be harmless, and we could use an `Arc` directly.

Keats:
> I'm not sure whether a new instance of Tera gets created for each rebuild in a zola serve session, but if not, this cycle would indeed be harmless, and we could use an Arc directly.

On template changes other than shortcodes, it only re-render the pages but re-uses the instance. On any other change content/shortcode change, it will rebuild completely the `Site` struct from scratch which contains the Tera instance.
Now that you mention it, I'm using `Arc` in a couple of functions and I'm wondering if https://github.com/getzola/zola/issues/1515 is a symptom.

Keats:
Closing as it will be addressed by #1564 in the end. Thanks though @NieDzejkob !

Issue #1197, opened 2020-10-05
Need examples of code using each named highlight scheme


tbrowder:
# Documentation issue

There is a long list of named highlight schemes but the links for some of them don't seem to work. In particular, the following cannot be seen:

+ base16-ocean-light
+ base16-ocean-dark

Keats:
Looks like the site is not working with spaces in names anymore? It definitely used to :/ and it doesn't seem maintained anymore...

rootkea:
It's been almost 5 years now since upstream has removed `base16-ocean-light` and `base16-ocean-dark` themes. https://github.com/chriskempson/base16-textmate/commit/d5e178c2780c36bc62bf34db9131df092bfc0d27

There have been `base16-ocean` and `base16-oceanicnext` along with other base16-textmate themes though. See https://github.com/chriskempson/base16-textmate/tree/master/Themes

We should remove these 2 defunct themes and also select some other upstream-supported theme instead of `highlight_theme = "base16-ocean-dark"` in config.toml docs https://www.getzola.org/documentation/getting-started/configuration/

Keats:
When https://github.com/getzola/zola/pull/1242 is merged in it should be easy to build a page to preview them.
Let's update those themes

rootkea:
Does Zola support Sublime Text 3 themes too? Or just v2 themes? v2 themes were in xml format while v3 themes are in json format so may be not?

In that case this sentence needs to be updated on https://www.getzola.org/documentation/getting-started/configuration/

> Zola uses the Sublime Text themes, making it very easy to add more.

I'm following up on the current list and found that some themes have moved to v3 format and have removed the v2 ones from upstream.

Keats:
It should work for both but I'm not 100% sure

rootkea:
So how should I check that?

I think I need to:
1. drop v3 theme file in sublime/themes directory
2. How to add to binary file `all.themedump`?
3. cargo build
4. use freshly built zola binary e..g `zola serve`

Is that correct? I'm not sure how to go about step 2 though...

Keats:
See https://github.com/getzola/zola/blob/master/CONTRIBUTING.md#adding-a-theme for step 2

rootkea:
Oops! Sorry, I should have read the CONTRIBUTING.md. Thanks!

rootkea:
`generate_sublime` didn't add v3 theme files to `all.themedump`. So I guess Zola supports only Sublime Text 2 themes and not v3 themes.

[...]

PR #1482, opened 2021-05-23
Docs: Unset highlight_theme variable in sample config.toml

rootkea:
Do not set `highlight_theme` like `theme`. See https://github.com/getzola/zola/issues/1197#issuecomment-846606390

### DIFF
--- a/docs/content/documentation/getting-started/configuration.md
+++ b/docs/content/documentation/getting-started/configuration.md
@@ -107,7 +107,7 @@ highlight_code = false
 
 # The theme to use for code highlighting.
 # See below for list of allowed values.
-highlight_theme = "base16-ocean-dark"
+highlight_theme = ""
 
 # When set to "true", emoji aliases translated to their corresponding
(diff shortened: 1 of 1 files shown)


Keats:
I would still default to a theme, you want something that works on the first try

rootkea:
We are talking about the sample config.toml in docs https://www.getzola.org/documentation/getting-started/configuration/
Besides, it has `highlight_code = false` so highlighting is off by default already.

Also, the immediate previous line says "# See below for list of allowed values." so user wouldn't be lost for what to put in for `highlight_theme` if she decides to set `highlight_code` true.

Also, there is the larger question of which highlight theme to select in docs. The current highlight theme `base16-ocean-dark` has been removed from upstream a long back.

rootkea:
My bad!

My previous comment was all mixed up and didn't consider `base16-ocean-dark` being default for `highlight_theme` in code too. So deleted it. I'll get back to this discussion later.

rootkea:
I'll get more clarity once #1242 gets merged in so closing this one for now. Related #1197

Issue #788, opened 2019-08-27
Can't get directory path to collocated assets when using YYYY-MM-DD_ page prefix


zdenek-crha:
# Bug Report

## Environment

Zola version: 0.8.0
Features used:
* shortcode
* collocated assets for page
* `YYYY-MM-DD_` prefix for page
* `resize_image()` function

## Setup

Directory structure like this:
```
content/section/2000-01-01_page/index.md
content/section/2000-01-01_page/image1.jpg
content/section/2000-01-01_page/image2.jpg
```

Attempt to create shortcode that would convert this (in `index.md`):
```md
{{ render_image(path="image1.jpg", width=100, height=200) }}
```
Into this:
```html
<div class="image">
<a href="/section/page/image1.jpg"><img src="static/processed_images/foo.jpg"></a>
</div>
```

## Expected Behavior

I've expected that when I pass relative asset name to shortcode, I can use `page` from Tera context to compute both:

* path to image source to pass into `resize_image()`
* path to deployed image to pass into link

## Current Behavior

The `page.path` points to path used in `public` directory (`section/page/` - ie without prefix). It can't be used to construct path for asset source location needed by `resize_image()`.

The `page.relative_path` points `section/2000-01-01_page/index.md` file in _correct_ - directory. But points to *file* not directory and can't be used for computing source path for `render_image()`

_Note: May be it could be use by sequence of `split`, remove last element, `join`, ... operations,  but I would not consider that very user friendly, especially for beginners :-)_

## Proposed solution

I can see two solutions:

1) Add `page.relative_dir` that would point to post directory (similar to `page.relative_path`)
2) Add `dirname`, `basename`, `file_path` filters to Tera to allow dissection and concatenation of paths

Personally, I find the second approach more desirable as it would benefit to Tera too and would be usable in wider range of applications, instead of fixing just one corner-case

## Workaround

I've resorted into passing `directory` and `file` into shortcode separately:
```md
{{ render_image(dir="section/2000-01-01_page", "file="image1.jpg", width=100, height=200) }}
```

And shortcode:
```html
<div>
    {% set uri  = page.permalink ~ "/" ~ file %}
    {% set path = dir ~ "/" ~ file %}
    <a href="{{ uri }}"><img src="{{ resize_image(path=path, ....) }}" /></a>
</div>
```

[...]

PR #1484, opened 2021-05-23
imageproc things for paths refactor

vojtechkral:
Let me know if this makes sense & is decent enough.
I'm not sure my code style is in line with the project, I can fix this if need be...


components/site/tests/site.rs
```diff
@@ -767,6 +770,91 @@ fn can_get_hash_for_static_files() {
...
+        height: 380,
+        format: Some("png".into()),
+    });
+    assert_eq!(content[1], ImageMetaResponse {
+        width: 300,
+        height: 380,
+        format: Some("jpg".into()),
+    });
+}
+
```
Keats: It all looks good to me except that - can we somehow move the tests in the imageproc module? This component already so many random tests that it's a bit hard to follow

vojtechkral: Sure! Will do :slightly_smiling_face:

vojtechkral: @Keats Moved the `imageproc` tests into its crate.
(But tbh it's kind of late over here and I should probably double-check everything tomorrow :slightly_smiling_face:)

components/imageproc/src/lib.rs
```diff
@@ -23,9 +28,37 @@ lazy_static! {
...
+        let size = reader.into_dimensions()?;
+
+        Ok(Self { size, format })
+    }
+
+    fn is_lossy(&self) -> bool {
+        use ImageFormat::*;
+
+        // We assume lossy by default / if unknown format
+        match self.format.unwrap_or(Jpeg) {
```
Keats: I think clippy will tell you to use `matches!` here

components/imageproc/tests/test_site.rs
```diff
@@ -0,0 +1,227 @@
...
+    date_prefix: HashMap<String, json::Value>,
+}
+
+impl TestSite {
+    fn read_json<T: DeserializeOwned>(site: &Site, path: &str) -> T {
+        let content = fs::read_to_string(site.output_path.join(path)).unwrap();
+        json::from_str(&content).unwrap()
+    }
+
+    fn build() -> Self {
```
Keats: Can we actually do the test calling only things from the imageproc module or is it too complicated?

vojtechkral: Probably, yeah, most of it anyway. Will try to do that. Except for the date-prefix test, that one requires an actual site. Would it be ok to put that one into `site` tests?

Keats: Why is the prefix needing a special test?

vojtechkral: My idea was to make sure #788 is fixed, but no pressure, I don't have to include it...

Keats: It's easy to test in the shortcode tests (and I might have already added one?) so let's leave it out there

vojtechkral: ok :slightly_smiling_face:

Keats: I think that + fixing CI and it's good to go!

vojtechkral: Allright, I'll try to update later today...

vojtechkral: Made some progress but not complete enough to update, brb tomorrow...

Keats: Don't worry, take your time

vojtechkral:
@Keats The PR is updated, imageproc integration tests are now site-less.
There's still a few bits which I'd like to improve, such as removing hashes from the templates image fns tests, adding a hash stability test, and improving docs, but if you'd be ok with that I'd do that as part of another PR so that we don't get mission creep in this one...

Keats:
>  if you'd be ok with that I'd do that as part of another PR so that we don't get mission creep in this one...

Definitely, let's get all those path changes into `next` as they are starting to be chunky.

components/imageproc/src/lib.rs
```diff
@@ -23,9 +29,34 @@ lazy_static! {
...
+impl ImageMeta {
+    fn read(path: &Path) -> ImageResult<Self> {
+        let reader = ImgReader::open(path).and_then(ImgReader::with_guessed_format)?;
+        let format = reader.format();
+        let size = reader.into_dimensions()?;
+
+        Ok(Self { size, format })
+    }
+
+    fn is_lossy(&self) -> bool {
```
Keats, line 48: Interesting, I thought `image` would have something like that already

components/imageproc/src/lib.rs
```diff
@@ -173,8 +235,8 @@ impl Format {
                 "png" => Some(false),
                 "gif" => Some(false),
                 "bmp" => Some(false),
-                // It is assumed that webp is lossless, but it can be both
-                "webp" => Some(false),
+                // It is assumed that webp is lossy, but it can be both
+                "webp" => Some(true),
```
Keats, line 239: Do we still need that function with `ImageMeta::is_lossy`?

components/imageproc/src/lib.rs
```diff
@@ -459,9 +529,89 @@ impl Processor {
...
+                .map(ImageMetaResponse::from)
+                .ok_or_else(|| Error::msg(format!("Failed to decode WebP image: {}", path.display())))
+        }
+        _ => ImageMeta::read(path).map(ImageMetaResponse::from).map_err(|e| error(e.into())),
+    }
+}
+
+/// Assert that `address` matches `prefix` + RESIZED_FILENAME regex + "." + `extension`,
+/// this is useful in test so that we don't need to hardcode hash, which is annoying.
+pub fn assert_processed_path_matches(path: &str, prefix: &str, extension: &str) {
```
Keats, line 603: can that be moved to tests or with a cfg[test]?

Keats (approved):
I like what I see! I'll merge that now to avoid staleness

components/imageproc/src/lib.rs
vojtechkral: `cfg[test]` can't be used here as integration tests build against a non-test library, iirc. But it could be moved to tests maybe, yeah, I'll try...

Issue #1458, opened 2021-05-15
Hide search-results on blur


benbro:
# Bug Report

## Environment
Online Zola docs

Zola version: Zola version on the official docs

## Expected Behavior
Hide searchresults on input blur.

## Current Behavior
The search-results div isn't hidden on blur.

## Step to reproduce
1. Open https://www.getzola.org/documentation/getting-started/overview/
2. Type "Template" in the search input.
3. Click somewhere to blur the input. search-results div should be hidden.

PR #1488, opened 2021-05-25
hide search-results on blur

mohsen-alizadeh:

Sanity check:

### DIFF
--- a/docs/static/search.js
+++ b/docs/static/search.js
@@ -168,6 +168,12 @@ function initSearch() {
       $searchResultsItems.appendChild(item);
     }
   }, 150));
+
+  window.addEventListener('click', function(e) {
+    if ($searchResults.style.display == "block" && !$searchResults.contains(e.target)) {
+      $searchResults.style.display = "none";
+    }
+  });
 }
 
 


mohsen-alizadeh:
Closes https://github.com/getzola/zola/issues/1458

docs/static/search.js
```diff
@@ -168,6 +168,12 @@ function initSearch() {
       $searchResultsItems.appendChild(item);
     }
   }, 150));
+
+  window.addEventListener('click', function(e) {
+    if ($searchResults.style.display == "block" && !$searchResults.contains(e.target)) {
+      $searchResults.style.display = "none";
+    }
```
Keats, line 175: You need to make sure clicking in the search results div will make it `block` again, otherwise it doesn't pop out anymore

Keats (changes requested):

docs/static/search.js
mohsen-alizadeh: How you can click `search results div` while it is "none" ?

I tested couple of times. it works fine. if you click outside, the `search results div` will be set to `none`, if you click on searchbox and type something again, `search results div` will be set to "block" and will be shown.

mohsen-alizadeh: check this line https://github.com/getzola/zola/blob/4f023d8c7037a01c1b34136ea204acf84d319f59/docs/static/search.js#L152

mohsen-alizadeh: @Keats

Keats (approved):

Issue #746, opened 2019-07-14
How to use subsections in templates


manu-cyber:
# Documentation issue

## Summary

I’m having a hard time figuring out how to use subsections, and the documentations of zola and tera don’t help me much. Also the non-existent `get_section` function is still referenced in some places.

## Detail

I’m trying to build a website with one tree-like, hierarchical section, similar to a wiki for example.

```
$ tree content/
content/
├── _index.md
├── [...]
├── weblog
│   └── […]
└── wiki
    ├── _index.md
    └── software
        └── _index.md
```

In order to navigate through them, I need to make use of the `subsections` array in my `section.html` template.  
The Zola documentation on sections in themes says this:

```
[...]
// Direct subsections to this section, sorted by subsections weight
// This only contains the path to use in the `get_section` Tera function to get
// the actual section object if you need it
subsections: Array<String>;
[...]
```

The `get_section` function doesn’t exist in Tera, and I guess it has been replaced by the `get` function which is documented like this:

```
get
Access a value from an object when the key is not a Tera identifier. Example: {{ sections | get(key="posts/content") }}
```

Now in my `section.html` I would like to have the url and names of a section’s subsections. Those should be associated with the objects I have to get. I try to achieve that like this:

```
{% for subsection in section.subsections %}
  {{ section | get(key=subsection) }<br>
  <a href="{{ subsection | safe }}">some label</a><!-- this doesn't work of course -->
{% endfor %}
```

The build fails with the following error:

```
Failed to build the site
Error: Failed to render section '/home/mgr/create/web/manuelgross.de/content/wiki/_index.md'
Reason: Failed to render 'section.html'
Reason: Filter `get` tried to get key `wiki/software/_index.md` but it wasn't found
```

From the `__tera_context` I can’t see which object I need to call `get` on, and from the available documentations I’m fairly uncertain if I’m using it correctly in the first place.

If someone can confirm to me how it’s supposed to be used or how I otherwise achieve the desired tree-like structure, I’m happy to provide pull-requests for the documentation myself.

## Proposed solution

Add a quick example on how to use subsections and especially how to get the object from the keys in the `subsections` array.  
Or just let me know how it’s meant to be used and I’ll try to improve the documentation myself via pull-requests :)

Keats:
Zola defines a few additional Tera functions like get_section: https://www.getzola.org/documentation/templates/overview/#get-section

The docs definitely need to be improved though

manu-cyber:
[...]

PR #1492, opened 2021-05-26
Docs: weight ties sort sections in random order

rootkea:
Also https://www.getzola.org/documentation/content/section/#sorting has example for how to sort pages.

How about adding an example template for sorting subsections under https://www.getzola.org/documentation/content/section/#sorting-subsections

I could come up with this and it works but I am struggling with the exact words to put into at the moment!
```
{% for section_path in section.subsections %}
{% set section = get_section(path=section_path) %}
<h1><a href="{{ section.permalink | safe }}">{{ section.title }}</a></h1>
{% endfor %}
```


Keats:
let's hold off a bit the example, it's not super efficient currently to do that

rootkea:
Found an old issue #746 looking for the same (subsection example)...

rootkea:
Maybe I should open another issue or we should just push this above example and then later on when things get efficient the example can be refined.

I just don't want this to be forgotten...

PR #1495, opened 2021-05-27
Docs: Don't suggest/use a CSS framework

### LINKED ISSUE #1494: Docs: class-based CSS framework Bulma and Markdown
Hello!

In "Overview" section https://www.getzola.org/documentation/getting-started/overview/ CSS framework Bulma is used in `base.html`.

Bulma is a class-based CSS framework which removes all of the styles of naked HTML elements (e.g. h1, h2 etc) and then uses classes to style those elements.

This makes following md file to display all headings as normal text:
```
Hello
# Foo

## Bar
```

Unaware of this Bulma behavior I spent a considerable time debugging why the markdown headings weren't getting rendered wondering whether I was violating CommonMark spec etc.

Possible solutions:
1. Continue suggesting Bulma but add a note saying user has to put `{{ page.content | safe }}` and `{{ section.content | safe }}` in `div` with class `content` for markdown to get rendered as expected. Ref: https://github.com/jgthms/bulma/issues/1603

    Also, making the same change in `blog-page.html` on that same page:
    ```
    <div class="content">
        {{ page.content | safe }}
    </div>
    ```
2. Suggest any other CSS framework than Bulma which doesn't remove the default styles of naked HTML elements.
3. Just don't suggest/use any CSS framework

rootkea:
Closes #1494

### DIFF
--- a/docs/content/documentation/content/section.md
+++ b/docs/content/documentation/content/section.md
@@ -7,7 +7,7 @@ A section is created whenever a directory (or subdirectory) in the `content` sec
 `_index.md` file.  If a directory does not contain an `_index.md` file, no section will be
 created, but Markdown files within that directory will still create pages (known as orphan pages).
 
-The index page (i.e., the page displayed when a user browses to your `base_url`) is a section,
+The homepage (i.e., the page displayed when a user browses to your `base_url`) is a section,
 which is created whether or not you add an `_index.md` file at the root of your `content` directory.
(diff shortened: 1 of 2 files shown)


rootkea:
Should CSS classes be removed too since they were meant for bulma to act upon?

Keats (approved):

Keats:
I think it's fine to leave them

Issue #419, opened 2018-09-10
Allow user to add custom highlighting themes


Bobo1239:
It would be great if users couldn't just add custom highlighting syntaxes (possible since 0.4.2) but also custom highlighting themes without the need to add them to gutenberg itself.

Keats:
I'm not entirely against that but I would rather have as many themes as possible in Gutenberg. Custom syntaxes make sense because some people are developing their own language but it's not the case really for themes.
I do plan to add all of https://boostlog.io/@junp1234/10-best-sublime-text-themes-of-2018-5b23805a44deba0054046809 for example though

chris-morgan:
There are a great many themes out there, that came about because people *do* develop their own themes. I interact with more people that I know have modified text editor themes or written themes from scratch than I know have written syntax highlighters. (I’ve done all three many times over.)

I think this variation may actually be a bigger deal for websites than for text editors: in the text editor the theme exists in relative isolation, but inside Zola syntax highlighting exists inside the rest of the site theme. The likes of Zola attract power users that care about such details. In my own fork of Zola (I’m sure to want other variations of stock functionality), I’m going to need to do this.

I believe that syntax highlighting themes should be part of the regular theme system. I don’t think that’s too radical a notion. If you want something more radical, I believe that including *any* highlighting themes out of the box is a misfeature—just as no site themes are included out of the box. There are decent repositories of TextMate themes out there already, and it’s easy to drop the .tmTheme.txt file in.

Keats:
If someone wants to send a PR, code changes can probably almost follow the syntax changes. In the `config` and `rendering` subcrates.

>  If you want something more radical, I believe that including any highlighting themes out of the box is a misfeature—just as no site themes are included out of the box. There are decent repositories of TextMate themes out there already, and it’s easy to drop the .tmTheme.txt file in.

I disagree on that, it would make syntax highlighting unusable without going on some random websites.

chris-morgan:
Correct, it would make syntax highlighting unusable without work. But Zola already requires you to find or make a theme; that’s why I don’t consider it as outlandish as all that, merely fairly radical.

Keats:
>  But Zola already requires you to find or make a theme

It also requires you to write some content :)
Themes are something I consider to be personal. I don't use pre-made themes: the only reason they are supported in Zola is because many people don't feel the same way.
There is an issue https://github.com/getzola/zola/issues/534 to help with the setup but I don't really to build/support n built-in themes for n usecases right now.

[...]

PR #1499, opened 2021-05-28
Custom highlighting themes

drmason13:
Related to #419

Introduces once_cell dependency to store extra SyntaxSet, ThemeSet as statics.
Option<SyntaxSet> is no longer stored in the Config Markdown struct.

Gruvbox tmTheme added to test_site, it is taken from
https://github.com/Colorsublime/Colorsublime-Themes (MIT licensed)

It looks like I need to fix this up to merge into next (I wrote it against master originally, woops!).
I would appreciate a hand/pointer updating the documentation.

Sanity check:

## Code changes


drmason13:
There! Conflicts resolved, fingers crossed for the tests... Works on my machine ;)

This touched a lot more things that I wanted/expected, some things were broken which I didn't change and I did my best to fix them to run tests.

In particular Config.parse lost a check re. default languages, because the tests panicked on it config.languages contained "en" and default_language was set to "en". Possibly the fix should be changing Config's default.
```
Default language `{}` should not appear both in `config.default_language` and `config.languages`
```
and some tests had a function missing a parameter, so I bluffed my way fixing that 😄 
```
    let context = RenderContext::new(&tera, &config, "", "", &permalinks_ctx, InsertAnchor::None);
                                                        ^^^^ I added this
```

components/config/src/config/markup.rs
```diff
@@ -29,23 +35,79 @@ pub struct Markdown {
 
     /// A list of directories to search for additional `.sublime-syntax` files in.
     pub extra_syntaxes: Vec<String>,
-    /// The compiled extra syntaxes into a syntax set
-    #[serde(skip_serializing, skip_deserializing)] // not a typo, 2 are need
-    pub extra_syntax_set: Option<SyntaxSet>,
+    /// A list of directories to search for additional `.tmTheme` files in.
+    pub extra_highlight_themes: Vec<String>,
```
Keats: I'm wondering if we should just add a single folder and search in that for additional syntaxes and themes instead of having 2 different vecs...

components/config/src/config/markup.rs
```diff
@@ -29,23 +35,79 @@ pub struct Markdown {
...
 }
 
 impl Markdown {
+    /// Gets the configured highlight theme from the BUILTIN_HIGHLIGHT_THEME_SET or the EXTRA_HIGHLIGHT_THEME_SET
+    pub fn get_highlight_theme(&self) -> &'static syntect::highlighting::Theme {
+        if let Some(theme) = &BUILTIN_HIGHLIGHT_THEME_SET.themes.get(&self.highlight_theme) {
+            theme
+        } else {
+            &EXTRA_HIGHLIGHT_THEME_SET.get().unwrap().themes[&self.highlight_theme]
```
Keats: We should probably invert that if/else in case someone wants to use a newer version of an existing theme for example.

components/config/src/config/markup.rs
```diff
@@ -29,23 +35,79 @@ pub struct Markdown {
...
+        }
+    }
+
+    /// Attempt to load any theme sets found in the extra highlighting themes of the config
+    /// TODO: move to markup.rs in 0.14
```
Keats: 0.14 is the branch you did this PR on  and it is on markup.rs so you can remove that comment.

components/config/src/config/markup.rs
```diff
@@ -29,23 +35,79 @@ pub struct Markdown {
...
+    pub fn load_extra_highlight_themes(&self, base_path: &Path) -> Result<Option<ThemeSet>> {
+        let extra_highlight_themes = self.extra_highlight_themes.clone();
```
Keats: I don't think you need to clone?

components/config/src/config/mod.rs
```diff
@@ -156,7 +148,12 @@ impl Config {
         let path = path.as_ref();
         let content =
             read_file(path).map_err(|e| errors::Error::chain("Failed to load config", e))?;
-        Config::parse(&content)
+
+        let config = Config::parse(&content)?;
+        let config_dir = path.parent().unwrap();
```
Keats: let's not unwrap if possible

components/config/src/highlighting.rs
```diff
@@ -7,49 +8,60 @@ use syntect::parsing::SyntaxSet;
...
+    let mut source = SyntaxSource::Plain;
+    if let Some(lang) = language {
+        let syntax = EXTRA_SYNTAX_SET
+            .get()
+            .and_then(|extra| {
+                source = SyntaxSource::Extra;
+                extra.find_syntax_by_token(lang)
+            })
+            .or_else(|| {
+                let hacked_lang = if lang == "js" || lang == "javascript" { "ts" } else { lang };
```
Keats: can you keep the comment?

docs/content/documentation/content/syntax-highlighting.md
```diff
@@ -171,38 +171,85 @@ If your site source is laid out as follows:
...
+OneHalfLight
+railsbase16-green-screen-dark
+solarized-dark
+solarized-light
+subway-madrid
+subway-moscow
+Tomorrow
+two-dark
+visual-studio-dark
+zenburn
```
Keats: there is already a list on the site: https://www.getzola.org/documentation/getting-started/configuration/#syntax-highlighting

Keats (changes requested):
What do you think of having a single folder for both additional syntaxes and themes rather than each having a vec?

drmason13:
That sounds better from a user's perspective, they're related things so I'd want them in the same folder. There are enough folders as it is :)

It should be fine to filter by extension when building the actual sets. I'll have a go at implementing it.

components/config/src/highlighting.rs
```diff
@@ -7,49 +8,60 @@ use syntect::parsing::SyntaxSet;
...
     };
-    pub static ref THEME_SET: ThemeSet =
+    pub static ref BUILTIN_HIGHLIGHT_THEME_SET: ThemeSet =
         from_binary(include_bytes!("../../../sublime/themes/all.themedump"));
 }
 
-pub enum HighlightSource {
-    Theme,
+pub static EXTRA_SYNTAX_SET: OnceCell<SyntaxSet> = OnceCell::new();
+pub static EXTRA_HIGHLIGHT_THEME_SET: OnceCell<ThemeSet> = OnceCell::new();
```
meithecatte: It is my understanding that you went for `OnceCell` because `ThemeSet` is not cloneable. I believe there is an easier way of resolving this problem: wrap it in an `Arc` instead.

drmason13: Hi, yes you are correct. I was working around clone. An Arc should work just as well, and without awkwardness around ensuring you only set it once.

drmason13: Ah, except `Arc<SyntaxSet>` doesn't impl Serialize/Deserialize. I don't know how to I would implement it manually, and I suspect it wouldn't be worth the hassle in this case?

meithecatte:
This is the same problem as with the existing `SyntaxSet` — add a
`#[serde(skip_serialize, skip_deserialize)]` annotation, or whatever the
syntax was.

Keats:
@drmason13 can you rebase and address comments?

drmason13:
Yes I can, thank you for the nudge.

EDIT: This is taking a while, but I'll carry on tomorrow. Exciting to see the option to use css for themes now :smile:

drmason13:
Windows has failed on the OnceCell set unwrap. Which surprises me because it checks that get returns None before setting it.
So there's probably something fishy with my use of once_cell ☹️ 

I will try the switch to `Arc<SyntaxSet>` in the `Config::Markdown` struct as @NieDzejkob suggested. I was only putting it off until I'd caught up with the next branch.

drmason13:
> I will try the switch to `Arc<SyntaxSet>` in the `Config::Markdown` struct as @NieDzejkob suggested. I was only putting it off until I'd caught up with the next branch.

Although hang on, The Config can store an `Option<Arc<ThemeSet>>` and yay that can be cloned, but this is only a reference to the extra theme set. The owned value still has to live somewhere. It can't (shouldn't?) live in a lazy_static because we need to have the config ready to know which folders to load custom syntaxes from. I don't want to initialize all of that in a lazy static.

So I still end up reaching for something like OnceCell to set the extra highlight theme during runtime. I feel stuck with OnceCell still. I want to try and figure out why the Windows test failed exactly.

drmason13:
> @drmason13 can you rebase and address comments?

@Keats - All comments addressed, how are we looking here?
I'll have a think about the windows test failure - all I have at the moment is to check over the once cell docs again and see if it would be OK (not great, but OK) to largely ignore the Err if it's already been set.

EDIT: something like this perhaps:
```rust
if let Some(extra_syntax_set) = loaded_extra_syntaxes {
    if EXTRA_SYNTAX_SET.get().is_none() {
        // This _shouldn't_ be an error, but it is safe to continue if it is
        let set_result = EXTRA_SYNTAX_SET.set(extra_syntax_set);
        debug_assert!(
            set_result.is_ok(),
            "Inserted into an empty OnceCell while initializing EXTRA_SYNTAX_SET but got an error - possible race condition?"
        );
    }
}
```

meithecatte:
> Although hang on, The Config can store an `Option<Arc<ThemeSet>>` and yay that can be cloned, but this is only a reference to the extra theme set. The owned value still has to live somewhere. It can't (shouldn't?) live in a lazy_static because we need to have the config ready to know which folders to load custom syntaxes from. I don't want to initialize all of that in a lazy static.

What? The `Arc` itself has ownership, shared across all the clones. That's the point of reference counting. Note [`Arc::new`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.new) doesn't take a `&T`, but `T`.

drmason13:
> What? The `Arc` itself has ownership, shared across all the clones. That's the point of reference counting. Note [`Arc::new`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#method.new) doesn't take a `&T`, but `T`.

D'oh! You're right, what was I thinking?

drmason13:
> It is my understanding that you went for OnceCell because ThemeSet is not cloneable. I believe there is an easier way of resolving this problem: wrap it in an Arc instead.

> > @drmason13 can you rebase and address comments?
> 
> @Keats - All comments addressed, how are we looking here?

Update: Switched over to an Arc version - many thanks to @NieDzejkob reminding me how Arc actually works - and it has worked out really nicely I think 😃

components/config/src/highlighting.rs
drmason13: Done, sorted with skip_serialize/deserialize and `Arc<>`

Keats:
I'll put it in the version after 0.14, I want to test it a bit and not add more things to the 0.14 relese

Keats:
@drmason13 
can you rebase?

drmason13:
I'm a little worried about the rendering/benches/all.rs file. I haven't been able to run it (I actually get an error for unresolved crate:
```
#![feature(test)]
extern crate test;
```

I've left it so far as a "not my problem" kind of thing, I hope that's OK! 😄

drmason13:
@Keats Hopefully that's all good. I should have rerun the tests locally to double check but the rebase was pretty smooth, so fingers crossed!

Keats:
> I'm a little worried about the rendering/benches/all.rs file. I haven't been able to run it (I actually get an error for unresolved crate:

The default test runner requires nightly so you would need to do something like `cargo +nightly bench` to have it working. Someone is working on using criterion which would work in stable.

drmason13:
uh oh! This is definitely an issue with my changes. Let me take a look over the tests again.
EDIT: Ah, I used the wrong version of one of the merge conflicts, my bad. Fixed now.

drmason13:
@Keats good to go :)

components/config/src/config/markup.rs
```diff
@@ -43,26 +49,92 @@ pub struct Markdown {
...
+        }
+        if let Some(extra_theme_set) = loaded_extra_highlight_themes {
+            self.extra_theme_set = Arc::new(Some(extra_theme_set));
+        }
+
+        // validate that the chosen highlight_theme exists in the loaded highlight theme sets
+        if !THEME_SET.themes.contains_key(&self.highlight_theme) {
+            if let Some(extra) = &*self.extra_theme_set {
+                if !extra.themes.contains_key(&self.highlight_theme) {
+                    bail!(
```
Keats, line 128: I think it doesn't need to be a different error than the one below

components/rendering/benches/all.rs
```diff
@@ -106,29 +106,30 @@ fn bench_render_content_without_highlighting(b: &mut test::Bencher) {
     let mut config = Config::default();
     config.markdown.highlight_code = false;
     let current_page_permalink = "";
+    let lang = "";
     let context = RenderContext::new(
         &tera,
         &config,
-        "",
+        lang,
```
Keats, line 113: too many arguments eh :(

components/config/src/config/markup.rs
```diff
@@ -43,26 +49,92 @@ pub struct Markdown {
...
+        }
+        let ss = ss.build();
+
+        Ok((
+            if ss.syntaxes().is_empty() { None } else { Some(ss) },
+            if ts.themes.is_empty() { None } else { Some(ts) },
+        ))
+    }
+
+    pub fn export_theme_css(&self, theme_name: &str) -> String {
```
Keats, line 104: I would prefer if that stayed with the rest of the highlighting code so syntect doesn't bleed too much in the config but it doesn't seem doable anymore :(

components/site/src/lib.rs
```diff
@@ -74,7 +73,7 @@ impl Site {
         let path = path.as_ref();
         let config_file = config_file.as_ref();
         let mut config = get_config(config_file)?;
-        config.markdown.load_extra_syntaxes(path)?;
+        config.markdown.load_extra_syntaxes_and_highlight_themes(path)?;
```
Keats, line 76: it should already be called by the line above no?

Keats (changes requested):
Just  a few things, I can merge and do it myself if you want

components/rendering/benches/all.rs
drmason13: It's a tad confusing yes!

drmason13:
> Just  a few things, I can merge and do it myself if you want

I'm away from PC for another couple of days, so please go ahead. Appreciate the comments though :)

PR #1509, opened 2021-05-31
Include path in the search index with include_path

### LINKED ISSUE #1481: Include path in the search index
Would be useful on my wiki as pages don't have a title, instead the path serves as the replacement for a title. I'd be willing to PR if you agree to add this feature (in form of a `search/include_path` switch).

timakro:
This way URLs can be searched conveniently on wiki- / knowledge base- / docs-style sites. E.g. the path `/documentation/getting-started/overview/` at https://getzola.org would be matched by queries like "documentation" or "getting started".

Closes #1481.

Sanity check:

## Code changes

### DIFF
--- a/components/search/src/lib.rs
+++ b/components/search/src/lib.rs
@@ -1,6 +1,8 @@
 use std::collections::{HashMap, HashSet};
 
 use elasticlunr::{Index, Language};
+use elasticlunr::pipeline;
+use elasticlunr::pipeline::TokenizerFn;
 use lazy_static::lazy_static;
 
 use config::{Config, Search};
@@ -36,6 +38,10 @@ fn build_fields(search_config: &Search) -> Vec<String> {
         fields.push("description".to_owned());
     }
 
+    if search_config.include_path {
+        fields.push("path".to_owned());
+    }
+
     if search_config.include_content {
         fields.push("body".to_owned());
     }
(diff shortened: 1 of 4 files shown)


components/search/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Vincent Prouillet <prouillet.vincent@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-elasticlunr-rs = {version = "2", default-features = false, features = ["da", "de", "du", "es", "fi", "fr", "it", "pt", "ro", "ru", "sv", "tr"] }
+elasticlunr-rs = {version = "2.3.12", default-features = false, features = ["da", "de", "du", "es", "fi", "fr", "it", "pt", "ro", "ru", "sv", "tr"] }
```
Keats: you don't need to change that, Cargo will automatically use the latest version in this case

components/search/src/lib.rs
```diff
@@ -36,17 +38,57 @@ fn build_fields(search_config: &Search) -> Vec<String> {
...
+
+    if search_config.include_path {
+        tokenizers.push(path_tokenizer);
+    }
+
+    if search_config.include_content {
+        tokenizers.push(text_tokenizer);
+    }
+
+    tokenizers
```
Keats, line 84: You can end up with the same tokenizer multiple times, is that not an issue?

Keats (changes requested):

Keats:
>  the path /documentation/getting-started/overview/ at https://getzola.org would be matched by queries like "documentation" or "getting started".

Does that mean that the index will be essentially filled with "useless" data in this case? Every doc page starts with documentation

components/search/src/lib.rs
timakro: The same tokenizer, that is the default tokenizer, should be used for all fields except the path field which gets it's own tokenizer that splits on `/`. Those are just function pointers so having two pointers point to the same function is not an issue.

timakro:
On my wiki I have a bunch of pages in the `hack` section. All of them are indexed under "hack":

image

So when you search for "hack" you find those pages. To me this seems like the behavior you'd expect.

You probably only want to use path search when the path of the pages is exposed to your users in some way. For example if you show the path of the pages in the search results and highlight the matches. Or if you show the path on top of the page e.g. in breadcrumbs style. Then matching all docs pages when searching for "documentation" makes sense.

image

Keats (approved):

timakro:
Thanks for taking the time to review my code. And thanks for creating this terrific piece of software! :)

Keats:
You're welcome!

PR #1523, opened 2021-06-10
Fixed  documentation code for deployment in Github Action

hbina:
It seems like you need to append with your username to access the secret token.
See: https://githubmemory.com/repo/shalzz/zola-deploy-action/issues/13

Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>

Sanity check:

* [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

### DIFF
--- a/docs/content/documentation/deployment/github-pages.md
+++ b/docs/content/documentation/deployment/github-pages.md
@@ -52,7 +52,7 @@ jobs:
           # Target branch
           PAGES_BRANCH: gh-pages
           # Provide personal access token
-          TOKEN: ${{ secrets.TOKEN }}
+          TOKEN: $GITHUB_ACTOR:${{ secrets.GITHUB_TOKEN }}
 ```
 
 This script is pretty simple, because the [zola-deploy-action](https://github.com/shalzz/zola-deploy-action) is doing everything for you. You just need to provide some details. For more configuration options check out the [README](https://github.com/shalzz/zola-deploy-action/blob/master/README.md).


Keats:
I am not sure if it's the right fix since you might use a token that is not bound to the actor?

Edit: imagine publishing the docs for zola which is in an org with many people editing the docs

hbina:
Hmm, I don't know enough about how any of this work so I'll leave it up to you.

Keats:
I think we should probably remove that documentation and just point to the action repo. Otherwise it can get out of sync easily.

prrao87:
I think this particular page is causing a lot of confusion to the community -- I agree with @Keats that the documentation should just point to the action repo. It's already out of sync and I had to resolve a host of issues in deploying my first Zola blog recently because it's misleading and the provided action in the Zola docs didn't work for me.

There are a host of issues related to gh-pages even here in this repo because some people (but not others) are facing weird issues similar to the one I documented here:
https://github.com/shalzz/zola-deploy-action/issues/81

The action author has published a hotfix that addresses this issue for some users and released `v0.17.2-1` that seems to do the job. As such, it makes sense to close this PR and clean up the documentation to just point to the action repo to reduce the maintenance burden on this end.

Happy to send a PR here if this works.

Keats:
Yeah, go for it. New version of Zola is going to be released either this weekend (unlikely due to RL) or next week, I had a look through the open bugs yesterday.

Keats:
Looks like it's not needed according to the README?

PR #1557, opened 2021-07-21
Added sidenote on serving to local devices

### LINKED ISSUE #1556: Document serving over local network
# Documentation issue

## Summary
Currently, interface `127.0.0.1` is used. I can imagine this is on purpose, since it will not ask for network access on both Windows and Mac. This, however, leads to the problem of not being able to reach websites you are testing from other devices. For example, testing to see if the mobile view is correct and usable (outside your browser's DevTools). The way around this is using both interface `0.0.0.0` and your local IP for your base_url as flags.

Since interface `0.0.0.0` is mostly standard for `serving`, this confused me massively when I was first creating a website using Zola. Documenting this could save others in the future massive amounts of time and confusion.

## Proposed solution
I would add a small sidenote to the CLI section on the serve command.

coastalwhite:
Sanity check:

## Code changes
I added a sidenote to the CLI usage section in the documentation on how to serve to devices on the local network. Something which is done standard in a lot of SSGs and caused many headaches for me. Hopefully, with this documentation not anymore or less for others.

Fixes #1556.

### DIFF
--- a/docs/content/documentation/getting-started/cli-usage.md
+++ b/docs/content/documentation/getting-started/cli-usage.md
@@ -72,6 +72,10 @@ the interface/port combination to use if you want something different than the d
 
 You can also specify different addresses for the interface and base_url using `--interface` and `-u`/`--base-url`, respectively, if for example you are running Zola in a Docker container.
 
+> By default, devices from the local network **won't** be able to access the served pages. This may be of importance when you want to test page interaction and layout on your mobile device or tablet. If you set the interface to `0.0.0.0` however, devices from your local network will be able to access the served pages by requesting the local ip-address of the machine serving the pages and port used.
+>
+> In order to have everything work correctly, you might also have to alter the `base-url` flag to your local ip.
+
 Use the `--open` flag to automatically open the locally hosted instance in your
 web browser.
 


Keats (approved):

Keats:
Thanks!

docs/content/documentation/getting-started/cli-usage.md
```diff
@@ -72,6 +72,10 @@ the interface/port combination to use if you want something different than the d
 
 You can also specify different addresses for the interface and base_url using `--interface` and `-u`/`--base-url`, respectively, if for example you are running Zola in a Docker container.
 
+> By default, devices from the local network **won't** be able to access the served pages. This may be of importance when you want to test page interaction and layout on your mobile device or tablet. If you set the interface to `0.0.0.0` however, devices from your local network will be able to access the served pages by requesting the local ip-address of the machine serving the pages and port used.
+>
+> In order to have everything work correctly, you might also have to alter the `base-url` flag to your local ip.
```
vadz, line 77: Sorry for barging in, but I'd like to suggest making this suggestion more forceful, because you definitely _have_ to use the `--base-url` option to make it work, otherwise even though the HTML is served from the IP specified with `--interface`, it uses `127.0.0.1` to reference to CSS, JS etc, which is not going to work (as I've just discovered).

Alternatively I wonder if `zola serve` could be smarter and use the correct interface automatically if the base URL is a local IP?

Keats: It would be nice to avoid having to specify multiple parameters, i'd take a PR for that

PR #1558, opened 2021-07-21
Added prompt for when zola build output-dir mentions an existing directory.

### LINKED ISSUE #1378: zola build --output-dir Document Deleting Behavior
# Documentation issue

## Summary
`zola build` has a helpful `--output-dir` flag. However the documentation from the `--help` command:

> -o, --output-dir <output_dir>    Outputs the generated site in the given path [default: public]

Doesn't make it clear that the target directory contents will be deleted. I accidentally wiped out my entire Zola project by specifying `--output-dir ../ZolaProject/`. (Thankfully it was all saved on Github). 

## Proposed solution
- Document fact flag will delete contents of target directory.
- Maybe if target directory is something other than `public/` give a warning if the directory is non-empty?

coastalwhite:
I went about implementing this. It wasn't too hard, especially since `ask_bool` and `Path::exists` exist.

There is one concern I ran into after implementing this. It may need to be discussed still. Is it okay to require CI's to first delete the folder in question? This may break CI's and automated scripts, and therefore it may be interesting to add to the 0.15 release.

Sanity check:

## Code changes

Added `ask_bool` prompt for `--output-dir` for when the output directory targeted already exists in the file system. Updated the documentation to match this behaviour.

Fixes #1378.


src/cmd/build.rs
```diff
@@ -14,6 +15,17 @@ pub fn build(
 ) -> Result<()> {
     let mut site = Site::new(root_dir, config_file)?;
     if let Some(output_dir) = output_dir {
+        // Check whether output directory exists or not
+        // This way we don't replace already existing files.
+        if output_dir.exists() {
+            console::warn("The directory you are about build to already exists. Building to this directory will delete files contained within this directory.");
```
Keats: can we add the path to the message?

src/cmd/build.rs
```diff
@@ -14,6 +15,17 @@ pub fn build(
...
+
+            // Prompt the user to ask whether they want to continue.
+            if !ask_bool("Are you sure you want to continue?", false)? {
```
Keats: This makes `zola build` potentially interactive with no way of skipping it. Either we add a flag to skip prompts or we add a timeout to the prompts

Keats (changes requested):
Seems ok but I'll wait to see if there is a need for a 0.14.1 before merging that.

src/cmd/build.rs
coastalwhite: Will do!

src/cmd/build.rs
coastalwhite: As I mentioned in the PR:

>There is one concern I ran into after implementing this. It may need to be discussed still. Is it okay to require CI's to first delete the folder in question? This may break CI's and automated scripts, and therefore it may be interesting to add to the 0.15 release.

A timeout would seem to be okay for now, but this would still change default behavior if the default value is `false`. I feel like `false` as default value is the correct option, though.

Keats: Yep I've seen that comment but I think there needs a way to just ignore prompts. I lean towards timeout

coastalwhite:
There is now a 10s prompt timeout using tokio's timeout.

Keats:
Thanks!

Yaroslav-95:
So I am back here to complain about another change that breaks my use-case :) I want to propose a flag that will force the build without prompting. My rationale is the following: I have a git hook that builds my site on every push to master and, although I could edit the script to delete the folder prior to building, I don't want to that, because I need to preserve the mode bits and ownership of the directory. What do you say @Keats ?

Actually I already have a patch, so I'll create a PR in a little bit ;)

(Edited to add some clarity into my words)

Issue #1559, opened 2021-07-23
Rewriting shortcode logic


coastalwhite:
In #1500 it came to light that there are a considerable amount of cases in which the current pipeline for inserting shortcodes shows UB. Because of this is was the conclusion to rewrite this area of the codebase.

This issue aims to discuss what this rewrite might look like.

# Current problems

Currently, many problems arise from the fact that multiline shortcodes are hard to handle. Especially in a inline or HTML context. It is very difficult to known when the codebase will or will not insert `<p>` tags and where it will insert these tags.

It is also awkward at the moment how these shortcodes are divided into two categories –namely markdown and html shortcodes – and these are handled by entirely the same code. This is awkward because in the end these have quite different behaviour.

# Proposed solution

My proposed solution is divided into three parts:

1. Have a look at ironing out some of the specifics of shortcodes. For example, question such as the following
   * What does an inline multiline markdown shortcode evaluate to?
   * When specifically will html shortcodes have `<p>` tags around them.
   * Will shortcodes in markdown shortcodes be evaluated?
   * Probably some more…
2. Following from step 1, write a set of tests this new rewrite should pass.
3. Discuss the structure of this rewrite.
4. Actually rewrite this piece of code.

coastalwhite:
To leave my own two cents:

I personally think shortcodes should act almost a `{% include ‘…’ %}` with variables. This means that all the markdown will be evaluated after the shortcodes are inserted. 

As a consequence everything that has no whiteline between in it after the shortcodes are inserted will be inline. With the exception of div tags which span from the beginning to the end of a paragraph.

To give some examples.

First some shortcodes to define:
*shortcodes/world.html*
```html
<span>World</span>
```

*shortcodes/multiline.html*
```html
Hello

World
```

*shortcodes/divmultiline.html*
```html
<div>
Hello

World
</div>
```

Then the following content:
```markdown
Hello {{ world() }}

Hello
{{ world() }}

Hello

{{ world() }}

{{ multiline() }}

{{ divmultiline() }}
```

Will evaluate too:
```html
<p>
Hello <span>World</span>
</p>

<p>
Hello
<span>World</span>
</p>

<p>
Hello
</p>

<p>
<span>World</span>
</p>

<p>
Hello
</p>

<p>
<span>World</span>
</p>

<div>
Hello

World
</div>
```

I also think recursive shortcodes in markdown shortcodes should be possible.

Keats:
[...]

Issue #1500, opened 2021-05-30
Incorrect HTML generated for shortcode inside `<div>`


rootkea:
Hello!

This is my shortcode template:
```
$ cat templates/shortcodes/google.html 
<div>
<a href="https://google.com/search?q={{query}}">Google Search</a>
</div>
$
```

This is how I invoke the shortcode:
```
foo {{ google(query="apple") }} bar.
````

And this is what I get:
```
  <p>foo <div>
<a href="https://google.com/search?q=apple">Google Search</a></p>
</div> bar.
```

I expected `</p>` after `bar.`

**Comment by rootkea** [comment]
Also I don't understand this from shortcode docs:
> The Markdown renderer will wrap an inline HTML node such as `<a>` or `<span>` into a paragraph. If you want to disable this behaviour, wrap your shortcode in a `<div>`.

But the markdown renderer does not wrap `<a>` into paragraph.

e.g.
This is my shortcode template:
```
$ cat templates/shortcodes/google.html 
<a href="https://google.com/search?q={{query}}">Google Search</a>
$
```

This is how I invoke shortcode:
```
{{ google(query="apple") }}
```

And this is the HTML generated:
```
<a href="https://google.com/search?q=apple">Google Search</a>
```

No `<p>` so I don't understand the point of wrapping shortcode inside `<div>`

**Comment by Keats** [comment]
I believe it's referring to the shortcodes with `{% %}`

**Comment by rootkea** [comment]
Umm.. I modified the above shortcode to use `{% %}` but still no `<p>`
```
$ cat templates/shortcodes/google.html 
<a href="https://google.com/search?q={% if query %}{{query}}{% endif %}">Google Search</a>
$
```
What am I missing?

Edit - I even tried encompassing whole shortcode inside `{% if query %}` and `{% endif %}` but still no `<p>`.

**Comment by coastalwhite** [comment]
Can confirm some weird behavior. Will do some more investigation, hopefully tomorrow.

From what I found now, this is a bug that happens when you have a shortcode surrounded by a div on the same line as some other text. While testing I found some even weirder possibly more problematic stuff.

So using the following shortcode:

*shortcodes/txt.html*
```html
<div>
     Some text
</div>
```

Here are some extremely strange outputs.

> *Note:* Sorry for the hard to read format.

# Test 1
## Input
```
{{ txt() }}
```

## Output
```html
<div>
    Some text
</div>
```

## Conclusion
As expected.

# Test 2
## Input
```
foo
[...]

PR #1564, opened 2021-07-28
Rewrite of Shortcode Logic

coastalwhite:
This is the PR devoted to the rewriting of the shortcode parsing and insertion. This is discussed in depth in #1559 and was first suggested in #1500. The PR should also replace #1475 which also addressed #515.

There are a couple of steps to this process:
- [x] Write the specifications of how shortcodes are going to be parsed in the form of unit tests
  - [x] Write specification
  - [x] Write summary unit tests
  - [x] Write TOC unit tests
  - [x] Write internal & external link unit tests
  - [x] Write a few complete integration tests
- [x] Rewrite the parser using [Logos](https://github.com/maciejhirsz/logos)
- [x] Rewrite the insertion code considering the specification and recursive shortcodes
- [x] Update the `markdown_to_html` function
- [ ] Make sure no massive performance problems arise 
  - [ ] Benchmark (Before PR vs. after)
  - [ ] Optimize functions


coastalwhite:
Currently, the codebase is extremely messy. I will clean this up later.

But... the parsing of shortcodes using [Logos](https://github.com/maciejhirsz/logos) is pretty much done. It was much more fun than I expected. I had never really worked with a proper Lexer. A few things are left to do before I will check that box though, but I am out of time for today.

Here is a list of things which still need to be done because the Logos box can be checked:
- [x] Properly handle errors
  - [x] Convert between parsing errors
  - [x] Properly show and format errors to give some proper information
- [x] Parse expressions instead of variables for `ArgValue`
  - [x] Add square bracket notation
  - [ ] ~Add filter notation~ **Will not do this, since I don't feel like it adds much**
- [x] Fix the escaped strings in `string_from_quoted`
- [x] Move `QuoteType` to be in the `ArgValueToken` definition
- [x] Make context dependent variables to not return the parsing function, but instead to continue normally
- [x] Organize the codebase
  - [x] Organize the `shortcode` submodule
  - [ ] ~Rip out the old parsing code~ **I am going to leave this for a bit, since I will probably use it for inspiration.**
  - [x] Write comments in modules signalling what they are doing, since that will probably become hard to understand

All of these should be relatively small tasks, though. Except for maybe the parsing of expressions instead of variables.

coastalwhite:
The [Logos](https://github.com/maciejhirsz/logos) parsing is now finished. Probably some more changes will happen whilst working on the rest of the codebase, but the functionality should be complete for now.

What I am going to do next is expanding the `ShortcodeContext` to include whether it is inline or standalone and to have information about it spans which can be updated as shortcodes get inserted.

Following this, I will have a look at redoing the insertion code.

coastalwhite:
The insertion code is pretty much done. Only two parts remain:

- [x] Make the code generic over HTML as well
- [ ] Add the original context to the context used at the start.

coastalwhite:
Going away for a week or so, if someone wonders why there is a lack of updates on this PR next week.

coastalwhite:
Although most of the functionality is now here, there are quite some kinks to work out. The codebase now doesn't use placeholders for the HTML shortcodes anymore. There are still some things which are not finished, closely relating to the things described above:

- [ ] It needs to be figured out how to deal with the content of bodied shortcodes so that they are not automatically considered by the `markdown_to_html` function to be turned into html even if that is not wanted. This should still allow for recursive and embedded shortcodes (and embedded bodied shortcodes).

This is what I am currently imagining for this. The function `fetch_shortcodes`
will now also return a `String`. This is basically the source but without the
shortcodes in it (probably there being some text placeholder for a shortcode).
This way bodies won't be processed any more. There now will need to be some
logic to run `insert_shortcodes` on the bodies and the results.

- [ ] The shortcodes should be fetched only once instead of the 3 times it does now.
- [ ] The builtin functions should be readded. This could be quite chaotic so I am holding of on this until I have access to a proper pc again.
- [ ] The errors should be properly handled. This is mostly internally since those errors will almost always result in the ignorance of that case.
- [ ] Some parts of the code are not very optimized
  - [ ] The decision whether to p-tag or not could be sorted beforehand and then pop-ed every time. This removes the search for every p. I am quite sure a popping process is faster, but maybe some benchmarking has to be done.
  - [ ] There are a lot of unnecessary clones done, especially during parsing. This could mostly be fixed by using the Cow type and figuring a way around the `lex.clone()` when it is not needed.
- [ ] Some of the code is now very badly documented, this should be gone over again.
- [ ] There are now a lot of weirdly named structs which could have better names. Again holding of on this until I have access to a proper pc again, since those can use mass-rename.
- [ ] Probably some other things I forgot but will add later.

coastalwhite:
Having worked on this PR for a couple of hours today and essentially making zero progress, has made me realize I need to step back a second and think of some kind of better architecture here. At the moment, everything by itself works really well, and I would also imagine pretty fast, but it is just really difficult to make work together. I even had a look at maybe creating a new PR just containing some of the smaller fixes contained in this PR, but turns out to be almost impossible, since making it interopt with the older code is going to be a nightmare.

The main problem stems from have recursive shortcodes with two different rendering cycli. This needs to have very formal restrictions. Personally, it is not completely clear anymore after working for so long what the problems are and thus what the restrictions should be. I have simply had a bit of short-circuit in my head thinking about this problem the entire day. So, I will take some time to actually map out the entire process again, and first make a list of problems and try to find a solution which solves them.

At the moment, my perspective involves creating some kind of huge datastructure from all parsed shortcodes.

Here is a list of basically everything done by this PR currently:
1. Replace Pest parsing with Logos parsing
2. Allow variables to be used as argument values in shortcodes
3. Formalize, when HTML shortcodes are surrounded with a `<p>`-tag and with that remove the `<pre>` placeholders. Currently, this already fixes a lot of the bugs found when inserting shortcodes.
4. Formalize the shortcode insertion process.
5. A lot of new tests
6. There is the foundations for a lot of the insertion code, containing all the logic for shortcodes within other shortcodes and embedding shortcodes in bodies of shortcodes.

@Keats if you have any input on the following question, it would be greatly appreciated:

*What should the actual scope of this PR be? Ideally of course, this would tackle all the problems which may require some specific code-architecture. I feel like the scope should be better declared.*

coastalwhite:
Having worked on this PR for a couple of hours today and essentially making zero progress, has made me realize I need to step back a second and think of some kind of better architecture here. At the moment, everything by itself works really well, and I would also imagine pretty fast, but it is just really difficult to make work together. I even had a look at maybe creating a new PR just containing some of the smaller fixes contained in this PR, but turns out to be almost impossible, since making it interopt with the older code is going to be a nightmare.

The main problem stems from have recursive shortcodes with two different rendering cycli. This needs to have very formal restrictions. Personally, it is not completely clear anymore after working for so long what the problems are and thus what the restrictions should be. I have simply had a bit of short-circuit in my head thinking about this problem the entire day. So, I will take some time to actually map out the entire process again, and first make a list of problems and try to find a solution which solves them.

At the moment, my perspective involves creating some kind of huge datastructure from all parsed shortcodes.

Here is a list of basically everything done by this PR currently:
1. Replace Pest parsing with Logos parsing
2. Allow variables to be used as argument values in shortcodes
3. Formalize, when HTML shortcodes are surrounded with a `<p>`-tag and with that remove the `<pre>` placeholders. Currently, this already fixes a lot of the bugs found when inserting shortcodes.
4. Formalize the shortcode insertion process.
5. A lot of new tests
6. There is the foundations for a lot of the insertion code, containing all the logic for shortcodes within other shortcodes and embedding shortcodes in bodies of shortcodes.

@Keats if you have any input on the following question, it would be greatly appreciated:

*What should the actual scope of this PR be? Ideally of course, this would tackle all the problems which may require some specific code-architecture. I feel like the scope should be better declared.*

Keats:
> The main problem stems from have recursive shortcodes with two different rendering cycli.

Let's forget recursive shortcodes for now then unless you have a breakthrough, it would be nice but not essential.
Just fixing all the known other issues with shortcodes and having more tests would be a huge benefit.

coastalwhite:
I think most of the functionality is there. Four more things to do:
- [x] Implement loading the builtins
- [x] Implement loading the shortcodes from files
- [ ] Clean up code
- [ ] Benchmark code

Keats:
@coastalwhite Do you have time to fins that PR?

coastalwhite:
@Keats I have been thinking about this PR a lot and am planning to finish it this weekend.

coastalwhite:
Currently, everything is working as it is supposed to as far as I can tell. I will give a major disclaimer that there is big chance that this PR will also introduce some bugs. I tested it on all the unit tests and that seemed to work fine. When I tested it on my website, which contains some basic shortcodes, it also seemed to work fine.

There is one thing that is important to fix before you actually want to merge. Currently, escaped shortcodes (e.g. `{{/* a() */}}` will not become unescaped, instead they will just be displayed as their escaped form. I added a test for it, but sadly, I don't have the time today any more to fix it. This basically involves adding rules for that in the Logos parsing openers and changing the `fetch_shortcodes` function. All the work for it should be doable in the `components/src/shortcode/parse.rs` file.

Since I am currently quite in the trenches of my studies with midterms happening in the coming weeks and I already had to majorly plan just to have the time to have a look at this PR today, I think the chance is very low that I will get another proper chance to look at this PR. If there are any questions about the code thus far, I am always prepared to answer. I am sorry if this is maybe a sad conclusion to this PR. If someone else wants to finish off this PR instead of @Keats, I would also be open to that.

Keats:
Thanks for all the work! I'll have a look asap and can finish it myself, unless someone wants to do it

Keats:
Merged into https://github.com/getzola/zola/pull/1564 so I can review without conflicts + finish it

PR #1573, opened 2021-08-01
Support standard byte order sorting

adriangoransson:
This PR adds the `sort_order` property to sections that defaults to
lexical in order to keep backwards compatability and adds "standard" for
byte order sort.

This is useful if you have a list of non-english names and you want them
sorted correctly. My site is in Swedish and with natural sort åäö (our
last three letters of the alphabet) end up in the same position as aao
respectively.

*Didn't realize this was a requirement for feature additions (as well as requests).. Sorry!*

Sanity check:

This PR is a continuation of #1315 which added natural sort order. Unfortunately that sort order wasn't perfect for my use case.

## Code changes

### DIFF
--- a/components/front_matter/src/lib.rs
+++ b/components/front_matter/src/lib.rs
@@ -60,6 +60,15 @@ pub enum SortBy {
 
 #[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
+pub enum SortOrder {
+    /// Lexicographical sorting provided by the crate lexical_sort
+    Lexical,
+    /// Standard byte-order sorting
+    Standard,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
 pub enum InsertAnchor {
     Left,
     Right,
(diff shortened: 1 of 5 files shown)


Keats:
Thanks, that's definitely something of interest.

There was also https://github.com/Keats/tera/issues/460 for sorting in Tera and I'm wondering how that works for non alphabetic languages.
I feel like it should probably be more a different `sort_by` value than a new configuration to keep things simple, unless we need to add a third order type.

What do you think of https://zola.discourse.group/t/remove-lighter-heavier-and-earlier-later/767/8 ?

adriangoransson:
Thanks for taking the time to review!

> I feel like it should probably be more a different sort_by value than a new configuration to keep things simple, unless we need to add a third order type.

If you mean something like `sort_by = "title_standard"` , I considered it but think having a sort order separate is more flexible if at some point `order_by` would support `extra.*`. Though I don't know if that would be of interest to have.

> What do you think of https://zola.discourse.group/t/remove-lighter-heavier-and-earlier-later/767/8 ?

I make no use of next/prev references but greater/smaller sounds great and easy to use to me!

Keats:
I've finally added that in the `next` branch: https://github.com/getzola/zola/pull/1741/commits/5873e0319c2eabc5c64fcbb560fe072baec9e53c
Started the commit from scratch as the code changed a lot since that PR.

Better late than never!

adriangoransson:
Thanks for your hard work! It's appreciated. :)

PR #1579, opened 2021-08-05
Give 404 error instead of empty response when file is not found

vilcans:


vilcans:
Closing this as I based it on the master branch. I'll create a new pull request on the next branch instead.

PR #1582, opened 2021-08-08
Support colocating subfolders

### LINKED ISSUE #894: Unable to use asset colocation with subfolders
# Bug Report
I'm attempting to use asset colocation on my blog to keep all the relevant images and such in the same directory as the post itself. However, it seems like I can't have a subdirectory inside that post folder, and have it still work.

Running Arch Linux, Zola version: 0.9.0

Example folder structure:

```
content/
- blog/
  - my_post/
    - index.md
    - images/
      - myimage.png
      - otherimage.png
```

Referring to an image inside `index.md` with `My alt-text` instead makes an absolute path reference to `/images/myimage.png` rather than the expected relative referance to `/blog/my_post/images/myimage.png`. Moving the images into `my_post/` and just referring to them as `myimage.png` in the markdown file works as expected.

timakro:
Closes #894.

### DIFF
--- a/components/library/src/content/mod.rs
+++ b/components/library/src/content/mod.rs
@@ -3,8 +3,9 @@ mod page;
 mod section;
 mod ser;
 
-use std::fs::read_dir;
 use std::path::{Path, PathBuf};
+
+use walkdir::WalkDir;
 
 pub use self::file_info::FileInfo;
 pub use self::page::Page;
@@ -32,7 +33,7 @@ pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
 pub fn find_related_assets(path: &Path, config: &Config) -> Vec<PathBuf> {
     let mut assets = vec![];
 
-    for entry in read_dir(path).unwrap().filter_map(std::result::Result::ok) {
+    for entry in WalkDir::new(path).into_iter().filter_map(std::result::Result::ok) {
         let entry_path = entry.path();
         if entry_path.is_file() {
             match entry_path.extension() {
@@ -46,18 +47,11 @@ pub fn find_related_assets(path: &Path, config: &Config) -> Vec<PathBuf> {
     }
 
     if let Some(ref globset) = config.ignored_content_globset {
-        // `find_related_assets` only scans the immediate directory (it is not recursive) so our
-        // filtering only needs to work against the file_name component, not the full suffix. If
-        // `find_related_assets` was changed to also return files in subdirectories, we could
-        // use `PathBuf.strip_prefix` to remove the parent directory and then glob-filter
-        // against the remaining path. Note that the current behaviour effectively means that
-        // the `ignored_content` setting in the config file is limited to single-file glob
-        // patterns (no "**" patterns).
         assets = assets
             .into_iter()
-            .filter(|path| match path.file_name() {
-                None => false,
-                Some(file) => !globset.is_match(file),
+            .filter(|p| match p.strip_prefix(path) {
+                Err(_) => false,
+                Ok(file) => !globset.is_match(file),
             })
             .collect();
     }
@@ -68,7 +62,7 @@ pub fn find_related_assets(path: &Path, config: &Config) -> Vec<PathBuf> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::fs::File;
+    use std::fs::{File, create_dir};
 
     use config::Config;
     use tempfile::tempdir;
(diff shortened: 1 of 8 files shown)


components/library/src/content/mod.rs
```diff
@@ -46,18 +47,11 @@ pub fn find_related_assets(path: &Path, config: &Config) -> Vec<PathBuf> {
     }
 
     if let Some(ref globset) = config.ignored_content_globset {
-        // `find_related_assets` only scans the immediate directory (it is not recursive) so our
-        // filtering only needs to work against the file_name component, not the full suffix. If
-        // `find_related_assets` was changed to also return files in subdirectories, we could
-        // use `PathBuf.strip_prefix` to remove the parent directory and then glob-filter
-        // against the remaining path. Note that the current behaviour effectively means that
-        // the `ignored_content` setting in the config file is limited to single-file glob
-        // patterns (no "**" patterns).
```
timakro, line 55: The changes are centered around this advice. Above, I replaced `read_dir` by `WalkDir` to collect assets recursively. And everywhere where the collected assets are accessed I replaced `.file_name()` with `.strip_prefix(&page.file.path.parent().unwrap())`. Note that I wanted to use `.strip_prefix(&page.file.parent)` originally but there is a special case for pages, where `parent` is actually `grand_parent`, that breaks this.

components/library/src/content/page.rs
```diff
@@ -282,24 +282,12 @@ impl Page {
...
-                path.pop();
-                path.push(filename);
-                path = path
-                    .strip_prefix(&base_path.join("content"))
-                    .expect("Should be able to stripe prefix")
-                    .to_path_buf();
-                path
-            })
-            .map(|path| format!("/{}", path.display()))
+            .map(|filename| format!("{}{}", self.path, filename))
```
timakro: This is assuming that the logic for pages which used to be much more complex than the logic for sections was unnecessarily complex even before my changes. Both functions are using the same logic after my changes.

components/rendering/src/markdown.rs
```diff
@@ -74,13 +74,6 @@ fn starts_with_schema(s: &str) -> bool {
     PATTERN.is_match(s)
 }
 
-/// Colocated asset links refers to the files in the same directory,
-/// there it should be a filename only
-fn is_colocated_asset_link(link: &str) -> bool {
-    !link.contains('/')  // http://, ftp://, ../ etc
-        && !starts_with_schema(link)
-}
-
```
timakro, line 83: We could adjust this to match files in colocated subfolders as well. But I'm proposing to get rid of this completely instead. Why change relative links to absolute links?

timakro:
I've missed some tests which are failing for obvious reasons now. (I don't know how to run all tests combined instead of each component separately.) Before resolving this let's first talk about the proposed change itself: if links to colocated files should be left relative or converted to absolute links.

timakro:
@Keats I'm worried you're holding off reviewing this because the checks are failing. Please take a look so we can decide on my proposed change. I will either undo my proposed change or adjust the unit tests then.

Keats:
> I'm worried you're holding off reviewing this because the checks are failing

I will have a look after 0.14.1, just focusing on fixing bugs rather than new features first. That and also holidays.

components/rendering/src/markdown.rs
Keats: I cannot remember

Keats:
Can you rebase your commit on the latest next branch?

timakro:
@Keats Done

timakro:
With the proposed change two tests would have to be adjusted 1890e387b1026f1556d5dc50c53586dfb25a897c. Still not sure if I'm missing something.

components/library/src/content/mod.rs
Keats: > there is a special case for pages, where parent is actually grand_parent, that breaks this.

Do you mean when you have an asset folder? That's expected then since otherwise the parent would refer to the page itself

Keats (approved):
Thanks!

components/library/src/content/mod.rs
timakro: Yeah, that's what I meant.

eberkund:
Amazing, this is the one feature I needed to migrate a site I'm working on to Zola 🎉

manu-cyber:
Hey @timakro , unfortounately this pull request seems to break a template of mine. It makes use of the child-page’s assets which cannot be rendered because their path is relative to the child page instead of the section. Could you please have a look?

The full discussion is in the forum: https://zola.discourse.group/t/assets-from-child-page-content-have-wrong-relative-paths-when-shown-in-sections/1181/

Issue #1570, opened 2021-07-31
Cachebust for Sass files broken in 0.14.0?


InputUsername:
# Bug Report

## Environment

Zola version: 0.14.0

## Expected Behavior
Linking to a Sass stylesheet `style.scss` that is compiled to `style.css`, using `get_url(path="style.css", cachebust=true)`, should produce a URL to this file. This works in 0.12.2.

## Current Behavior
In Zola 0.14.0, the exact same `get_url(path="style.css", cachebust=true)` gives the following error:
```text
Building site...
Checking all internal links with anchors.
> Successfully checked 0 internal link(s) with anchors.
-> Creating 3 pages (0 orphan), 1 sections, and processing 0 images
Error: Failed to render page '<site path>/content/about.md'
Reason: Failed to render 'page.html' (error happened in a parent template)
Reason: Function call 'get_url' failed
Reason: `get_url`: Could not find or open file style.css
```

The error disappears when removing `cachebust=true`.

## Step to reproduce
- Create Sass stylesheet `style.scss`;
- Link to it in a template using `<link rel="stylesheet" href="{{ get_url(path='style.css', cachebust=true) }}">`

Keats:
And it works without the cachebust? weird

InputUsername:
I ran a `git bisect` and found this:
```
9e0cac6d4d1623918f63e08848b9a7a219621289 is the first bad commit
commit 9e0cac6d4d1623918f63e08848b9a7a219621289
Author: Vincent Prouillet <balthek@gmail.com>
Date:   Thu May 13 22:20:47 2021 +0200

    Cleanup of get_url and get_file_hash

 CHANGELOG.md                                     |   4 +-
 components/site/src/tpls.rs                      |  12 +-
 components/templates/src/global_fns/files.rs     | 435 +++++++++++++++++++++
 components/templates/src/global_fns/helpers.rs   |  36 ++
 components/templates/src/global_fns/images.rs    |   5 +-
 components/templates/src/global_fns/load_data.rs |  36 +-
 components/templates/src/global_fns/mod.rs       | 473 +----------------------
 test_site/themes/sample/templates/index.html     |   2 +-
 8 files changed, 501 insertions(+), 502 deletions(-)
 create mode 100644 components/templates/src/global_fns/files.rs
 create mode 100644 components/templates/src/global_fns/helpers.rs
```

lcowell:
[...]

PR #1586, opened 2021-08-14
Fix cachebust for files in the output path

InputUsername:
The `search_for_file` helper function now accepts an ~optional~ output path. If passed, the file will also be searched there.
This is used in the `get_url` Tera function to search in the `Site::output_path`.

In practice, this means cachebust works again for files in the output path.
I've tested this with one of my sites and it seems to work perfectly.

This *should* fix the main issue of #1570. That issue's thread also discusses other things related to getting urls of files in the output path though, so you might want to keep it open.

Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/global_fns/helpers.rs
+++ b/components/templates/src/global_fns/helpers.rs
@@ -19,12 +19,9 @@ pub fn search_for_file(
     base_path: &Path,
     path: &str,
     theme: &Option<String>,
-    output_dir: Option<&Path>,
+    output_path: &Path,
 ) -> Result<Option<(PathBuf, String)>> {
-    let mut search_paths = vec![base_path.join("static"), base_path.join("content")];
-    if let Some(o) = output_dir {
-        search_paths.push(base_path.join(o));
-    }
+    let mut search_paths = vec![base_path.join("static"), base_path.join("content"), base_path.join(output_path)];
     if let Some(t) = theme {
         search_paths.push(base_path.join("themes").join(t).join("static"));
     }
--- a/components/templates/src/global_fns/helpers.rs
+++ b/components/templates/src/global_fns/helpers.rs
@@ -5,11 +5,12 @@ use errors::{bail, Result};
 use utils::fs::is_path_in_directory;
 
 /// This is used by a few Tera functions to search for files on the filesystem.
-/// This does try to find the file in 3 different spots:
+/// This does try to find the file in 5 different spots:
 /// 1. base_path + path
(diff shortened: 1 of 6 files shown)


components/templates/src/global_fns/helpers.rs
```diff
@@ -18,8 +19,12 @@ pub fn search_for_file(
     base_path: &Path,
     path: &str,
     theme: &Option<String>,
+    output_dir: Option<&Path>,
```
Keats: in which case do we not have an output_dir?

Keats (commented):
I think we should always pass the output_dir so we remove the Option

components/templates/src/global_fns/helpers.rs
InputUsername: I assumed it would only be needed for `get_url`. Removing the option requires changing a lot more Tera functions (like `load_data`) that don't use the `output_dir`. But if you prefer removing the option, I will do that

Keats: I think all the functions operating on files should behave the same so it would be better to change all of them

InputUsername: Fair enough! I've added this now.

Keats (changes requested):
Can you add a test? It could be added in one of the existing ones

InputUsername:
> Can you add a test? It could be added in one of the existing ones

Sure thing. Where should I add it exactly? Do I add a test to `components/templates/src/global_fns/files.rs`? Or should I add one for `load_data`, `get_file_hash` etc too?

I was thinking a test like this:
```rust
#[test]
fn can_link_to_file_in_output_path() {
    let dir = create_temp_dir();
    let public = dir.path().join("public");
    create_dir(&public).expect("Failed to create output directory");
    create_file(&public.join("style.css"), "// Hello world")
        .expect("Failed to create file in output directory");

    let static_fn = GetUrl::new(dir.path().to_path_buf(), Config::default(), HashMap::new(), public);
    let mut args = HashMap::new();
    args.insert("path".to_string(), to_value("style.css").unwrap());
    assert_eq!(static_fn.call(&args).unwrap(), "http://a-website.com/style.css");
}
```

Sorry for all the questions :sweat_smile:

Keats:
Just one is fine, the test you wrote seems ok

InputUsername:
> Just one is fine, the test you wrote seems ok

Thanks, done.

Keats (approved):

Keats:
Thanks!

PR #1606, opened 2021-08-31
Fix docs by moving `extra_syntaxes` to `[markdown]`

marier-nico:
Sanity check:

**Description**
After some trial and error and issue-browsing, it looks like the correct location in the configuration for `extra_syntaxes` is actually the `[markdown]` section, and not the main section, so I just wanted to update the docs to make sure they were accurate :smile:


Keats (approved):

Keats:
Thanks!

PR #1608, opened 2021-09-01
[bugfix] config.default_language is exposed to templates

southerntofu:
Hello! I noticed a regression since v0.14.0: `config.default_language` was no longer usable in templates. That was not listed in breaking changes. Is there a reason for not including it? Otherwise, here's a patch.

I personally use that in [my theme](https://tildegit.org/southerntofu/zola-water) and in a [shortcode](https://codeberg.org/joinjabber/website/src/branch/main/templates/shortcodes/server.html#L14).


Keats:
Definitely an oversight!

PR #1611, opened 2021-09-04
Add OrgaTalk wiki to list of example sites

homeworkprod:
We're trying to use Zola as a wiki. Seems to work well so far! :D


Keats:
That's cool! I want to make my personal wiki with Zola but I need to add the backlinks first

PR #1620, opened 2021-09-14
Strip all occurrences of `<pre data-shortcode>...</pre>`

hannobraun:
Fix the code that strips `<pre data-shortcode>...</pre>` from HTML code that was generated from Markdown. Currently, if there are multiple occurrences of that `pre` element in the same paragraph, only the first one is stripped. This pull request extracts the offending code into a separate function, adds unit tests, and fixes the function so it strips all occurrences.

This pull request conflicts with #1564, which is a larger rewrite of the shortcode logic. I haven't tested whether that PR has the same bug, but since it's a larger change that might not be merged soon (I don't know), I think it makes sense to fix this issue separately in any case. It seems that #1564 removes the code I'm fixing here, so it should be easy to merge this change.


Keats:
I'll be merging #1564 for the next release so let's not add potential merge conflicts/changes that won't be used.

Issue #1361, opened 2021-02-18
Custom site root doesn't work on Windows


southerntofu:
From [the forums](https://zola.discourse.group/t/first-step-failed/794)

Windows 10, zola 0.13.0

It appears the `-r` flag is broken on Windows, at least with absolute links, like so:

```
zola.exe -r "D:\Dev\Zola\zola-clean-blog" build
```

Keats:
This feels like https://github.com/getzola/zola/blob/master/src/main.rs#L15-L20 should work 🤔

Keats:
Is it still happening with 0.14?

warsus:
I had the same problem on master branch [0] and did a little analysis.
It appears to me that (at least in my case) the problem might be that the windows canonical path looks like this:

root_dir: \\?\C:\Users\user\my_site

My guess is the questionmark in the prefix will later be interpreted as a globbing character. 
Tera::parse will create no matches then.

I kinda confirmed this by removing the "canonicalize" call in the root_dir assignment, which fixed the problem for me. 

    let root_dir = match matches.value_of("root").unwrap() {
        "." => env::current_dir().unwrap(),
        path => PathBuf::from(path)
            .canonicalize()
            .unwrap_or_else(|_| panic!("Cannot find root directory: {}", path)),
    }; 

[0] commit 9946506e1c363a35f097a7b220c613f77a989c43 (HEAD -> master, origin/master, origin/HEAD)

Keats:
@warsus thanks for digging!
I can't remember why we're canonicalizing it to be honest. If you can do a PR with that change on the `next` branch, I'll test on macos.

warsus:
@Keats 
Sure i will try to do a PR later today. Also i think my assumption about the questionmark was not completly right.
But i still think it's something about the \\? prefix, might do some further digging in the tera lib afterwards.

PR #1624, opened 2021-09-22
fix passing a site root with -r/--root parameter on windows

warsus:
#1361 

the "\\\\?" prefix from the canonical path seems to break wildcard matching


Keats:
Can you do the PR on the `next` branch instead?

warsus:
Sure, sorry!
i just saw that this will not work with relative paths. For example "cargo run -- -r ..\my_site serve"
Serve command prints the following, where the path does not look right to me:
Listening for changes in ..\my_site{..\my_site\config.toml, content, sass, static, templates, themes}

warsus:
I think i'm gonna have to do a new PR, unless you want to take it from here. 
Windows user here obviously 😆

PR #1636, opened 2021-10-10
Add corona-effekt.orgatalk.de to list of example sites

homeworkprod:
The site offers a manually maintained list of LAN party cancellations due to COVID-19.


Keats:
A very sad website! Thanks.

PR #1640, opened 2021-10-19
Shortcodes

Keats:


components/rendering/src/range_relation.rs
```diff
@@ -0,0 +1,78 @@
+use std::ops::Range;
```
Keats: @coastalwhite this file doesn't seem used?

components/rendering/src/context.rs
```diff
@@ -28,6 +31,35 @@ impl<'a> RenderContext<'a> {
     ) -> RenderContext<'a> {
         let mut tera_context = Context::new();
         tera_context.insert("config", &config.serialize(lang));
+
+        let mut shortcode_definitions = HashMap::new();
+
+        for (identifier, template) in tera.templates.iter() {
```
Keats: we can do that only once, no need to do it on every render

components/rendering/src/markdown.rs
```diff
@@ -178,43 +202,85 @@ pub fn markdown_to_html(content: &str, context: &RenderContext) -> Result<Render
...
+                            if range.contains(&span.start) {
+                                if range.start != span.start {
+                                    events.push(return_text(
+                                        text[..(span.start - range.start)].into(),
+                                        &mut code_block,
+                                        context,
+                                    ));
+                                }
+
+                                //TODO: remove unwrap
```
Keats: TODO: set the error value if there is one when rendering the shortcode

Keats (commented):
It's pretty long! I will need to block some weekend to work on that

components/rendering/src/range_relation.rs
coastalwhite: Range relation is indeed not used anymore. I believe Transform is also either not used anymore or rarely used

Keats:
I think it's possible to unify all the lexers into a single one (or two). I'll try that next time I have some free time

coastalwhite:
Just to give some context on why I used multiple lexers, instead as little as possible. When looking at logos, I was doubting whether either approach was better, so I had a look at how the logos main contributor used it themselves. It turned out they has a very similar scenario where he used multiple lexers. I do believe it provides some readability and modularity, and thus maintainability. That is just my two cents.

You decide what you do with that information; either way it would work quite efficiently probably and would be a huge improvement over the old code.

Keats:
I've removed the string_literal file for now, `'he'ho'` is something that should not work. My current fix do lose the possibility of escaping strings manually by the user but that's why we have 3 types of string markers, so you can pick one that won't cause issues.

Keats:
Looks like quite a few tests were failing but they were not ran before the unignore/integration tests so it wasn't shown. I'm looking into those. I also have the ignored shortcode working in a branch but fixing all the tests first.

Keats:
@coastalwhite I ended up rewriting the lexer/parser to something that I understood a bit better since I'll probably need to extend it soon-ish.
The next step is to move all the shortcode tests to one file so they are easy to find but then it should be close to ready.

Keats:
I added tests for the issues I found on the repo and they all passed 🎊 
Overall fantastic job @coastalwhite it looks like it behaves perfectly.
I think i'll merge it as is in the `next` branch and maybe tweak some of the tests slightly and see if i can reduce the duplication in `markdown.rs` between Event::Text and Event::Html

coastalwhite:
Perfect to hear! It looks really good! Hopefully this will make the shortcode system less buggy. Especially in the long run. Did you perform a benchmark by any chance. I am quite curious how the performance difference is after this change.

Keats:
It was faster from what I remember but I'll run the benchmarks before merging

Keats:
One issue I just found while trying to render the docs: it doesn't ignore the shortcode-like things in code blocks. I'm going to see if it's hard to add that.

Keats:
So we have 2 choices:

1. Ignore everything in code span/blocks, this means users cannot use shortcode in there
2. Do like the previous parser was doing and only try to render shortcodes if they are valid, otherwise return the output as is. This means we cannot give errors back

What do you think? I don't think many people would care if 1 is implemented but I'm not sure, that's a change from how it was before.

coastalwhite:
> One issue I just found while trying to render the docs: it doesn't ignore the shortcode-like things in code blocks. I'm going to see if it's hard to add that.

I actually took that to be the expected behaviour. Probably for backwards compatibility it should be fixed though.

coastalwhite:
> So we have 2 choices:
> 
> 
> 
> 1. Ignore everything in code span/blocks, this means users cannot use shortcode in there
> 
> 2. Do like the previous parser was doing and only try to render shortcodes if they are valid, otherwise return the output as is. This means we cannot give errors back
> 
> 
> 
> What do you think? I don't think many people would care if 1 is implemented but I'm not sure, that's a change from how it was before.

I would imagine, since it preprocesses, cases in which it is ignored should be minimized. I am not sure how many problems this will cause though.

Keats:
So what the current parser does is only consider something as a shortcode if it's a valid one. The docs have a lot of example Tera snippet where `{{` and `{%` are present a lot and we can't really ask people to not use those characters in their text.

After sleeping on it, I'm thinking maybe the move to Logos was jumping the shark a bit as the current parser does _exactly_ what we want, and it was the replacing in the source content that was the main issue. We can probably reuse the pest parser to create the `Shortcode` like the current parser does and we would have the parsing "for free". I'll think about it a bit more but that seems like a more reasonable approach than rewriting the full parser from scratch.

eberkund:
> So we have 2 choices:
> 
> 1. Ignore everything in code span/blocks, this means users cannot use shortcode in there
> 2. Do like the previous parser was doing and only try to render shortcodes if they are valid, otherwise return the output as is. This means we cannot give errors back
> 
> What do you think? I don't think many people would care if 1 is implemented but I'm not sure, that's a change from how it was before.

I am not a heavy user of code blocks but the first option is how I would expect it to work.

Keats:
https://github.com/getzola/zola/pull/1665 seems to work well and people with `{%` or `{{` in their text do not need to escape it all the time

PR #1650, opened 2021-10-30
Add check for export highlight theme CSS in Config

### LINKED ISSUE #1635: Crash when using a theme that doesn't exist in `highlight_themes_css`
# Bug Report

Like #1299 but for `highlight_themes_css`

## Environment

Zola version: 0.14.1

## Expected Behavior

```toml
[markdown]
highlight_code = true
highlight_theme = "css"
highlight_themes_css = [
  { theme = "one-dark", filename = "syntax-theme-dark.css" },
  { theme = "one-light", filename = "syntax-theme-light.css" },
]
```

Since one of the above themes doesn't exist, the program should gracefully exit after displaying an error message (i.e., not by panicking).

## Current Behavior

```
Building site...
> Successfully checked ██ internal link(s) with anchors.
-> Creating ███ pages (0 orphan), ██ sections, and processing ██ images
thread 'main' panicked at 'no entry found for key', components/config/src/highlighting.rs:93:18
```

peterkos:
## Background

This PR fixes #1635.

The markup init method `init_extra_syntaxes_and_highlight_themes` checked to see if the specified `highlight_theme` existed, but did not check any attempts to export theme CSS, in `highlight_themes_css`.

(Notably, this logic used to be in `Config::parse()`, but was refactored in #1499 to accomodate custom themes.)

## Contents
- Check `THEME_SET` and `extra_theme_set` for each theme attempted to export
- Give user readable message if this is not the case.

## Notes
- I didn't notice any tests in the `config` module; probably best for a different PR, to prevent regressions?
- I'm not too sold on the the 3 nested `if` logic, coming from a Swift background; if there's a cleaner way in Rust I'm happy to change it :)

## Template Checks
Sanity check:

N/A

### DIFF
--- a/components/config/src/config/markup.rs
+++ b/components/config/src/config/markup.rs
@@ -139,7 +139,7 @@ impl Markdown {
         // Validate that all exported highlight themes exist as well
         for theme in self.highlight_themes_css.iter() {
             let theme_name = &theme.theme;
-            if !(THEME_SET.themes.contains_key(theme_name)) {
+            if !THEME_SET.themes.contains_key(theme_name) {
                 // Check extra themes
                 if let Some(extra) = &*self.extra_theme_set {
                     if !extra.themes.contains_key(theme_name) {
--- a/components/config/src/config/markup.rs
+++ b/components/config/src/config/markup.rs
@@ -121,7 +121,7 @@ impl Markdown {
             self.extra_theme_set = Arc::new(Some(extra_theme_set));
         }
 
-        // validate that the chosen highlight_theme exists in the loaded highlight theme sets
+        // Validate that the chosen highlight_theme exists in the loaded highlight theme sets
         if !THEME_SET.themes.contains_key(&self.highlight_theme) {
             if let Some(extra) = &*self.extra_theme_set {
                 if !extra.themes.contains_key(&self.highlight_theme) {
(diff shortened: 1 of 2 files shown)


components/config/src/config/markup.rs
```diff
@@ -136,6 +136,20 @@ impl Markdown {
             }
         }
 
+        // Validate that all exported highlight themes exist as well
+        for theme in self.highlight_themes_css.iter() {
+            let theme_name = &theme.theme;
+            if !(THEME_SET.themes.contains_key(theme_name)) {
```
Keats: you don't need the braces

components/config/src/config/markup.rs
```diff
@@ -136,6 +136,20 @@ impl Markdown {
...
+                // Check extra themes
+                if let Some(extra) = &*self.extra_theme_set {
```
Keats, line 144: You can probably do something like:

```
if self.extra_theme_set.map(|e| extra.themes.contains_key(theme_name)).unwrap_or(false) {
   bail!(...)
}
```

Keats (changes requested):
There are tests in `config/mod.rs`, you can definitely add one there

components/config/src/config/markup.rs
peterkos: I'm a little confused on how this cleans things up -- imo its a little cleaner to unwrap the `self.extra_theme_set` and _then_ check if it contains our specified element; how does the `.map` work here?

Keats: It will run what's in the `map` only if `self.extra_theme_set` is not `None`. Unless i missed something it should be equivalent to your code. It's up to personal preference after that :)

Keats (approved):

Keats:
Thanks!

peterkos:
Likewise! Apologies, forgot about this PR. I played around with the map but this just felt easier :)

Issue #1630, opened 2021-10-01
Generic taxonomy template


marcodpt:
This is by far in my opinion the best static site generator that i worked with.

Thanks a lot for the awesome work!

In the documentation it says : 
Zola will look up the following files in the templates directory:

    $TAXONOMY_NAME/single.html
    $TAXONOMY_NAME/list.html

I want to know if it is possible to use a fallback to ./single.html and ./list.html in case $TAXONOMY_NAME is not present.
For example, look at this theme:
[Zulma](https://github.com/Worble/Zulma/tree/master/templates)

It would work out of the box perfectly with a **movie** site with **directors** and **release** instead of **authors**, **catogories** and **tags**

Themes working out of the box to most use cases imaginable is a great thing for a static site generator

Keats:
Thanks!

This is not possible currently but I agree it would make sense.

BezPowell:
> Thanks!
> 
> This is not possible currently but I agree it would make sense.

I've just had a go at implementing this, and while it does work, the code is a bit messy. It also relies on running a string.ends_with() each time render_template() is called which I imagine would slow it down. My updated code for render_template() is below, if there's some interest in merging it I am happy to open a pull request.
```
pub fn render_template(
    name: &str,
    tera: &Tera,
    context: Context,
    theme: &Option<String>,
) -> Result<String> {
    // check if it is in the templates
    if tera.templates.contains_key(name) {
        return tera.render(name, &context).map_err(std::convert::Into::into);
    }

    // check if it is part of a theme
    if let Some(ref t) = *theme {
        let theme_template_name = format!("{}/templates/{}", t, name);
        if tera.templates.contains_key(&theme_template_name) {
            return tera.render(&theme_template_name, &context).map_err(std::convert::Into::into);
        }
    }

    // check if it is part of ZOLA_TERA defaults
    let default_name = format!("__zola_builtins/{}", name);
    if tera.templates.contains_key(&default_name) {
        return tera.render(&default_name, &context).map_err(std::convert::Into::into);
    }

    // check if it is a taxonomy list
    if name.ends_with("list.html") {
        // Check for generic list.html
        if tera.templates.contains_key("list.html") {
            return tera.render("list.html", &context).map_err(std::convert::Into::into);
        }
        // Check for theme generic list.html
        else if let Some(ref t) = *theme {
            let theme_template_name = format!("{}/templates/list.html", t);
            if tera.templates.contains_key(&theme_template_name) {
                return tera
[...]

PR #1651, opened 2021-11-01
Add generic template fallbacks to taxonomies

BezPowell:

Sanity check:

## Code changes

If the change is a new feature or adding to/changing an existing one: adding to an existing feature.

## Description
This PR adds support for generic fall-back templates for taxonomies, as per issue #1630. It currently splits off the template fall-back logic into a separate function, and then utilises that when rendering a taxonomy to check whether there is a taxon-specific template available and, if not, fallback to a generic `list.html` or `single.html` file in the `template` directory root.

Getting it to work with paginated taxonomies did require that I extend the number of arguments on the `Paginator::from_taxonomy()` function which may not be desirable. Hopefully other might be able to suggest and alternate implementation, or comment on whether what I have done is worth it for the extra utility.

### DIFF
--- a/components/utils/src/templates.rs
+++ b/components/utils/src/templates.rs
@@ -95,6 +95,8 @@ pub fn check_template_fallbacks<'a>(
 
 #[cfg(test)]
 mod tests {
+    use crate::templates::check_template_fallbacks;
+
     use super::rewrite_theme_paths;
     use tera::Tera;
 
@@ -118,6 +120,25 @@ mod tests {
         assert_eq!(
             tera.templates["hyde/templates/child.html"].parent,
             Some("index.html".to_string())
+        );
+    }
+
+    #[test]
+    fn template_fallback_is_successful() {
+        let mut tera = Tera::parse("test-templates/*.html").unwrap();
+        tera.add_raw_template(&"hyde/templates/index.html", "Hello").unwrap();
+        tera.add_raw_template(&"hyde/templates/theme-only.html", "Hello").unwrap();
+
+        // Check finding existing template
+        assert_eq!(check_template_fallbacks("index.html", &tera, &None), Some("index.html"));
+
+        // Check trying to find non-existant template
+        assert_eq!(check_template_fallbacks("not-here.html", &tera, &None), None);
+
+        // Check theme fallback
+        assert_eq!(
+            check_template_fallbacks("theme-only.html", &tera, &Some("hyde".to_string())),
+            Some("hyde/templates/theme-only.html")
         );
     }
 }
--- a/components/utils/src/templates.rs
+++ b/components/utils/src/templates.rs
@@ -25,23 +25,8 @@ pub fn render_template(
     context: Context,
     theme: &Option<String>,
 ) -> Result<String> {
-    // check if it is in the templates
-    if tera.templates.contains_key(name) {
-        return tera.render(name, &context).map_err(std::convert::Into::into);
-    }
-
-    // check if it is part of a theme
-    if let Some(ref t) = *theme {
-        let theme_template_name = format!("{}/templates/{}", t, name);
-        if tera.templates.contains_key(&theme_template_name) {
-            return tera.render(&theme_template_name, &context).map_err(std::convert::Into::into);
-        }
-    }
-
-    // check if it is part of ZOLA_TERA defaults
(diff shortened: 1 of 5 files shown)


components/utils/src/templates.rs
```diff
@@ -78,6 +63,32 @@ pub fn rewrite_theme_paths(tera_theme: &mut Tera, theme: &str) {
     tera_theme.templates.extend(new_templates);
 }
 
+/// Checks for the presence of a given template. If none is found, also looks for a 
+/// fallback in theme and default templates. Returns the path of the most specific
+/// template found, or none if none are present.
+pub fn check_template_fallbacks(name: &str, tera: &Tera, theme: &Option<String>) -> Option<String> {
```
Keats: can you return a `&str` instead?

docs/content/documentation/templates/taxonomies.md
```diff
@@ -3,11 +3,15 @@ title = "Taxonomies"
...
 
-Zola will look up the following files in the `templates` directory:
+Zola will look up the following, taxon-specific files in the `templates` directory:
 
 - `$TAXONOMY_NAME/single.html`
 - `$TAXONOMY_NAME/list.html`
 
+if they are not found, it will attempt to fall back on the following generic template files:
+- `single.html`
+- `list.html`
```
Keats: can we call them `taxonomy_single.html` and `taxonomy_list.html`? So you know exactly what they are from the name

components/utils/src/templates.rs
```diff
@@ -25,23 +25,8 @@ pub fn render_template(
...
-        if tera.templates.contains_key(&theme_template_name) {
-            return tera.render(&theme_template_name, &context).map_err(std::convert::Into::into);
-        }
-    }
-
-    // check if it is part of ZOLA_TERA defaults
-    let default_name = format!("__zola_builtins/{}", name);
-    if tera.templates.contains_key(&default_name) {
-        return tera.render(&default_name, &context).map_err(std::convert::Into::into);
+    if let Some(template) = check_template_fallbacks(name, tera, theme) {
```
Keats, line 28: I'm wondering if adding an optional `fallback` param to that function would simplify things, so you don't have to call `check_template_fallbacks` manually would be worth it. What do you think?

BezPowell: That does sound like a good idea. Would you envisage passing an `Option<&str>` as the fallback and just running the tests twice? It would be nice to avoid duplicating the code, but aside from making a `Vec<&str>` and iterating over that I can't think of a way of doing so off the top of my head.

docs/content/documentation/templates/taxonomies.md
BezPowell: Absolutely, just pushed a commit now.

components/utils/src/templates.rs
BezPowell: I'd tried doing that initially, and after wrestling with the borrow checker gave up. Clearly, trying to make a pull request when ill is not a good idea as it worked much better this morning.

Thank you for reminding me, commit just pushed now.

components/utils/src/templates.rs
Keats: I'm trying to remember what I had in mind... but I think it was `Option<&str>`

BezPowell: Having a go at that now. If we implement a fallback would we throw and error when constructing the Paginator?
Thinking of these lines in components/library/src/pagination/mod.rs:
```
// Check for taxon-specific template, or use generic as fallback.
let specific_template = format!("{}/single.html", taxonomy.kind.name);
let template = match check_template_fallbacks(&specific_template, tera, theme) {
    Some(template) => template,
    None => "taxonomy_single.html",
};
```
With a fallback param check_template_fallbacks will now presumably always return the template if it exists. Currently paginated taxonomies only error when attempting to render. I could convert the match into an unwrap, but that would move the location where the logic fails.

Sorry to be a pain, just don't want to do something that might make later re-factoring more difficult.

Keats:
Sorry I missed that before, can you add tests?

BezPowell:
> Sorry I missed that before, can you add tests?

I've just added some to the `check_template_fallbacks()` function. Were there any more you feel would help?

Keats:
That's good thanks!

PR #1652, opened 2021-11-05
update hugo graph

bacongobbler:
Hugo has [internal link check](https://gohugo.io/content-management/cross-references/) support through `ref` and `relref`, and there are community projects available for [external link checking](https://github.com/marketplace/actions/hugo-broken-link-check).

> Hugo emits an error or warning if a document cannot be uniquely resolved.

Search is available through third party integrations like [agolia](https://www.algolia.com/).

Breadcrumb support is implemented using a code snippet that is available through the [Hugo Codex](https://hugocodex.org/add-ons/breadcrumbs/).

Signed-off-by: Matthew Fisher <matt.fisher@fermyon.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,19 @@ in the `docs/content` folder of the repository and the community can use [its fo
 | Themes                          | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Shortcodes                      | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Internal links                  | ![yes] | ![no]  | ![yes] | ![yes]  |
-| Link checker                    | ![yes] | ![no]  | ![no]  | ![yes]  |
+| Link checker                    | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Table of contents               | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Automatic header anchors        | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Aliases                         | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Pagination                      | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Custom taxonomies               | ![yes] | ![no]  | ![yes] | ![no]   |
-| Search                          | ![yes] | ![no]  | ![no]  | ![yes]  |
+| Search                          | ![yes] | ![no]  | ![ehh] | ![yes]  |
 | Data files                      | ![yes] | ![yes] | ![yes] | ![no]   |
 | LiveReload                      | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Netlify support                 | ![yes] | ![no]  | ![yes] | ![no]   |
 | Vercel support                  | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Cloudflare Pages support        | ![yes] | ![no]  | ![yes] | ![yes]  |
-| Breadcrumbs                     | ![yes] | ![no]  | ![no]  | ![yes]  |
+| Breadcrumbs                     | ![yes] | ![no]  | ![yes] | ![yes]  |
 | Custom output formats           | ![no]  | ![no]  | ![yes] | ![no]   |
 
 ### Supported content formats


Keats:
The README only contains things that don't rely on third parties.
The link checker row should probably be split in two: internal and external,  Hugo should get a check for internal but not for external.
Search is requiring a third party so they would still get a cross.
As for breadcrumbs, the template https://raw.githubusercontent.com/jhvanderschee/hugocodex/main/layouts/partials/breadcrumbs.html relies on the URL and the site organization might not match the URL so it's probably a "eh". Can you get the parent sections in Hugo?

bacongobbler:
> Can you get the parent sections in Hugo?

Yes: https://gohugo.io/content-management/sections/#section-page-variables-and-methods

> The README only contains things that don't rely on third parties.

The title is literally "[Comparisons with other static site generators](https://github.com/getzola/zola#comparisons-with-other-static-site-generators)". To me that means if the static site generator can support these features either through first-party features or plugins provided by the community, then it should be checked off.

When approaching the graph, the question I asked myself was "Is it built into core? Have developers found solutions through other means?" in which case you could argue that Hugo supports everything listed in the graph.

There was no mention that the graph only contains features that don't rely on third parties. So in your view, plugins or community projects don't count?

> The link checker row should probably be split in two: internal and external

Please feel free to do that. I do not have the context necessary to know what features Zola, Cobalt, and Pelican support. I'm just commenting on perceived misinformation about Hugo.

An aside: the row about having a "Sane & powerful template engine" is clearly biased because it's hard to objectively state whether it's "good" or not. The comparison is subjective in nature. If you're preparing a comparison chart, I'd suggest sticking to things you can objectively compare. Commenting on subjective topics like how "sane" a certain feature works only discredits the rest of the graph.

Keats:
> There was no mention that the graph only contains features that don't rely on third parties. So in your view, plugins or community projects don't count?

If it works with the binary sure and I can run it on my computer, sure. https://github.com/marketplace/actions/hugo-broken-link-check is a GH action, that's not a plugin. If you look at the pelican side, it does mention lots of the features come from plugins.
I don't consider third party services to be a check on that graph, otherwise it would be all checks for everything since you can always hack things.
For the search for example: Algolia for example will not let you use their services for free if you generate a lot of requests but you could set up your own Solr/ES instance instead. You could also export your data as JSON and build the index in JS but that would not be super speedy depending on its size. So I guess for the search it can be checked everywhere by that logic?

> An aside: the row about having a "Sane & powerful template engine" is clearly biased because it's hard to objectively state whether it's "good" or not. The comparison is subjective in nature. If you're preparing a comparison chart, I'd suggest sticking to things you can objectively compare. Commenting on subjective topics like how "sane" a certain feature works only discredits the rest of the graph.

But the Golang template engine is _clearly_ insane. I would pick regex replace over it. It does mention it is biased, this tool wouldn't exist otherwise after all. Imagine my immeasurable sadness when I realized Helm was using go templates.

bacongobbler:
> https://github.com/marketplace/actions/hugo-broken-link-check is a GH action, that's not a plugin.

That github action uses a library called [broken-link-checker](https://github.com/stevenvachon/broken-link-checker), which is available in binary form (`blc`). So if we're debating your original argument...

> If it works with the binary sure and I can run it on my computer, sure.

You can run `blc` on your computer. By extension, Hugo supports external URL checks through community projects. To that end, Hugo (and Cobalt) should either get an ✔️ or an 🟡 at the very least. But not an ❌.

Again, the question I asked myself was "can a developer do X with Y?". To which the answer is "yes, developers can check for broken external links with Hugo. It's not built into Hugo, but here's how other community members have done it."

> But the Golang template engine is _clearly_ insane.

It's not clear to me what is perceived about text/template as "insane". It has its limitations, sure, but so does every other template language. Every time I've used it in other projects, text/template accomplished the job it was sought out to achieve, which is to provide input data to a template and print the results. It only gets complicated when you start to write Turing-complete templates. And in those cases, the developer is using the wrong tool for the job. Writing complicated branch logic in a Helm chart when you should be using other tools is one of them.

> Imagine my immeasurable sadness when I realized Helm was using go templates.

Huh... I guess you picked up on the fact that I'm a Helm maintainer.

If you have feedback about the project, I invite you to join the conversation on Slack.

I'm really upset because I was really looking forward to using Zola on our next project. I just came here to update a few points on a graph that appeared incorrect since I've used Hugo for a number of years. But I don't want to be part of this community if this is how other community projects are perceived by the maintainers.

Keats:
> Huh... I guess you picked up on the fact that I'm a Helm maintainer.

Haha not at all, I didn't look at your profile or anything. It's just that we moved some stuff from kustomize to helm at work last week so it's fresh on my mind.

Keats:
Either way, I've removed the comparison entirely in https://github.com/getzola/zola/pull/1597/commits/ea3e782dfb8504a9b96138ebf676d356228ce692 it's not fair if I don't update it anyd I haven't looked at Hugo, Pelican or Cobalt in years.

PR #1672, opened 2021-12-01
fixes #1599

### LINKED ISSUE #1599: Error: Was not able to copy file  **
# Bug Report

## Environment

windows 10
Zola version:0.14.0

## Expected Behavior
filename.md add many images_resize short_code:
```
{% set image = resize_image(path=path, width=width, op=op,format="jpg") %}
<img src="{{ image.url }}" />
```

filename.md
```
{{ resize_image(path="/entertainment/filename/1.png", width=728, op="fit_width") }}
{{ resize_image(path="/entertainment/filename/2.png", width=728, op="fit_width") }}
...
{{ resize_image(path="/entertainment/filename/12.png", width=728, op="fit_width") }}
png , image size > 1M

```

## Current Behavior
Failed to build the site
Error: Was not able to copy file D:/web/www/temp/site_framework/content/entertainment\filename\2.png to D:\web\www\temp\site_framework\public\entertainment\filename\2.png

Reason:Another program is using this file and the process cannot access it(os error 32)

## Step to reproduce
Build site succeed by chance, I guess image is not normal close or async.
But zola serve is ok.

mtolk:
no longer associate assets from subfolders, only the current folder.
This caused two threads writing creating the same file. when this happens at the same time zola build on windows crashes.

I've ran the test-suite about 10 times no longer seen the error.

Sanity check:

* [ x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* [ x ] Are you doing the PR on the `next` branch?

### DIFF
--- a/components/library/src/content/mod.rs
+++ b/components/library/src/content/mod.rs
@@ -3,9 +3,8 @@ mod page;
 mod section;
 mod ser;
 
+use std::fs::read_dir;
 use std::path::{Path, PathBuf};
-
-use walkdir::WalkDir;
 
 pub use self::file_info::FileInfo;
 pub use self::page::Page;
@@ -32,19 +31,32 @@ pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
 /// file. Those will be copied next to the rendered .html file
 pub fn find_related_assets(path: &Path, config: &Config) -> Vec<PathBuf> {
     let mut assets = vec![];
-
-    for entry in WalkDir::new(path).into_iter().filter_map(std::result::Result::ok) {
-        let entry_path = entry.path();
-        if entry_path.is_file() {
-            match entry_path.extension() {
-                Some(e) => match e.to_str() {
-                    Some("md") => continue,
-                    _ => assets.push(entry_path.to_path_buf()),
-                },
-                None => continue,
-            }
-        }
-    }
+	
+	if path.is_dir() {
+		match read_dir(path) {
+			Ok(d) => {
+				for entry_path_res in d {
+					match entry_path_res {
+						Ok(entry) => {
+							let entry_path = entry.path();
+							if entry_path.is_file() {
+								match entry_path.extension() {
+									Some(e) => match e.to_str() {
+										Some("md") => {},
+										_ => assets.push(entry_path.to_path_buf()),
+									},
+									None => {},
+								}
+							}
+						}
+						_ => {}
+					}
+					
+				}
+			},
+			_ => {},
+		}	
+	}
 
     if let Some(ref globset) = config.ignored_content_globset {
         assets = assets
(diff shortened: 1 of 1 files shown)


Keats:
cc @timakro looking at that change it does make sense to me to ignore the subdirectories and it wouldn't break anything, correct?

timakro:
@mtolk @Keats On first glance it seems to me this effectively reverts #1582. Changing the asset scanning from flat (`read_dir`) to recursive (`WalkDir`) [was the first step to making subfolder colocation work](https://github.com/getzola/zola/pull/1582#discussion_r684779229). Note also that I [added two dummy files in the tests to specifically test subfolder colocation](https://github.com/getzola/zola/pull/1582/files#diff-3a59dd52e702c05926e72f5dceee4f81565a4b3fb5eacc25829f5c858926089fL77-L90), increasing the number of assets from 3 to 4, which this PR reverts as well. I might be missing something; only way to know for sure is testing.

Keats:
It looks like it still works: https://github.com/getzola/zola/pull/1597#issuecomment-985246142

But I see the issue. The issue is that section assets are poorly defined so using walkdir on a section would just mark every single page asset as their own.
I think the right fix is to use `Walkdir` for pages and `read_dir` for sections? Honestly at some point we should just ignore assets in the same folder as the section and force the use of a folder like we do for pages.

Keats:
I think https://github.com/getzola/zola/pull/1677 might work?

eberkund:
Do I understand correctly that the difficulty with this feature is that it's a little tricky to figure out how a file should be treated because it differs based on (not sure which factors?)

Keats:
So a section and a page can have colocated assets. A page has to be in its own folder while assets for a sections are at its level (not consistent, it will need to be fixed, in an ulterior version).
Imagine you have the following folders:

```
content
   - blog
      - index.md
      - banner.gif
      -  an_article
          - index.md
          - image.jpg
          -  example_js
            - data.js
```

When looking for the assets for the `an_article` page, it can search recursively in the folder thanks to @timakro PR since it's a page, there's nothing below in terms of content.
However, if you do a recursive search for assets for the blog section, it will find `banner.gif` but also all the assets from every article below which would cause some files to be copied several times (twice or more, it depends if you have subsection).
I believe #1677 solves that by only searching recursively if we are looking for the assets of a page.

eberkund:
Gotcha, so this could potential break if you had a subfolder of blog called `blog_assets` with perhaps `banner.gif` inside it right? I only tested the page scenario.

I think we need to check if the subfolders do not contain an `index.md` file then the folder can be treated as a colocated asset of the parent otherwise it would be treated as a page or section.

Keats:
>  this could potential break if you had a subfolder of blog called blog_assets with perhaps banner.gif inside it right? 

This is more an issue with duplicated assets in the output dir and threads trying to copy the files at the same time which seems to be an issue on Windows.

mtolk:
The problem i tried to solve was that, because 1 file could be an asset of multiple sections and a page, it was copied multiple times. It's actually the multiple copies that cause erros that was my problem.

I think it makes sense that an asset is associated with only 1 page or section, but that's not a requirement i think. I used it as a solution to the multiple copying problem. 

If you would allow different pages and section to share an assets another solution would have to found to prevent it being copied multiple times, at the same time.

#1677 is already a more elegant solution that this pull request. I had never realised people would place assets in subfolders.

Keats:
> I had never realised people would place assets in subfolders.

It's a new feature, it hasn't been released yet ;)

I'll close this in favour of #1677 .
@timakro if you can try the `next` branch again that would be great, it's good to release after that I think.

Issue #1429, opened 2021-04-08
Feature Request: Image Optimization with Pio


coastalwhite:
Maybe a good feature to have optional automation asset image optimization with [pio](https://github.com/siiptuo/pio). This could really reduce a lot of the size of the resulting website. I was busy creating my own pushing pipeline, so I guess I will just suggest the feature.

I would like to know what others think about this.

Note: I am in no way associated with the pio project, and other projects could also work.

Keats:
It could be a nice part of some automatic pipeline. It should probably do like the `imageproc` module does and save the output in `static` so we don't re-process it on every build though.

adworacz:
This seems like an interesting project. For now I optimize images using `mozjpeg/jpegtran` and `advpng` as part of my website publishing script (basically, `zola build` then find images and pass them to optimizers using `fd`/`find`).

This works pretty well, and has a nice "separation of concerns". Not sure how much benefit is brought by adding `pio` directly to Zola vs encouraging users to add it to their own scripts.

Keats:
If the library is in Rust/not too hard to build and the results are good, it's nice to have this kind of stuff built-in since we already do some image manipulation and I can see many people (including me) using that.

BezPowell:
I’ve just started work on a simple perceptual encoding library that uses image-rs for encoders: [BezPowell/perceptual-image - Codeberg.org](https://codeberg.org/BezPowell/perceptual-image). There's still work to be done, such as adding webp support, but is there still interest in adding something like this to Zola?

I'm envisaging a config setting to toggle this on and off, as well as a max_iterations variable, as it will be much slower than regular encoding. As it uses image-rs, jpegs still aren't progressive, but this seemed like a sensible way of adding the feature in without having to completely rewrite Zola's image processing code or adding in tonnes of extra dependencies.

Keats:
Yeah it would be nice. Although one thing is puts into question: should this be done via the template engine as a fn or some kind of pass after rendering is done, like minifiying. I'm leaning towards the additional pass at the end.

BezBIS:
BezPowell here, just commenting from my work account. I do like the idea of running it as a post-processing step, as that should avoid slowing down `serve`. Doing it that way, however, does create 2 potential problems we would have to work around:

[...]

PR #1685, opened 2021-12-07
Oxipng

mtolk:

Sanity check:

* [ x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* [ x ] Are you doing the PR on the `next` branch?

* [ x ] Have you created/updated the relevant documentation page(s)?

Updated my oxipng branch with latest changes from next. new pull request of the addition of configurable png optimization to zola.


Keats:
https://github.com/shssoichiro/oxipng for people other than me not knowing what it was. I'll have a look at this PR as soon as the next bugfix version is released

components/site/src/lib.rs
```diff
@@ -1127,6 +1128,18 @@ impl Site {
             })
             .collect::<Result<()>>()
     }
+	
+	/*
+		when running serve we do not want png optimizations, because that's too slow.
```
Keats, line 1133: This should only ever ran once for a file, after that it should be stored in `static` like the rest of the image processing so it's ok to run it on `serve`

components/config/src/config/mod.rs
```diff
@@ -62,6 +62,10 @@ pub struct Config {
     pub compile_sass: bool,
     /// Whether to minify the html output
     pub minify_html: bool,
+	/// Whether to optimize png files
+	pub optimize_png: bool,
+	/// level to optimize png files if applicable
+	pub optimize_png_level: u8,
```
Keats, line 68: I believe this field can be an `Option<u8>` and you remove the need for `optimize_png`

Keats (changes requested):
I think the optimization should be run before copying the assets so it happens only once but it means it would overwrite the existing file but I'm not sure how to recognize that a file has already been processed other than keeping some state file or something.

mtolk:
Hi @Keats,

I agree it would be nicer if a file would only have to proces a file once. But then we would need some way of keeping track of which files are already processed and which not.

I'm not sure if that would be worth the effort. 
Depending on the chosen solution it would make the code more complex, maybe also make the experience more complex for users. The way you would keep track of which file is optimized and which not, would that have to be under sourcecontrol? Do the optimized files have to be under source control( so that in my case netlify publishes the optimized versions).

It opens a whole can of worms, maybe it would be better to use webp with png as a fallback.

Keats:
Yeah I'm not too sure what to do honestly... Re-processing the files everytime is not great and you probably want to at least see what they look like on `zola serve`. 

> The way you would keep track of which file is optimized and which not, would that have to be under sourcecontrol? 

Probably. Some simple json file or whatever that just stores a list of transformation/input and their hashes to see if anything is needed. Code wise, it would be a bit more complicated but not much more so. I expect there will be other transformations that will be needing the same kind of check, I think?

mtolk:
From what i understand from your comment is that optimized png's, the assets, would have to be under source control as well.
That would a problem for some sites. I have a site that generates multiple sizes from 1 source image using resize_image.
I would like to have all these generated images under source control, just the source image. 

So I'm not sure we should proceed with this. It seems to add more complexity than that it adds value. I dont think it's worth it in this form.

Keats:
> I would like to have all these generated images under source control, just the source image.

Do you mean you want only the source one or also the generated ones?

mtolk:
in my case i have just one image version under source control. during `zola build` other versions(different sizes but also webp copies) are created.
this way i dont have to keep track of all the different version of the image. should i make a change to it.

should you put the generated under source control then you could end up with lots of obsolete images in git, that you would have manually remove from the static\processed_images folder

mwcz:
It would be really surprising to me if the generated images were under source control.

Keats:
> It would be really surprising to me if the generated images were under source control.

It's already the cases with resized images

mwcz:
> > It would be really surprising to me if the generated images were under source control.
> 
> It's already the cases with resized images

Consider me surprised. :sweat_smile:

Keats:
To be fair they are all put in the same folders so you can gitignore it if you want.

mtolk:
The folder for resized images is always in my gitignore file. But then any administration you keep about files being optimized becomes irrelevant, because these files are generated and, if configured, optimized on every build.

Jieiku:
oxipng is fast! so happy I found this pull request so I could learn of this tool!

ISSOtm:
If I may, it seems more appropriate to me to store the images separately (e.g. in a `/src-imgs/` directory), and manually move them to `/static/`/`/content/`.

This takes care of the "have the images been processed already?" question, while already integrating nicely with the existing workflow/model (if you modify the source image but forget to compress it, then it will simply not update in the render, which should immediately signal that something is wrong. This also lets the user pick the optimisation settings that work best for them. (I've opted for the most aggressive settings since I'm only doing this once, but that's *really* slow.)

Also, on [my personal website](//github.com/ISSOtm/issotm.github.io), I have PNGs, but also [JPEGs](https://github.com/ISSOtm/issotm.github.io/blob/d29d0ab9fb961bf3e42aacb24108f8693307298e/static/assets/banner.jpg) and [SVGs](https://github.com/ISSOtm/issotm.github.io/blob/d29d0ab9fb961bf3e42aacb24108f8693307298e/static/assets/mastodon-icon.svg), which need to be processed [by other tools](https://github.com/ISSOtm/issotm.github.io/blob/d29d0ab9fb961bf3e42aacb24108f8693307298e/optimise_imgs.sh#L4). [^bulk] (Tools list sourced from [Squoosh](https://github.com/GoogleChromeLabs/squoosh).)

[^bulk]: Yeah, that script re-processes all images in the site; I have in mind to only process images that have been modified since a certain (known-good) commit instead, but I've lazed out and not written that functionality yet.

The obvious downside is [a lack of discoverability](https://github.com/getzola/zola/pull/1685#issuecomment-1140228858); this could be solved by adding a page to the docs, I think.

Keats:
Closing this PR as it's very stale and there is the https://github.com/getzola/zola/issues/1429 issue. It refers to a different library but the concept is the same

Issue #1689, opened 2021-12-09
thread '<unnamed>' panicked at 'byte index 1116 is out of bounds of `


Folyd:
# Bug Report

## Environment

Zola version: zola (edge) 0.15.0 

## Panic backtrace
```
thread '<unnamed>' panicked at 'byte index 1116 is out of bounds of `, ||ZOLA_SC_PLACEHOLDER||, ||ZOLA_SC_PLACEHOLDER||.`', components/rendering/src/markdown.rs:217:45
stack backtrace:
   0:     0x56477cfc6d5a - std::backtrace_rs::backtrace::libunwind::trace::hf6a6dfd7da937cb0
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x56477cfc6d5a - std::backtrace_rs::backtrace::trace_unsynchronized::hc596a19e4891f7f3
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x56477cfc6d5a - std::sys_common::backtrace::_print_fmt::hb16700db31584325
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x56477cfc6d5a - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h231c4190cfa75162
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x56477cac77fc - core::fmt::write::h2a1462b5f8eea807
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/fmt/mod.rs:1163:17
   5:     0x56477cfc5c54 - std::io::Write::write_fmt::h71ddfebc68685972
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/io/mod.rs:1696:15
   6:     0x56477cfc6330 - std::sys_common::backtrace::_print::hcc197d4bebf2b369
                               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/sys_common/backtrace.rs:49:5
[...]

PR #1690, opened 2021-12-09
Don't panic on HTML short codes

ralfbiedert:
Full disclosure, this is me mostly trying to help out debugging #1689. I don't quite understand the code I patched and just followed your advice making it run `Shortcode::update_range`. Feel free to close this PR.


Keats:
I think the first step would be to add a minimal failing test. This is not the right fix though, this will render the HTML shortcodes before rendering the markdown. Thanks for having a look though!

PR #1710, opened 2021-12-30
templates/load_data: add an optional parameter `headers`

liushuyu:

Sanity check:

## Code changes

* [X] New tests are added as well

------

This pull request adds an optional parameter `headers` to the `load_data` function.

This is useful for accessing protected resources (either by setting `Authorization` header or `Cookie`) or handling some anomalies like GitHub Markdown rendering API (it expect you to set `Accept` to `application/json` but it will return a response with a type of `text/html`).

Two new tests have been added to ensure the newly added logic does not interfere with the existing `load_data` logic and works as expected. Due to limitations with `mockito`, one of the tests has relaxed passing conditions.

Two worked examples have been added to the documentation, they have been tested to work correctly.

### DIFF
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -440,8 +440,7 @@ headers=["accept=application/json,text/html"]
 
 Which is equivalent to two `Accept` headers with `application/json` and `text/html`.
 
-In rare cases where the target resource backend is not implemented correctly, you can instead specify
-the headers multiple times to achieve the same effect:
+If it doesn't work, you can instead specify the headers multiple times to achieve a similar effect:
 
 ```
 headers=["accept=application/json", "accept=text/html"]
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -432,6 +432,21 @@ environment variable to the access token you have obtained.
 {{postdata|safe}}
 ```
 
+In case you need to specify multiple headers with the same name, you can specify them like this:
+
+```
+headers=["accept=application/json,text/html"]
+```
+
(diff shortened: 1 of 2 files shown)


liushuyu:
I have updated documentation with more comprehensive explanations. Also, Hugo supports this feature as well: https://gohugo.io/templates/data-templates/#add-http-headers

liushuyu:
Do we need a discussion for this new parameter? Is this considered a new feature?

Keats:
No sorry, I just missed that. I'll have a look next week

docs/content/documentation/templates/overview.md
```diff
@@ -409,6 +409,44 @@ This example will make a POST request to the kroki service to generate a SVG.
...
+```
+
+Which is equivalent to two `Accept` headers with `application/json` and `text/html`.
+
+In rare cases where the target resource backend is not implemented correctly, you can instead specify
+the headers multiple times to achieve the same effect:
+
+```
+headers=["accept=application/json", "accept=text/html"]
+```
```
Keats, line 447: The last part deviates a bit from the point of the example, can we just use that one with two headers instead so we don't have to worry about the backend?

Keats (changes requested):

docs/content/documentation/templates/overview.md
liushuyu: > The last part deviates a bit from the point of the example, can we just use that one with two headers instead so we don't have to worry about the backend?

This is actually debatable. Because some servers may accept `Key: Value1,Value2`, some may accept `Key: Value1` `Key: Value2` sent separately, and some may accept both.

I feel like my wording may lead to some confusion, so I re-worded this segment.

Keats (approved):

Issue #1707, opened 2021-12-25
Build fails because of the [false] anchor validation error


biodranik:
# Bug Report

## Environment

Zola version: 0.15.2 (mac os x)

## Expected Behavior

`zola serve` should work.

## Current Behavior

```
Error: The anchor in the link `@/_index.md#install` in organicmaps.github.io/content/_index.md does not exist.
```

## Step to reproduce

The markdown source contains a link in the form `[Install here](#install)`.

Neither `<p id="install">` nor `<a name="install">` in the same markdown file are detected by Zola. It only works with markdown sections started from 2 or more `#` characters.

Keats:
> Neither <p id="install"> nor <a name="install"> in the same markdown file are detected by Zola

We are not parsing HTML for anchors in markdown so that's expected.

biodranik:
Well, that's not cool. Especially if you have a multilingual site, where the same `## Title` header is translated differently (and has a different `#id`), and can't be universally referenced.

Also, what about links from shortcodes?

Keats:
> Especially if you have a multilingual site, where the same ## Title header is translated differently (and has a different #id), and can't be universally referenced.

You would refer to the one from the current language? Or do you mean you would have only anchors in say English for all languages?
The vast majority of people use anchors to refer to headers so it hasn't really been a problem so far.

> Also, what about links from shortcodes?

MD shortcodes links would still be checked, HTML ones no.

biodranik:
> You would refer to the one from the current language? Or do you mean you would have only anchors in say English for all languages?
> The vast majority of people use anchors to refer to headers so it hasn't really been a problem so far.

Example:
```
This is index.md, it generates `<h1 id="#welcome">Welcome</h1>`

## Welcome

[Link to the Welcome section](#welcome)
```

```
This is index.fr.md, it generates `<h1 id="#bienvenue">Bienvenue</h1>`

## Bienvenue

[This link to the Welcome section should be also modified and can not use #welcome](#Bienvenue)
```

> MD shortcodes links would still be checked, HTML ones no.

Any chance to check the whole generated HTML for valid links/ids? It obviously looks wrong: the id is there, but the validator says "no, it's not there".

Keats:
> It obviously looks wrong: the id is there, 

[...]

PR #1716, opened 2022-01-03
Add heuristic checking for HTML anchors

phillord:
As a starter for discussion, not finished yet.

Previously only anchors specified or generated in markdown could be
linked to, without complaint from the link checker. We now use a
simple heuristic check for `name` or `id` attributes.

Addresses #1707

Sanity check:

## Code changes

### DIFF
--- a/components/library/src/content/mod.rs
+++ b/components/library/src/content/mod.rs
@@ -14,6 +14,7 @@ pub use self::ser::{SerializingPage, SerializingSection};
 
 use config::Config;
 use rendering::Heading;
+use utils::links::anchor_id_checks;
 
 pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
     for heading in headings {
@@ -26,6 +27,12 @@ pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
     }
 
     false
+}
+
+
+pub fn has_anchor_id(content: &str, anchor: &str) -> bool {
+    let checks = anchor_id_checks(anchor);
+    checks.is_match(content)
 }
 
 /// Looks into the current folder for the path and see if there's anything that is not a .md
--- a/components/library/src/content/mod.rs
+++ b/components/library/src/content/mod.rs
@@ -14,7 +14,6 @@ pub use self::ser::{SerializingPage, SerializingSection};
 
 use config::Config;
 use rendering::Heading;
-use utils::links::anchor_id_checks;
 
 pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
     for heading in headings {
@@ -27,12 +26,6 @@ pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
     }
 
     false
-}
-
-
-pub fn has_anchor_id(content: &str, anchor: &str) -> bool {
-    let checks = anchor_id_checks(anchor);
-    checks.is_match(content)
 }
 
 /// Looks into the current folder for the path and see if there's anything that is not a .md
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -63,7 +63,8 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
             let page = library
                 .get_page(&full_path)
                 .expect("Couldn't find section in check_internal_links_with_anchors");
-            !page.has_anchor(anchor)
+
+            !(page.has_anchor(anchor)||page.has_anchor_id(anchor))
         }
     });
 
--- a/components/utils/Cargo.toml
+++ b/components/utils/Cargo.toml
@@ -9,6 +9,7 @@ include = ["src/**/*"]
 tera = "1"
 unicode-segmentation = "1.2"
 walkdir = "2"
+regex="1"
 toml = "0.5"
 serde = { version = "1.0", features = ["derive"] }
 slug = "0.1"
--- a/components/utils/src/links.rs
+++ b/components/utils/src/links.rs
@@ -1,9 +1,14 @@
 use regex::Regex;
 
 
-pub fn anchor_id_checks(anchor:&str) -> Regex {
+pub fn has_anchor_id(content: &str, anchor: &str) -> bool {
+    let checks = anchor_id_checks(anchor);
+    checks.is_match(content)
+}
+
+fn anchor_id_checks(anchor:&str) -> Regex {
     Regex::new(
-        &format!(r#" (?i)(id|ID|name|NAME) *= *("|')*{}("|'| |>)+"#, anchor)
+        &format!(r#" (?i)(id|name) *= *("|')*{}("|'| |>)+"#, anchor)
     ).unwrap()
 }
 
--- a/components/utils/src/links.rs
+++ b/components/utils/src/links.rs
@@ -1,0 +1,42 @@
+use regex::Regex;
+
+
+pub fn anchor_id_checks(anchor:&str) -> Regex {
+    Regex::new(
+        &format!(r#" (?i)(id|ID|name|NAME) *= *("|')*{}("|'| |>)+"#, anchor)
+    ).unwrap()
+}
+
+
+#[cfg(test)]
+mod tests{
+    use super::anchor_id_checks;
+
+    fn check(anchor:&str, content:&str) -> bool {
+        anchor_id_checks(anchor).is_match(content)
+    }
+
+    #[test]
+    fn matchers () {
+        let m = |content| {check("fred", content)};
+
+        // Canonical match/non match
+        assert!(m(r#"<a name="fred">"#));
+        assert!(m(r#"<a id="fred">"#));
+        assert!(!m(r#"<a name="george">"#));
+
+        // Whitespace variants
+        assert!(m(r#"<a id ="fred">"#));
+        assert!(m(r#"<a id = "fred">"#));
+        assert!(m(r#"<a id="fred" >"#));
+        assert!(m(r#"<a  id="fred" >"#));
+
+        // Quote variants
+        assert!(m(r#"<a id='fred'>"#));
+        assert!(m(r#"<a id=fred>"#));
+
+        // Case variants
+        assert!(m(r#"<a ID="fred">"#));
+        assert!(m(r#"<a iD="fred">"#));
+    }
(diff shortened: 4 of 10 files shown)


components/utils/src/links.rs
```diff
@@ -0,0 +1,16 @@
+pub fn anchor_id_checks(anchor:&str) -> Vec<String> {
+    vec![
+        format!(" id={}", anchor),
```
biodranik: Will regex be slower or faster here?

Keats: Not sure, this is a lot of allocation, i'm guessing the regex would be faster? But hard to say without a benchmark

phillord: Possibly faster and probably more maintainable. I just copied what was already there, but happy to refactor it out to use a regexp. This code only runs if other anchor checks fail, so any performance hit will only be on sites that are broken or have these anchors in.

The regexp crate is already a dependency, so nothing wrong with trying. Am guessing we'd need a few hundred links to make a benchmark informative.

Keats: let's go with the regex approach then

phillord: Okay, pushed. Performance testing results.

I compiled site with a two small files but one contain ~2k links to an XML anchor. I then tested various options, in all cases using `cargo run` but taking the time reported at the end.

Comparing regexp to direct string matching

```
regexp:
   --release: 1.3s
   no --release: 39s

non-regexp:
   --release: 1.0s
   no --release: 19.4s
```

Conclusion: Regexp is slower esp for non-optimized builds.

Then comparing the non-regexp implementation with heuristic_link_check false or true

```
heuristic_link_check = false --release   0.5s 
heuristic_link_check = false no --release 1.1s
```
Conclusion: The  heuristic link check causes a 100% in build times (for release build) and a much bigger increase for non-optimized.

However, in this case zola produces lots of error messages and printouts can be slow. So, I have also tried fixing the broken links by adding an markdown generated anchor (with the same ~2k links to it). I also tried comparing this situation with and without the heuristic link checker.

```
--release, no failures: 0.5s
no -- release, no failures: 1.1s

heuristic_link_check = false --release no failures 0.5s
heuristic_link_check = false no --release no failures 1.1s
```

Conclusion 1: printing errors doesn't make much of a difference to the performance.
Conclusion 2: if the heuristic link checker is on, it does not make that much difference if most links are not broken. This would be predicted from the code.

So, I would say the regexp code is a bit neater, the multiple format! checking is a bit faster, so either implementation is fine. I'd probably got with the regexp cause neat trumps fast for me, but happy with what ever you prefer.

And while the heuristic_link_checker is significantly slows things down, you only pay for it if you are using XML anchors. As zola current refuses to build a site under these circumstances, it seems a reasonable price to pay.

components/utils/src/links.rs
```diff
@@ -1,16 +1,8 @@
...
-        format!(" NAME='{}'", anchor),
-        format!(r#" name="{}""#, anchor),
-        format!(r#" NAME="{}""#, anchor),
-    ]
+use regex::Regex;
+
+
+pub fn anchor_id_checks(anchor:&str) -> Regex {
+    Regex::new(
+        &format!(r#" (id|ID|name|NAME) *= *("|'){}("|')"#, anchor)
```
Keats: you're missing the ones where they have no quotes

docs/content/documentation/getting-started/configuration.md
```diff
@@ -130,6 +130,9 @@ skip_anchor_prefixes = [
     "https://caniuse.com/",
 ]
 
+# Check for links to anchors defined in HTML rather than markdown
+heuristic_link_check = true
```
Keats: in the end since it's only ran on local failures, it's probably fine to not need an option

components/utils/src/links.rs
biodranik: I would also make regex case-insensitive to handle Name, iD, etc.

components/site/src/link_checking.rs
```diff
@@ -63,7 +63,10 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
             let page = library
                 .get_page(&full_path)
                 .expect("Couldn't find section in check_internal_links_with_anchors");
-            !page.has_anchor(anchor)
+
+            !(page.has_anchor(anchor)||
+              site.config.link_checker.heuristic_link_check &&
+              page.has_anchor_id(anchor))
```
biodranik: Is regex created for each page? Would caching it make processing faster?

docs/content/documentation/getting-started/configuration.md
phillord: Your call, but it would give the option of disabling something that can produce failures. Switching the default around might be easier?

components/site/src/link_checking.rs
phillord: For each anchor and each page. I would guess that it would only make a minor difference in most cases.

components/utils/src/links.rs
phillord: Now done, hopefully, with any obscure side effects.

components/site/src/link_checking.rs
mwcz: I think caching the regex wouldn't be as useful because the regex pattern contains the text of each anchor, so the likelihood any particular regex would be needed again is slim.

Unless the method was flipped around, so there was a single cached regex behind a mutex, with a pattern like `r"(?i) (id|name) *= *"[^"]"? "` (case insesitively match _all_ name or id attributes (I used only double quoting for simplicity)).  Then there's only one regex, which can find all matches in a given page, then you can check those matches for the anchor in question.

I think there's some chance that would be faster than creating so many fresh regexes.

phillord: @mwcz I would agree that this might be faster. Then I thought regexs would be faster than exact string matching and was wrong. We don't have a good test realistic test case for any kind of real world benchmarking. My natural inclination would be to leave the simplest implementation till it's clear that this is a problem.

mwcz: For sure, +1 to keeping it simple at first.

For benchmarking, I might be able to help with that.  The site I'm starting to build with zola has 130,000 pages of markdown, with a lot of name/id anchors, so it would be highly sensitive to any performance issues.  I can run benchmarks myself with that content.  Maybe if I can find a way to scrub the content (lorem-ipsumify it but keep the markdown layout) then I could share it.

phillord: @mwcz That would be a good benchmark but only to see it it this code does not affect it at all. This code only runs if the current link checker fails. So if your current site builds (i.e. all the links work), then this code should never run. That's a useful thing to check, but doesn't give any indication of the performance of this code. I'm assuming your anchors here are either zola generated section slugs or manually specified anchors.

What you need could do is take half the markdown files, turn them into header/footer-less HTML with pandoc or equivalent. Now you should have 65,000 markdown files with pointers to 65,000 HTML files, so half of your outlinks would be to XML specified anchors and therefore use this code.

Quite how we would interpret the results I don't know. This check is surely slower than checking for markdown anchors which have to be parsed out anyway. So, it's not clear what would be an acceptable slowdown.

biodranik: The fastest implementation would be integrated into the markdown=>HTML convertor, where it's easy to collect all ids/links at the conversion stage and then to finally check at the end.

phillord: @biodranik Who knows what would be fastest, without a good test. Collecting all the ids/links would potentially mean checking a lot more links, so it might be slower. Of course, it does mean that the link checker could check all links. At the moment HTML -> HTML links that are broken would not be picked up.

It's a question of what the link checker is there for. I think it's fine that it is limited to markdown links, so long as it does not preclude me putting in HTML anchors.

docs/content/documentation/getting-started/configuration.md
Keats: There's already the workaround of creating the link in HTML directly

phillord:
Think this is ready to go now!

components/utils/src/links.rs
```diff
@@ -0,0 +1,42 @@
+use regex::Regex;
+
+
+pub fn anchor_id_checks(anchor:&str) -> Regex {
+    Regex::new(
+        &format!(r#" (?i)(id|ID|name|NAME) *= *("|')*{}("|'| |>)+"#, anchor)
```
Keats: Since we're setting the `?i` flag, we can get rid of the multiple cases in the first part of the regex

components/library/src/content/mod.rs
```diff
@@ -28,6 +29,12 @@ pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
     false
 }
 
+
+pub fn has_anchor_id(content: &str, anchor: &str) -> bool {
```
Keats: I would probably move this function to utils directly since we will only ever do `is_match`

Keats (changes requested):

Keats:
That looks ok to me, anyone else having comments?

components/utils/Cargo.toml
```diff
@@ -9,6 +9,7 @@ include = ["src/**/*"]
 tera = "1"
 unicode-segmentation = "1.2"
 walkdir = "2"
+regex="1"
```
biodranik, line 12: nit: space

components/utils/src/links.rs
```diff
@@ -0,0 +1,47 @@
...
+        assert!(m(r#"<a id="fred" >"#));
+        assert!(m(r#"<a  id="fred" >"#));
+
+        // Quote variants
+        assert!(m(r#"<a id='fred'>"#));
+        assert!(m(r#"<a id=fred>"#));
+
+        // Case variants
+        assert!(m(r#"<a ID="fred">"#));
+        assert!(m(r#"<a iD="fred">"#));
```
biodranik, line 45: Maybe add some other tags, so readers can easily understand, that the code also works with any element with id?

biodranik (approved):

mwcz:
> That looks ok to me, anyone else having comments?

Looks good to me.  It works well in my two trivial test case and in @mwcz's Giant Site, which had 493 broken anchors in 0.15.2 and with this patch is down to 209 (which seem to be real).  And there was no noticable performance difference at all.

PR #1717, opened 2022-01-03
Update arg parsing to clap v3

tranzystorekk:

Switches arg parsing to the modern derive style via the newly released clap 3.0.0

Important/to discuss:

- Now the help is colored by default
- The `base_path`/`base_url` parameters for the `check` subcommand were read from `ArgMatches` but never defined in the clap app, so I set them to `None` for now

Sanity check:

## Code changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,14 +13,15 @@ keywords = ["static", "site", "generator", "blog"]
 include = ["src/**/*", "LICENSE", "README.md"]
 
 [build-dependencies]
-clap = "2"
+clap = "3"
+clap_complete = "3"
 
 [[bin]]
 name = "zola"
 
 [dependencies]
 atty = "0.2.11"
-clap = { version = "2", default-features = false }
+clap = { version = "3", features = ["derive"] }
 chrono = "0.4"
 lazy_static = "1.1"
 termcolor = "1.0.4"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,11 +6,11 @@ use clap::{Parser, Subcommand};
 #[clap(version, author, about)]
 pub struct Cli {
     /// Directory to use as root of project
-    #[clap(short, long, default_value = ".")]
+    #[clap(short = 'r', long, default_value = ".")]
     pub root: PathBuf,
 
     /// Path to a config file other than config.toml in the root of project
-    #[clap(short, long)]
+    #[clap(short = 'c', long)]
     pub config: Option<PathBuf>,
 
     #[clap(subcommand)]
@@ -26,7 +26,7 @@ pub enum Command {
         name: String,
 
         /// Force creation of project even if directory is non-empty
-        #[clap(short, long)]
+        #[clap(short = 'f', long)]
         force: bool,
     },
 
@@ -37,7 +37,7 @@ pub enum Command {
         base_url: Option<String>,
 
         /// Outputs the generated site in the given path (by default 'public' dir in project root)
-        #[clap(short, long)]
+        #[clap(short = 'o', long)]
         output_dir: Option<PathBuf>,
 
         /// Include drafts when loading the site
@@ -48,16 +48,16 @@ pub enum Command {
     /// Serve the site. Rebuild and reload on change automatically
     Serve {
         /// Interface to bind on
-        #[clap(short, long, default_value = "127.0.0.1")]
+        #[clap(short = 'i', long, default_value = "127.0.0.1")]
         interface: String,
 
         /// Which port to use
-        #[clap(short, long, default_value = "1111")]
+        #[clap(short = 'p', long, default_value_t = 1111)]
         port: u16,
 
         /// Outputs assets of the generated site in the given path (by default 'public' dir in project root).
         /// HTML/XML will be stored in memory.
-        #[clap(short, long)]
+        #[clap(short = 'o', long)]
         output_dir: Option<PathBuf>,
 
         /// Changes the base_url
@@ -73,7 +73,7 @@ pub enum Command {
         open: bool,
 
         /// Only rebuild the minimum on change - useful when working on a specific page/section
-        #[clap(short, long)]
+        #[clap(short = 'f', long)]
         fast: bool,
     },
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,102 +1,86 @@
-use clap::{crate_authors, crate_description, crate_version, App, AppSettings, Arg, SubCommand};
+use std::path::PathBuf;
 
-pub fn build_cli() -> App<'static, 'static> {
-    App::new("zola")
-        .version(crate_version!())
-        .author(crate_authors!())
-        .about(crate_description!())
-        .setting(AppSettings::SubcommandRequiredElseHelp)
-        .arg(
-            Arg::with_name("root")
-                .short("r")
-                .long("root")
-                .takes_value(true)
-                .default_value(".")
-                .help("Directory to use as root of project")
-        )
-        .arg(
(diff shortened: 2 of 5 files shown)


src/cli.rs
```diff
@@ -1,102 +1,86 @@
...
+
+        /// Include drafts when loading the site
+        #[clap(long)]
+        drafts: bool,
+
+        /// Open site in the default browser
+        #[clap(short = 'O', long)]
+        open: bool,
+
+        /// Only rebuild the minimum on change - useful when working on a specific page/section
```
Keats, line 75: is it automatically grabbing the comments to use as help text?

src/cli.rs
```diff
@@ -1,102 +1,86 @@
...
+    },
+
+    /// Deletes the output directory if there is one and builds the site
+    Build {
+        /// Force the base URL to be that value (defaults to the one in config.toml)
+        #[clap(short = 'u', long)]
+        base_url: Option<String>,
+
+        /// Outputs the generated site in the given path (by default 'public' dir in project root)
+        #[clap(short, long)]
```
Keats: Can the `short` always be named? I don't really want to rely on struct field names because those could be changed and it would only break silently, while the flags are the public api from the cli pov.

src/main.rs
```diff
@@ -1,124 +1,96 @@
...
-        Some(path) => PathBuf::from(path)
-            .canonicalize()
-            .unwrap_or_else(|_| panic!("Cannot find config file: {}", path)),
-        None => root_dir.join("config.toml"),
-    };
+    let cli = Cli::parse();
+    let root_dir = cli
+        .root
+        .canonicalize()
+        .unwrap_or_else(|_| panic!("Cannot find root directory: {}", cli.root.display()));
```
Keats, line 18: that's not handling the `.` case?

src/main.rs
```diff
@@ -1,124 +1,96 @@
...
-                Ok(()) => (),
-                Err(e) => {
-                    console::unravel_errors("Failed to create the project", &e);
-                    ::std::process::exit(1);
-                }
-            };
-        }
-        ("build", Some(matches)) => {
+    match cli.command {
+        Command::Init { name, force } => match cmd::create_new_project(&name, force) {
```
Keats: looks like it would be easier to use `if let Err(e) = ...` than match

src/main.rs
```diff
@@ -1,124 +1,96 @@
...
-                    console::error("The request port needs to be an integer");
-                    ::std::process::exit(1);
-                }
-            };
-            let open = matches.is_present("open");
-            let include_drafts = matches.is_present("drafts");
-            let fast = matches.is_present("fast");
-
-            // Default one
+        Command::Serve { interface, mut port, output_dir, base_url, drafts, open, fast } => {
```
Keats, line 51: what error message do users get when the port is not a valid number for example with clap3?

src/cli.rs
```diff
@@ -1,102 +1,86 @@
...
+        fast: bool,
+    },
+
+    /// Try to build the project without rendering it. Checks links
+    Check {
+        /// Include drafts when loading the site
+        #[clap(long)]
+        drafts: bool,
+    },
```
Keats, line 85: can we be explicit with all the default values as well? eg set `false` for drafts etc. I guess `Subcommand` relies on default but I'd like to be able to see the full set of option values at a glance

Keats (changes requested):
Thanks for the PR!

src/cli.rs
tranzystorekk: Sure, sounds reasonable.

src/cli.rs
tranzystorekk: Exactly! 😄 Only the doc comments, to be specific.

src/main.rs
tranzystorekk: by default whatever the error from the `FromStr` impl for the type (here `u16`) displays, e.g.:

image

src/main.rs
tranzystorekk: Here `cli.root` already has the `.` value if it wasn't specified on the command line, so we handle only the canonicalize operation result, which I think should work for `.`

src/main.rs
tranzystorekk: True, I wanted to avoid doing refactorings unrelated strictly to the clap porting

src/cli.rs
tranzystorekk: Additionally, unless disabled, it does typical markdown preprocessing like joining together text from adjacent lines etc.

src/cli.rs
tranzystorekk: I've looked at the documentation, but especially for bool flags this is neither elegant to write nor widely accepted style to specify the default explicitly.

src/cli.rs
Keats: Nice

src/main.rs
Keats: I'm not entirely sure it does work on all platforms, I'd like to be certain before changing it

src/main.rs
Keats: That would be a fine change if you don't mind

src/cli.rs
Keats: oh :(
Just passing true/false to default doesn't work? It's kind of weird that non bool gets to have their default value in the struct but not bool since there are only 2 possible values...

tranzystorekk: Did a quick check and clap protests with a "default value is meaningless for bool" error.

I'm guessing the rationale is that bool flags are simple on-off switches and don't require any special handling.

src/main.rs
tranzystorekk: I built a small program to verify that canonicalizing `.` works on windows, but if we wanted to be extra careful, there's the [normpath](https://github.com/dylni/normpath) crate that seems to be saner (also personally confirmed that it works).

Keats: Let's keep like you have for now, we'll see if it causes an issue

Keats (approved):

Cargo.toml
```diff
@@ -13,14 +13,15 @@ keywords = ["static", "site", "generator", "blog"]
 include = ["src/**/*", "LICENSE", "README.md"]
 
 [build-dependencies]
-clap = "2"
+clap = "3"
+clap_complete = "3"
```
Keats, line 17: what does this do and where is it used?

tranzystorekk:
The remaining issue is that clap v3 uses the `include_str!` macro to add README.md contents to its docs, which effectively makes it MSRV 1.54, while linux-pinned uses 1.52

Keats:
Bumping the MSRV is not an issue but can that feature be disabled as it wouldn't be used anyway?

Cargo.toml
tranzystorekk: In v3 the completion functionality has been split off to the separate `clap_complete` package, I noticed that there is some completion code in build.rs that is commented out, so admittedly this is a bit preemptive.

I guess we could instead add it in a separate PR focused on bringing back completion

tranzystorekk:
> but can that feature be disabled as it wouldn't be used anyway?

I'm afraid building docs for dependencies can only be disabled in `cargo doc`, and the error occurs in `cargo build` in general.

Keats:
> I'm afraid building docs for dependencies can only be disabled in cargo doc, and the error occurs in cargo build in general.

Oh I see, I thought clap was somehow pulling the readme in the built cli. This makes more sense.

Keats:
Can you resolve the conflicts? Or i'll do it later today

Keats (approved):

Keats:
Thanks!

Issue #1723, opened 2022-01-08
Extra syntaxes not working


arnau:
# Bug Report

## Environment

Zola version: 0.15.2

## Expected Behavior

I would expect a `config.toml` like:

```toml
[markdown]
extra_syntaxes = ["syntaxes"]
```

to load and apply the syntaxes in the `syntaxes` folder as described in the [config documentation](https://www.getzola.org/documentation/getting-started/configuration/).

In case it helps:

- [the full config file](https://github.com/arnau/seachess/blob/main/website/config.toml)
- [the extra syntaxes](https://github.com/arnau/seachess/tree/main/website/syntaxes)

Note that the exact code linked above works well with zola version **0.14.1**.

## Current Behavior

Both `zola --root website serve` and `zola --root website build` work as expected except for warnings such as:

```
Warning: Highlight language rfc8288 not found in notes/dive-into-the-link-header.md
Warning: Highlight language abnf not found in notes/dive-into-the-link-header.md
```

With the resulting HTML not highlighting codeblocks labelled as `rfc8288` or `abnf`.

## Step to reproduce

A simple `zola serve` on [this minimal test](https://github.com/arnau/zola-test) should suffice.

Keats:
Ah the snippet hasn't been updated, it should be `extra_syntaxes_and_themes` as mentioned later on the page

arnau:
Ah! I saw it, tried it on my website but didn't work and forgot to try it on the minimal test.

So, when I tried on my website it failed again so after digging a bit I found that if I add `highlight_theme = "css"` it fails again. I updated the repo with the small test which reproduces the issue.

jk-ozlabs:
Seeing the same issue here, and thought I'd have a go at fixing it. Let me know if you need any details.

PR #1732, opened 2022-01-13
Fix extra syntaxes when highlight_theme = "css"

jk-ozlabs:
Currently, init_extra_syntaxtes_and_highlight_themes will supress
loading anything if highlight_theme is "css". This makes sense for the
highlight themes, but will also skip loading any extra syntax
definitions too.

This change defers the theme = css check, so we still load syntaxes in
this case.

Fixes: #1723

Signed-off-by: Jeremy Kerr <jk@codeconstruct.com.au>

Sanity check:

## Code changes

### DIFF
--- a/components/config/src/config/markup.rs
+++ b/components/config/src/config/markup.rs
@@ -107,16 +107,17 @@ impl Markdown {
     }
 
     pub fn init_extra_syntaxes_and_highlight_themes(&mut self, path: &Path) -> Result<()> {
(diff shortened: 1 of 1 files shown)


jk-ozlabs:
We may still want to suppress the highlight theme load with highlight_theme = "css" - let me know if there's a better approach here.

Keats:
I think this is fine as is. It would be nice to have a test for it but I can't remember if that's easy to set up

PR #1734, opened 2022-01-13
fix #1731 (asset colocation does not detect files without an extension)

### LINKED ISSUE #1731: Asset colocation does not copy files without an extension
# Bug Report

## Environment

Zola version: zola 0.15.2

## Expected Behavior

Asset/section colocation should work when the assets in question have no file extension.

## Current Behavior

Files inside a section with no extension are not copied into the output folder

## Step to reproduce

Create two files in a section; one with an extension (i used *.jpg) and one without and then build.

tsheinen:
Pretty minor change; I just adjusted the match statement so that if Path::extension returned None it would include that asset -- only Some("md") should be ignored. I also added a file without an extension to the test cases for this function. 

## Code changes


Keats (approved):
Thanks for the quick fix!

Issue #1686, opened 2021-12-07
[Feature Request] Publish Zola to Winget


Kixiron:
Publishing Zola to [Winget](https://docs.microsoft.com/en-us/windows/package-manager/) would be really neat as an alternative way to install it on Windows ([Guide](https://docs.microsoft.com/en-us/windows/package-manager/package/))

Keats:
It looks like anyone can do it via a PR on their repo right?

Kixiron:
I believe so, you just make a manifest and then PR it into the repo

leotsarev:
Hi!
I created PR for this 
https://github.com/microsoft/winget-pkgs/pull/79060

But winget now can't extract binaries from zip.
For winget working, you have either:
1. Publish .exe in release (not zipped .exe) — seems to be easy
2. Or publish full .msi / .msix — seems to be hard.

Winget will add "zip + portable exe" support, but in unspecified future version.

Please add .exe into release and I'll update PR.

@Keats 

P.S. Thank you for your awesome project!

Keats:
I don't use Azure enough to know the answer but it might be just adding another section like https://github.com/getzola/zola/blob/master/azure-pipelines.yml#L117-L123 where it's just copying the exe on release?

leotsarev:
@Keats hm, I'm not sure, but probably it's something like https://github.com/getzola/zola/pull/1987

leotsarev:
Seems that we just could put this on hold. winget soon will be able to extract from zip

Keats:
Looks like waiting wins again

leotsarev:
Ok, I managed to get .zip working. But here is another issue. 
https://github.com/microsoft/winget-pkgs/pull/79060#issuecomment-1456455967

zola requires Microsoft.VCRedist.2015+.x64 and if there is some dependency (even so trivial and common), it not considered "single-file" .exe by winget standarts.

What if CRT will be statically linked on Windows?  How it will increase file size? 
https://stackoverflow.com/questions/31770604/how-to-generate-statically-linked-executables/44387312#44387312

leotsarev:
and please assign this issue to me

Keats:
Hmm where is that coming from though. I didn't think we needed anything special after replacing libsass. I think we can try the stackoverflow comment but I don't have a Windows machine to check the new file size

leotsarev:
Rust std requires some libc. And for Windows, libc is VCRedist.

[...]

PR #1735, opened 2022-01-14
build: include metadata in Windows binary

liushuyu:

Sanity check:

## Code changes

* [N/A] Have you created/updated the relevant documentation page(s)?

------

This is a prerequisite for anyone who wishes to take a stab at #1686.

By adding metadata to the resulting binary, it will fill Microsoft's requirements pretty nicely (also Zola binary will have an icon in the Explorer!)

To support cross-compiling from Unix systems, some of the dependencies are not gated to a specific OS platform.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ include = ["src/**/*", "LICENSE", "README.md"]
 [build-dependencies]
 clap = "3"
 clap_complete = "3"
+winres = "0.1"
+time = "0.3"
 
 [[bin]]
 name = "zola"
@@ -64,3 +66,7 @@ codegen-units = 1
 # Disabling debug info speeds up builds a bunch,
 # and we don't rely on it for debugging that much.
 debug = 0
+
+[package.metadata.winres]
+OriginalFilename = "zola.exe"
+InternalName = "zola"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3688,6 +3688,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "winres"
+version = "0.1.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


Keats:
Sorry, can you rebase?

liushuyu:
> Sorry, can you rebase?

Sure. Rebased.

Cargo.toml
```diff
@@ -15,6 +15,8 @@ include = ["src/**/*", "LICENSE", "README.md"]
 [build-dependencies]
 clap = "3"
 clap_complete = "3"
+winres = "0.1"
+chrono = "0.4"
```
Keats: Can those 2 be added with `[target.'cfg(windows)'.build-dependencies]` or whatever would be the right TOML key to only require those when building on Windows?

Keats (changes requested):

Cargo.toml
liushuyu: > Can those 2 be added with `[target.'cfg(windows)'.build-dependencies]` or whatever would be the right TOML key to only require those when building on Windows?

No, due to issues in `cargo`, this is currently not possible and will break `build.rs` build script. See https://github.com/rust-lang/cargo/issues/4932 for more information (specifically, this comment: https://github.com/rust-lang/cargo/issues/4932#issuecomment-357163633).

Keats: oh ok :(
Can we use `time` instead of chrono?

liushuyu: > oh ok :( Can we use `time` instead of chrono?

Sure thing. Which version of `time` do you want me to use? `0.1` series or the latest `0.3` series?

liushuyu: Rebased and switched to `time 0.3`.

Keats (approved):

Keats:
perfect thanks!

PR #1738, opened 2022-01-17
Make ignored_content work with nested paths and directories

### LINKED ISSUE #1701: Nested directory globs with ignored_content
What I'd like to do is have this work: `ignored_content = ["albums/**/*.{jpg,jpeg,png}"]`. Basically ignore any images in `content/albums` _and_ all sub-folders of that. This does not work.

`ignored_content = ["*.{jpg,jpeg,png}"]` works, but it ignores all images in `content` basically.

I think https://github.com/getzola/zola/issues/481 says this should work, but it may just be referring to specifying a particular folder.

I'm using `v0.15.2`.

kartva:

resolves #1701 

Sanity check:

## Code changes

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -561,21 +561,24 @@ ignored_content = []
         let config_str = r#"
 title = "My site"
 base_url = "example.com"
-ignored_content = ["*.{graphml,iso}", "*.py?"]
+ignored_content = ["*.{graphml,iso}", "*.py?", "**/{target,temp_folder}"]
         "#;
 
         let config = Config::parse(config_str).unwrap();
         let v = config.ignored_content;
-        assert_eq!(v, vec!["*.{graphml,iso}", "*.py?"]);
+        assert_eq!(v, vec!["*.{graphml,iso}", "*.py?", "**/{target,temp_folder}"]);
 
         let g = config.ignored_content_globset.unwrap();
-        assert_eq!(g.len(), 2);
+        assert_eq!(g.len(), 3);
         assert!(g.is_match("foo.graphml"));
         assert!(g.is_match("foo.iso"));
         assert!(!g.is_match("foo.png"));
         assert!(g.is_match("foo.py2"));
         assert!(g.is_match("foo.py3"));
         assert!(!g.is_match("foo.py"));
+        assert!(g.is_match("foo/bar/target"));
+        assert!(g.is_match("foo/bar/baz/temp_folder"));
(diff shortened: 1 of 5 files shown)


DesmondWillowbrook:
resolves #1701

kartva:
This may break some config files if they had a filter which included directories (like `"foo/bar/*.xlsx"`) - previously Zola would ignore _all_ `*.xlsx` but now only `foo/bar/*.xlsx`. I don't think anyone relied on this behavior, though.

Keats:
Can you rebase the PR?

kartva:
Better now?

Keats:
Hm no, it still contains lots of unrelated commits from master. The easiest way would be reset --hard on master and cherrypick the 2 relevant commits on it rather than merging

DesmondWillowbrook:
I should be cherry-picking commits on top of next, right?

kartva:
Is it somewhat better now?

components/config/src/config/mod.rs
```diff
@@ -561,21 +561,24 @@ ignored_content = []
...
+        assert_eq!(g.len(), 3);
         assert!(g.is_match("foo.graphml"));
         assert!(g.is_match("foo.iso"));
         assert!(!g.is_match("foo.png"));
         assert!(g.is_match("foo.py2"));
         assert!(g.is_match("foo.py3"));
         assert!(!g.is_match("foo.py"));
+        assert!(g.is_match("foo/bar/target"));
+        assert!(g.is_match("foo/bar/baz/temp_folder"));
+        assert!(g.is_match("temp_folder"));
```
Keats, line 583: can you add some more test examples?

```
assert!(g.is_match("my/isos/foo.iso"));
assert!(g.is_match("content/poetry/zen.py"));
```

Just to assert that single files glob still work without striping the prefix.

Keats (changes requested):
Looks good, just a small tweak needed to a test

components/config/src/config/mod.rs
kartva: `assert!(g.is_match("content/poetry/zen.py"));` actually won't match since the pattern is `"*.py?"` (another test case `assert!(!g.is_match("foo.py"));` confirms that.

But yes, I'll add the rest.

kartva:
Done? GitHub still shows "Changes requested" even if I marked it as resolved, so I'm not sure.

Keats (approved):
Yep sorry it's good

PR #1741, opened 2022-01-23
Next version

Keats:


src/main.rs
```diff
@@ -1,124 +1,93 @@
...
-    let config_file = match matches.value_of("config") {
-        Some(path) => PathBuf::from(path)
-            .canonicalize()
-            .unwrap_or_else(|_| panic!("Cannot find config file: {}", path)),
-        None => root_dir.join("config.toml"),
-    };
+    let root_dir = cli_dir
+        .ancestors()
+        .find_map(|a| if a.join(&cli.config).exists() { Some(a) } else { None })
+        .unwrap_or_else(|| panic!("could not find directory containing config file"));
```
mwcz: @Keats I think something in here breaks `zola init`.  I tried init today on the next branch and it immediately panics with this error.

mwcz:
I tried `zola init` on the next branch today and it panics.  Left a comment in main.rs where I think it happens. Want me to open a new issue or is this sufficient?

src/main.rs
Keats: Hm, I'll fix it but we should really have some test for the cli :(

mwcz:
@Keats Just in case you hadn't noticed, there are some errors related to the backlinks test.

Keats:
Thanks, I missed that. Fixed now

Keats:
Last blockers are 2 perf issues: https://github.com/getzola/zola/issues/1902 and https://github.com/getzola/zola/issues/1897

For #1902, anyone seeing a big perf difference between 0.15.3 and this branch?

PR #1749, opened 2022-01-29
Include path separator in log message

srstevenson:
This adds a path separator between the directory and filenames in the message printed by `zola serve` to indicate it is listening for changes.

Before:

    Listening for changes in /path/to/dir{config.toml, content, sass, static, templates}

After:

    Listening for changes in /path/to/dir/{config.toml, content, sass, static, templates}
                                         ^


Keats (approved):
Thanks!

PR #1760, opened 2022-02-11
fix crash in build.rs when installing from source on linux

aslowwriter:
The current `build.rs` fails to compile "widows resources" when installing from source on Linux, so I added a `!cfg(windows)` to fix that. That's all. 

Sanity check:

## Code changes


Keats (approved):

Keats (changes requested):
I believe you can wrap L32 in the if cfg(windows) instead?

Keats (approved):

Keats:
Thanks!

PR #1761, opened 2022-02-11
Make sure that zola serve/build can start from anywhere inside dir tree

aslowwriter:
in response to request here: https://zola.discourse.group/t/allow-starting-zola-from-anywhere-in-the-site-directory/1164 Quite a slimple PR all it does is check parents of `root_dir` for `cli.config` and picks those if it finds them instead of only checking `root_dir` itself.

### DIFF
--- a/components/library/src/pagination/mod.rs
+++ b/components/library/src/pagination/mod.rs
@@ -99,10 +99,12 @@ impl<'a> Paginator<'a> {
         let paginate_by = taxonomy.kind.paginate_by.unwrap();
         // Check for taxon-specific template, or use generic as fallback.
         let specific_template = format!("{}/single.html", taxonomy.kind.name);
-        let template = match check_template_fallbacks(&specific_template, tera, theme) {
-            Some(template) => template,
-            None => "taxonomy_single.html",
-        };
+        let template =
+            if let Some(template) = check_template_fallbacks(&specific_template, tera, theme) {
+                template
+            } else {
+                "taxonomy_single.html"
+            };
         let mut paginator = Paginator {
             all_pages: Cow::Borrowed(&item.pages),
             pagers: Vec::with_capacity(item.pages.len() / paginate_by),
--- a/components/library/src/pagination/mod.rs
+++ b/components/library/src/pagination/mod.rs
@@ -99,12 +99,8 @@ impl<'a> Paginator<'a> {
         let paginate_by = taxonomy.kind.paginate_by.unwrap();
         // Check for taxon-specific template, or use generic as fallback.
         let specific_template = format!("{}/single.html", taxonomy.kind.name);
-        let template =
-            if let Some(template) = check_template_fallbacks(&specific_template, tera, theme) {
-                template
-            } else {
-                "taxonomy_single.html"
-            };
+        let template = check_template_fallbacks(&specific_template, tera, theme)
+            .unwrap_or("taxonomy_single.html");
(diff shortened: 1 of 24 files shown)


aslowwriter:
whoops, meant to push the linting commit to different branch, but it's very insubstantial so if you don't mind I'll just leave it here

components/library/src/pagination/mod.rs
```diff
@@ -99,10 +99,12 @@ impl<'a> Paginator<'a> {
...
-        let template = match check_template_fallbacks(&specific_template, tera, theme) {
-            Some(template) => template,
-            None => "taxonomy_single.html",
-        };
+        let template =
+            if let Some(template) = check_template_fallbacks(&specific_template, tera, theme) {
+                template
+            } else {
+                "taxonomy_single.html"
+            };
```
Keats: that seems worse than the match?

components/library/src/taxonomies/mod.rs
```diff
@@ -232,12 +235,15 @@ impl Taxonomy {
...
-        let template = match check_template_fallbacks(&specific_template, tera, &config.theme) {
-            Some(template) => template,
-            None => "taxonomy_list.html",
+        let template = if let Some(template) =
+            check_template_fallbacks(&specific_template, tera, &config.theme)
+        {
+            template
+        } else {
+            "taxonomy_list.html"
         };
```
Keats: same

src/main.rs
```diff
@@ -12,17 +13,13 @@ mod prompt;
...
-        .config
-        .map(|path| {
-            path.canonicalize()
-                .unwrap_or_else(|_| panic!("Cannot find config file: {}", path.display()))
-        })
-        .unwrap_or_else(|| root_dir.join("config.toml"));
+    let cli_dir: PathBuf =
+        cli.root.canonicalize().unwrap_or_else(|_| panic!("Could not find canonical root dir"));
+
+    let root_dir = cli_dir
```
Keats, line 19: we should probably display a message if the root dir is not the current dir

Keats (changes requested):

src/main.rs
aslowwriter: like this?

components/library/src/taxonomies/mod.rs
aslowwriter: checked it again and now clippy gives different suggestions. Not sure why, but I do agree that it's less good than a match. This time clippy suggested replacing it with `unwrap_or` which seems even better to me

components/library/src/pagination/mod.rs
aslowwriter: see other comment

components/library/src/taxonomies/mod.rs
```diff
@@ -204,12 +204,15 @@ impl Taxonomy {
...
         let specific_template = format!("{}/single.html", self.kind.name);
-        let template = match check_template_fallbacks(&specific_template, tera, &config.theme) {
-            Some(template) => template,
-            None => "taxonomy_single.html",
+        let template = if let Some(template) =
+            check_template_fallbacks(&specific_template, tera, &config.theme)
+        {
+            template
+        } else {
+            "taxonomy_single.html"
```
Keats, line 212: it could be `unwrap_or` like the line below but i can do that myself

Keats (approved):

Issue #1705, opened 2021-12-23
Links to images broken in summary


Yaroslav-95:
# Bug Report

Before version 0.15.0 zola would correctly generate links to any images that belonged to a page (e.g. a blog post) that were contained in a summary and rendered outside that page (e.g. the index page). From version 0.15.0 onward this is broken. 

Inside the page, the link is generated correctly, but outside it, in the summary, it is not; instead of something like `/blog/post/image.jpg` you get `image.jpg`.

## Environment

Zola version: 0.15.0, 0.15.1

## Expected Behavior

The link in the summary should be something like `/blog/post/image.jpg`.

## Current Behavior

zola renders something like `image.jpg` instead.

## Step to reproduce

Build a site with a post that has a picture in its summary and whose template renders summaries. An example of such a site: <https://git.yaroslavps.com/yaroslavps.com/> under `/food/`.

theowenyoung:
Same as you.

Keats:
Closing as there is also https://github.com/getzola/zola/issues/1772 and #1779 about this

Yaroslav-95:
After finding myself with some free time, I decided to try to tackle
this issue on my own. So far I have found the culprit to be commit
4086b075 merged in PR #1582 "Support colocating subfolders". The commit
for this feature broke the functionality that this issue is about.

I'll soon open a PR with a fix.

Keats:
That should be fixed in 0.16?

Issue #977, opened 2020-03-26
Unify the usage of `path` parameter in functions


Lesik:
The `path` parameter is used in multiple functions that Zola adds to the Tera templating engine, like `get_url()` and `resize_image()`, as well as in Markdown links like `[link](path)`. However, the way that they resolve the path seems to differ between them.

For example, `get_url()` seems to do a lookup in `static/` for paths that have no leading `/` or `@`, and for paths that start with an `@` it does a lookup in `content/`. On the other hand, both `get_page()` and `get_section()` start in `content/` for paths without leading `/` or `@`. It is unclear from the documentation what happens with paths with a leading `/` in any of these functions. 

| function | `mypath` | `@mypath` | `/mypath` |
|---|---|---|---|
| \[link](mypath) | `$PWD/mypath` | `content/mypath` | ? | 
| get_url() | `static/mypath` | `content/mypath` | ? |
| get_page() | `content/mypath` | ? | ? |
| get_section() | `content/mypath` | ? | ? |
| resize_image() | `content/mypath` | ? | ? |

I propose the unification of the `path` parameter in some or all of those cases. Let's discuss!

Lesik:
By the way, I've stumbled upon this while creating a shortcode that resizes images. I wanted to link to the original image as well as the resized one, so both `get_url()` and `resize_image()` came into play. But since the path parameters are resolved differently, I couldn't just pass the same path to both functions, which I thought was pretty confusing.

Keats:
Related: https://github.com/getzola/zola/issues/911

Keats:
Can anyone spend some time to figure a nice way to unify them? No need to actually implement it, just think things through

https://github.com/getzola/zola/pull/1044 took the approach of actually allowing basic path and `@` path.

vojtechkral:
I ran into this issue - and #788 - during refactoring of the `imageproc` code. I'm not yet sure how to solve this but I would like to make `resize_image()` consistent with the others as part of the refactor... I'll try to think of something...

Keats:
Yep, hopefully the next version will have fully consistent paths, I'm thinking about it too.

Lesik:
[...]

PR #1779, opened 2022-02-24
Discussion: Reenable link rewriting for colocated assets

timakro:
In #1582 I broke the asset colocation example from the documentation:
https://www.getzola.org/documentation/content/overview/#asset-colocation

The breakage was discovered by irimi1:
https://github.com/getzola/zola/pull/1582#issuecomment-1042300973

When I wrote #1582 I was not aware of the old behavior: links to assets
used to be rewritten to absolute links, so that colocated assets could
be linked to from `pagename/index.md` simply as `asset.xyz` instead of
`pagename/asset.xyz`. I unintentionally changed this, so that currently
you have to write `pagename/asset.xyz`.

Read this commit as a partial revert of #1582. It restores the old
behavior as described above.

Caveats:
1. This will break lots of sites, just like #1582 unintentionally broke
   lots of sites (but no one seemed to notice).
2. The old behavior is restored, including the quirk, that when linking
   from a page `pagename.md` to a section asset `asset.xyz` you need to
   write `../asset.xyz`, even though both files are in the same folder
   on disk. You can make sense of it in the following way: the asset
   belongs to the section so that is one level up from the page.

---

@irimi1

Please discuss this before merging. Some might prefer the current behavior over the old behavior.

### DIFF
--- a/components/rendering/src/markdown.rs
+++ b/components/rendering/src/markdown.rs
@@ -2,6 +2,7 @@ use libs::gh_emoji::Replacer as EmojiReplacer;
 use libs::once_cell::sync::Lazy;
 use libs::pulldown_cmark as cmark;
 use libs::tera;
+use libs::regex::Regex;
 
 use crate::context::RenderContext;
 use crate::table_of_contents::{make_table_of_contents, Heading};
@@ -18,6 +19,13 @@ use crate::shortcode::{Shortcode, SHORTCODE_PLACEHOLDER};
 const CONTINUE_READING: &str = "<span id=\"continue-reading\"></span>";
 const ANCHOR_LINK_TEMPLATE: &str = "anchor-link.html";
 static EMOJI_REPLACER: Lazy<EmojiReplacer> = Lazy::new(|| EmojiReplacer::new());
+
+/// Although there exists [a list of registered URI schemes][uri-schemes], a link may use arbitrary,
+/// private schemes. This regex checks if the given string starts with something that just looks
+/// like a scheme, i.e., a case-insensitive identifier followed by a colon.
+///
+/// [uri-schemes]: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
+static STARTS_WITH_SCHEMA_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"^[0-9A-Za-z\-]+:").unwrap());
 
 #[derive(Debug)]
 pub struct Rendered {
@@ -62,6 +70,11 @@ fn find_anchor(anchors: &[String], name: String, level: u16) -> String {
     find_anchor(anchors, name, level + 1)
 }
 
+/// Colocated asset links refers to the files in the same directory.
+fn is_colocated_asset_link(link: &str) -> bool {
+    !link.starts_with('/') && !link.starts_with('#') && !STARTS_WITH_SCHEMA_RE.is_match(link)
+}
+
 /// Returns whether a link starts with an HTTP(s) scheme.
 fn is_external_link(link: &str) -> bool {
     link.starts_with("http:") || link.starts_with("https:")
@@ -92,6 +105,8 @@ fn fix_link(
                 return Err(format!("Relative link {} not found.", link).into());
             }
         }
+    } else if is_colocated_asset_link(link) {
+        format!("{}{}", context.current_page_permalink, link)
     } else {
         if is_external_link(link) {
             external_links.push(link.to_owned());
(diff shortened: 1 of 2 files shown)


southerntofu:
> #1582 unintentionally broke lots of sites (but no one seemed to notice)

Hello i definitely noticed [and had to work around it ](https://tildegit.org/southerntofu/staticadventures/commit/58bc88b47e90a1df7b53ab87499b1bfb6e8c2ee8). Sorry i didn't catch it earlier that's because i don't have time to update my sites all of the time, but i'll soon setup an automated CI to ensure that my sites continue to build against zola next branch.

I personally believe rewriting those URLs is a good thing as it's consistent with the folder view. To be fair, path rules are a fairly complex topic and there's been [long debates](https://github.com/getzola/zola/issues/977) about how to make them more consistent. I personally haven't followed the conclusions of those debates (i'll get up to speed), but i've personally advocated for searching for [simpler rules for path resolution](https://github.com/getzola/zola/issues/977#issuecomment-759725671) which would support all the types of links we need: standard HTML links (relative/absolute with `[text](link or /link)`) as well as links from the zola site root (`[text](@/.../page.md)`) or even relative to the current page (`[text](@asset_relative_to_markdown.png)`)

I'm not sure about the best syntax or the best choice of building blocks, but i'm sure there's currently a lot of corners we're cutting: for example how can i link to some resource from the site root from markdown? an absolute URL will only work with sites living on the webroot of the domain, so i had to wrap that into a shortcode calling `get_url` which is not the most intuitive.

Keats:
> I personally haven't followed the conclusions of those debates (i'll get up to speed)

That's the status in the end: https://github.com/getzola/zola/blob/master/components/templates/src/global_fns/helpers.rs#L7 but for now it only applies in Tera functions, not in the markdown rendering.

Keats:
Any idea on how to solve this neatly @southerntofu ?

components/rendering/src/markdown.rs
```diff
@@ -62,6 +70,11 @@ fn find_anchor(anchors: &[String], name: String, level: u16) -> String {
     find_anchor(anchors, name, level + 1)
 }
 
+/// Colocated asset links refers to the files in the same directory.
+fn is_colocated_asset_link(link: &str) -> bool {
+    !link.starts_with('/') && !link.starts_with('#') && !STARTS_WITH_SCHEMA_RE.is_match(link)
```
manu-cyber, line 75: Does it make sense to consider `../` as well? Those assets also wouldn’t be colocated, but I don’t know if those links could occur while using zola in a reasonable way.

manu-cyber:
The old behaviour makes more sense for me since when creating a page using `pagename.md`, a subdirectory `pagename` is created nonetheless. To me having to use a quirk for referencing parent section assets for pages created using somewhat of a “shorthand” does not way as heavy as not being able to use subpage assets at all. But then again, I might just be one of the very few people actually making use of this.

components/rendering/src/markdown.rs
timakro: For consistency we want to change links starting with `../` as well. Imagine you want to link to `section/asset.png` from `section/page/index.md`, then you would write `../asset.png`. That would be rewritten to `section/page/../asset.png`, which is correct. Without link rewriting the link would be broken.

manu-cyber: Okay, makes sense. What exactly is `'/'` for then?

timakro: `/` is for absolute links. We should probably prepend `base_url`, but not the page permalink.

manu-cyber: Okay, thanks for explaining! :)

Yaroslav-95:
I just recently decided to try to fix the problem I had in #1705 and was about to make a PR when I stumbled upon this. In my use-case it also makes sense to rewrite the links as absolute, since I need them to be displayed in other parts of the webpage, e.g. as part of previews/summaries of posts. Unfortunately it was broken by #1582, like other use-cases of this functionality as it seems from this discussion.

Keats:
Ok I think this is the last big thing to solve for 0.16.

Does this patch satisfy everyone?

Keats:
ping @irimi1  @timakro @southerntofu @Yaroslav-95 to confirm, not 100% sure what the emoji reactions mean :)

Yaroslav-95:
On 22/06/09 12:53PM, Vincent Prouillet wrote:
> ping @irimi1  @timakro @southerntofu @Yaroslav-95 to confirm, not 100%
> sure what the emoji reactions mean :)
> 

Tested it on my end, and it definitely satisfies me :) Will be very
glad to see this patch on 0.16 so I can finally remove zola from
my package manager's ignore list.

manu-cyber:
> ping @irimi1 @timakro @southerntofu @Yaroslav-95 to confirm, not 100% sure what the emoji reactions mean :)

I’m also happy with that patch :)

Keats:
This is added to the `next` branch.

Keats:
I think we forgot non-images while fixing that? See https://github.com/getzola/zola/issues/1993 and https://github.com/getzola/zola/pull/1965/commits/19b3376317f42c1cd399ab7064c079ba8b13e946 for the fix.

Keats:
Just a quick ping @irimi1 @timakro @Yaroslav-95 to make sure it works for you. I'm getting confused again with that :)

Yaroslav-95:
On 22/10/11 10:55AM, Vincent Prouillet wrote:
> Just a quick ping @irimi1 @timakro @Yaroslav-95 to make sure it works for you. I'm getting confused again with that :)
> 
Works for me™ alright.

-- 

Ярослав де ла Пенья Смирнов
Yaroslav de la Peña Smirnov

manu-cyber:
> Just a quick ping @irimi1 @timakro @Yaroslav-95 to make sure it works for you. I'm getting confused again with that :)

Working for me, too :)

PR #1780, opened 2022-02-25
Bugfix check links i18n (close #1694)

### LINKED ISSUE #1694: check_internal_links_with_anchors crashes on multilinguial website
```
thread 'main' panicked at 'Couldn't find section in check_internal_links_with_anchors', components/site/src/link_checking.rs:66:18
```

The error happens here

https://github.com/getzola/zola/blob/1ed1fdbee7e580375367e5df00f57af86f116842/components/site/src/link_checking.rs#L65

I think this is caused by ` if md_path.contains("_index.md") {` in my case, the file is named `_index.de.md` as it is in german.

If I change the `contains` call for `if md_path.contains("_index.") {` the issue seems to be fixed but I am not fully sure this is the correct fix.

southerntofu:
I faced the same problem as described in #1694. I added the fix suggested over there, as well as a test: internal link in blog/_index.fr.md panics the whole build on the current next branch but passes with this test

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -47,7 +47,7 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
 
     // Check for targets existence (including anchors), then keep only the faulty
     // entries for error reporting purposes.
-    let missing_targets = links_with_anchors.filter(|(_, md_path, anchor)| {
+    let missing_targets = links_with_anchors.filter(|(page, md_path, anchor)| {
         // There are a few `expect` here since the presence of the .md file will
         // already have been checked in the markdown rendering
         let mut full_path = site.base_path.clone();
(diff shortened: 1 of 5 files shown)


components/site/src/link_checking.rs
```diff
@@ -47,23 +47,26 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
...
         let mut full_path = site.base_path.clone();
         full_path.push("content");
         for part in md_path.split('/') {
             full_path.push(part);
         }
-        if md_path.contains("_index.md") {
+        // NOTE: This will also match _index.foobar.md where foobar is not a language
+        // I don't think this is supported by zola so nothing should be broken
+        // Will also prevent having a section like content/_index.foobar/_index.md
+        if md_path.contains("/_index.") {
```
Keats: Does the root section starts with `/`? I can't remember

southerntofu: Very good point i removed the `/` as initially suggested on #1694. I added a note in the docs that any **page** whose path contains `_index.` is unsupported.

Keats (approved):

PR #1790, opened 2022-03-07
switch from chrono to time

mwcz:
This PR replaces the `chrono` crate with the `time` crate.

It also makes a subtle internal change in how times are used within Zola.  Previously, NaiveDateTime from the chrono crate was used everywhere, which omits timezone offset.  If a timezone offset is given to Zola in a post's front matter, it will be flattened into a NaiveDateTime by adding the offset to the datetime.  If you tell Zola "this post happened at 4:00+2:00", Zola would convert that to "6:00".  Now, the offset is preserved, and UTC is assumed when no offset is provided.  The old system was implicitly "convert everything to UTC" and the new system is "assume UTC unless otherwise specified".  There's no change in Zola's behavior, though.

Sanity check:

## Code changes

  - I don't think any documentation needs to be updated, but there is a random typo fix in the docs in a section about time.


src/cmd/serve.rs
```diff
@@ -529,7 +530,12 @@ pub fn serve(
                         if path.is_dir() && is_folder_empty(&path) {
                             continue;
                         }
-                        println!("Change detected @ {}", Local::now().format("%Y-%m-%d %H:%M:%S"));
+
+                        let format = format_description!("[year]-[month]-[day] [hour]:[minute]:[second] UTC");
+                        println!(
+                            "Change detected @ {}",
+                            OffsetDateTime::now_utc().format(&format).unwrap().to_string()
+                        );
```
mwcz: @Keats This timestamp is the only reason the root crate needs a dependency on `time`.  It also includes very very minor regression because with `chrono`, Zola could print the change detected timestamp in the user's local timezone.  [`time` can't do that](https://github.com/time-rs/time/issues/427) in multithreaded applications.  There's some kind of UB/unsoundness around getting local time in a thread, which is why this is now printing UTC times.

What do you think of just doing `println!("Change detected")` and dropping the timestamp?  Then the root crate wouldn't need a dependency on time.

components/front_matter/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 serde = {version = "1.0", features = ["derive"] }
-time = { version = "0.3", features = ["serde", "parsing", "local-offset", "formatting", "std", "macros"] }
```
mwcz: I thought these features were required but it still compiles and runs without them so :shrug:

components/libs/Cargo.toml
```diff
@@ -18,7 +18,7 @@ globset = "0.4"
 unic-langid = "0.9"
 image = "0.24"
 regex = "1"
-time = { version = "0.3", features = ["serde", "parsing", "local-offset", "formatting", "std", "macros"] }
```
mwcz: Ditto here. :shrug:

components/front_matter/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 serde = {version = "1.0", features = ["derive"] }
-
+time = { version = "0.3", features = ["macros"] }
```
Keats, line 8: are the macros only used for tests? If so, we should probably remove that dep entirely from this crate and build them manually

src/cmd/serve.rs
Keats: Is it only on MacOS? I kind of like having the datetime in there to make sure I'm getting the changes when I expect them (eg I see nothing changing on the site and I check the the terminal to see if it has been picked up). 
At the same time, if it's not showing the local timezone it's not worth it.
It's kinda weird it's not possible though, can we get the user offset on startup before starting any threads and re-using it?

mwcz: Seems like that should be possible, I'll look into it.

mwcz: That issue indicates it is only on Mac OS but it was happening to me on Fedora. Calling now_local was Err.

components/front_matter/Cargo.toml
mwcz: I'll check.

src/cmd/serve.rs
mwcz: @Keats Grabbing the UTC offset before threading starts worked like a charm.  patch pushed to this branch.

mwcz: I did have to add another argument to `serve()`, but thankfully `#[allow(clippy::too_many_arguments)]
` was already there. :sweat_smile:

components/front_matter/Cargo.toml
mwcz: We can't remove the crate, or the macro feature, because they're used in for example, front_matter's `parse_datetime`:

https://github.com/getzola/zola/pull/1790/files#diff-313e3525c8ae290a23638c1248f2514eada393c295f005ec3c8702c2f7c521a4R73-R79

Keats (approved):

Keats:
> This is a scheduled windows-2016 brownout. The windows-2016 environment is deprecated and will be removed on April 1st, 2022.

Damn looks like we need to fix the Windows CI build soon.

mwcz:
> > This is a scheduled windows-2016 brownout. The windows-2016 environment is deprecated and will be removed on April 1st, 2022.
> 
> Damn looks like we need to fix the Windows CI build soon.

😬

PR #1794, opened 2022-03-11
Allow `load_data` to take a literal

asimpletune:
Relevant discussion: https://zola.discourse.group/t/allow-load-data-to-take-a-literal/1165

I posted on the Zola discourse site but I don't think anyone responded. In any case, I've being using `POST` + an "echo server" to essentially do the same thing, which is kind of less than ideal. The change was relatively easy to make, so I just did it. 

Basically, I just think this is relatively harmless and would make Zola a tiny bit more flexible and easy to use. It could be something as simple as prototyping and not wanting to have to load from an actual file, or there could be something very complicated that you're doing perhaps, for example maybe you need to deviate from something that's stored in a file, like it's beyond your control, but you still need it in object format. 

In any case, I ran into some circumstances that weren't pleasant to deal with without it, and my **best** solution was the echo server I mentioned above, which isn't super fun to use.

Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/global_fns/load_data.rs
+++ b/components/templates/src/global_fns/load_data.rs
@@ -19,7 +19,7 @@ use utils::fs::{get_file_time, read_file};
 use crate::global_fns::helpers::search_for_file;
 
 static GET_DATA_ARGUMENT_ERROR_MESSAGE: &str =
-    "`load_data`: requires EITHER a `path` or `url` argument";
+    "`load_data`: requires EITHER a `path`, `url`, or `literal` argument";
 
 #[derive(Debug, PartialEq, Clone, Copy, Hash)]
 enum Method {
@@ -82,6 +82,7 @@ impl OutputFormat {
 enum DataSource {
     Url(Url),
     Path(PathBuf),
+    Literal(String),
 }
 
 impl DataSource {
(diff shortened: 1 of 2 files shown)


Keats:
Can you remove the unrelated commits?

asimpletune:
Sure, I’ll rebase when I get home.

asimpletune:
@Keats I fixed the commit history like you asked, so should be ready to review. Thanks.

components/templates/src/global_fns/load_data.rs
```diff
@@ -93,11 +94,12 @@ impl DataSource {
     fn from_args(
         path_arg: Option<String>,
         url_arg: Option<String>,
+        literal_arg: Option<String>,
         base_path: &Path,
         theme: &Option<String>,
         output_path: &Path,
     ) -> Result<Option<Self>> {
-        if path_arg.is_some() && url_arg.is_some() {
+        if !(path_arg.is_some() ^ url_arg.is_some() ^ literal_arg.is_some()) {
```
Keats: let's keep using `&&`

components/templates/src/global_fns/load_data.rs
```diff
@@ -117,6 +119,10 @@ impl DataSource {
                 .map_err(|e| format!("`load_data`: Failed to parse {} as url: {}", url, e).into());
         }
 
+        if let Some(str) = literal_arg {
```
Keats: I thought `str` was a keyword, I didn't realise it could be used like that :o

Keats (changes requested):

components/templates/src/global_fns/load_data.rs
asimpletune: Ok refactored representing "Exactly one of `path`, `url`, or `literal`" from using XOR to using AND's + OR's.

components/templates/src/global_fns/load_data.rs
asimpletune: Um, haha, I read the rust book a few years ago, so I don't really remember a lot. Hopefully `string_literal` is ok? I'm open to learning the naming convention of variables/constants that are strings.

Keats (approved):

Keats:
Thanks!

PR #1827, opened 2022-04-14
Refactor start

Keats:
Still lots of work to do

PR #1833, opened 2022-04-25
apply skip_prefixes before parsing external link domain

mwcz:
From [Handling improperly formed external links](https://zola.discourse.group/t/handling-improperly-formed-external-links/1268) on the forum.

This PR applies `skip_prefixes` before attempting to parse the URL.

Sanity check:

## Code changes

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -96,6 +96,10 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Result<()> {
     }
 }
 
+fn should_skip_by_prefix(link: &String, skip_prefixes: &Vec<String>) -> bool {
+    skip_prefixes.iter().any(|prefix| link.starts_with(prefix))
+}
+
 fn get_link_domain(link: &str) -> Result<String> {
     return match Url::parse(link) {
         Ok(url) => match url.host_str().map(String::from) {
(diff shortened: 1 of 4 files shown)


components/site/src/link_checking.rs
```diff
@@ -109,33 +113,50 @@ fn get_link_domain(link: &str) -> Result<String> {
 pub fn check_external_links(site: &Site) -> Result<()> {
     let library = site.library.write().expect("Get lock for check_external_links");
 
-    let mut all_links: Vec<(PathBuf, String, String)> = vec![];
+    struct LinkDef {
+        file_path: PathBuf,
+        external_link: String,
+        domain: String,
```
mwcz, line 119: I created this struct because I find named fields easier to follow than indexed ones, but I'm happy to switch back to the tuple if that's preferred.

components/site/src/link_checking.rs
```diff
@@ -109,33 +113,50 @@ fn get_link_domain(link: &str) -> Result<String> {
...
+                continue;
+            }
+
             let domain = get_link_domain(&external_link)?;
-            all_links.push((s.file.path.clone(), external_link, domain));
+            all_links.push(LinkDef::new(s.file.path.clone(), external_link, domain));
         }
     }
 
     println!("Checking {} external link(s).", all_links.len());
```
mwcz, line 152: Should we also print a count of skipped links?

mwcz: Should we also print a count of skipped links?

components/site/src/link_checking.rs
Keats: This is fine

components/site/src/link_checking.rs
Keats: Yes

components/site/src/link_checking.rs
```diff
@@ -109,33 +113,50 @@ fn get_link_domain(link: &str) -> Result<String> {
...
     for s in library.sections_values().into_iter() {
         for external_link in s.clone().external_links.into_iter() {
+            if should_skip_by_prefix(&external_link, &site.config.link_checker.skip_prefixes) {
+                continue;
+            }
+
             let domain = get_link_domain(&external_link)?;
-            all_links.push((s.file.path.clone(), external_link, domain));
+            all_links.push(LinkDef::new(s.file.path.clone(), external_link, domain));
         }
```
Keats, line 150: those 2 sections should really be refactored but I'll do that in my branch.

Keats (approved):
LGTM, do you want to do the strictness change in the same PR?
Any chance you can have a look at https://github.com/getzola/zola/pull/1827#issuecomment-1099624251 as well?

components/site/src/link_checking.rs
mwcz: Yeah, I tried combining the iterators with [chain](https://doc.rust-lang.org/std/iter/struct.Chain.html) so there would only need to be one loop, but couldn't get it to work.

components/site/src/link_checking.rs
mwcz: Done https://github.com/getzola/zola/pull/1833/commits/b4fda105ac8ad83b5e7dd7a3a02c99f76664eb9a

mwcz:
> LGTM, do you want to do the strictness change in the same PR?

Hmm, I think a separate PR would be good, since strictness is a new feature and this one is kinda-sorta bugfixy.

Keats (approved):

PR #1841, opened 2022-04-28
Build taxonomies as pages are added

Keats:
In theory at least.


Keats:
cc @mwcz if you are using taxonomies on your site it might help. From my benchmarks it's a tiny tiny bit faster but barely out of the margin of error.

mwcz:
@Keats Taxonomies aren't a good fit for our site since we have paths like `/$foo/$bar/$baz` and I don't think there's currently a way to nest taxonomies like that.

PR #1848, opened 2022-05-03
add link_checker settings for external_level and internal_level

mwcz:
Based on [this convo](https://zola.discourse.group/t/link-checker-warnings-vs-errors/1287).

This PR adds new link checker configuration settings: `internal_level` and `external_level`, which can both be set to `"warn"` or `"error"`.  The current behavior of zola is to always error on broken internal links and on external links with unparsable URLs, so these new settings allow those to be treated as warnings instead.  In other words, it allows zola to build sites even though they have broken links.  The current behavior is still the default.  

Sanity check:

## Code changes


components/console/src/lib.rs
```diff
@@ -0,0 +1,56 @@
+use std::env;
```
mwcz, line 1: This crate has the color stuff from `src/console.rs`.

components/console/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "console"
```
mwcz, line 2: This crate has the color stuff from `src/console.rs`.

components/markdown/src/markdown.rs
```diff
@@ -139,7 +139,27 @@ fn fix_link(
...
+                        console::warn(
+                            format!(
+                                "{}Dead relative link `{}` in {}",
+                                config::LinkCheckerLevel::WarnLevel.log_prefix(),
+                                link,
+                                context.current_page_path.unwrap_or("unknown"),
+                            )
+                            .as_str(),
+                        );
+                        link.to_string() // TODO rendering the broken internal relative link may not be optimal, what href should be rendered to HTML when relative links are invalid but we're in warn mode?
```
mwcz: This needs a sanity check.  In the case of a broken relative link like `@/foobar`, previously the build would panic but if we're in `warn` mode, we need to render something into HTML.  Currently I'm writing the relative link directly into the href, so `<a href="@/foobar">` is what comes out the other end.  There's nothing we can really do to fix the link, but that seemed like the least surprising behavior.  Thoughts?

src/messages.rs
```diff
@@ -0,0 +1,79 @@
+use libs::time::Duration;
+use std::convert::TryInto;
+use std::time::Instant;
+
+use errors::Error;
+use site::Site;
+
+/// Display in the console the number of pages/sections in the site
+pub fn notify_site_size(site: &Site) {
```
mwcz, line 9: This messages.rs module contains the stuff from `src/console.rs` that's about formatting log messages.  The colorization stuff moved to a new crate: `components/console`

components/config/src/config/link_checker.rs
```diff
@@ -1,10 +1,37 @@
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum LinkCheckerLevel {
+    #[serde(rename = "error")]
+    ErrorLevel,
+    #[serde(rename = "warn")]
+    WarnLevel,
```
Keats: Let's call them `Error` and `Warn`, no need to repeat Level

components/console/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+atty = "0.2.11"
+termcolor = "1.0.4"
```
Keats: let's put those deps in `libs` since they don't use macros

components/markdown/src/markdown.rs
```diff
@@ -139,7 +139,27 @@ fn fix_link(
                 resolved.permalink
             }
             Err(_) => {
-                return Err(anyhow!("Relative link {} not found.", link));
+                match context.config.link_checker.internal_level {
+                    config::LinkCheckerLevel::ErrorLevel => {
+                        return Err(anyhow!(
+                            "Dead relative link `{}` in {}",
```
Keats: this one is missing the prefix?
I think it would be easier to add a method to `LinkCheckerLevel` to build the error string.

components/markdown/src/markdown.rs
Keats: I think your behaviour is fine if it's documented

components/site/src/link_checking.rs
```diff
@@ -104,27 +117,32 @@ fn should_skip_by_prefix(link: &str, skip_prefixes: &[String]) -> bool {
...
-fn get_link_domain(link: &str) -> Result<String> {
+fn get_link_domain(link: &str) -> Result<String, String> {
     return match Url::parse(link) {
         Ok(url) => match url.host_str().map(String::from) {
             Some(domain_str) => Ok(domain_str),
-            None => bail!("could not parse domain `{}` from link", link),
+            None => Err(format!("could not parse domain `{}` from link", link)),
         },
-        Err(err) => bail!("could not parse domain `{}` from link: `{}`", link, err),
+        Err(err) => Err(format!("could not parse domain `{}` from link: `{}`", link, err)),
```
Keats: Why that change? It should be equivalent to `bail!`

components/site/src/link_checking.rs
```diff
@@ -148,24 +166,61 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
 
+    // separate the links with valid domains from the links with invalid domains
+    let (checked_links, invalid_url_links): (Vec<&LinkDef>, Vec<&LinkDef>) =
+        checked_links.iter().partition(|link| link.domain.is_ok());
+
+    // get any domains that failed to parse and log them at the configured log level
+    let invalid_link_errs: Vec<String> = invalid_url_links
+        .iter()
+        .map(|link: &&LinkDef| link.domain.as_ref().unwrap_err().clone())
+        .collect();
```
Keats: You can do that without cloning the error by iterating on them when printing

test_site/config.toml
```diff
@@ -22,6 +22,8 @@ taxonomies = "on"
 anchors = "on"
 
 [link_checker]
+internal_level = "error"
+external_level = "error"
```
Keats: That's the default, so let's not add those in the test site

Keats (changes requested):

components/markdown/src/markdown.rs
mwcz: Error messages are prefixed with `Error: `  by anyhow, so they don't need our custom prefix.  Except in the cases where an error message vec is joined with `\n`, then every line except the first one needs `Error: `  prefixed.  It definitely makes handling the messages more awkward and verbose.  If the error/warn messages were identical except for the prefix, the code would be a lot cleaner.  I'm not sure how to achieve that though, with anyhow's automatic prefixing of `Error: `.

components/site/src/link_checking.rs
mwcz: I made that change so it doesn't bail on the first unparseable URL.  It tries to parse everything, and then if there were any failures, it prints out all the unparseable URLs and then bails.

components/site/src/link_checking.rs
```diff
@@ -148,24 +166,61 @@ pub fn check_external_links(site: &Site) -> Result<()> {
...
+            format!(" {} link(s) had unparseable URLs.", invalid_link_errs.len())
+        }
     );
 
+    if !invalid_link_errs.is_empty() {
+        match site.config.link_checker.external_level {
+            // panic if the link checker level is set to error.  bail! can only take one error
+            // message, and it prefixes "Error: ", but we may have accumulated many errors, and it
+            // loos weird if only the first line says "Error: ", so we use join() here to add the
+            // ErrorLevel's log_prefix to each line.
```
mwcz: @Keats here's a description of the discrepancy between prefixing on errors and warnings.

components/site/src/link_checking.rs
mwcz: I'm happy to reverse that change though, I can see how bailing on the first error is more consistent and leads to less complexity.

components/markdown/src/markdown.rs
Keats: ah I forgot anyhow was prefixing Error:

components/site/src/link_checking.rs
Keats: but I believe the code of `get_link_domain` before and the one with your changes is exactly the same no? You're just returning the error either way in `get_link_domain`

It's fine to report all of them though

test_site/config.toml
mwcz: Done :+1:

components/site/src/link_checking.rs
mwcz: Whoops, for some reason I was under the impression that bail caused an immediate panic.  Whoops, I'll change it back.

mwcz: Done 05dde1e2

components/markdown/src/markdown.rs
```diff
@@ -139,7 +139,32 @@ fn fix_link(
...
+                    config::LinkCheckerLevel::Warn => {
+                        console::warn(
+                            format!(
+                                "{}Dead relative link `{}` in {}",
+                                config::LinkCheckerLevel::Warn.log_prefix(),
+                                link,
+                                context.current_page_path.unwrap_or("unknown"),
+                            )
+                            .as_str(),
+                        );
```
Keats: that's duplicated code with `context.config.link_checker.internal_level.log` above no?

components/config/src/config/link_checker.rs
```diff
@@ -1,10 +1,48 @@
...
+    pub fn log(&self, msg: String) {
+        match self {
+            LinkCheckerLevel::Error => {
+                console::error(format!("{}{}", self.log_prefix(), msg).as_str())
+            }
+            LinkCheckerLevel::Warn => {
+                console::warn(format!("{}{}", self.log_prefix(), msg).as_str())
+            }
+        }
+    }
```
Keats: hm having the config depend on the console module is not ideal, maybe it was a bad idea

mwcz: What issue do you see with config depending on console?

components/markdown/src/markdown.rs
mwcz: Yeah, thanks, I missed that one.

components/console/Cargo.toml
mwcz: Done :+1:

components/site/src/link_checking.rs
mwcz: Done :+1:

components/markdown/src/markdown.rs
mwcz: I'm kinda thinking that the `console::warn` and `console:error` functions should be prefixing `"Warning: "` and `"Error: "` instead of putting that in LinkCheckerLevel.

Keats: Would make sense imo

components/config/src/config/link_checker.rs
Keats: Mostly trying to keep down the amount of internal dependencies, the `config` module should not be printing things to stdout/stderr imo, it should just be a place to keep config struct and logic about it (syntect already is way too present in that module but nothing to be done about that)

mwcz: Gotcha, what would you suggest then for logging warnings?  Should I duplicate the terminal colorization code into link_checker to avoid having a dependency on `console`?  Or not colorize warnings at all?  I'm open to whatever.

mwcz: I think https://github.com/getzola/zola/pull/1848#discussion_r866069170 will resolve that problem.  I'll give it a try.

components/site/src/link_checking.rs
mwcz: > It's fine to report all of them though

Question for you.  Since anyhow always prints an error message, and I'm trying to print multiple error messages at a time, how do you feel about these choices:

1. (Current Zola behavior) Bail out when the very first error is encountered (error message is printed by anyhow / unravel errors).  Looks like:
```
Failed to build the site
Error: The anchor in the link `@/post.md#foo` in content/post.md does not exist.
```
2. (Current behavior in this PR) Collect related errors and print them with `console::error`, then bail, passing a "summary" message to anyhow.
```rust
for err in errors { console::error(err) }
return Err(anyhow!("Broken anchor links were found"));
```
Looks like:
```
Failed to build the site
Error: The anchor in the link `@/post.md#foo` in content/post.md does not exist.
Error: The anchor in the link `@/post.md#bar` in content/post.md does not exist.
Error: The anchor in the link `@/post.md#baz` in content/post.md does not exist.
Error: Broken anchor links were found
```
4. (Previous behavior in this PR) Go back to my previous implementation which used `join` judiciously to apply prefixes to 
```rust
return Err(anyhow!(errors.join("\nError: ")));
```
Looks like:
```
Failed to build the site
Error: The anchor in the link `@/post.md#foo` in content/post.md does not exist.
Error: The anchor in the link `@/post.md#bar` in content/post.md does not exist.
Error: The anchor in the link `@/post.md#baz` in content/post.md does not exist.
```
5. Some other solutions I haven't thought of?

Sorry for writing so much, this PR has gone down several different paths and I'm trying to get my bearings.

components/config/src/config/link_checker.rs
mwcz: Done :+1:

components/site/src/link_checking.rs
Keats: > Sorry for writing so much, this PR has gone down several different paths and I'm trying to get my bearings.

No worries!
I was thinking about it today as well and my thoughts might be a bit out of scope for the PR.

What if the link checker was not using `Error`s but was instead using a `Vec<String>` string for the errors.
This way we can build the error message ourselves nicely, something like:

```
Failed to build the site
Error: Found 3 broken anchor links
  1. The anchor in the link `@/post.md#foo` in content/post.md does not exist.
  2. The anchor in the link `@/post.md#bar` in content/post.md does not exist.
  3. The anchor in the link `@/post.md#baz` in content/post.md does not exist.
```

We can/should obviously apply the same kind of .formatting to external link errors obviously.
What do you think?

mwcz: I can imagine `check_external_links` and `check_internal_links_with_anchors` emitting a `Vec<String>` which contains the text for all the errors or warnings that were encountered.  Those functions are called from `site/src/lib.rs`, so I suppose that's where the logging could happen.  Seems like a clean solution; the logging code would only need to exist in `site/src/lib.rs` instead of being duplicated at every early-return site.  It would potentially change the ordering of error/warning logs relative to regular printlns since the errors/warnings would be delayed until the function returns, but the link checker hardly has any printlns so I don't see a problem there.  Seems good to me, so I guess I'll give it a try next week.

mwcz: The internal link checking that happens in `markdown.rs` would need a similar treatment I suppose, so the code would be more-or-less duplicated, but only in two places. :thinking:

Keats: Yes the markdown one is unfortunate... maybe keep it as is for now and see if it feels odd when trying it out.

mwcz: It seems pretty good.  The `markdown.rs` link checking errors/warnings aren't formatted exactly the same, but it doesn't feel odd when using zola.  Here's a demo.

https://asciinema.org/a/Ructp25lMiFB1ztSatUr74ASn

components/config/src/config/link_checker.rs
mwcz: :+1:

Keats (approved):
Sweet thanks

Keats:
FYI I just pushed a commit to make the external link checking a bit more DRY: https://github.com/getzola/zola/pull/1741/commits/1a3b78312082d0a6054c9a09370937b085764081

PR #1853, opened 2022-05-07
Enables setting base-url to / for relative URLs in serve mode

### LINKED ISSUE #1568: Relative Urls support
I have been using zola for creating a simple static blog and it is working out great. However, I don't see any way to have relative links for pages, only the permalink variable is available. 
Hugo has a **relativeURLs** configuration option which basically converts absolute links to relative links. For example, if your `/posts/first/` page contains a link to `/about/`, Hugo will rewrite the URL to `../../about/`.

# Usage
I am currently hosting my blog on ipfs using [Fleek](https://fleek.co/). 
If I am using a browser like Brave, then the HTTP address will resolve and directly browse the site from ipfs. For example:
```
https://example.com -> ipns://[HASH]
```
But the base_url is `https://example.com `, so anytime a permalink comes up, first the HTTP site opens which then resolves to ipfs. This just adds unnecessary steps, especially for things like the links to full post from post preview. It is significantly slower compared to a relative link.

So, is it already possible to do so or are there any plan to support relative URLs in Zola? I think there might be some other use cases where relative links are more useful than absolute links.

colindean:
Having to change the base URL to whatever ngrok URL was provisioned for
me was frustrating. This patch enables setting it to `/`, which will
then make the `get_url()` function simply return `/`.

Fixes #1568 for at least `zola serve` in my local testing.

Sanity check:

## Code changes

### DIFF
--- a/docs/content/documentation/getting-started/cli-usage.md
+++ b/docs/content/documentation/getting-started/cli-usage.md
@@ -74,7 +74,7 @@ You can also specify different addresses for the interface and base_url using `-
 
 > By default, devices from the local network **won't** be able to access the served pages. This may be of importance when you want to test page interaction and layout on your mobile device or tablet. If you set the interface to `0.0.0.0` however, devices from your local network will be able to access the served pages by requesting the local ip-address of the machine serving the pages and port used.
 >
-> In order to have everything work correctly, you might also have to alter the `base-url` flag to your local ip.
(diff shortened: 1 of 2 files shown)


Keats (approved):

colindean:
Thank you!

Keats:
Thank you!

PR #1862, opened 2022-05-16
Build static binary on linux

bemyak:

Sanity check:

## Code changes

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,9 +60,9 @@ stages:
               rustup_toolchain: stable
               target: 'x86_64-apple-darwin'
             linux-stable:
-              imageName: 'ubuntu-20.04'
+              imageName: 'rust:alpine'
               rustup_toolchain: stable
-              target: 'x86_64-unknown-linux-gnu'
+              target: 'x86_64-unknown-linux-musl'
         pool:
           vmImage: $(imageName)
         steps:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -60,17 +60,17 @@ stages:
               rustup_toolchain: stable
               target: 'x86_64-apple-darwin'
             linux-stable:
-              imageName: 'ubuntu-20.04'
+              imageName: 'rust:alpine'
               rustup_toolchain: stable
-              target: 'x86_64-unknown-linux-gnu'
+              target: 'x86_64-unknown-linux-musl'
```
Keats, line 65: that would break all the package managers, if you want to add musl, it needs to be another target, not replace the existing one.

azure-pipelines.yml
```diff
@@ -60,17 +60,17 @@ stages:
...
         pool:
           vmImage: $(imageName)
         steps:
           - script: |
               curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain $RUSTUP_TOOLCHAIN
               echo "##vso[task.setvariable variable=PATH;]$PATH:$HOME/.cargo/bin"
             displayName: Install rust
-            condition: ne( variables['Agent.OS'], 'Windows_NT' )
+            condition: eq( variables['Agent.OS'], 'Darwin' )
```
Keats, line 73: ?

Keats (changes requested):

azure-pipelines.yml
bemyak: Why would it? All the distros compile it themselves, IMO the only use case for a pre-compiled binary is downloading it to some arbitrary system, which we can't make any assumptions about. So it seems to me that for Linux we should ship only static binaries.

azure-pipelines.yml
bemyak: For Linux the official `rust:alpine` image is used, which has all the needed stuff installed. So the only target that needs explicit toolchain installation is `Darwin`

azure-pipelines.yml
Keats: > All the distros compile it themselves, IMO the only use case for a pre-compiled binary is downloading it to some arbitrary system, which we can't make any assumptions about

I would rather have both options than only one

> So it seems to me that for Linux we should ship only static binaries.

Last time I checked musl, performance was significantly worse. It might have been fixed since then but my results were kinda similar to https://andygrove.io/2020/05/why-musl-extremely-slow/

bemyak: Thanks for pointing this out, I didn't know. I'll close this PR then.

PR #1873, opened 2022-05-23
Update to elasticlunr-rs 3.0.0

mattico:
I'm planning to release elasticlunr-rs 3.0.0 soon. The one noticeable change for you is the new `IndexBuilder::add_field_with_tokenizer` function.

I wanted to give a chance for feedback before I release the final version.

### DIFF
--- a/components/search/src/lib.rs
+++ b/components/search/src/lib.rs
@@ -1,9 +1,7 @@
 use std::collections::{HashMap, HashSet};
 
 use libs::ammonia;
-use libs::elasticlunr::pipeline;
-use libs::elasticlunr::pipeline::TokenizerFn;
-use libs::elasticlunr::{Index, Language};
+use libs::elasticlunr::{lang, Index, IndexBuilder};
 use libs::once_cell::sync::Lazy;
 
 use config::{Config, Search};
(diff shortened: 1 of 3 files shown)


components/search/src/lib.rs
```diff
@@ -55,34 +52,6 @@ fn path_tokenizer(text: &str) -> Vec<String> {
...
 }
 
-fn build_tokenizers(search_config: &Search, language: Language) -> Vec<TokenizerFn> {
-    let text_tokenizer = match language {
-        #[cfg(feature = "indexing-zh")]
-        Language::Chinese => pipeline::tokenize_chinese,
-        #[cfg(feature = "indexing-ja")]
-        Language::Japanese => pipeline::tokenize_japanese,
-        _ => pipeline::tokenize,
-    };
```
Keats, line 65: Additional languages gated by features work automatically now?

Keats (commented):
lgtm

components/search/src/lib.rs
mattico: Tokenization is now included in the language trait.

Keats: Looks good then, I'll merge it when 3.0 is released

Keats (approved):

PR #1874, opened 2022-05-23
Little exactness tweaks

jonjensen:
Sanity check:

## Code changes

* [N/A] Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -419,7 +419,7 @@ pub fn serve(
         broadcaster
     };
 
-    println!("Listening for changes in {}{{{}}}", root_dir.display(), watchers.join(", "));
+    println!("Listening for changes in {}{{{}}}", root_dir.display(), watchers.join(","));
 
     println!("Press Ctrl+C to stop\n");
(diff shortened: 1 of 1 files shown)


src/cmd/serve.rs
```diff
@@ -419,7 +419,7 @@ pub fn serve(
         broadcaster
     };
 
-    println!("Listening for changes in {}{{{}}}", root_dir.display(), watchers.join(", "));
```
Keats, line 422: Why? It's not an actual regex but the space is there to make it easier to read

Keats (changes requested):

src/cmd/serve.rs
jonjensen: It's written like a shell brace-expansion (not a regex) and I copied and pasted it to a shell command (`ls`, if I remember right) to look at the timestamps on those files, and it didn't work because the spaces end up being wrong extra characters in the filesystem paths.

For someone just looking at it the extra space doesn't matter, but in that case, could leave off the `{}` so it isn't taken literally.

Keats (approved):

PR #1875, opened 2022-05-24
Add `charset=utf-8` when serving site HTML files

lnicola:
This is useful when the site uses emoji or other non-Latin characters, which can get jumbled by the browser. We're not setting the charset for non-Zola-generated pages, but that's probably less important.

image

image

Sanity check:

## Code changes

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -158,12 +158,9 @@ fn livereload_js() -> Response<Body> {
 
 fn in_memory_content(path: &RelativePathBuf, content: &str) -> Response<Body> {
     let content_type = match path.extension() {
-        Some(ext) => match ext {
-            "xml" => "text/xml",
-            "json" => "application/json",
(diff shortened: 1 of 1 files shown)


Keats:
Wouldn't it work if you set the charset in your template HTML instead?

lnicola:
It would, but it seems pointless if the server already send the right header, and it seems like a useful feature anyway for someone less used to the intricacies of HTTP.

But it's fine if you think that's a good enough workaround and you don't want to merge this.

Keats:
> It would, but it seems pointless if the server already send the right header, and it seems like a useful feature anyway for someone less used to the intricacies of HTTP.

By doing that, you end up in a situation where `zola serve` displays correctly but the output of `zola build` might not. It's better to have the charset in the HTML so it works in both ways. Making it work silently in `zola serve` would actually potentially bite users.

Issue #1878, opened 2022-05-27
Unable to reliably get the current URL on a multilang site


bemyak:
# Bug Report

There two ways of getting the current web-page URL:

* Use `current_url` variable
* Call `get_url(path=current_path, lang=lang)` function

Both of these work identically for the default language, however, the behavior is different and sometimes unexpected for a secondary language (`fi` in the examples below).

## Environment

Zola version: 0.15.3, `next`

## Expected Behavior

URLs reported by `current_url` and `get_url(path=current_path, lang=lang)` match each other and the actual URL

## Current Behavior

* `/fi/`
  *  get_url = /fi/fi
  *  current_url = /fi/
* `/fi/post`
  *  get_url = /fi//fi/post
  *  current_url = /fi/post/
* `/fi/tags` **this is the most problematic part**
  * get_url = /fi//tags
  * current_url = /tags/  <-- lang is missing
* `/fi/tags/fi-tag/`
  *  get_url = /fi//tags/fi-tag
  *  current_url = /tags/fi-tag/ <-- lang is missing again

For most pages, `current_url` gives the correct result, except for taxonomies, where it misses the language part, so `get_url` is _more_ correct here.

This messes up setting the "canonical URLs" for webpages, so some search engines are not happy with it.

## Step to reproduce

Checkout the [attached site](https://github.com/getzola/zola/files/8786693/test-site.zip)

Keats:
This should be an easy fix if someone wants to try.

PR #1882, opened 2022-05-28
Fix taxonomy current path

bemyak:
Fixes https://github.com/getzola/zola/issues/1878

---

### Sanity check:

### DIFF
--- a/components/content/src/taxonomies.rs
+++ b/components/content/src/taxonomies.rs
@@ -119,6 +119,7 @@ pub struct Taxonomy {
     pub kind: TaxonomyConfig,
     pub lang: String,
     pub slug: String,
+    pub path: String,
     pub permalink: String,
     // this vec is sorted by the count of item
     pub items: Vec<TaxonomyTerm>,
@@ -159,6 +160,7 @@ impl Taxonomy {
             slug,
             lang: tax_found.lang.to_owned(),
             kind: tax_found.config.clone(),
+            path,
             permalink,
             items: sorted_items,
         }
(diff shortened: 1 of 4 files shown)


components/templates/src/global_fns/files.rs
```diff
@@ -104,7 +104,7 @@ impl TeraFn for GetUrl {
             // anything else
             let mut segments = vec![];
 
-            if lang != self.config.default_language {
+            if lang != self.config.default_language && !path.starts_with(&format!("/{}", lang)) {
```
Keats: you can do that without allocating by slicing I believe?

components/templates/src/global_fns/files.rs
```diff
@@ -399,6 +399,29 @@ title = "A title"
         );
     }
 
+    #[test]
+    fn does_not_duplicate_lang() {
```
Keats, line 403: what does it mean to duplicate lang?

components/content/src/taxonomies.rs
```diff
@@ -204,8 +203,8 @@ impl Taxonomy {
         context.insert("terms", &terms);
         context.insert("lang", &self.lang);
         context.insert("taxonomy", &self.kind);
-        context.insert("current_url", &config.make_permalink(&self.kind.name));
-        context.insert("current_path", &format!("/{}/", self.kind.name));
+        context.insert("current_url", &self.permalink);
+        context.insert("current_path", &self.path);
```
Keats, line 207: so we already had them eh

Keats (changes requested):

components/templates/src/global_fns/files.rs
bemyak: Can you elaborate please?

components/templates/src/global_fns/files.rs
bemyak: It's described in https://github.com/getzola/zola/issues/1878.

TL;DR: currenly `get_url(path='/lang/something', lang='lang')` will return a URL with duplicated `lang` part: `/lang//lang/something`

components/content/src/taxonomies.rs
bemyak: Yes, we indeed had both `permalink` and `path` already computed in the `new` function, but `path` was discarded and `permalink` wasn't used. I've made it to match other similar structs e.g., `Path`

components/templates/src/global_fns/files.rs
Keats: Something like `path[1..].starts_with(lang)`. Haven't tried it so it might not work

bemyak: Indeed, this works! Thanks for the handy trick :)

I've amended the last commit.

Keats (approved):

Issue #503, opened 2018-10-29
Follow symbolic link directories


getreu:
Gutenberg follows symbolic links to files already, but not yet symbolic links to directories.
This feature would be very handy, because then you could just link assets under `/static` wherever they live. The same reasoning applies to `/content` and the other directories,

Keats:
I can see the point for the `static` folder but I'm not sure I see it for `content`. When would you have gutenberg md files used outside of the `content` folder?

getreu:
A useful consistent policy could be:
1) All absolute and relative symbolic links (dirs and files) referencing assets *outside* the Gutenberg tree are followed: The referenced assets are copied into `/public/`.
2) All relative symbolic links (dirs and files in `/content` or `/static`) referencing assets  *inside*  `/content` or `/static` not followed. The relative symbolic links are copied *as they are* into `/public`. They will be followed later by the Web-server, when the web page is accessed.

Advantages and use case:
1) This feature would be very handy, because then you could just link assets under /static wherever they actually live. The same reasoning applies to `/content` and the other directories.  Use case: documentation and blog content is kept in each project folder and are linked into the Gutenberg tree with symbolic folders. No need to copy assets from outside into `/content` and `/static` before launching `gutenberg`.
2) Internal symbolic links are useful to avoid redundancy when you have identical content on different pages.

The most important is to implement 1) the "follow links behaviour" for symbolic directory links (for files it works already).

No 2) is a *nice to have*

getreu:
BTW: I am using a script to assemble assets to be published dynamically using symbolic directory links. If you are interested how it works see 1.3 in [A collection of Bash scripts - bash-scripts.pdf](https://blog.getreu.net/_downloads/bash-scripts.pdf).

Keats:
@getreu are you willing to work on a PR to add that?

getreu:
@Keats : I would like to and had a look into your code by curiousity (knowing that I do not have any free resources until the end of the year). I like the way you modulerized your code base.
Concerning the work load for a PR on [this](https://github.com/Keats/gutenberg/issues/503#issuecomment-434392896): 
(Policy 1): you follow symbolic file links already, following symbolic directory links should be straightforward: *2-3 lines of code* 
(Policy 2): To decide if a symbolic link points inside the "path" tree  and - if it does - create  the same relative symbolic link in `./public/` is not very complicated either (special case here is Windows OS): *2-3 lines of code.*
More work is to enhance `test-site` and to add some appropriate tests. 

[...]

PR #1883, opened 2022-05-29
Always follow symlinks

bemyak:
Despite what I've said [here](https://github.com/getzola/zola/issues/503#issuecomment-699448917), symlinks don't work ATM, only an empty dir is created without contents.

---

Sanity check:


Keats:
Do we have no tests for that?

bemyak:
We don't. I'm not sure if we should though, symlinking is a file-system dependent feature, I don't know what FS CI is using

Issue #1572, opened 2021-08-01
Provide a Flatpak as software package for Linux systems


Ayush1325:
Flatpak like snaps are containerized software packages that are simple to create and install. It also works on all major Linux 
systems without modification.

The one major area where flatpak and snap differ is that snap needs to create a symlink in root filesystem which is not possible in immutable OS like [Fedora Silverblue](https://silverblue.fedoraproject.org/) or MicroOS. See [here](https://github.com/coreos/rpm-ostree/issues/1711) for more information. This pretty much renders usage of snaps impossible. Also, the Fedora package is also currently at 0.12.2 version in Fedora 34, which is about 10 months old at this point. So I think a Flatpak version would be a good idea.

I have a basic version of the manifest file [here](https://github.com/Ayush1325/zola/blob/21889805f84e7808ad2630f978485011020f0c26/flatpak/org.getzola.zola.yml) which works for the most part but has a few problems with git submodules in this project (so syntax highlighting not working yet).

If having a flatpak version is acceptable to everyone, then I can create a Draft pull request and iron out the problems with submodules? 

Also, I am not an expert on flatpak. This is my first attempt at creating a flatpak so it would be great if there was someone already familiar with it and can catch mistakes and problems which might not be obvious to me.

doamatto:
Imo, a Flatpak version isn't necessary per-say. It's already a one bin solution that has your main binaries pre-built in the releases (x64), with it being a simple one command to build (bar installing things like MSVC, g++, make, et al).

In other words, I don't know if a Flatpak would really be used, but I wouldn't be opposed to it, either. Kind of a could take it or leave it opinion. That, and its possible sandboxing could cause issues, not to mention having to use `flatpak run org.getzola.zola [command]` for everything (afaik).

Ayush1325:
Well, in my current flatpak build, all commands seem to be working in the blog. 
The only thing that doesn't work is syntax highlighting since flatpak-builder was having some problem with downloading git submodules.

The reason I prefer flatpak is because I am using Fedora silverblue as my daily driver. I started off with using Zola inside a toolbox (they are just fancy containers if u haven't used silverblue). Anyway, as I said, the snap isn't an option in silverblue and the version in Fedora repo is old. So I thought if there was a snap, we can have a flatpak too.

As for the flatpak run thing, I'm kinda used to at this point, so don't really find it problematic, I mostly just create alias when required, but yeah, that's why I thought creating an issue might be better to see if a flatpak version is helpful.

[...]

PR #1900, opened 2022-06-15
Add AppData

Ayush1325:
Required for publishing to Flathub

Sanity check:

Related to #1572

### DIFF
--- a/zola.metainfo.xml
+++ b/zola.metainfo.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<component type="console-application">
+  <id>org.getzola.zola</id>
+  <icon type="remote">https://github.com/getzola/zola/blob/72461a1fc9a767d4eadb6be48a0c5ba35e974049/docs/static/favicon.ico</icon>
+
+  <name>Zola</name>
+  <summary>A fast static site generator in a single binary with everything built-in</summary>
+  <developer_name>Vincent Prouillet and contributors</developer_name>
+
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
+
+  <description>
+    <p>
+      Zola comes as a single executable with Sass compilation, syntax highlighting, table of contents and many other features 
+      that traditionally require setting up a dev environment adding some JavaScript libraries to your site.
+    </p>
+    <p>
+      The average site will be generated in less than a second, including Sass compilation and syntax highlighting.
+    </p>
+    <p>
+      Zola renders your whole site as static files, making it trivial to handle any kind of traffic you will throw at it at no 
+      cost without having to worry about managing a server or a database.
+    </p>
+    <p>
+      From the CLI to the template engine, everything is designed to be intuitive. Don't take my word for it though, look at the
+      documentation and see for yourself.
+    </p>
+    <p>
+      Zola gets out of your way so you can focus on your content, be it a blog, a knowledge base, a landing page or a combination 
+      of them.
+    </p>
+    <p>
+      Zola comes with shortcodes and internal links to make it easier to write your content.
+    </p>
+  </description>
+
+  <categories>
+    <category>Development</category>
+    <category>WebDevelopment</category>
+  </categories>
+
+  <url type="homepage">https://www.getzola.org/</url>
+  <url type="bugtracker">https://github.com/getzola/zola/issues</url>
+  <url type="help">https://www.getzola.org/documentation/</url>
+
+  <provides>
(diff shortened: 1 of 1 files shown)


zola.appdata.xml
```diff
@@ -0,0 +1,63 @@
...
+    <binary>zola</binary>
+  </provides>
+
+  <content_rating type="oars-1.1" />
+
+  <releases>
+    <release version="0.15.3" date="2022-01-22">
+      <url>https://github.com/getzola/zola/releases/tag/v0.15.3</url>
+      <description>
+        Zola Static Site Generator v0.15.3
```
Keats: Any way to fill that automatically on each version? I'm already forgetting to update the snapcraft file half of the time.
Does this need to be in this repo or could someone maintain it in their own repo?

Ayush1325: It is possible to keep it in the flathub repo but it is encouraging to put it upstream (https://github.com/flathub/flathub/wiki/App-Maintenance#acceptable-but-should-be-submitted-upstream). 

Currently, it is present in the flathub pull request as well (https://github.com/flathub/flathub/pull/3273). I will remove it from there if this gets merged.

Newbytee:
Please rename the file to zola.metainfo.xml. appdata.xml is an old name. See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

zola.appdata.xml
Newbytee: Some people invent scripts for doing it automatically, but other than that, no.

Ayush1325:
> zola.metainfo.xml

Done

zola.metainfo.xml
```diff
@@ -0,0 +1,63 @@
...
+  </provides>
+
+  <content_rating type="oars-1.1" />
+
+  <releases>
+    <release version="0.15.3" date="2022-01-22">
+      <url>https://github.com/getzola/zola/releases/tag/v0.15.3</url>
+      <description>
+        Zola Static Site Generator v0.15.3
+      </description>
```
Newbytee: Since this section doesn't really contain any information about the release, I would suggest you remove it.

Ayush1325: Should I paste the changelog there instead?

Newbytee: Sure, you can do that. Here's an example if it's helpful: https://github.com/flathub/com.jpexs.decompiler.flash/blob/master/com.jpexs.decompiler.flash.metainfo.xml#L66

Ayush1325: @Newbytee I think it would be better to remove the description. Since the changelog is autogenerated, it's not all that useful.

Newbytee: Sounds good to me.

zola.metainfo.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<component type="console-application">
+  <id>org.getzola.zola</id>
+  <icon type="local">docs/static/favicon.ico</icon>
+
+  <name>Zola</name>
+  <summary>A fast static site generator in a single binary with everything built-in</summary>
+  <developer_name>Vincent Prouillet</developer_name>
```
Newbytee: @Keats What do you think about this? We could also put "Vincent Prouillet and contributors" or "Zola contributors".

Ayush1325: I think "Vincent Prouillet and contributors" should be good. That's what is present in the website footer.

zola.metainfo.xml
```diff
@@ -0,0 +1,60 @@
...
+  <provides>
+    <binary>zola</binary>
+  </provides>
+
+  <content_rating type="oars-1.1" />
+
+  <releases>
+    <release version="0.15.3" date="2022-01-22">
+      <url>https://github.com/getzola/zola/releases/tag/v0.15.3</url>
+    <release>
```
Newbytee:
```suggestion
    </release>
```

Keats:
Just to double-check: this can be closed?

Ayush1325:
> Just to double-check: this can be closed?

If the metadata file cannot be used by any other zola packages, then I guess it doesn't need to be merged.

Newbytee:
> Just to double-check: this can be closed?

We currently have the metainfo downstream in the Flathub Flatpak manifest, but it would be preferable (albeit not strictly necessary) to have it here.

Keats:
I don't mind having it here as long as someone does a PR on new releases to update it if I forget.

Newbytee:
> I don't mind having it here as long as someone does a PR on new releases to update it if I forget.

I can make sure that happens.

Issue #1904, opened 2022-06-20
`resize_image` doesn't work with WebP images


bemyak:
# Bug Report

## Environment

Zola version: `0.15.3`

## Expected Behavior

`resize_image` function is able to process WebP input images

## Current Behavior

When trying to supply a WebP image to `resize_image` function I get this error:

```bash
$ zola build
Building site...
Failed to build the site
Error: Failed to render content of <path>/content/post.md
Reason: Failed to render img shortcode
Reason: Failed to render 'shortcodes/img.html'
Reason: Function call 'resize_image' failed
Reason: `resize_image`: Failed to read image: <path>/content/zola.webp
```

## Step to reproduce

Try to build the [attached website](https://github.com/getzola/zola/files/8943469/zola-webp-bug.zip). Then replace `zola.webp` to `zola.png` and observe the expected behavior.

Keats:
Hmm it works for me?

bemyak:
I messed something up a bit when I was testing this. The issue **doesn't** occur on the `next` version, only on `0.15.3` (haven't tried earlier versions).

Guess it can be closed then.

PR #1912, opened 2022-06-26
Apply orientation transformation based on EXIF data

bemyak:
Some images (especially photos) are stored with incorrect orientation. There is a EXIF metadata attached, so browsers and image viewers know how to display it properly.

Unfortunately, when applying transformations, this metadata isn't copied (see https://github.com/image-rs/image/issues/1045#issuecomment-733790099), so the only way to fix it is to apply the corresponding transform on the Zola's side.

---

## Sanity check:

## Code changes

* [?] Have you created/updated the relevant documentation page(s)? (Should I?)


bemyak:
Just to clarify: all 8 new images are differently oriented, but they have the correct EXIF tag set, so the browser displays them properly. The issue can be reproduced if you try to `resize_image` any of them (except the first one, it has no rotation).

Keats (changes requested):
That seems good. The GH preview shows all image with the top left being red, can you add a detail in one of the test what is the exif data in each?

bemyak:
The number in an image filename indicates the value of the `Orientation` tag. For `exif_0.jpg` the tag is not set. I've added the corresponding descriptions.

For reference, the image description is taken from an output of this command:
```bash
$ exiftool -Orientation exif_5.jpg
Orientation                     : Mirror horizontal and rotate 270 CW
```

You can get the raw value like this:
```bash
$ exiftool -b -Orientation exif_5.jpg
5
```

components/imageproc/src/lib.rs
```diff
@@ -343,6 +346,35 @@ impl ImageOp {
     }
 }
 
+/// Apply image rotation based on EXIF data
+pub fn fix_orientation(img: DynamicImage, path: &Path) -> DynamicImage {
+    let file = std::fs::File::open(path).unwrap();
```
Keats: Can we avoid unwrap?

components/imageproc/src/lib.rs
```diff
@@ -343,6 +346,35 @@ impl ImageOp {
...
+    let mut buf_reader = std::io::BufReader::new(&file);
+    let exif_reader = exif::Reader::new();
+    let exif = exif_reader.read_from_container(&mut buf_reader);
+    let exif = match exif {
```
Keats: you can put ` exif_reader.read_from_container(&mut buf_reader)` after `match` and you avoid one variable named exif

Keats (changes requested):
Small nits.
Do you have an example site where Zola was displaying things incorrectly after an op? Just to easily try this branch somewhere

bemyak:
> Do you have an example site where Zola was displaying things incorrectly after an op? Just to easily try this branch somewhere

Yep, you can clone [this](https://gitlab.com/bemyak/blog.git), and then open `/ru/finland/tuomarilanrinne/` url. The first image (with a baby) is oriented incorrectly, but if you click on it, the original will be opened, which didn't pass the transformation, so it displays correctly.

To build the blog you'll probably need [git-lfs](https://git-lfs.github.com/) installed and zola `next`, it doesn't build on stable because of https://github.com/getzola/zola/issues/1904

Also when trying different versions remember to delete `static/processed_images` dir.

bemyak:
Alternatively, you can just use the test images from this PR, it'll probably be faster. The blog is pretty heavy.

Keats (approved):
Perfect, thanks!

PR #1917, opened 2022-07-08
Add sorting by path

bemyak:
This adds sorting by file system path. Very useful when you name your pages like `00_post.md`, `01_post.md`.

---

Sanity check:

### DIFF
--- a/components/content/src/sorting.rs
+++ b/components/content/src/sorting.rs
@@ -202,6 +202,16 @@ mod tests {
         assert_eq!(pages[1], page1.file.path);
         assert_eq!(pages[2], page2.file.path);
         assert_eq!(ignored_pages.len(), 0);
+
+        // 10 should come after 2
+        let page1 = create_page_with_title("1");
+        let page2 = create_page_with_title("10");
+        let page3 = create_page_with_title("2");
+        let (pages, ignored_pages) = sort_pages(&[&page1, &page2, &page3], SortBy::Path);
+        assert_eq!(pages[0], page1.file.path);
+        assert_eq!(pages[1], page3.file.path);
+        assert_eq!(pages[2], page2.file.path);
+        assert_eq!(ignored_pages.len(), 0);
     }
 
     #[test]
--- a/components/content/src/sorting.rs
+++ b/components/content/src/sorting.rs
@@ -1,5 +1,5 @@
 use std::cmp::Ordering;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::{Page, SortBy};
 use libs::lexical_sort::natural_lexical_cmp;
@@ -16,6 +16,7 @@ pub fn sort_pages(pages: &[&Page], sort_by: SortBy) -> (Vec<PathBuf>, Vec<PathBu
             }
             SortBy::Title | SortBy::TitleBytes => page.meta.title.is_some(),
             SortBy::Weight => page.meta.weight.is_some(),
+            SortBy::Path => true,
             SortBy::None => unreachable!(),
         });
 
@@ -32,6 +33,8 @@ pub fn sort_pages(pages: &[&Page], sort_by: SortBy) -> (Vec<PathBuf>, Vec<PathBu
                 a.meta.title.as_ref().unwrap().cmp(b.meta.title.as_ref().unwrap())
             }
             SortBy::Weight => a.meta.weight.unwrap().cmp(&b.meta.weight.unwrap()),
+            SortBy::Path => compare_by_path_lexically(&a.file.path, &b.file.path)
+                .unwrap_or_else(|| a.file.path.cmp(&b.file.path)),
             SortBy::None => unreachable!(),
         };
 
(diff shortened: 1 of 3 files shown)


bemyak:
Lexical sorting was chosen because most of the modern file explorers (windows explorer, dolphin), and text editors (vs code, haven't checked others) sort entries like this, so the idea is to have the same order in Zola.

BTW, not sure why mac tests have failed, there are no changes that could cause this. Can you please rerun it somehow?

Keats:
I'm not sure it makes sense to add that, by default the path of a path will come from the filename if it's not set explicitly and we can already sort on that.
If you specify a path/slug for each, then I think it's fine to set a weight in each as well.

bemyak:
> we can already sort on that.

I haven't figured out how. The default sorting seems to put pages in random order 🤔

c-git:
> > we can already sort on that.
> 
> I haven't figured out how. The default sorting seems to put pages in random order thinking

Yes the default is not sorted. According to the [docs](https://www.getzola.org/documentation/content/section/#sorting-pages) "If sort_by is not set, the pages will be sorted in the none order, which is **not intended for sorted content**.".

Would setting `sort_by` to `title` meet your needs?

bemyak:
The title is meant to be meaningful, I don't want to prefix it with numbers just for the sake of sorting.

Currently I'm using weights, but it's a bit tedious to put them everywhere, while page filenames already have these numbers.

Since sorting by path / slug is always possible I would even consider it a candidate for the default sorting, I can't imagine a real use case when you want your pages being sorted randomly.

c-git:
I agree the title should be meaningful. I also feel like the slug should be meaningful but that's a personal thing I suppose.

I see, I think I understand now. You have these numbers as prefixes already prefixed to the file names and right now are manually putting them into the weights but the problem is the effort to put them into the weights. Is that correct? If so I would say that a solution may be to have a small program/script that passes over you directory structure and updates the weights using the prefixes.

I haven't looked at any implementations for zola but I can see that a default of `none` sorting having optimizations benefits when no sorting is desired (Don't pay for what you don't use kind of thing). That said I don't know the zola culture but even if a new sort method were to be supported I would think it in poor taste to change the default because that unnecessarily breaks backward compatibility. As for a use case of not needing sorting I don't readily have one but it's not hard to imagine where you only access a set of pages by their links and never use the table of contents and therefore having it sorted provides no value.

Keats:
The fact that nothing is sorted by default is not for optimization, it's mostly so people don't rely on it and are explicit about what they want.

We could imagine a way to set up weights from filename though, like we do for dates as mentioned in https://www.getzola.org/documentation/content/page/

bemyak:
> You have these numbers as prefixes already prefixed to the file names and right now are manually putting them into the weights but the problem is the effort to put them into the weights. Is that correct?

Yes. My main concern here is keeping them in sync. And I don't want to put an additional burden of running scripts on an artist who is the one updating the site.

> We could imagine a way to set up weights from filename though, like we do for dates as mentioned in

I think if we do this, it'll be quite a breaking change, sine `slug`s of existing pages will change. Users may rely on the current behavior and would like to get it back, so we'll have to introduce a new configuration option just to turn it off.

To me, having this sorting options seems much more transparent.

Keats:
> I think if we do this, it'll be quite a breaking change, sine slugs of existing pages will change. Users may rely on the current behavior and would like to get it back, so we'll have to introduce a new configuration option just to turn it off.

We can make sure so it's only if the path starts with eg 4 digits followed by `-` like we do for datetimes. In practice adding that didn't cause any issues.

bemyak:
This is how I see pros and cons of these two approaches:

### `sort_by = "path"`
Pros:
* User-controllable. You can opt-in or opt-out transparently.
* Goes inline with other sorting options (except date-prefixed, but that's a special case)
* Easily discoverable by looking for values accepted by `sort_by`. This is the first thing people do when they want the change the default random sorting.

Cons:
* ?

### `0000-post.md`
Pros:
* No need to modify existing posts' **content**

Cons:
* Need to rename all the files to use
* Not flexible (e.g., you can't have two digits as prefix)
* Discoverable only after reading documentation thoroughly

---

I really fail to see what downsides of introducing this there are. To me, it seems less niche than `title_bytes` for example :shrug:

Keats:
It's not sorting by path but by filename isn't it? As for the cons it adds another sorting method.

For the pros of  0000-post.md: it is symmetric to the existing date shorthand in files and re-use the `weight` sorting method.

I mean, I would use 0000-post.md for eg Zola docs but probably not `sort_by = "filepath"` because it feels to me like a hack. Obviously this is only my opinion

bemyak:
> It's not sorting by path but by filename isn't it?

Indeed. Initially I named it `path`, but then realized it could be ambiguous, so I switched to `filepath`, but the first name still sticks with me :)

> As for the cons it adds another sorting method.

Why do you consider this as a bad thing? It doesn't slow anything down or bloats the documentation , it just gives people more options.

Since Zola doesn't have plugins, the only (reasonable) way to have this kind of options is to contribute them upstream. This works great with syntax highlighting addition, they seem to be desireble. I don't think sorting, or template filters, or other options are any different.

> it feels to me like a hack

If I remake this PR into sorting by `slug`, will it be better?

Keats:
> Why do you consider this as a bad thing? It doesn't slow anything down or bloats the documentation , it just gives people more options.

I don't see a usecase other than numbering files, which is equivalent to the weight sorting method which sorts the same way. So you get 2 sorting methods doing essentially the same thing, one from the filename and one from the front-matter which is redundant.

> If I remake this PR into sorting by slug, will it be better?

I think that one is ok? I would like to see if someone else wants it but it sounds ok.

bemyak:
Okay, I'll close this, let's continue the discussion in https://github.com/getzola/zola/pull/1926

Issue #1765, opened 2022-02-13
minify_html=true does not minify embedded CSS and JS


JeremyBarbosa:
# Bug Report

## Environment

Zola version: 0.15.3

## Expected Behavior
When setting `minify_html = true` in `config.toml`, it is expected that the entire generated HTML file would be minified, as mentioned in the documentation for the `minify_html`:
"When set to "true", the generated HTML files are minified."

 For example, the following code block:
```html
<!doctype html>
<html>
<head>
  <meta charset="utf-8">
  <style>
    p {
      color: white;
      margin-left: 10000px;
    }
  </style>
  <script>
    alert("Hello World!");
    console.log("Some information: %o", information);
  </script>
</head>
<body>


    <p>Example blog post</p>

  FOO BAR
</body>
</html>
```

Should create this:

```html
<!doctype html><html><head><meta charset=utf-8><style>p{color:#fff;margin-left:10000px}</style><script>alert("Hello World!"),console.log("Some information: %o",information);</script><body><p>Example blog post</p> FOO BAR
```

## Current Behavior
Currently, embedded CSS and JavaScript (in things such as <style> and <script> tags) are not minified. For example, the following code block:

```html
<!doctype html>
<html>
<head>
  <meta charset="utf-8">
  <style>
    p {
      color: white;
      margin-left: 10000px;
    }
  </style>
  <script>
    alert("Hello World!");
    console.log("Some information: %o", information);
  </script>
</head>
<body>


    <p>Example blog post</p>

  FOO BAR
</body>
</html>
```

Creates this:

```html
<!doctype html><html><head><meta charset=utf-8><style>p {
      color: white;
      margin-left: 10000px;
    }</style><script>alert("Hello World!");
    console.log("Some information: %o", information);</script><body><p>Example blog post</p> FOO BAR
```

## Step to reproduce
1. Include embedded CSS or JS in an HTML template
2. Set `minify_html = true` in `config.toml`
3. Inspect the output of `zola build`

Keats:
This won't be fixed for now. The minifier lib we're using requires bundling Go.

JeremyBarbosa:
Great news! The minifier library no longer relies on `js-esbuild` and therefore the Go compiler! It instead uses its own `js-minify` library now.

PR #1922, opened 2022-07-15
Minify embedded css and js with additional lib flags

JeremyBarbosa:

Sanity check:

## Code changes

^ I don't believe any documentation changes are required with the existing documentation mentioning that "the generated HTML files are minified."

## Notes
A small fix for [#1765](https://github.com/getzola/zola/issues/1765). Just two extra config options so that the already included minify-html library also minifies CSS and JS.

### DIFF
--- a/components/site/src/minify.rs
+++ b/components/site/src/minify.rs
@@ -6,6 +6,8 @@ use libs::minify_html::{minify, Cfg};
 pub fn html(html: String) -> Result<String> {
     let mut cfg = Cfg::spec_compliant();
     cfg.keep_html_and_head_opening_tags = true;
+    cfg.minify_css = true;
+    cfg.minify_js = true;
 
     let minified = minify(html.as_bytes(), &cfg);
     match std::str::from_utf8(&minified) {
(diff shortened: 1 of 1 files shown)


Keats:
Did they change the backing compiler in a minor version?

JeremyBarbosa:
It appears that in version 0.9.0, minify-html now uses the [minify-js](https://github.com/wilsonzlin/minify-js) rust library to minify CSS and JS. Before 0.9.0, minifing CSS and JS was handled by [esbuild-rs](https://github.com/wilsonzlin/esbuild-rs), which relied on the Go compiler. 

So minify-html just seems to have switched to a lighter weight dependency that removes the Go requirement.

[Here is the commit](https://github.com/wilsonzlin/minify-html/commit/c7d0652fbcf4a961063baaf21f22841808cfe526) in minify-html where they changed the dependency,

Keats:
Thanks. I think I will wait a bit till those crates are a bit more tested and will merged that for 0.17 rather than 0.16

PR #1949, opened 2022-08-01
Strip binary artifact when building for release

apiraino:
As discussed on Discourse ([thread](https://zola.discourse.group/t/stripping-binaries-in-releases/1396)).

thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ members = ["components/*"]
 [profile.release]
 lto = true
 codegen-units = 1
+strip = true
 
 [profile.dev]
 # Disabling debug info speeds up builds a bunch,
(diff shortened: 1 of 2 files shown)


apiraino:
mmm ... I didn't notice that the CI uses an old version of Cargo (1.57) which does not yet support the `strip` parameter ([stabilized in 1.59](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-strip-option)).

@Keats opinions? Should we hold this change until Cargo is updated or would it make sense to update cargo in the CI (assuming it can be done, I'm not familiar with Azure stuff)

Keats:
We can update the pinned version

apiraino:
"we" as in "us Zola maintaners"? If so, that would be great.

Or is this an actionable for me? :)

Keats:
The royal we ;) 
You can update the pinned version

PR #1950, opened 2022-08-01
The Glob library does not work correctly on windows.

nickdarnell:
The Glob library does not work correctly on windows.  This changes the approach to use the normal WalkDir crate, and the Globset crate instead to find the filter.  This new code correctly compiles sass files across platforms now.

### DIFF
--- a/components/site/src/sass.rs
+++ b/components/site/src/sass.rs
@@ -1,7 +1,8 @@
 use std::fs::create_dir_all;
 use std::path::{Path, PathBuf};
 
-use libs::glob::glob;
+use libs::walkdir::{WalkDir, DirEntry};
+use libs::globset::{Glob};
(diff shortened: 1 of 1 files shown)


Keats:
Reference issue in glob: https://github.com/rust-lang/glob/pull/112

Keats:
Can we add a test for it? On the integration site in `components/libs/site/tests/site.rs` to make sure the sass content is correctly compiled.

Keats (approved):

ErichDonGubler:
In case it's interesting,the [`wax` crate](https://docs.rs/wax) trades some minor behaviors that are difficult to keep cross-platform for portable behavior across all of them (see the `Unsupported Path Features`), in addition to some other features.

Keats:
That could be nice!

PR #1959, opened 2022-08-09
parse also *.yml file suffix as YAML; closes #1958

### LINKED ISSUE #1958: Allow to load_data from yml in addition to yaml?
This is not a bug report but rather a change proposal which I could possibly code/send.

Zola allows to `load_data` from yaml files (in addition to other formats) and this is a really nice recent addition (thanks!):
https://www.getzola.org/documentation/templates/overview/#load-data

But many YAML files end with `*.yml` and not `*.yaml` and Zola does not seem to recognize these as YAML files.

Would it make sense to `load_data` from files/URLs ending with either `*.yaml` or `*.yml` via the YAML parser? If yes, I could try to send a patch.

But maybe this already works and I just misunderstood the documentation and/or sources and/or my local testing of a page.

bast:
previously only files with *.yaml suffix were recognized as YAML

Questions to the reviewer:
- Should I add more to the documentation?
- Should I add anything to the changelog or change commit message format?
- I have only tested this locally with `load_data` from a path ending with *.yml (and this worked). Should I add/extend some test?
- I tried to also test this for URLs but I am unsure how the parsing works there. It does not seem to look at the URL suffix (tested with .toml and .yaml) but perhaps at the header. So I am unsure whether more is needed for URLs or whether they already worked and I did not have a good test. I tried with pages served via https://raw.githubusercontent.com/ but maybe they had wrong header.

Sanity checks:
- [x] Are you doing the PR on the `next` branch?
- [x] Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/components/templates/src/global_fns/load_data.rs
+++ b/components/templates/src/global_fns/load_data.rs
@@ -61,7 +61,7 @@ impl FromStr for OutputFormat {
             "bibtex" => Ok(OutputFormat::Bibtex),
             "xml" => Ok(OutputFormat::Xml),
             "plain" => Ok(OutputFormat::Plain),
-            "yaml" => Ok(OutputFormat::Yaml),
+            "yaml" | "yml" => Ok(OutputFormat::Yaml),
             format => Err(format!("Unknown output format {}", format).into()),
         }
     }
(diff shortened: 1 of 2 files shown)


Keats:
I believe load_data with urls will basically require you to be explicit about the format you're loading rather than relying on headers. I'll double-check tomorrow.

No need to add tests, it's just a simple match.

Keats:
Yep, it looks like it just defaults to Plain for urls

bast:
Thanks for double-checking it. And the setting makes sense.

PR #1962, opened 2022-08-13
add get_hash function

ghost:
Hello,

I saw the [Discourse topic](https://zola.discourse.group/t/hash-of-variable-rather-than-hash-of-file/1168) which got no answers yet, so I decided to give it a try (as I need this feature myself).

It’s basically a rough copy of `get_file_hash` but it takes a string literal in parameter instead of a file.
It was easier to adapt this function (and tests/documentation) than writing a filter instead. I hope this implementation still satisfies your expectations.

Sanity check:

## Code changes


Keats:
Can you rebase the PR?

ghost:
Done!

components/templates/src/global_fns/files.rs
```diff
@@ -228,9 +245,52 @@ impl TeraFn for GetFileHash {
     }
 }
 
+#[derive(Debug)]
+pub struct GetHash {}
```
Keats: you can use a basic fn like https://github.com/Keats/tera/blob/master/src/builtins/functions.rs#L31 no need to impl the Trait explicitly if there's no state

Keats (changes requested):
At this point I think I would merge get_file_hash and get_hash and just have a `get_hash` that takes either a `path` or `literal` arg since the logic/arguments is identical

components/templates/src/global_fns/files.rs
ghost: Nice! I’ve applied the change.

ghost:
I reworked the PR to merge get_hash and get_file_hash into a common function (`process_hashes`) to minimize the amount of duplicated lines of code.

Though I couldn’t completely get rid of `compute_file_hash` because [another function](https://github.com/getzola/zola/blob/next/components/templates/src/global_fns/files.rs#L131) is using it and I have no clue how to get around those `and_then()`… Sorry for my lack of experience.

Keats:
Any chance you can rebase + fix conflict? I'll merge it after, it looks good.

ghost:
Should be fine now :)

docs/content/documentation/templates/overview.md
```diff
@@ -258,6 +258,19 @@ integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Int
...
+
+It can take the following arguments:
+- `literal`: mandatory, the value to be hashed
+- `sha_type`: optional, one of `256`, `384` or `512`, defaults to `384`
+- `base64`: optional, `true` or `false`, defaults to `true`. Whether to encode the hash as base64
+
+```jinja2
+{{/* get_hash(literal="Hello World", sha_type=256) */}}
+```
+
```
Keats, line 264: Let's merge `get_file_hash` and `get_hash` into a single function `get_hash` that takes either a literal or a path

Keats (changes requested):

docs/content/documentation/templates/overview.md
ghost: Okay, I’ll work on it tomorrow (30/08). I guess I’ll have to make the `get_hash` function into a `GetHash` struct with a `TeraFn` impl to make it work.

Are you sure about the deprecation of `get_file_hash` ? That’s quite a big breaking change (even if I agree with the move).

ghost:
As requested, I removed the `get_file_hash` function and added `path` argument to `get_hash`. I’ve updated the documentation and tests to reflect this change.

(I also managed to [get rid of `compute_file_hash`](https://github.com/getzola/zola/pull/1962/files#diff-7ef99150fce5f78d571b96bc3338ee4243645099ceaae5a894c38caf74c70bceR132) somehow)

docs/content/documentation/templates/overview.md
Keats: It's a pretty simple fix though, it can be done with a basic search & replace.

Issue #1897, opened 2022-06-14
Site rendering is much slower with taxonomies


scouten:
# Bug Report

## Environment

Zola version: 0.15.3

## Expected Behavior
Site generation remains speedy.

## Current Behavior
I have a (closed source, sorry) site with about 800 pages, 120 sections, and two taxonomies. There is at least one entry per taxonomy per page.

When I added the taxonomies, the site rendering time dropped from about 1.5 seconds to ~20 seconds on a very beefy M1 MacBook Pro laptop. Running with `zola serve` often fails to reload properly.

## Step to reproduce
If it would be useful, I could take the time to build an obfuscated version of the site. It would take me some hours, so please don't ask lightly.

```sh
$ zola build
Building site...
Checking all internal links with anchors.
> Successfully checked 0 internal link(s) with anchors.
-> Creating 791 pages (0 orphan) and 124 sections
Done in 20.9s.
```

Keats:
Can you try with the `next` branch?

scouten:
No appreciable difference. Same site just rendered in 20.3 seconds using the latest `next` branch.

Keats:
Hm I think I would need the obfuscated version :/ Is it not possible to just replace the content in .md files after the front-matter with lorem ipsum? That should be pretty easy with a small script and you can send the archive to my email directly so it's not public.
I can write a script for that if you want too

Keats:
Something like

```python
import glob
import sys
import shutil

NEW_TEXT = """

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus quis dui tellus. Vestibulum consequat arcu eu porta egestas. Donec urna ligula, convallis sit amet convallis sed, ultrices vel ligula. Integer congue vitae felis ut ultricies. Duis a dolor ornare lectus fermentum laoreet condimentum vitae est. Praesent arcu nibh, pulvinar in congue at, semper sed arcu. Aliquam erat volutpat. Cras at sollicitudin libero, quis vestibulum est. Phasellus sagittis bibendum metus, sit amet laoreet nibh auctor eu. Maecenas mauris urna, auctor sed volutpat tempor, fermentum ut nisi. Morbi ut dolor nisl. Fusce pulvinar sem id rhoncus aliquet. Etiam pellentesque condimentum blandit. 
"""

source = sys.argv[1]
destination = sys.argv[2]

try:
    shutil.rmtree(destination)
except:
    pass

shutil.copytree(source, destination)
[...]

PR #1964, opened 2022-08-13
Get term fn

scouten:
As discussed in #1897, this adds a new `get_term` global function which returns a single `TaxonomyTerm` item instead of the entire site. In the personal site discussed in that issue, this results in a 4x improvement in rendering speed (5.4s vs 22.1s).

## Code changes

### DIFF
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -191,7 +191,7 @@ The type of the output is a single `TaxonomyTerm` item.
 
 `lang` (optional) default to `config.default_language` in config.toml
 
-`include_pages` (optional) default to true. If false, the `pages` item in the `TaxonomyTerm` is replaced with `page_count`, an integer with the number of pages for this term.
+`include_pages` (optional) default to true. If false, the `pages` item in the `TaxonomyTerm` will be empty, regardless of what pages may actually exist for this term. `page_count` will correctly reflect the number of pages for this term in both cases.
 
 `required` (optional) if a taxonomy or term is not found`.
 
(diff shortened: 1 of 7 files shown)


docs/content/documentation/templates/overview.md
```diff
@@ -180,6 +180,23 @@ items: Array<TaxonomyTerm>;
 
 See the [Taxonomies documentation](@/documentation/templates/taxonomies.md) for a full documentation of those types.
 
+### `get_term`
```
Keats: can we call it `get_taxonomy_term`?

Keats (commented):
All good except the name and that it needs a rebase

docs/content/documentation/templates/overview.md
scouten-adobe: Sure. I'll rename in next day or so.

scouten:
@Keats updated as requested.

docs/content/documentation/templates/overview.md
```diff
@@ -180,6 +180,23 @@ items: Array<TaxonomyTerm>;
...
+
+```jinja2
+{% set categories = get_taxonomy_term(kind="categories", term="term_name") %}
+```
+
+The type of the output is a single `TaxonomyTerm` item.
+
+`lang` (optional) default to `config.default_language` in config.toml
+
+`include_pages` (optional) default to true. If false, the `pages` item in the `TaxonomyTerm` is replaced with `page_count`, an integer with the number of pages for this term.
```
Keats: looking at it again, I think `page_count` should be additional field that is always passed and this arguments just make pages an empty array.

Keats (changes requested):

Keats (approved):

Keats:
Thanks!

PR #1968, opened 2022-08-15
Add completion subcommand

tranzystorekk:
There seem to have been a few attempts at completions in the past, but generating from a subcommand seems to be the most versatile for both users and package maintainers. Maybe the old manually written completion scripts could be ditched as well for simplicity?

Sanity check:

## Code changes


Keats:
Honestly I'm kind of wondering if it's even worth adding considering how few commands zola has.

tranzystorekk:
IMO even two commands is a good reason to have them, and `zola serve` in particular has quite a few flags that also get completed.

c-git:
If I may be permitted to add my opinion. It is always wonderful if a program supports completions. It's not required but a pleasant surprise if you find out it's available. If there are no material downsides I would say this is a nice thing to add. Are there any notable disadvantages to having this added? I did note that some things needed to be commented out but I didn't quite understand what they were.

PS. I've really come to appreciate how careful @Keats is with regard to what is added. I can't tell for sure but it makes it seem like you have a vision for the project and you try to keep it on that vision while being open to other people's ideas. You never just shoot things down without a reason. Even like this you gave your reasoning why you are reluctant. Kudos to you.

src/main.rs
```diff
@@ -28,6 +28,13 @@ fn get_config_file_path(dir: &Path, config_path: &Path) -> (PathBuf, PathBuf) {
 
 fn main() {
     let cli = Cli::parse();
+
+    if let Command::Completion { shell } = &cli.command {
```
Keats: can that be moved in the match below?

Keats (commented):
I'll try locally later to see how it works in practice

src/main.rs
tranzystorekk: I wanted to make sure that `zola completion` would never fail, but yeah, probably would be a really weird edge case to make it fail anyway, like `zola --root nonexistent_dir completion bash`

tranzystorekk:
If we merge this, I'd point it out in the changelog that the scripts in `completion/` are obsolete so the package maintainers know what to use, or even remove that whole dir in this PR

Keats:
Let's remove them in that PR, they haven't been updated in years anyway.

src/main.rs
tranzystorekk: Still, to make sure completion just works I'd keep it as it is

Keats: I'd still prefer it in the match, I don't mind a user passing an invalid flag erroring

tranzystorekk: Done, then 😄

Keats (approved):

Keats:
Thanks!

PR #1969, opened 2022-08-15
Wrote test for `is_colocated_asset_links`

JoelMon:
The `is_colocated_assets_links` makes sure the links are in the same directory. 

- I added a check for `../` which would check to see if the file is in a _sub_ directory.
- I wrote tests for the function
- Reworded the docstring slightly to improve grammar/readability


Keats:
Can you rebase just so my commit is not in there?

JoelMon:
> Can you rebase just so my commit is not in there?

I'll give it a shot when I arrive home from work. No problem 👍

JoelMon:
I messed up the rebase. I'm going to send another PR because I have no idea how to recover from what I did... :disappointed:

PR #1978, opened 2022-09-01
Tera date-locale feature

### LINKED ISSUE #1103: `date(format=)`-Filter is not localized
Zola version: zola-0.10.1-1.fc32.x86_64

## Expected Behavior

in config.toml:
```
default_language = 'de'
```

```jinja
{{ "2020-07-27"  | date(format="%B") }}
```
is not localized:  It returns "July" and not "Juli".

I am certainly missing some setting to make it all work.

ZettaScript:
## Code changes

Add tera's feature `date-locale`, to enable setting the locale when using the `date` filter in templates.

Fixes #1103

### DIFF
--- a/components/libs/Cargo.toml
+++ b/components/libs/Cargo.toml
@@ -34,7 +34,7 @@ sha2 = "0.10"
 slug = "0.1"
 svg_metadata = "0.4"
 syntect = "5"
-tera = { version = "1", features = ["preserve_order"] }
+tera = { version = "1.17", features = ["preserve_order", "date-locale"] }
 termcolor = "1.0.4"
 time = "0.3"
 toml = "0.5"


Keats (approved):

Keats:
Thanks!

Keats:
Just to double check, have you tried that this works for you?

ZettaScript:
I've just checked with next HEAD, it works. To check: `{{page.date | date(locale="fr_FR", format="%A %d %B, %H:%M")}}`

However it's not very convenient because you have to use the `fr_FR` format instead of only `fr`. I'll see how to fix that in Tera. (maybe there is a canonical way to convert `xx` to `xx_XX`)

Keats:
It's a bit annoying but I wouldn't bother with changing that

PR #1985, opened 2022-09-11
add preserve_hidden_files configuration option

### LINKED ISSUE #1976: Only delete contents of public folder to avoid breakage when the folder is a git submodule
# Bug Report

## Environment
fedora 36

Zola version:
0.16.1

## Expected Behavior
The zola build should only delete the contents of the `public` directory instead of deleting it completely. 

## Current Behavior
The entire public folder is deleted. When the public directory is a submodule pointing to another repository (e.g. when the site is deployed via github pages), currently it then breaks the git submodule. I believe this could be prevented when only the contents of the directory are deleted.

EDIT: to be more precise, the issue seems to be that it removes the `.git` file from the submodule directory. It might be preferable to ignore it, or ignore all hidden files?

## Step to reproduce
- have the public directory as a git submodule
- run `zola build`

flxzt:
This PR introduces a `preserve_hidden_files` configuration option. When this option is set to true, zola preserves all hidden ( starting with `.`) files and folders in the output directory when building the site.
It also changes that all files inside the output directory, but not the folder itself are deleted.

closes #1976 when merged.

Sanity check:

* [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* [x ] Are you doing the PR on the `next` branch?

* [x ] Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -23,7 +23,7 @@ use std::time::Instant;
 use templates::{load_tera, render_redirect_template};
 use utils::fs::{
     copy_directory, copy_file_if_needed, create_directory, create_file, ensure_directory_exists,
-    file_hidden,
+    is_dotfile,
 };
 use utils::net::get_available_port;
 use utils::templates::{render_template, ShortcodeDefinition};
@@ -584,10 +584,11 @@ impl Site {
         imageproc.do_process()
     }
 
-    /// Deletes the contents in the `public` directory if it exists
+    /// Deletes the `public` directory if it exists and the `preserve_dotfiles_in_output` option is set to false,
+    /// or if set to true: its contents except for the dotfiles at the root level.
     pub fn clean(&self) -> Result<()> {
         if self.output_path.exists() {
-            if !self.config.preserve_hidden_files {
+            if !self.config.preserve_dotfiles_in_output {
                 return Ok(remove_dir_all(&self.output_path)
                     .context("Couldn't delete output directory")?);
             }
@@ -598,8 +599,8 @@ impl Site {
             ))? {
                 let entry = entry.context("Couldn't read entry in output directory")?.path();
 
-                // Skip hidden files and folders if the preserve_hidden_file configuration option is set
-                if file_hidden(&entry) {
+                // Skip dotfiles if the preserve_dotfiles_in_output configuration option is set
+                if is_dotfile(&entry) {
                     continue;
                 }
 
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -589,8 +589,8 @@ impl Site {
     pub fn clean(&self) -> Result<()> {
         if self.output_path.exists() {
             if !self.config.preserve_dotfiles_in_output {
-                return Ok(remove_dir_all(&self.output_path)
-                    .context("Couldn't delete output directory")?);
+                return remove_dir_all(&self.output_path)
+                    .context("Couldn't delete output directory");
             }
 
             for entry in self.output_path.read_dir().context(format!(
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -587,7 +587,11 @@ impl Site {
     /// Deletes the contents in the `public` directory if it exists
     pub fn clean(&self) -> Result<()> {
         if self.output_path.exists() {
-            // Delete contents of current `public` directory so we can start fresh
+            if !self.config.preserve_hidden_files {
+                return Ok(remove_dir_all(&self.output_path)
+                    .context("Couldn't delete output directory")?);
+            }
+
             for entry in self.output_path.read_dir().context(format!(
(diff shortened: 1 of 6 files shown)


components/site/src/lib.rs
```diff
@@ -583,11 +584,29 @@ impl Site {
...
     }
 
-    /// Deletes the `public` directory if it exists
+    /// Deletes the contents in the `public` directory if it exists
     pub fn clean(&self) -> Result<()> {
         if self.output_path.exists() {
-            // Delete current `public` directory so we can start fresh
-            remove_dir_all(&self.output_path).context("Couldn't delete output directory")?;
+            // Delete contents of current `public` directory so we can start fresh
+            for entry in self.output_path.read_dir().context(format!(
```
Keats, line 596: No need to iterate on it if `self.config.preserve_hidden_files` is `false`. You can return the original line from the diff on the left in that case.

components/site/src/lib.rs
```diff
@@ -583,11 +584,29 @@ impl Site {
         imageproc.do_process()
     }
 
-    /// Deletes the `public` directory if it exists
+    /// Deletes the contents in the `public` directory if it exists
```
Keats: The behaviour should be the same if the option is not set, the folder should just be removed entirely

components/site/src/lib.rs
```diff
@@ -583,11 +584,29 @@ impl Site {
...
+                }
+
+                if entry.is_dir() {
+                    remove_dir_all(entry)
+                        .context("Couldn't delete folder while cleaning the output directory")?;
+                } else {
+                    remove_file(entry)
+                        .context("Couldn't delete file while cleaning the output directory")?;
+                }
+            }
```
Keats, line 614: That doesn't work recursively so it only preserves hidden files at the root of the dir?

Keats (changes requested):

components/site/src/lib.rs
flxzt: Should it work recursively? It would result that regular directories are kept when they contain any hidden file inside them so it could potentially result in odd looking behaviour from a user perspective.
My personal intuition would be that this configuration option only affects files and folders at the root level. I can make that more clear in the docs

Keats: I'm wondering if root should appear in the config name to be more explicit

flxzt: I was thinking about it too but I could not come up with a concise name

Keats: Let's think about it for a few days and see if anything comes up

tosti007:
I personally would like the feature where the public folder is kept and it's content are removed, but don't need the preservation of hidden files. Shouldn't be keeping the public folder be a seperate flag where the preserce_hidden flag implies keep_public? Another option would be make preserving the public folder default behavior, whcih ij turn would make the code here a bit simpler.

flxzt:
I would prefer that too and implemented it this way initially, but @Keats reviewed that the original code path should be preserved when the option is not set. My guess is to keep backwards compatibility in cases the folder is watched by other tools.

tosti007:
That makes sense! So would it be a good option to add a separate flag to keep the public folder?

Keats:
What's the usecase for wanting the public folder preserved but empty?

tosti007:
I am running zola as a docker image to rebuild the site on file-change. However, things tend to break when a folder mounted to a docker image is removed. Hence, by removing the files rather than the entire folder, the public folder itself can be mounted. Currently I am running a `rm` and `mv` command after build to achieve the same.

components/site/src/lib.rs
flxzt: How about `preserve_dotfiles_in_public` ?

flxzt:
what are your thoughts about this @Keats ?

Keats:
I think the current PR is fine, I'm just wondering about the naming. Files starting with `.` are not hidden on Windows so I think it might be confusing to people not used to Unix/Linux

flxzt:
How about renaming the option to "preserve_dotfiles" instead of "preserve_hidden_files"? Thats a term that is associated with configuration files, and does not imply anything about visibility. Is also a little shorter

Keats:
`preserve_dotfiles_in_output`? Just to give some context on the location

flxzt:
Okay, sounds good to me! I'll change it

Keats:
Looks like you need to do a manual rebase from the next branch

flxzt:
ok, done

Keats:
So i believe just turning on this option will mean that the public folder will not be deleted, solving a few people issues as well.

flxzt:
Nice :) I should probably add this behaviour to to the documentation to make it more discoverable. Something like `Enabling this also prevents the deletion of the output folder itself on rebuilds.`

Edit: done.

Keats:
Thanks!

Issue #1849, opened 2022-05-04
Alternative Search Engines, elasticlunr last updated 2019


Jieiku:
I just finished implementing the elasticlunr search for the [abridge theme](https://abridge.netlify.app/).

So far I like elasticlunr, but it was last updated 2019.

It would be really cool if Zola supported multiple search engines.

I am going to attempt implementing [stork](https://github.com/jameslittle230/stork) and [tinysearch](https://github.com/mre/tinysearch)

Stork seems to provide VERY relevant results and shows suggestions based on partial words which is nice!

Keats:
stork and tinysearch look very nice! Thanks for finding those. Looking at those 2 though, they only support western languages and not CJK? I don't see specific stemmers/stopword list for various languages so they might only work well with English?

It's not a problem if the elasticlunr hasn't been updated since 2019 if it works well.

Jieiku:
@Keats If I can add stork and tinysearch index generation to zola, is that something you would accept pull requests for so long as it works without causing any issues for the existing code base?

I understand these two don't have as many features as elasticlunr for CJK, however if zola supported the index generation for these, then other developers more familiar with those languages (CJK) would probably submit improvements to those projects (stork and tinysearch)

Keats:
Both stork and tinysearch only support English so I would say their scope is too limited for Zola. IMO if they don't support CJK (fair enough) they should at least support most languages using spaces like Italian/German/Dutch/French/Spanish/Portuguese and scandinavian/slavic languages. It's pretty much just adding stopwords for each which are easy to find, no reasons not to support them.

As for adding alternatives to Zola, before doing so it would need a few things:

- benchmarks for index sizes & client side libs
- sample where we can test all of them to compare accuracy/response time for the same dataset

Alternatives will be added if they are giving compelling results in the above, elasticlunr.js might not be maintained anymore but it still works perfectly fine and is not going to suddenly decay.

Jieiku:
Ok, very cool! I am not sure when I will have time to work on this, but when I do I will be sure to make some benchmarks and put some samples and test sites available online.

Jieiku:
I have been working on this, I still need to find some example content to use for the sample/demo sites.

Keats:
The ability to add multiple formats has been added in the `next` branch.

andreacfromtheapp:
hi @Jieiku and @Keats :)

[...]

Issue #1745, opened 2022-01-26
 [Feature Request] Support Algolia


azzamsa:
Hi.

I saw several project weight other SSG more just because they have [Algolia](https://www.algolia.com/doc/guides/getting-started/how-algolia-works/) support.

Would you like to support Algolia for this awesome SSG?

Related:
- https://github.com/getzola/zola/issues/1349
- https://github.com/codewars/docs/issues/248

Keats:
Yep, I would take a PR that output the search index in the Algolia format instead of elasticlunr

Keats:
FYI we've added support for formats for the search engine output so it's just a matter of someone adding it.

PR #1998, opened 2022-09-29
Search json index

sgued:
Based on [this discussion](https://zola.discourse.group/t/make-zola-ability-to-export-search-index-into-json-instead-of-js/1032/3)

Sanity check:

## Code changes

Description
=========

This PR adds an `index_format` configuration item.
```
[search]
index_format = "javascript" | "json"
```

When set to Javascript, a JSON file is created instead of a `js` file for the index.

This is:

- More convenient. Lazy-loading Json is easier than a script
- More flexible. The JSON file can be loaded with the `fetch` API and be manipulated more easily than a script that sets a global variable for which you don't control the name. While it's not a use case I encountered, I don't think the current setup makes it easy to load multiple search index at the same time, for example for multiple languages.

The docs are updated to 
- Document the feature
- Use the feature and lazy-load the index, saving more than 1MB (200KB with compression) from loading when just loading the site.

### DIFF
--- a/components/config/src/config/search.rs
+++ b/components/config/src/config/search.rs
@@ -1,6 +1,20 @@
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum IndexFormat {
+    #[serde(rename = "json")]
+    Json,
+    #[serde(rename = "javascript")]
+    Javascript,
+}
+
+impl Default for IndexFormat {
+    fn default() -> IndexFormat {
+        IndexFormat::Javascript
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(default)]
 pub struct Search {
     /// Include the title of the page in the search index. `true` by default.
@@ -15,6 +29,8 @@ pub struct Search {
     pub include_description: bool,
     /// Include the path of the page in the search index. `false` by default.
     pub include_path: bool,
+    /// Foramt of the search index to be produced. Javascript by default
+    pub index_format: IndexFormat,
 }
 
 impl Default for Search {
@@ -25,6 +41,7 @@ impl Default for Search {
             include_description: false,
             include_path: false,
             truncate_content_length: None,
+            index_format: Default::default(),
         }
     }
 }
--- a/components/config/src/config/search.rs
+++ b/components/config/src/config/search.rs
@@ -1,16 +1,15 @@
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
 pub enum IndexFormat {
-    #[serde(rename = "json")]
-    Json,
-    #[serde(rename = "javascript")]
-    Javascript,
+    ElasticlunrJson,
+    ElasticlunrJavascript,
 }
 
 impl Default for IndexFormat {
     fn default() -> IndexFormat {
-        IndexFormat::Javascript
+        IndexFormat::ElasticlunrJavascript
     }
 }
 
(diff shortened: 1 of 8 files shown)


Keats:
This is an interesting feature.
There has been some request to generate in other formats (https://github.com/getzola/zola/issues/1745, https://github.com/getzola/zola/issues/1849).

components/config/src/config/search.rs
```diff
@@ -1,5 +1,19 @@
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum IndexFormat {
+    #[serde(rename = "json")]
+    Json,
+    #[serde(rename = "javascript")]
+    Javascript,
```
Keats: We should likely prefix them with elasticlunr as it would the format for those and we want to avoid breaking changes later if we add more format.

components/config/src/config/search.rs
```diff
@@ -1,5 +1,19 @@
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum IndexFormat {
+    #[serde(rename = "json")]
+    Json,
+    #[serde(rename = "javascript")]
```
Keats: you can tell serde to change casing for everything in one attribute: https://serde.rs/container-attrs.html#rename_all

Keats (changes requested):

components/config/src/config/search.rs
sgued: Done in c716bfcdba9ff69e097b7372b59fe192435a7ae7

components/config/src/config/search.rs
sgued: Done in c716bfcdba9ff69e097b7372b59fe192435a7ae7

Jieiku:
I really like this idea. It also makes me wonder if this could be leveraged to generate a css file?

https://github.com/Jieiku/abridge/issues/72

At the very least I can take a look at what you have done here to get some ideas.

sgued:
> I really like this idea. It also makes me wonder if this could be leveraged to generate a css file?
> 
> [Jieiku/abridge#72](https://github.com/Jieiku/abridge/issues/72)
> 
> At the very least I can take a look at what you have done here to get some ideas.

A css file for what?

Edit: I think you mean csp

Jieiku:
My idea was to create a way for shortcodes to be able to write a portion of their output into a css file.

The issue I linked discusses it in more detail, this is unrelated to search functionality, the only connection is being able to write to a file other than html.

https://github.com/Jieiku/abridge/issues/72

The problem is that you don't want style tags in the middle of an html page if you have a strict content security policy, because it would disallow this, so instead you would make the shortcode write the style related stuff directly to a css file.

Writing style tags to a page via a shortcode is probably not a common use for a shortcode, but it is how I facilitate changing the image on mouse hover (useful thing for image comparison, especially in photography and video compression related blogs).

Keats:
> My idea was to create a way for shortcodes to be able to write a portion of their output into a css file.

That's not going to happen I think but it would be unrelated to that

Keats (approved):

Jieiku:
I am in the process of [refactoring the Abridge theme](https://github.com/Jieiku/abridge/issues/105), I got this json index search feature working a few moments ago, but noticed one thing... there is no cachebust hash on the json index

so if you add new posts... repeat visitors may not have those posts in their index if their browser still has the old index cached, correct?

Other than that works great!

One very nice thing about fetching json instead of loading the javascript later is that it will not violate a strict csp since its simply json, before I was using this:

https://github.com/Jieiku/abridge/blob/master/static/search_facade.js

(with the above script, I update the hashes using an npm script)

Keats:
Can you open a new issue for the missing cachebust?

PR #2005, opened 2022-10-14
Quotation mark bug fix in docs

YesSeri:
Changed double quotation mark " to single quotation ', otherwise the code doesn't work.

Sanity check:

### DIFF
--- a/docs/content/documentation/getting-started/overview.md
+++ b/docs/content/documentation/getting-started/overview.md
@@ -220,7 +220,7 @@ The `index.html` file inside the `templates` directory should be:
 <h1 class="title">
   This is my blog made with Zola.
 </h1>
-<p>Click <a href="{{/* get_url(path="@/blog/_index.md") */}}">here</a> to see my posts.</p>
+<p>Click <a href="{{/* get_url(path='@/blog/_index.md') */}}">here</a> to see my posts.</p>
 {% endblock content %}
 ```  
 


Keats:
Why would double quotes not work? I'm only using them everywhere

YesSeri:
I should have been more clear in my communication.

```
"{{/* get_url(path="@/blog/_index.md") */}}"
```

You have nested quotation marks here. That means that the  code above gets interpreted as 

```html
// Text
"{{/* get_url(path="
//Html
@/blog/_index.md
// Text
") */}}"
```

Does my example make sense? 

I think it is only a problem here, but if you accept my pull request I can look through all of the documentation for similar problems.

What do you think?

Keats:
Why would you have newlines in there?

YesSeri:
There should not be newlines. I was just trying to emphasize my point. Do you understand why the code doesn't work?

Keats:
I don't sorry, I must be dense 😿

jtojnar:
Perhaps @YesSeri means syntax highlighting is broken in some editors?

jtojnar:
Actually, it even breaks syntax highlighting on the website:

Bottom of the Overview page

Keats (approved):

YesSeri:
> Actually, it even breaks syntax highlighting on the website:
> 
> Bottom of the Overview page

Well, the syntax highlightning is broken because the code is broken. If you want to nest double quotation marks you need a way to escape them or you can use single quotation marks.

PR #2017, opened 2022-11-05
Filter more than 1 footnote in summary

mscherer:
If the summary contains more than 1 footnote, only the 1st is removed, and so dangling links exists for all the others.

Sanity check:

## Code changes

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -229,7 +229,7 @@ impl Page {
         self.summary = res
             .summary_len
             .map(|l| &res.body[0..l])
-            .map(|s| FOOTNOTES_RE.replace(s, "").into_owned());
+            .map(|s| FOOTNOTES_RE.replace_all(s, "").into_owned());
         self.content = res.body;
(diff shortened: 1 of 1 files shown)


Keats (approved):
Thanks!

components/content/src/page.rs
```diff
@@ -515,13 +515,17 @@ Hello world
...
+
 <!-- more -->
 
-And here's another. [^2]
+And here's another. [^3]
 
 [^1]: This is the first footnote.
 
-[^2]: This is the second footnote."#
+[^2]: This is the secund footnote.
```
BartG95, line 526: Is the spelling error in "second" intentional?

c-git: secund is a word... but it doesn't fit in that sentence. Maybe auto correct?

mscherer: Nah, likely just a typo in my patch, sorry. i was more focused on the Rust code than the text.

PR #2039, opened 2022-11-26
Merge default language data with section language data

### LINKED ISSUE #1918: Merge default data language data in config.toml
Right now we can put the data in 2 different places for the default language in config.toml: the base file and `language.DEFAULT_LANGUAGE`. Those are not currently merged and zola will just take `language.DEFAULT_LANGUAGE` data if present, resulting in missing data and hard to diagnose error.  See https://github.com/justint/papaya/blob/44be3a1b39a8bad5b33e55e8f6de8953d0ca2e20/config.toml for an example

vcrn:
Resolves #1918

When there is a merge conflict, I went with causing a panic right away and not returning anything if everything went well,  instead of `bail` at conflict and returning an `Ok()` if everything went well. If you want this changed to the latter approach, just let me know.

I don't seem to be able to directly manipulate `languagesi[default_language]`, which is why I'm cloning it, manipulating it via `.merge()` and then using `.insert()`.

The test `add_default_language_skips_insert()` doesn't check that insert actually hasn't been performed, it is only when running it manually that a person can check and see that a warning message hasn't been printed. I used it for my testing and left it, but we can remove it if you want to.

Also, under method `add_default_language()` there was some TODO comments. This PR might solve it but I'm not sure, so I've left it.


components/config/src/config/languages.rs
```diff
@@ -30,6 +30,113 @@ pub struct LanguageOptions {
...
 }
 
+impl LanguageOptions {
+    /// Merges self with another LanguageOptions, panicking if 2 equivalent fields are not None,
+    /// empty or of default value.
+    pub fn merge(&mut self, other: &LanguageOptions) {
+        match &self.title {
+            None => self.title = other.title.clone(),
+            Some(self_title) => match &other.title {
+                Some(other_title) => panic!(
```
Keats: We can't use panic like that, it should an error.

components/config/src/config/languages.rs
```diff
@@ -30,6 +30,113 @@ pub struct LanguageOptions {
...
+        match &self.translations.is_empty() {
+            true => self.translations = other.translations.clone(),
+            false => match &other.translations.is_empty() {
+                true => (),
+                false => panic!("Error: `translations` for default language is specified twice, as {:?} and {:?}.", self.translations, other.translations),
+            },
+        };
+    }
+
+    /// Checks whether self is of default value.
```
Keats: You can just instantiate a default struct and compare them

components/config/src/config/mod.rs
```diff
@@ -205,25 +205,36 @@ impl Config {
...
+            description: self.description.clone(),
+            generate_feed: self.generate_feed,
+            feed_filename: self.feed_filename.clone(),
+            build_search_index: self.build_search_index,
+            taxonomies: self.taxonomies.clone(),
+            search: self.search.clone(),
+            translations: self.translations.clone(),
+        };
+
+        if !base_default_language_options.is_default() {
```
Keats: This logic is wonky i think? you're checking whether some values are set on the struct you just made up.

You should check grab the existing struct in `self.languages` if it exists and handle all potential conflicts in this function to keep it simple.
I don't know if i'm clear, let me know if it isn't and I can do a quick version of what I think

Keats (changes requested):
I think that's not the right way to do it. We should handle that before adding the default language struct in this line: https://github.com/getzola/zola/blob/next/components/config/src/config/mod.rs#L128

components/config/src/config/languages.rs
vcrn: OK!

components/config/src/config/languages.rs
vcrn: Don't know how that slipped my mind, or to compare the complete structs with each other instead of field by field. It seems like with your other suggestions, we would be able to get rid of this kind of check altogether, which would be nice.

components/config/src/config/mod.rs
vcrn: I think I get what you mean. The issue with my logic is that it's a bit hard to follow, even if it would turn out to be sound. I'm checking whether the struct I just made up from the base language data is of default value in order to know if the user has specified any data at base, since this data always will be present.

We could get rid of this step, and just merge if data is specified at section. I'd have to change the warning message, because right now it warns the user if they've specified data for the default language at base _and_ under a section. The new warning would pop up every time the user has specified data under a section that is of the default language, since base data is always available as default. If specifying default language data under a section isn't frowned upon though, I can simply get rid of the warning and only raise an error if there's a merge conflict.

vcrn:
Thanks for the feedback! It seems clear for me how to proceed.

Edit:

> I think that's not the right way to do it. We should handle that before adding the default language struct in this line: https://github.com/getzola/zola/blob/next/components/config/src/config/mod.rs#L128

This one I'm a bit uncertain about though. Which part are you referring to that should be handled before adding the default language struct, @Keats?

components/config/src/config/mod.rs
vcrn: You're right, the logic is incorrect. I'll fix it, and I think I also have a way of getting the behavior you desire in a quite simple manner.

I am checking whether the struct I created from the default language data are all of the default value in order to print the warning, but like you pointed out, the logic is wrong. I'll give it another go and see how it turns out.

vcrn:
Fixing my broken logic, I went with a _slightly_ more complicated route than I first imagined, in order to not do a merge unless necessary. If you think we should go with a simpler route, I propose something like this:
```
        if let Some(section_default_language_options) = self.languages.get(&self.default_language) {
            if base_default_language_options != languages::LanguageOptions::default() {
                println!("Warning: config.toml contains both default language specific information at base and under section `[languages.{}]`, \
                    which may cause merge conflicts. Please use only one to specify language specific information", self.default_language);
            }
            base_default_language_options.merge(section_default_language_options)?;
        }
        self.languages.insert(self.default_language.clone(), base_default_language_options);

        Ok(()
```

vcrn:
Just wanted to check whether you believe you'll have time to check out the updated PR any time soon, @Keats ?

components/config/src/config/mod.rs
```diff
@@ -388,6 +399,75 @@ mod tests {
     use super::*;
     use utils::slugs::SlugifyStrategy;
 
+    #[test]
+    #[should_panic]
```
Keats: can you test for the error instead of testing the panic?

Keats (changes requested):
It's ok I think. I'll change some style stuff later after merging but I think we should at least correcly test the errors before merging.

components/config/src/config/mod.rs
vcrn: Sure!

vcrn:
@Keats, I've updated the test in question. I also took a second look at your test names and gave mine ones that are more in line with yours. I also changed some of the variable names in the new tests so they are more descriptive.

Keats:
Thanks!

PR #2073, opened 2023-01-17
Shorter cachebust string

orlp:
Using a full 256-bit hash for cachebusting is complete overkill (and a waste of pagesize bytes). It isn't for security, simply to invalidate the browser cache if the file changes.

I've changed it to 80 bits. Since you'll see on average a collision from a collection of `n` elements after `sqrt(n)` random samples, this means you'd have to do 2^40 edits on the same file, with the browser caching all of them, before you would load a stale version. That is, even if you edited your CSS files every second and cached all of them for eternity it would still take you ~34865 years before seeing a collision. I think that's plenty.


orlp:
I screwed up in the git voodoo, let me fix that.

PR #2074, opened 2023-01-17
Reduced cachebust fingerprint to be more reasonable

orlp:
Using a full 256-bit hash for cachebusting is complete overkill (and a waste of pagesize bytes). It isn't for security, simply to invalidate the browser cache if the file changes.

I've changed it to 80 bits. Since you'll see on average a collision from a collection of n elements after sqrt(n) random samples, this means you'd have to do 2^40 edits on the same file, with the browser caching all of them, before you would load a stale version. That is, even if you edited your CSS file every second and cached all of them for eternity it would still take you ~34865 years before seeing a collision. I think that's plenty.

So instead of this:

```html
<link rel="stylesheet" href="/assets/style.css?h=23d9476e2b2eb5b846066019f1ffecec1be0e80ce171161175b5441699b24ccf">
```

You would see:

```html
<link rel="stylesheet" href="/assets/style.css?h=23d9476e2b2eb5b84606">
```

### DIFF
--- a/components/site/tests/site.rs
+++ b/components/site/tests/site.rs
@@ -777,7 +777,7 @@ fn can_ignore_markdown_content() {
 fn can_cachebust_static_files() {
     let (_, _tmp_dir, public) = build_site("test_site");
     assert!(file_contains!(public, "index.html",
-        "<link href=\"https://replace-this-with-your-url.com/site.css?h=83bd983e8899946ee33d0fde18e82b04d7bca1881d10846c769b486640da3de9\" rel=\"stylesheet\">"));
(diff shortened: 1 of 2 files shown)


orlp:
Locally the tests pass, so not sure what the CI is up to.

Keats (approved):

Keats:
There is a hardcoded fingerprint in the `can_cachebust_static_files` test

orlp:
...I made a line less long and now the `cargo fmt` CI starts complaining? I can't be bothered with this.

c-git:
Yeah because now it should be part of the line above. It is not long enough to warrant a separate line

Keats (approved):
Don't worry about the CI failure, I'll fmt it later

Issue #1844, opened 2022-05-02
Allow redirections to external URLs


Lucidiot:
# Bug Report

## Environment

Zola version: 0.15.3

## Expected Behavior
Tell us what should have happened.

When setting `redirect_to` on a section's `_index.md` to an external URL, the redirection page should redirect to the external URL.

## Current Behavior
Tell us what happens instead of the expected behavior. If you are seeing an
error, please include the full error message and stack trace. You can get the
stacktrace of a panic by adding `RUST_BACKTRACE=1` when running a `zola` command.

The site's base URL gets concatenated to the external URL: with a base URL at `https://foo.com` and `redirect_to = "https://bar.com/something"`, the user gets redirected to `https://foo.com/https://bar.com/something`.

## Step to reproduce
Please provide the steps to reproduce the issue.
If the issue is hard to reproduce, please provide a sample repository or sample
that triggers the bug.

1. Set the site's base URL to `https://foo.com`
1. Create an `_index.md` anywhere in a site's content, and set a redirection:

   ```
   +++
   redirect_to = "https://bar.com/something"
   +++
   ```

2. Generate the site.
3. Open the section that you created.

You will get redirected to `https://foo.com/https://bar.com/something`.

Keats:
Wouldn't you want to use 301 or 302 in that case?

Lucidiot:
Well if we could get the server to redirect a few URLs, we would, but in many cases such as using GitHub Pages, you just do not have control over the server's responses. Zola already has `aliases` and `redirect_to` to work around this and use a small webpage to redirect, and redirecting to an external URL instead of an internal one does not seem to make much of a difference.

I made a workaround by copying the [built-in `internal/alias.html` template](https://github.com/getzola/zola/blob/96db5231e74599815df381fbf90d499c2d5b3814/components/templates/src/builtins/internal/alias.html) over to the site's templates and using a custom variable in `[extra]` instead of `redirect_to` to avoid getting the base URL appended. So I guess just not adding the site's base URL when the destination URL is already absolute would fix this?

kartva:
https://github.com/getzola/zola/blob/ede306cdd766bc53c590d44791b535e3e3309780/components/config/src/config/mod.rs#L173-L194

This seems like the code responsible for adding the base URL. Adding a check for `:` for an absolute path should be easy enough.

Keats:
It should be changed upstream of that function, not in there. This function is responsible for making permalinks in the current site.

[...]

PR #2080, opened 2023-01-20
feat: External URL redirect

kartva:
fixes: #1844

## Code changes
Moved `is_external_link` to `utils::net`. Not the best place, I'm open to suggestions for better named modules or putting `is_external_link` in `utils`.


Keats:
Can you do the PR on the `next` branch instead please?

kartva:
Done!

components/site/src/lib.rs
```diff
@@ -1114,7 +1115,11 @@ impl Site {
         }
 
         if let Some(ref redirect_to) = section.meta.redirect_to {
-            let permalink = self.config.make_permalink(redirect_to);
+            let permalink: Cow<String> = if is_external_link(redirect_to) {
+                Cow::Borrowed(redirect_to)
```
Keats, line 1147: This should be added to the list of external links to check for

Keats (changes requested):

components/site/src/lib.rs
kartva: I'm sorry, I don't understand what you're saying. By this, you mean `permalink`? Is there a list of external links somewhere? What would checking against it mean? I may be missing something critical here, since I'm not very familiar with Zola's codebase at the moment.

Keats: So Zola checks external links in your markdown automatically to see if they are still valid. Since we will essentially redirect to this page, we should that URL to the list of external URLs to check so if you make a typo or the site is not up anymore you can correct/remove.

Have a look at `components/link_checker` to see where the checks happen

kartva:
Ok! It took me some time to find the right place to insert the checks, but I think I've got it. Is it better now?

Keats:
Yep it looks ok. Just need a rebase and a test that the alias is added to the external links.

kartva:
Done, done.

Keats (approved):

kartva:
Looks like everything is failing because of `clippy` linting stuff that isn't related to my changes.

Keats:
Don't worry about that

Keats (approved):

PR #2086, opened 2023-01-25
Procimage refactor

Keats:


Keats:
Current bug: running zola build multiple times will erase the processed_images folder? That's me being silly with the pruning

PR #2087, opened 2023-01-26
Fix erroring suggestion in docs

InputUsername:
The docs ([link](https://www.getzola.org/documentation/templates/archive/)) suggest `reverse` after `group_by`, but that causes an error:
```
Error: Reason: Filter `reverse` received an incorrect type for arg `value`: got `...` but expected Array|String
```

https://tera.netlify.app/docs/#group-by has a solution, so I used that.

Sanity check:


Keats (approved):
Thanks!

Issue #2024, opened 2022-11-15
Zola Builds Broken RSS/atom feed


whoisYoges:
### Environment
Zola 0.16.1 running in Arch-Based Linux Distribution

### Configuration
generate_feed = true  
feed_filename = "atom.xml"

### Steps
Enable the above configurations in `config.toml`  
Run `zola build` or `zola serve`

## Current Behavior
When I try to add the output feed `atom.xml` in my rss reader (I am using [Thunderbird](https://www.thunderbird.net/en-US/)), It gives an error saying It's not a valid Feed.  
Validating the feed with [w3c feed validator](https://validator.w3.org/feed/), it shows following errors:

```
Missing entry element: author
</entry>
```

```
title should not be blank
<title></title>
```

### Expected Behavior
The feed should have been validated without any error.

### Detected Problem
It generates the feed (`atom.xml`) with Empty title.

```
...
...
<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
<title></title>
...
...
```

### Remark
Using `feed_filename = "rss.xml"` in place generates `rss.xml` with empty title and description.

```
...
...
<channel>
<title></title>
<link>http://127.0.0.1:1111</link>
<description></description>
<generator>Zola</generator>
<language>en</language>
...
...
```

### Temporary/Manual Solution
Adding the title manually in `atom.xml` solved the problem for me or adding title and description manually if you're using `rss.xml`.

```
...
...
<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
<title>My Blogs</title>
...
...
```

### Permanent Solution
Zola should detect the default title and description in following priority order or similar appropriate one and use it:  

1. /content/blog/_index.md
2. /templates/blog.html
3. config.toml

### Reference Project
<https://github.com/whoisYoges/website>

Keats:
The title can potentially be fixed (eg use url of the site if we have nothing) but what do we do for the author?

whoisYoges:
> The title can potentially be fixed (eg use url of the site if we have nothing) but what do we do for the author?

@Keats Thanks for the quick reply!

After a little research I found out that you'll need to place the following inside each `<entry>` tag.

```
<author>
   <name>Blog Author Name</name>
</author>
```

### Currently produced output:

```
...
...
<entry xml:lang="en">
<title>Create and delete GPG keypair | Introduction to GnuPG </title>
<published>2022-10-30T00:00:00+00:00</published>
[...]

PR #2092, opened 2023-02-04
Page and config authors (#2024)

sethm:
## Code changes

This PR is meant to address #2024 by introducing `authors` to Page frontmatter and a default `author` to Config. The built-in RSS and Atom templates have been updated to use either the first available author from the Page's frontmatter, or the default `author` from Config, or the fixed string "Unknown" if neither is available.


obi1kenobi:
Thanks for putting this together! I'm excited to use it as soon as it's merged. It seems like the tests have failed but only due to clippy, and I believe `cargo clippy --fix` will take care of it.

sethm:
This is a tough one... the clippy warnings that cause the build to fail are from code I didn't touch. It looks like Rust 1.64.0 is happy, but Rust 1.67.0 (current stable) is Running `cargo clippy --fix --workspace`  modifies 32 files but still doesn't clear up all the warnings, so I'm pretty wary of doing a clippy fix in this PR. Does anyone else have any opinions on it?

obi1kenobi:
They are newly-added lints encouraging use of the newer inline formatting ability: `"{foo}"` instead of `"{}", foo`. Many projects have been dealing with this kind of issue, and the lint is essentially cosmetic — not a correctness or API misuse lint.

It's safe to ignore, and it's also an easy fix. I'd be happy to open a PR to fix all the new lints, if you'd like.

components/templates/src/builtins/atom.xml
```diff
@@ -1,32 +1,43 @@
...
+    <updated>{{ last_updated | date(format="%+") }}</updated>
+    <id>{{ feed_url | safe }}</id>
+    {%- for page in pages %}
+    <entry xml:lang="{{ page.lang }}">
+        <title>{{ page.title }}</title>
+        <published>{{ page.date | date(format="%+") }}</published>
+        <updated>{{ page.updated | default(value=page.date) | date(format="%+") }}</updated>
+        <author>
+          <name>
+            {%- if page.authors[0] -%}
```
Keats: You don't need `[0]` in the expression I think

components/templates/src/builtins/rss.xml
```diff
@@ -6,25 +6,35 @@
...
+      <generator>Zola</generator>
+      <language>{{ lang }}</language>
+      <atom:link href="{{ feed_url | safe }}" rel="self" type="application/rss+xml"/>
+      <lastBuildDate>{{ last_updated | date(format="%a, %d %b %Y %H:%M:%S %z") }}</lastBuildDate>
+      {%- for page in pages %}
+      <item>
+          <title>{{ page.title }}</title>
+          <pubDate>{{ page.date | date(format="%a, %d %b %Y %H:%M:%S %z") }}</pubDate>
+          <author>
+            {%- if page.authors[0] -%}
```
Keats: same here

Keats (commented):
Don't worry about the clippy issue, it's just some annoying change

components/templates/src/builtins/rss.xml
sethm: I think you're right -- fixed in new push.

Keats:
Seems good to me. Do we need authors for the sections as well?

obi1kenobi:
The clippy lints that were raised have been downgraded to allowed-by-default in the new Rust 1.67.1 that was just released. They were annoying many projects like this, and the lang maintainers listened :)

Keats:
I've just seen that, very nice - it annoyed me in multiple repos already.

Keats (approved):

obi1kenobi:
Awesome, thank you! Do you happen to have a sense of when this might get released? A few people have reached out to tell me that my RSS is failing validation right now, and this PR is part of the fix so I'd love to deploy it on my site as soon as it's released.

Keats:
Not sure yet, I think not too long. I would have liked to fix some Windows bugs before releasing but I might not wait for contributions (no Windows machine here).

sethm:
> Awesome, thank you! Do you happen to have a sense of when this might get released? A few people have reached out to tell me that my RSS is failing validation right now, and this PR is part of the fix so I'd love to deploy it on my site as soon as it's released.

@obi1kenobi In the meantime, you can do what I've done. I copied the atom.xml (or, in your case, rss.xml) file from the Zola source into my site's `templates` directory, and then edited it to include a hard-coded "author". That will cause Zola to use the local copy instead of the built-in template. It's not a complete fix because you can't set authors on a per-page basis, but it will at least fix your RSS for now!

PR #2096, opened 2023-02-10
Extended documentation about relative image resizing and scaling of high…

jonasdiemer:
Hi,

it took me a bit to figure out how to "properly" (?) render high-resolution screenshots. I thought it might fit well into the documentation.

Cheers

### DIFF
--- a/docs/content/documentation/content/image-processing/index.md
+++ b/docs/content/documentation/content/image-processing/index.md
@@ -174,7 +174,37 @@ Here is the result:
 </small>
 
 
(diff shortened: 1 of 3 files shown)


Keats:
Seems like we could add a `scale` parameter to resize_image to handle that natively?

jonasdiemer:
> Seems like we could add a `scale` parameter to resize_image to handle that natively?

Yes, that would be nice (I actually considered a feature request before I realized I could solve it with a shortcode).

Issue #2053, opened 2022-12-14
word_count includes alt text and other things


classabbyamp:
# Bug Report

## Environment

Zola version: 0.16.1

## Expected Behavior

I'm not sure if this is the intended behaviour, but I expected alt text and other things to not count.

## Current Behavior

words in image alt text and other things count towards the word count

## Step to reproduce

for this page: 

```md
+++
title = "an example post"
date = "2021-10-10"
+++

one two three

<!-- more -->

![four five six](idk.png)
```
`page.word_count` will be 11

Keats:
Yes, it operates on the raw content. It's very rough and is mostly useful for an estimated reading time, not anything perfect.

PR #2107, opened 2023-02-18
Update word count implementation to handle raw HTML content

mufeedvh:

Sanity check:

## Code changes

---

Noticed the word count implementation treated HTML/CSS/JS inside the Markdown as words after unicode separated word counting. This PR introduces a way to extract words from raw HTML content by removing the `<script>` and `<style>` chunks along with HTML tags.

This introduces one new dependency [`fancy-regex`](https://docs.rs/fancy-regex/latest/fancy_regex/), [`once_cell`](https://docs.rs/once_cell/latest/once_cell/) is already being [used](https://github.com/getzola/zola/blob/next/components/libs/Cargo.toml#L22) in the codebase. Added `once_cell` since the regex shouldn't be compiled for every page it process.

This should address #2053! :raised_hands:

### DIFF
--- a/components/content/Cargo.toml
+++ b/components/content/Cargo.toml
@@ -6,6 +6,8 @@ edition = "2021"
 [dependencies]
 serde = {version = "1.0", features = ["derive"] }
 time = { version = "0.3", features = ["macros"] }
+once_cell = "1.17.1"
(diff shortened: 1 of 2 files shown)


mufeedvh:
Has some quirks, will revisit, closing this for now.

PR #2153, opened 2023-03-26
Allow empty templates dir

### LINKED ISSUE #2150: [Regression] Doesn't build without top-level templates directory
Since version 0.17.1 zola requires the top-level `/templates` directory to exist, even if a theme is used. Otherwise it will fail with:
```
Error: Failed to build the site
Error: Error parsing templates from the /templates directory
Error: Reason: Io error while writing rendered value to output: NotFound
Error: Reason: No such file or directory (os error 2)
```

This is because now tera handles globs differently, as discussed in https://github.com/Keats/tera/issues/819#issuecomment-1480392230.

I believe that this is not very user-friendly, as it can quickly lead to unexpected build failures with the default setup as follows:
- Run `zola init mysite`
- Run `zola serve`
- The official default index.html says that you should either add an `index.html` or add a theme to use
- You decide for the second option and add a theme, which leaves the top level templates directory empty
- It builds fine, so you push to a git repo
- Now pulling on a different machine or on CI will cause zola to fail building the site, because there is no templates directory checked into git

In my opinion this is not optimal, and zola should still work if there is no top-level templates directory.

The following patch fixes the problem from zola's side:
```diff
diff --git a/components/templates/src/lib.rs b/components/templates/src/lib.rs
index 6441ff49..e2747bfc 100644
--- a/components/templates/src/lib.rs
+++ b/components/templates/src/lib.rs
@@ -47,8 +47,7 @@ pub fn load_tera(path: &Path, config: &Config) -> Result<Tera> {
 
     // Only parsing as we might be extending templates from themes and that would error
     // as we haven't loaded them yet
-    let mut tera =
-        Tera::parse(&tpl_glob).context("Error parsing templates from the /templates directory")?;
+    let mut tera = Tera::parse(&tpl_glob).unwrap_or(Tera::default());
 
     if let Some(ref theme) = config.theme {
         // Test that the templates folder exist for that theme
```

However it's not really a good solution, because it will shadow actual parsing problems if `/templates` exists and there are real problems. This is why I believe this problem is better fixed on tera's side, i.e. making invalid globs just return the empty set instead of an error, which is how it is done in most languages.

iwahbe:

Sanity check:

## Code changes

This will fix #2150.

### DIFF
--- a/components/templates/src/lib.rs
+++ b/components/templates/src/lib.rs
@@ -42,13 +42,15 @@ pub fn render_redirect_template(url: &str, tera: &Tera) -> Result<String> {
 }
 
 pub fn load_tera(path: &Path, config: &Config) -> Result<Tera> {
-    let tpl_glob =
-        format!("{}/{}", path.to_string_lossy().replace('\\', "/"), "templates/**/*.{*ml,md}");
-
-    // Only parsing as we might be extending templates from themes and that would error
-    // as we haven't loaded them yet
-    let mut tera =
-        Tera::parse(&tpl_glob).context("Error parsing templates from the /templates directory")?;
+    let mut tera = if path.join("templates").exists() {
+        let tpl_glob =
+            format!("{}/{}", path.to_string_lossy().replace('\\', "/"), "templates/**/*.{*ml,md}");
+        // Only parsing as we might be extending templates from themes and that would error
+        // as we haven't loaded them yet
+        Tera::parse(&tpl_glob).context("Error parsing templates from the /templates directory")?
+    } else {
+        Tera::default()
+    };
 
     if let Some(ref theme) = config.theme {
         // Test that the templates folder exist for that theme
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.18.0 (unreleased)
 
+- Fix error on missing top-level `templates` directory.
+
 ## 0.17.2 (2023-03-19)
 
 - Fix one more invalid error with colocated directories
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+## 0.18.0 (unreleased)
+
 ## 0.17.2 (2023-03-19)
 
 - Fix one more invalid error with colocated directories


components/templates/src/lib.rs
```diff
@@ -47,8 +47,15 @@ pub fn load_tera(path: &Path, config: &Config) -> Result<Tera> {
 
     // Only parsing as we might be extending templates from themes and that would error
     // as we haven't loaded them yet
-    let mut tera =
-        Tera::parse(&tpl_glob).context("Error parsing templates from the /templates directory")?;
+    let mut tera = match Tera::parse(&tpl_glob) {
```
Keats: We have the path from above. We can just join `templates` and see if it exists. If it does then we load Tera with the glob otherwise Tera::default()

Keats (changes requested):

components/templates/src/lib.rs
iwahbe: That would be a little more straight forward.

vimpostor:
I still believe this issue is best fixed on tera's side. If you look at the bottom of https://github.com/Keats/tera/issues/819, you can already see quite an enormous amount of reverse references to repos where this change caused a regression.

@Keats Perhaps it is best to rethink this and to allow globs over non-existing directories to return the empty set again. After all this is how it is done in most of other languages and sometimes following widely established conventions is more important than a personal strong opinion about some behaviour. ;)

Keats:
We can fix it in Tera, I didn't see the number of reverse mentions

cwkang1998:
Hey, encountering issues with this. Since tera had already fixed and merged this in (and i believe its published with the latest version of tera), can we upgrade tera to latest to address the issue?

Keats:
It's already upgraded in the next branch

cwkang1998:
Currently still having issues with 0.17.2 on CI, is it scheduled for the next release?

vimpostor:
Currently zola is still using tera 1.18.1, which is the problematic version: https://github.com/getzola/zola/blob/9ae7d15139fe49319c081870ec456844a2709b11/Cargo.lock#L3537

You need to upgrade to tera 1.19.0 by running `cargo update`. Then afterwards make a zola release, so that the version with the fix can be used in CI and in distros.

PR #2157, opened 2023-03-29
Update documentation for multilingual use of taxonomies

lfxgroove:

Sanity check:

Kind of a issue/pr in the same thing unfortunately. As far as I can tell I haven't gotten taxonomies to work for a multilingual site with different kinds of errors when trying what the current docs say. With the patch I'm proposing here it seems to work, i.e. the taxonomies exist and `zola serve` doesn't give any errors.

Thanks for taking a look!


Keats (approved):
Thanks!

PR #2158, opened 2023-03-30
Documentation: minor updates to the GettingStarted/Overview page

felixwrt:
While working through the docs I noticed that the GettingStarted/Overview is based on Zola 0.9, which I guess is pretty old.

I went through the tutorial using the current version 0.17.1 and can confirm that it still works as described in the docs. The only difference was the output generated by `zola serve`, which I updated in the docs. I also put the instruction to `cd` into the `myblog` directory before running `zola serve` and removed the hint that talked about it afterwards. I think it's more user-friendly this way.

That's all! Let me know if you want something changed.

---

Sanity check:


Keats (approved):
Thanks!

PR #2163, opened 2023-04-02
Implement replace_re filter

cydave:
Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/filters.rs
+++ b/components/templates/src/filters.rs
@@ -4,8 +4,8 @@ use std::hash::BuildHasher;
 
 use config::Config;
 
-use libs::regex::Regex;
 use libs::base64::engine::{general_purpose::STANDARD as standard_b64, Engine};
+use libs::regex::Regex;
 use libs::tera::{
     to_value, try_get_value, Error as TeraError, Filter as TeraFilter, Result as TeraResult, Tera,
     Value,
--- a/components/templates/src/filters.rs
+++ b/components/templates/src/filters.rs
@@ -308,9 +308,11 @@ mod tests {
         let mut args = HashMap::new();
         args.insert("pattern".to_string(), to_value(pattern).unwrap());
         args.insert("rep".to_string(), to_value(rep).unwrap());
-        let result = ReplaceReFilter::new().filter(&to_value(value).unwrap(), &args);
+        let replace_re = ReplaceReFilter::new();
+        let result = replace_re.filter(&to_value(value).unwrap(), &args);
         assert!(result.is_ok());
         assert_eq!(result.unwrap(), to_value(expected).unwrap());
+        assert!(replace_re.re_cache.lock().unwrap().contains_key(pattern));
     }
 
     #[test]
--- a/components/templates/src/filters.rs
+++ b/components/templates/src/filters.rs
@@ -82,46 +82,48 @@ pub fn base64_decode<S: BuildHasher>(
 }
 
 #[derive(Debug)]
-pub struct ReplaceReFilter {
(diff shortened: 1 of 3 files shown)


components/templates/src/filters.rs
```diff
@@ -78,6 +80,22 @@ pub fn base64_decode<S: BuildHasher>(
...
+    let text = try_get_value!("replace_re", "value", String, value);
+    let pattern = match args.get("pattern") {
+        Some(val) => try_get_value!("replace_re", "pattern", String, val),
+        None => return Err(TeraError::msg("Filter `replace_re` expected an arg called `pattern`")),
+    };
+    let rep = match args.get("rep") {
+        Some(val) => try_get_value!("replace_re", "rep", String, val),
+        None => return Err(TeraError::msg("Filter `replace_re` expected an arg called `rep`")),
+    };
+    let pattern_re = Regex::new(&pattern)
```
Keats: Can you cache the compiled regex? If someone puts that filter in their base template and they have a lot of pages, re-compiling it could take a significant amount of time wrt the build time.

Keats (changes requested):

components/templates/src/filters.rs
cydave: I'm rather new to rust (and zola) but I'll try and give it a shot.
Is there a place in zola which does caching already?

Keats: You can have a struct that contains a hashmap regex str -> compiled_regex and implement the filter trait on it. It's the same as https://github.com/getzola/zola/blob/master/components/templates/src/global_fns/load_data.rs#L213

cydave: I wasn't sure if I needed the same sync primitives for this filter as the one you have linked to, so I just included it. I hope something along those lines in commit 
185856d should work?

components/templates/src/filters.rs
```diff
@@ -80,20 +81,49 @@ pub fn base64_decode<S: BuildHasher>(
...
+            Some(val) => try_get_value!("replace_re", "replace", String, val),
+            None => {
+                return Err(TeraError::msg("Filter `replace_re` expected an arg called `replace`"))
+            }
+        };
+
+        let mut cache = self.re_cache.lock().expect("re_cache lock");
+        let pattern_re = {
+            match cache.get(&pattern) {
+                Some(pat) => pat.clone(),
```
Keats: You don't need to clone it, you can call replace_all here and return replace in L112

components/templates/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub static ZOLA_TERA: Lazy<Tera> = Lazy::new(|| {
     .unwrap();
     tera.register_filter("base64_encode", filters::base64_encode);
     tera.register_filter("base64_decode", filters::base64_decode);
-    tera.register_filter("replace_re", filters::replace_re);
+    tera.register_filter("replace_re", filters::ReplaceReFilter::new());
```
Keats: Can you rename it to `regex_replace`?

Keats (changes requested):

components/templates/src/filters.rs
Keats: It's probably fine

components/templates/src/lib.rs
cydave: Sure thing, will do.

cydave:
I've renamed the filter to regex_replace and updated the documentation as well as tried removing calls to Regex.clone(). I hope it looks better now.

Keats (approved):
Looks good, thanks!

PR #2171, opened 2023-04-06
Speedup "zola check" command by reusing the Client

MTRNord:

Sanity check:

## Code changes
The code changes that `zola check` reuses the reqwest Client instead of creating it for every single request.
The benefit of this is a major speedup since it is able to reuse the connections. This is especially helpful for big pages like the one I am working on (17733 external links). The current stable way takes multiple hours, while this PR makes it go down to way less (as of writing it is still running but the progress bar I added in my local code predicts it to take 30-40 minutes instead of 14h+).

### DIFF
--- a/components/link_checker/src/lib.rs
+++ b/components/link_checker/src/lib.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::result;
 use std::sync::{Arc, RwLock};
+use std::time::Duration;
 
 use libs::once_cell::sync::Lazy;
 use libs::reqwest::header::{HeaderMap, ACCEPT};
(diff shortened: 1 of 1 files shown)


Keats (approved):

Keats:
I just realised, the commit says it adds timeout but the code doesn't?

MTRNord:
> I just realised, the commit says it adds timeout but the code doesn't?

Ooops that's 100% on me. I removed that one after I opened the PR after realizing reqwest actually already defaults to 30s these days (It didn't in the past. So I naturally just gravitated to add it, but while checking the Docs I saw they default to it now. The commit is wrong because I used amend... Sorry!)

PR #2194, opened 2023-04-27
Add 'this' var to alias page or section for templates

Raymi306:

Sanity check:

## Code changes

---
Fixes issue #687

If there is interest in this PR I can edit the documentation.

Also, this was only a few lines to implement, definitely let me know if I missed something or oversimplified this.

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -1,4 +1,4 @@
-/// A page, can be a blog post or a basic page
+// A page, can be a blog post or a basic page
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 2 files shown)


Keats:
That's something we want to do but not right now, it would make the current rendering quite slow. In the next Tera version we should be able to that without perf impact

Raymi306:
Closing, waiting for next Tera version

aterenin:
Gentle ping on this: has the Tera version been updated so that this is no longer a problem? Having a variable called `this` so that I don't have to add `{% set this = page | default(value = section) %}` to a bunch of template files would be extremely useful.

Keats:
It's not done yet: https://github.com/Keats/tera2

Next is making it so tests/filters/fns can access the values in the current scope so you don't need to do things like `lang=lang`

salif:
How to do this in tera2?

Keats:
That will be part of a bigger discussion: https://zola.discourse.group/t/section-vs-page/522
I'll start a RFC soonish

PR #2196, opened 2023-04-27
Add search.index_format into the serialized config (#2165)

### LINKED ISSUE #2165: [bug] config.search.index_format not accessible from templates.
# Bug Report

I am in the process of [refactoring the Abridge theme](https://github.com/Jieiku/abridge/issues/105).
One of the things I am doing is switching over to the [new json index format](https://github.com/getzola/zola/pull/1998)

When using the new json index format, you use a different search.js then you would with the regular js index. (this new version fetches the json file on interaction)

because of this I need to ensure or check that the right configuration is set.

```html
  {%- if config.search.index_format == "elasticlunr_json" %}
  <script defer src="{{ get_url(path='search_json.js', trailing_slash=false, cachebust=true) | safe }}"{%- if config.extra.integrity | default(value=true) %} integrity="sha384-{{ get_hash(path='search_json.js', sha_type=384, base64=true) | safe }}"{%- endif %}></script>
  {%- else %}
  <script defer src="{{ get_url(path='search_js.js', trailing_slash=false, cachebust=true) | safe }}"{%- if config.extra.integrity | default(value=true) %} integrity="sha384-{{ get_hash(path='search_js.js', sha_type=384, base64=true) | safe }}"{%- endif %}></script>
  {%- endif %}
```

## Environment

Zola version: 0.17.1

## Expected Behavior
I need for `config.search.index_format` to be accessible.

OR maybe `build_search_index = true` could also be `build_search_index = "elasticlunr_json"`  then there would not even be the need for the additional `index_format` field.

## Current Behavior
Currently you cannot check the value of `config.search.index_format` from templates

Raymi306:

Sanity check:

## Code changes

---
Fixes #2165 

If there is interest in continuing with this PR, I will update documentation

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -109,7 +109,7 @@ pub struct SerializedConfig<'a> {
     build_search_index: bool,
     extra: &'a HashMap<String, Toml>,
     markdown: &'a markup::Markdown,
-    search: &'a search::Search,
+    search_index_format: &'a search::IndexFormat,
 }
 
 impl Config {
@@ -332,7 +332,7 @@ impl Config {
             build_search_index: options.build_search_index,
             extra: &self.extra,
             markdown: &self.markdown,
-            search: &self.search,
+            search_index_format: &self.search.index_format,
         }
     }
 }
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -109,6 +109,7 @@ pub struct SerializedConfig<'a> {
     build_search_index: bool,
     extra: &'a HashMap<String, Toml>,
     markdown: &'a markup::Markdown,
+    search: &'a search::Search,
 }
 
 impl Config {
@@ -331,6 +332,7 @@ impl Config {
             build_search_index: options.build_search_index,
             extra: &self.extra,
             markdown: &self.markdown,
+            search: &self.search,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Raymi306:
@Jieiku This should enable you to do what you need

Jieiku:
> @Jieiku This should enable you to do what you need

I have a few open tickets, which one would this fix? would love to test it out.

Jieiku:
Thanks! Going to give it a try!

Keats:
I would start with just the index format serialized and then we can see if we need more. The rest feels like config that you don't need in the templates.

Raymi306:
> I would start with just the index format serialized and then we can see if we need more. The rest feels like config that you don't need in the templates.

This makes sense.

I looked in the docs for places where I want to mention this variable being available, perhaps [here](https://www.getzola.org/documentation/templates/overview/) in templates/overview?

Would you like any tests for this?

components/config/src/config/mod.rs
```diff
@@ -109,6 +109,7 @@ pub struct SerializedConfig<'a> {
     build_search_index: bool,
     extra: &'a HashMap<String, Toml>,
     markdown: &'a markup::Markdown,
+    search_index_format: &'a search::IndexFormat,
```
Keats: You can create a new struct (have a look at how we do it for page and sections) for the data serialized so if we add more items it's nicely contained.

Keats:
> I looked in the docs for places where I want to mention this variable being available, perhaps [here](https://www.getzola.org/documentation/templates/overview/) in templates/overview?

No need I think, it's going to be part of the config which is documented.

> Would you like any tests for this?

No need for just serialization

Keats (approved):

Keats:
Thanks!

Issue #2135, opened 2023-03-09
Warn when a page/section is using the default template


Keats:
So people don't have weird pages without realising.

Raymi306:
I'm taking a stab at this, the changes I'm having to make are non trivial though...
I suspect I'll be a tad busy this weekend though. We'll see if I get to finishing it.

PR #2200, opened 2023-04-30
Add default template warning

Raymi306:

Sanity check:

## Code changes

---
Issue #2135 
SO, I manually tested this a bit and it worked as best I could tell.
I ended up having to dig pretty far into code for rendering templates and then bubble information up about whether or not a default template was rendered.
Besides the arguable ugliness of this approach, it also means that it doesn't work for zola check as the build has to happen.
There should be some simpler way to check all of the templates without building...I will think about it now that I'm getting more familiar with the code and I am also open to suggestions. I may be able to work on this later today yet, we'll see.

IF this approach is acceptable, at the very least I want to do a cleanup pass and add some comments/documentation.

### DIFF
--- a/components/content/src/library.rs
+++ b/components/content/src/library.rs
@@ -26,6 +26,9 @@ macro_rules! set {
 pub struct Library {
     pub pages: AHashMap<PathBuf, Page>,
     pub sections: AHashMap<PathBuf, Section>,
+    // NOTE FOR REVIEW perhaps I can put a property on Page and Section and compute this collection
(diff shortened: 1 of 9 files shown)


Raymi306:
I've begun rewriting this.
My new plan is to add a method to Site, "get_default_templates".
This method will iterate through all pages and sites of the library.
I will write some functions to determine if a page or site is using a default template, collect the results, and return a vec either of pathbufs or strings

Raymi306:
See #2205

PR #2205, opened 2023-05-01
Add default template warning

Raymi306:

Sanity check:

## Code changes

---

Rewrite of #2200 
This should be less ugly with way fewer lock acquisitions, and it actually works for `zola check`.
I have only done some simple manual tests so far.

### DIFF
--- a/components/utils/src/templates.rs
+++ b/components/utils/src/templates.rs
@@ -101,7 +101,7 @@ pub fn render_template(
 }
 
 pub fn is_default_template(name: &str, tera: &Tera, theme: &Option<String>) -> Result<bool> {
-    if let Some(_) = check_template_fallbacks(name, tera, theme) {
+    if check_template_fallbacks(name, tera, theme).is_some() { 
         return Ok(false);
     }
     match name {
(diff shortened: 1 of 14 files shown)


components/utils/src/templates.rs
```diff
@@ -100,6 +100,16 @@ pub fn render_template(
     }
 }
 
+pub fn is_default_template(name: &str, tera: &Tera, theme: &Option<String>) -> Result<bool> {
```
Keats, line 103: That's not the right place to check. You essentially need to search for templates with that name in a theme or the templates directory Tera instances. Only if there are no overrides should the message be shown. Another way to do it would be to have a `log` Tera functions that does println so the template itself can warn

Keats (changes requested):

components/utils/src/templates.rs
Raymi306: > You essentially need to search for templates with that name in a theme or the templates directory Tera instances. Only if there are no overrides should the message be shown. 

Sorry, I thought I had accomplished this with this function which uses check_template_fallbacks.  I will think on this but I am currently confused as to how to do this differently without repeating the logic in check_template_fallbacks.

Raymi306: How do I do this without repeating logic in check_template_fallbacks?

Issue #2071, opened 2023-01-15
Some valid yaml dates are treated as invalid


jyn514:
# Bug Report

## Environment

Zola version: zola 0.16.1

## Expected Behavior

This front-matter should parse successfully (see https://www.yamllint.com/ for an example of another tool that can parse it; this was originally migrated from a Jekyll site which can also parse it):

```
---
layout:	post
title:	"Hosting an SSH Server"
date:	"2017-12-27 20:46:24 -0500"
excerpt: I decided one day that I really wanted a remote server.
audience: tech-savvy
---
```

## Current Behavior

```
PS C:\Users\jyn\src\third-website> zola serve                     
Building site...
Error: Failed to serve the site
Error: Error when parsing front matter of section `\\?\C:\Users\jyn\src\third-website\content\blog\2017-12-27-Hosting-an-SSH-Server.md`
Error: Reason: YAML deserialize error: Error("failed to parse datetime", line: 2, column: 1)
```

I think the problem is that dates are unconditionally parsed as toml dates, which is not correct: https://github.com/getzola/zola/blob/39bfbcd0199340ce5de4d0b951a2050cc9fc239e/components/content/src/front_matter/page.rs#L31-L33

## Step to reproduce

Add the given frontmatter to any .md file in `content/blogs`.

Keats:
Thanks for the clear bug report.
Anyone wants to grab that?

Kidsan:
@jyn514 Unless I'm mistaken this is an error with your date value and not to do with parsing the field. The supported date formats are YYYY-MM-DD and RFC3339. The field should also not be quoted. See the comment on the date field in the docs: https://www.getzola.org/documentation/content/page/#front-matter I have no experience with Jekyll but it looks like it does not use the same date formats in front matter: https://jekyllrb.com/docs/front-matter/

Incidentally, the page you linked isn't parsing the date, it is just formatting the document, so does not change this.

jyn514:
> Unless I'm mistaken this is an error with your date value and not to do with parsing the field. The supported date formats are YYYY-MM-DD and RFC3339.

These are valid YAML formats, so if zola is unable to parse them that means it's YAML parser isn't compliant with the specification. I'm not sure why Zola chose to use a custom date parser instead of the types that are built-in to both toml and yaml. See Dates and Times on https://toml.io/en/ and https://yaml.org/type/timestamp.html.

> The field should also not be quoted.

[...]

PR #2208, opened 2023-05-02
Add support for yaml datetimes

Raymi306:

Sanity check:

## Code changes

---

Issue #2071 
Please note that I wasn't fully sure how to incorporate my code with from_toml_datetime deserializer.
It looks like the time crate has [some sort of way to output rfc3339 strings](https://docs.rs/time/latest/time/format_description/well_known/index.html), and I'll try and wire that up if this PR is deemed acceptable.

I probably should write more tests if we want to keep this too.

I really ought to use verbose mode on the regex string, so I can add back in newlines and comments

### DIFF
--- a/components/content/src/front_matter/page.rs
+++ b/components/content/src/front_matter/page.rs
@@ -7,7 +7,7 @@ use time::macros::{format_description, time};
 use time::{Date, OffsetDateTime, PrimitiveDateTime};
 
 use errors::{bail, Result};
-use utils::de::{fix_toml_dates, from_toml_datetime};
+use utils::de::{fix_toml_dates, from_unknown_datetime};
 
 use crate::front_matter::split::RawFrontMatter;
 
@@ -20,7 +20,7 @@ pub struct PageFrontMatter {
     /// Description in <meta> that appears when linked, e.g. on twitter
     pub description: Option<String>,
     /// Updated date
-    #[serde(default, deserialize_with = "from_toml_datetime")]
+    #[serde(default, deserialize_with = "from_unknown_datetime")]
     pub updated: Option<String>,
     /// Datetime content was last updated
     #[serde(default, skip_deserializing)]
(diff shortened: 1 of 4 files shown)


Keats:
Sorry I completely missed that. Are you still interested or should I finish it myself?

Raymi306:
> Sorry I completely missed that. Are you still interested or should I finish it myself?

If you are interested in these changes, I should be able to clean up this PR and finish it.

> It looks like the time crate has [some sort of way to output rfc3339 strings](https://docs.rs/time/latest/time/format_description/well_known/index.html), and I'll try and wire that up if this PR is deemed acceptable.

I don't remember why I said this or if it was important.

Keats:
Yep definitely go for it

components/utils/src/de.rs
```diff
@@ -1,7 +1,82 @@
+use core::convert::TryFrom;
+use errors::Result;
+use libs::regex::Regex;
 use libs::tera::{Map, Value};
+use libs::time;
 use libs::toml;
 use serde::{Deserialize, Deserializer};
 
+pub fn parse_yaml_datetime(date_string: &str) -> Result<time::OffsetDateTime> {
```
Keats, line 10: Maybe the idea was to use the time crate to replace that?

Raymi306:
I rebased and wrote some more tests

Raymi306:
Some additional thoughts:
The spec I based my parsing off of was not finalized.
I used [this](https://yaml.org/type/timestamp.html) as a resource.
Since YAML is time format agnostic, it was valid to just specify that Zola only supported what it supports.

Jekyll is using something not quite the same as what was proposed for YAML:
> A date is specified in the format YYYY-MM-DD HH:MM:SS +/-TTTT; hours, minutes, seconds, and timezone offset are optional. 

This theoretically should be a simpler pattern to match, and would expand the supported time formats to be on par with Jekyll.
I guess it doesn't hurt to support Jekyll plus a little more which is what I ended up doing here.

components/content/src/front_matter/page.rs
```diff
@@ -129,6 +129,7 @@ impl PageFrontMatter {
     /// Converts the TOML datetime to a time::OffsetDateTime
     /// Also grabs the year/month/day tuple that will be used in serialization
     pub fn date_to_datetime(&mut self) {
+        println!("{:?}", self.date);
```
Keats, line 132: this can be removed

Keats (approved):
LGTM, I'll remove the print after merging

Keats:
Thanks!

Issue #2055, opened 2022-12-15
Feature request : Add support for lazy loading images


TheCapsLock:
To make our pages load faster and minimize the use of bandwidth/memory when a user does not scroll a whole page ; we can use the following attribute to make the image load only when needed (lazy loading): 

```
<img src="image.jpg" loading="lazy" />
```
references :

* https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes
* https://caniuse.com/loading-lazy-attr

c-git:
This can be done using a shortcode, I'm not sure how else it could be done. Maybe a setting in the config.toml.

Keats:
Oh that's nice, I didn't know that attribute. Right now you can do it with a shortcode.
Adding it to the parser is more involved as it's unlikely to be added to pulldown-cmark as they don't seem to add more features. Maybe it could be added to Zola.

Keats:
Fixed a while ago

PR #2211, opened 2023-05-04
Add support for lazy loading images

sinofp:

Sanity check:

## Code changes

---

Recently I learnt `loading="lazy" decoding="async"` and wanted to try them in my Zola-powered blog, then I found issue #2055 requesting `loading="lazy"`.
Given that issue has been marked by “help wanted”, I decided to give it a try, hence this PR.

This PR adds an option `lazy_async_image` to set `loading="lazy" decoding="async"` for images.
The `loading="lazy"` part is discussed in the issue, I can delete `decoding="async"` part if that’s too far.

The alt text in pulldown-cmark is enclosed within `Start(Image)` and `End(Image)`, and can contain styles (<https://github.com/raphlinus/pulldown-cmark/issues/394#issuecomment-527197423>).
I think most people will only use plain text in alt text so I didn’t bother cleaning the styles because “Having something simple and easy to use for 90% of the usecases is more interesting than covering 100% usecases after sacrificing simplicity.”
That said, there might be people want to use styles, links or even inline HTML for alt text, so the documentation includes a note on the limitation of the `lazy_async_image` option.

I added a small test to demonstrate the difference between the output of pulldown-cmark and mine. I also tested the program against my blog.

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -253,7 +253,6 @@ pub fn markdown_to_html(
     let mut stop_next_end_p = false;
 
     let lazy_async_image = context.config.markdown.lazy_async_image;
-    let mut is_image_alt = false; // For alt text between Event::Start(Tag::Image(..)) and Event::End(Tag::Image(..))
 
     let mut opts = Options::empty();
     let mut has_summary = false;
@@ -329,13 +328,7 @@ pub fn markdown_to_html(
         for (event, mut range) in Parser::new_ext(content, opts).into_offset_iter() {
             match event {
                 Event::Text(text) => {
-                    if lazy_async_image && is_image_alt {
-                        let mut alt = String::new();
-                        cmark::escape::escape_html(&mut alt, &text)
-                            .expect("Could not write to buffer");
-
-                        events.push(Event::Html(alt.into()));
-                    } else if let Some(ref mut _code_block) = code_block {
+                    if let Some(ref mut _code_block) = code_block {
                         if contains_shortcode(text.as_ref()) {
                             // mark the start of the code block events
                             let stack_start = events.len();
@@ -404,8 +397,6 @@ pub fn markdown_to_html(
                     };
 
                     events.push(if lazy_async_image {
-                        is_image_alt = true;
-
                         let mut img_before_alt: String = "<img src=\"".to_string();
                         cmark::escape::escape_href(&mut img_before_alt, &link)
                             .expect("Could not write to buffer");
@@ -423,7 +414,6 @@ pub fn markdown_to_html(
                     });
                 }
                 Event::End(Tag::Image(..)) => events.push(if lazy_async_image {
-                    is_image_alt = false;
                     Event::Html("\" loading=\"lazy\" decoding=\"async\" />".into())
                 } else {
                     event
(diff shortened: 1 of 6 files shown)


components/markdown/src/markdown.rs
```diff
@@ -326,7 +329,13 @@ pub fn markdown_to_html(
         for (event, mut range) in Parser::new_ext(content, opts).into_offset_iter() {
             match event {
                 Event::Text(text) => {
-                    if let Some(ref mut _code_block) = code_block {
+                    if lazy_async_image && is_image_alt {
```
Keats: Can we not just push back the Text event in that case and let pulldown-cmark handle it?

components/markdown/src/markdown.rs
```diff
@@ -387,13 +396,38 @@ pub fn markdown_to_html(
...
+                        cmark::escape::escape_href(&mut img_before_alt, &link)
+                            .expect("Could not write to buffer");
+                        if !title.is_empty() {
+                            img_before_alt
+                                .write_str("\" title=\"")
+                                .expect("Could not write to buffer");
+                            cmark::escape::escape_href(&mut img_before_alt, &title)
+                                .expect("Could not write to buffer");
+                        }
+                        img_before_alt.write_str("\" alt=\"").expect("Could not write to buffer");
```
Keats, line 410: It's not certain there will be an alt right?

components/markdown/tests/img.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    // alt text with link
+    let rendered = common::render_with_config(
+        "![ha[ha](https://example.com)](https://example.com/abc.jpg)",
+        config.clone(),
+    )
+    .unwrap()
+    .body;
+    assert_eq!(rendered, "<p><img src=\"https://example.com/abc.jpg\" alt=\"ha<a href=\"https://example.com\">ha</a>\" loading=\"lazy\" decoding=\"async\" /></p>\n");
+}
```
Keats, line 33: Can you add a test with an empty alt?

Keats (changes requested):
That's really cool, I didn't know about `decoding="async"`

components/markdown/src/markdown.rs
sinofp: Indeed, this addition is redundant. I have removed it.

components/markdown/src/markdown.rs
sinofp: Yes, the user might not input any alt text, but we still need to emit an empty alt attribute for the screen reader to ignore the picture: https://www.boia.org/blog/images-that-dont-need-alternative-text-still-need-alt-attributes

components/markdown/tests/img.rs
sinofp: Done! I added it for both pulldown-cmark and mine.

Keats: Sorry I forgot zola was already using [insta](https://crates.io/crates/cargo-insta) for snapshot testing. Can you check eg `can_render_basic_markdown` to see how it works? Otherwise I'll do it myself later

Keats (changes requested):

components/markdown/tests/img.rs
sinofp: No worries, I want to try it, so I’ve converted the test.

Keats (approved):

Keats:
Thanks!

PR #2217, opened 2023-05-17
remove libsass from build instructions

connorskees:
Removes reference to `libsass` in build instructions. Based on my reading, the `make`, `g++`, and `libssl-dev` dependencies are all required only by `libsass`, though I haven't tested a build without these installed.

I'll note I did also see these dependencies installed during some CI steps.


Keats (approved):
Thanks!

PR #2224, opened 2023-06-03
Updating hyper, tokio and open.

parasew:
Updating hyper, tokio and open. 
All tests passing.


Keats:
That's on the wrong branch and I believe we will automatically pick all those up by running `cargo update` so I'll close it. Thanks though.

PR #2230, opened 2023-06-09
docs: correct example use of Github GraphQL

### LINKED ISSUE #2229: Error in headers for Github GraphQL example
# Documentation issue

## Summary
The example using `load_data()`  to fetch from the Github GraphQL API has an `authentication` header. The header for sending the token should be `authorization` as [per the Github documentation](https://docs.github.com/en/graphql/guides/forming-calls-with-graphql).

## Proposed solution
Update [overview.md (line 472)](https://github.com/getzola/zola/blob/master/docs/content/documentation/templates/overview.md?plain=1#L472) to use `authorization` instead in case anyone else lazily cuts and pastes the example as a starting point (like I did!) and gets a failed query.

Aonrud:
Fixes #2229

### DIFF
--- a/docs/content/documentation/templates/overview.md
+++ b/docs/content/documentation/templates/overview.md
@@ -469,7 +469,7 @@ environment variable to the access token you have obtained.
 
 ```jinja2
 {% set token = get_env(name="GITHUB_TOKEN") %}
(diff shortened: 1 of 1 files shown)


Keats (approved):
Thanks!

PR #2236, opened 2023-06-29
always sort assets by filename (Windows/Linux difference)

wold5:
For me the `page.assets` items are sorted differently on Windows and Linux. Zola on Windows sorts assets alphabetically, for Linux the sorting is kinda arbitrary as shown below:
```
aaa.txt
ddd.txt
ccc.txt
bbb.txt
```

It's no big deal, and could also be fixed using `page.assets | sort`, but this differing  between platforms is not what one would expect.

A small patch is attached. The patch modifies `find_related_assets` to use the walkdir option [`sort_by_file_name`](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.sort_by_file_name) (docs hints that the order isn't guaranteed  otherwise). Feel free to accept or reject the patch.

## Using
Zola version: 1.7.2
Windows 11 + Debian (PopOS 22.04)

## Step to reproduce

Re-create the asset files using the scripts below. They're reversed, checking if Linux sorts by date (it does not).

### Windows powershell
```
New-Item ddd.txt
sleep 2
New-Item ccc.txt
sleep 2
New-Item bbb.txt
sleep 2
New-Item aaa.txt
```

### Linux/BSD
```
touch ddd.txt
sleep 2
touch ccc.txt
sleep 2
touch bbb.txt
sleep 2
touch aaa.txt
```

### DIFF
--- a/components/content/src/utils.rs
+++ b/components/content/src/utils.rs
@@ -87,13 +87,20 @@ mod tests {
         create_dir(path.join("subdir")).expect("create subdir temp dir");
         File::create(path.join("subdir").join("index.md")).unwrap();
         File::create(path.join("subdir").join("example.js")).unwrap();
+        File::create(path.join("FFF.txt")).unwrap();
+        File::create(path.join("GRAPH.txt")).unwrap();
+        File::create(path.join("subdir").join("GGG.txt")).unwrap();
 
         let assets = find_related_assets(path, &Config::default(), true);
-        assert_eq!(assets.len(), 4);
-        assert_eq!(assets.iter().filter(|p| p.extension().unwrap_or_default() != "md").count(), 4);
+        assert_eq!(assets.len(), 7);
+        assert_eq!(assets.iter().filter(|p| p.extension().unwrap_or_default() != "md").count(), 7);
 
-        for asset in ["example.js", "graph.jpg", "fail.png", "subdir/example.js"] {
+        for asset in ["example.js", "graph.jpg", "fail.png", "subdir/example.js", "FFF.txt", "GRAPH.txt", "subdir/GGG.txt"] {
             assert!(assets.iter().any(|p| p.strip_prefix(path).unwrap() == Path::new(asset)))
+        }
+
+        for w in assets.windows(2) {
+	        assert!(w[0].to_str().unwrap().to_ascii_lowercase() < w[1].to_str().unwrap().to_ascii_lowercase());
         }
     }
 
--- a/components/content/src/utils.rs
+++ b/components/content/src/utils.rs
@@ -122,12 +122,18 @@ mod tests {
         create_dir(path.join("subdir")).expect("create subdir temp dir");
         File::create(path.join("subdir").join("index.md")).unwrap();
         File::create(path.join("subdir").join("example.js")).unwrap();
-        let assets = find_related_assets(path, &Config::default(), false);
-        assert_eq!(assets.len(), 3);
-        assert_eq!(assets.iter().filter(|p| p.extension().unwrap_or_default() != "md").count(), 3);
+        File::create(path.join("FFF.txt")).unwrap();
+        File::create(path.join("GRAPH.txt")).unwrap();
+        File::create(path.join("subdir").join("GGG.txt")).unwrap();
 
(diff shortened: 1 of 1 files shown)


Keats:
Looking at the WalkDir function, I don't know if it does case insensitive ordering, I'm guessing it doesn't from the source? Might be needed for Mac, I'll check later tonight.

wold5:
> Looking at the WalkDir function, I don't know if it does case insensitive ordering, I'm guessing it doesn't from the source? Might be needed for Mac, I'll check later tonight.

Good point. It uses case-sensitive sorting (checked both Windows and Linux), i.e.

```
AAA.txt
BBB.txt
DDD.txt
aaa.txt
bbb.txt
```

Unpatched Windows builds do case-insensitive sorts.

There's currently no toggle for this in WalkDir . An [issue](https://github.com/BurntSushi/walkdir/issues/172) was followed by a [PR](https://github.com/BurntSushi/walkdir/pull/174#issuecomment-1424060261), that was [refused](https://github.com/BurntSushi/walkdir/pull/174#issuecomment-1424060261) and in limbo since. In summary, sorting Unicode is complex, and the dev doesn't want more dependencies and suggests using `sort_by` instead, with some supplied function. 
The PR provides many references with background info. For Rust there's `caseless` and `icu_collator`. 

Suggests using Walkdir [`sort_by`](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.sort_by), or sorting the `assets` vector outside WalkDir, or dropping the issue altogether...

Keats:
We can just keep the default unsorted walkdir and sort it ourselves after lowercasing the filenames. I _think_ it's what most people expect but I could be wrong!

wold5:
> We can just keep the default unsorted walkdir and sort it ourselves after lowercasing the filenames. I _think_ it's what most people expect but I could be wrong!

KISS then :)

So something like this before returning `assets`:
```
// Only ascii lowercase - 
assets.sort_by(|a, b| 
    a.file_name().unwrap().to_ascii_lowercase()
        .cmp(&b.file_name().unwrap().to_ascii_lowercase()
    )
);

// or for all lowercase convert to_str
assets.sort_by(|a, b| 
    a.file_name().unwrap().to_str().unwrap().to_lowercase()
        .cmp(&b.file_name().unwrap().to_str().unwrap().to_lowercase()
    )
);

```
I'm new to Rust BTW ;)

Keats:
KISS is my motto. First one is fine

wold5:
OK. It now uses `.to_str()` to also take the subdirectories into account.

Note sure about `unwrap()`. Given the checks in the function it should do (else use `unwrap_or_default()`). Tried `match`, but I really need to read up on Rust first, so I'll leave it at this for now.

Keats (approved):

Keats:
Thanks it looks good. Do you know how to write tests in Rust? Otherwise I can do it.

wold5:
> Thanks it looks good. Do you know how to write tests in Rust? Otherwise I can do it.

Great. Tests, no, but I can try. You can always replace it.

This would require extending the test below (at least), creating additional files like `aaa.txt bbb.txt CCC.txt ddd.txt` and then checking their order within `assets`, not?
https://github.com/getzola/zola/blob/a3dd45f9b42d24bcaeb042e87e9a64c76a5832f9/components/content/src/utils.rs#L78-L98

Keats:
Yep and since they will be sorted you can remove the loop in https://github.com/getzola/zola/blob/a3dd45f9b42d24bcaeb042e87e9a64c76a5832f9/components/content/src/utils.rs#L95 and compare by the array directly

wold5:
Just dawned on me what you meant... at least I hope so  :)

Alternatively, see the now removed standalone check using `windows()` in https://github.com/getzola/zola/pull/2236/commits/aeb9668eeece8a7661439b8783ccdaccb50ac81c 

~Apply this to `can_find_related_assets_non_recursive` as well?~
The change is now also applied to the non-recursive test function.
https://github.com/getzola/zola/blob/dec6815b49fe44d4ddbb03835ca50dff0577bd57/components/content/src/utils.rs#L108

Keats (approved):

Keats:
Nice, thanks!

PR #2237, opened 2023-07-04
Prevent spans crossing line boundaries in class-based code block formatter

### LINKED ISSUE #1942: minify_html = true causes Major page formatting issues to pages with numbered code blocks.
# Bug Report

`minify_html = true` causes Major page formatting issues to pages with numbered code blocks.

This was first reported to the abridge theme: https://github.com/Jieiku/abridge/issues/84

I implemented this in the [after-dark theme](https://github.com/getzola/after-dark) and seen the same effect (pay particular attention to the footer at the bottom of the pages)

with `minify_html = false` it looks normal:

## Environment

Zola version: 0.16.0

I also tested 0.15.3 with the same result.

## Expected Behavior

The page should look the same with or without `minify_html = true`

## Current Behavior
No Errors, just errors in page formatting when using `minify_html = true`

## Step to reproduce

```shell
git clone https://github.com/Jieiku/zola-minify-test
cd zola-minify-test
zola serve

```
open this page: http://127.0.0.1:1111/overview-code-blocks/

Notice the page section headings and page footer are all broken following the numbered code blocks.

To further confirm the issue you can removed the line numbered code blocks, then `minify_html = true` will no longer cause an issue.

TheOnlyMrCat:
The class-based code block formatter has bad interactions with other line-based highlighting effects (`linenos`, `hl_lines`) because the spans it generates cross multiple lines. This patch copies the `scope_to_classes` function from syntect and uses it to make each line independent of one another when using class-based highlighting.

The `SyntaxHighlighter::finalize` method has also been removed in this patch, as it has been rendered unnecessary.

This greatly increases the number of `<span>` tags generated by the formatter, even in code blocks that don't make use of line-based highlighting effects. I would be happy to implement a mechanism to only split the spans when it would otherwise cause problems, and keep the previous nesting behaviour elsewhere.

Fixes #1942

### DIFF
--- a/components/markdown/src/codeblock/highlight.rs
+++ b/components/markdown/src/codeblock/highlight.rs
@@ -6,7 +6,9 @@ use libs::syntect::highlighting::{Color, Theme};
 use libs::syntect::html::{
     line_tokens_to_classed_spans, styled_line_to_highlighted_html, ClassStyle, IncludeBackground,
 };
-use libs::syntect::parsing::{ParseState, ScopeStack, SyntaxReference, SyntaxSet};
+use libs::syntect::parsing::{
+    ParseState, Scope, ScopeStack, SyntaxReference, SyntaxSet, SCOPE_REPO,
+};
 use libs::tera::escape_html;
 
 /// Not public, but from syntect::html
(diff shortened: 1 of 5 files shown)


Keats:
Thanks. Can you add some test that would fail before this change?

TheOnlyMrCat:
I've added two snapshot tests for cases that would have exhibited incorrect behaviour (and extended `codeblocks`' testing framework to allow classed highlighting). Are there other tests that might suit this change better?

Keats:
Thanks that looks good!

PR #2241, opened 2023-07-05
Add 'serve --no-livereload-injection'

Hyask:
Hi Zola team,

Here is a small PR adding a `--no-livereload-injection` flag to the `serve` command.
Here is my use-case:
I maintain a website for an association that, among other things, needs a calendar. To share it easily with everyone, our website generates an ICS leveraging the templating system. You can see that [here](https://gitlab.com/ecole-lionel-oudart/ecole-lionel-oudart.gitlab.io/-/blob/main/templates/calendar.ics.html).
This works fine out-of-the-box, because zola always puts its generated files as `page_name/index.html`, meaning we have `calendar.ics/index.html` working fine from the browser when calling the `/calendar.ics` page. (The hack in [build.sh](https://gitlab.com/ecole-lionel-oudart/ecole-lionel-oudart.gitlab.io/-/blob/main/build.sh) is only here to help with `Content-type`).  
However, developing locally with such hack doesn't work well, because `zola serve` injects its `<script
src="livereload.js">` code in all pages, whether they are HTML or not, and thus breaks the ICS format.  
This simple PR helps with that use-case, the only drawback being the need to manually refresh the pages in the browser.

If you have other suggestions for fixing that use-case, I'd be willing to help and discuss it, but in the meantime, if
you're fine with the code, well, here it is.  
Sorry for not discussing it on Discourse before, but I discovered that only when opening the PR and I had already hacked
the code anyway to develop my website.

Sanity check:

## Code changes


Keats:
Hi

I'm not super keen on adding an argument for one specific hack. Templates right now in Zola are meant to be HTML, other formats are not supported. If/When we add other output format, this issue would be solved but with this hack you could just use zola build and a watcher and get the same thing.

Hyask:
Thanks for giving your position regarding whether or not to support the edgy use-case.  
In the meantime, here is the one-liner I now use for development, in case this helps anyone:  
`find . -not -path "./.git/*"  -not -path "./public/*" | entr -rs "zola build -u http://127.0.0.1:8000; cd public; python3 -m http.server"`

PR #2242, opened 2023-07-07
inject site config during Sass compilation

greenwoodcm:

Sanity check:

## Code changes


components/site/src/sass/mod.rs
```diff
@@ -52,6 +55,24 @@ pub fn compile_sass(base_path: &Path, output_path: &Path) -> Result<()> {
...
+/// and contained `.scss` file will be deleted on drop of the returned `TempDir`
+/// struct, which should happen after Sass compilation finishes.
+fn build_dependencies_dir_from_config(config: &Config) -> Result<TempDir> {
+    let dir = tempdir().context("failed to create tempdir for SASS dependencies")?;
+
+    let config_serialized = serde::serialize_config(config)?;
+
+    std::fs::write(dir.path().join("zola.scss"), format!("$config: {}", config_serialized))?;
+
+    Ok(dir)
```
Keats, line 73: What do you think of writing it to the sass directory so people can easily see what there is? It would be still overwritten on each serve/build but we can maybe not delete it

components/site/src/sass/serde.rs
```diff
@@ -0,0 +1,481 @@
...
+        let expected = "(\"a\":42,\"b\":\"abc\",\"c\":(\"a\":[6,7,8],\"b\":[\"def\",123.45]))";
+        verify_serialization(val, expected);
+    }
+
+    fn verify_serialization<T: Serialize>(val: T, expected: &str) {
+        let mut ser = SassMapSerializer::default();
+        val.serialize(&mut ser).unwrap();
+        assert_eq!(ser.output, expected);
+    }
+}
```
Keats, line 477: cc @connorskees Do you think it's something that would be interesting to have in grass directly? I can see injecting data being useful for more than Zola

test_site/sass/blog.scss
```diff
@@ -1,5 +1,21 @@
+@use 'zola';
+@use 'sass:map';
+
 body {
-  background: red;
+  background: 
+    // once we are using a more contemporary version of Dart/Sass we
+    // can use notation like:
+    //   map.get(zola.$config, "extra", "sass", "background_color")
```
Keats: Are we sure it's not supported yet? This seems..... very bad to use.

Keats (commented):
I like where this is going!

components/site/src/sass/serde.rs
connorskees: Interesting -- you mean defining variables in rust and exposing them to Sass?

The grass internals crate has an API for [defining builtin functions in rust and exposing them to Sass](https://docs.rs/grass_compiler/latest/grass_compiler/struct.Options.html#method.add_custom_fn). Something similar for variables would make a lot of sense. 

Exposing such an API in the main crate is hard because generally the AST nodes store their source span, but if I exposed a separate `Value` enum that was more minimal than the enum used internally, it should be possible to make something a bit more ergonomic. I also wouldn't have to worry about breaking changes to the internal representation.

test_site/sass/blog.scss
greenwoodcm: fixed in grass in the following PR

https://github.com/connorskees/grass/pull/83

components/site/src/sass/mod.rs
greenwoodcm: you're saying write it to the user's sass source directory?  i figured that would be undesirable because then maybe it gets erroneously checked in.  but i'm open to it if that seems better.

test_site/sass/blog.scss
greenwoodcm: i could patch in the latest grass release (1.13.0) in this PR, or i could do it as a follow up.  preference?

components/site/src/sass/serde.rs
greenwoodcm: i think the suggestion was to move the sass map serializer into grass, so that anything that implements `serde::Serialize` can be converted to a sass map literal string.  this would enable the functionality i've built in zola.

i don't have a feel for whether or not that functionality has other use cases, so i'm happy either way (that remaining in zola or being moved to grass).

connorskees: Ah. I would probably not implement this in grass directly to avoid the serde dependency, even if optional. I would say the more common solution would be to just define the config in Sass directly, which is what the major Sass libraries do AFAIK. 

I think zola is a special case here in needing to serialize TOML to Sass. Though I do think this code is quite cool. If there is another project out there that wants this functionality, I would be open to bringing it into grass-proper. It just currently seems a bit too niche.

components/site/src/sass/serde.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl<'a> serde::ser::SerializeMap for &'a mut SassMapSerializer {
+    type Ok = ();
+    type Error = SassMapSerializerError;
+
+    /// The Serde data model allows map keys to be any serializable type. SCSS
+    /// only allows string keys so the implementation below will produce an invalid
+    /// SCSS map if the key serializes as something other than a string.  That
+    /// should be fine here as the source TOML only allows string keys as well.
```
connorskees: I will note that Sass maps do permit any value as a key: 
[Sass docs](https://sass-lang.com/documentation/values/maps/#:~:text=Maps%20allow%20any%20Sass%20values%20to%20be%20used%20as%20their%20keys)
[example in playground](https://connorskees.github.io/grassmeister/?v=YSB7CiAgICBjb2xvcjogaW5zcGVjdCgoMTogMiwgImEiOiAiYiIsIGM6IGQsIFsxLCAyLCAzXTogWzQsIDVdKSk7Cn0=)

test_site/sass/blog.scss
Keats: Let's update in this PR

components/site/src/sass/mod.rs
Keats: I think it would be ok to check it in? It's not going to change very often

components/site/src/sass/serde.rs
greenwoodcm: ah ok thanks.  i'll just remove this comment.  it's actually not necessary because the implementation seems to do fine serializing non-string keys (though i don't think TOML supports that, so it's moot).

test_site/sass/blog.scss
greenwoodcm: updated to 1.13.0 in the latest

components/site/src/sass/serde.rs
greenwoodcm: ok I'm going to leave the functionality in Zola for now

components/site/src/sass/mod.rs
greenwoodcm: the more i think about it, the more i prefer _not_ having the injected config in source.  i don't particularly like it being checked in when the author didn't write it and when it's liable to get overwritten by the Zola compiler.  that creates a scenario where the author is looking through the sass directory and erroneously edits that file, only to have their edits confusingly wiped out on the next zola build.  we could solve that with a warning comment at the top of that file, but i prefer just hiding the source since the author isn't writing it.  i added site docs explaining this feature, in my opinion that makes clear to the author what's going on.  what do you think?

greenwoodcm:
@Keats thoughts on this latest rev?

components/site/src/sass/mod.rs
```diff
@@ -52,6 +55,24 @@ pub fn compile_sass(base_path: &Path, output_path: &Path) -> Result<()> {
...
 
+/// write out a subset of the Zola config document to a temporary SCSS file
+/// as an SCSS map variable literal.  this will allow parts of the site's
+/// config to be usable during Sass compilation.  this enables theme configuration
+/// like allowing the site owner to change header color.  this function returns
+/// a tempdir holding a single `.scss` file.  the tempdir should then be used as
+/// a load directory above when compiling the site's Sass files.  the tempdir
+/// and contained `.scss` file will be deleted on drop of the returned `TempDir`
+/// struct, which should happen after Sass compilation finishes.
+fn build_dependencies_dir_from_config(config: &Config) -> Result<TempDir> {
```
Keats, line 66: I don't understand the name of that function.

components/site/src/sass/mod.rs
Keats: It's fine to not add it the directory but it's a bit annoying that you can't see it to look at what exactly is defined

components/site/src/sass/serde.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    /// serialize any chars as if they were single-character strings
+    fn serialize_char(self, v: char) -> std::result::Result<Self::Ok, Self::Error> {
+        self.serialize_str(&v.to_string())
+    }
+
+    fn serialize_str(self, v: &str) -> std::result::Result<Self::Ok, Self::Error> {
+        self.output += "\"";
+        self.output += v;
```
Keats, line 117: what if the string contains `"`?

components/site/src/sass/serde.rs
```diff
@@ -0,0 +1,477 @@
...
+}
+
+impl std::error::Error for SassMapSerializerError {}
+
+impl<'a> Serializer for &'a mut SassMapSerializer {
+    type Ok = ();
+    type Error = SassMapSerializerError;
+
+    type SerializeSeq = Self;
+    type SerializeTuple = Self;
```
Keats, line 55: You can use the type here so you don't repeat multiple times the same impl for different kinds like tuple/seq

components/site/src/sass/serde.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    fn end(self) -> SassMapSerializerResult {
+        self.output += "]";
+        Ok(())
+    }
+}
+
+// Same thing but for tuple structs.
+impl<'a> serde::ser::SerializeTupleStruct for &'a mut SassMapSerializer {
```
Keats, line 308: See my comment above, you don't need to repeat it like that

Keats (changes requested):

components/site/src/sass/serde.rs
Keats: See https://github.com/Keats/tera2/blob/master/src/value/ser.rs#L16 for an example

components/site/src/sass/mod.rs
greenwoodcm: here we take the config document and construct a directory that contains files that are ultimately added as dependencies when compiling Sass.  the Sass compiler asks for a "load dir" during compilation, so instead of providing a single file we must provide a single directory containing a single file.  so the function name is basically "builds a dependencies directory from the config document".  happy to rename, do you have a suggestion?

components/site/src/sass/mod.rs
greenwoodcm: any suggestions on how to surface this content to the site builder without putting it in the directory?  i could create a non-tmp directory in `/tmp` and log the location of it?

components/site/src/sass/serde.rs
greenwoodcm: great point.  it seems like the most straightforward way to make this more robust is to first convert the `String` to a `serde_json::Value::String` and then write that out.  `Value::String` will do the right thing regarding escapes, which should function the same for a Sass literal as for JSON.

the downside here is obviously taking a dependency on the `serde_json` crate.  the alternative would be to simply check for the presence of characters that would require an escape and fail with a helpful error message.  that also seems acceptable, as I can't think of a compelling reason why someone would need quotes or other escape characters in this context.

@Keats do you have a recommendation on which route to go?

components/site/src/sass/mod.rs
Keats: Honestly I'm not too sure. Maybe it's ok to just show an example of how it looks like in the docs

components/site/src/sass/mod.rs
Keats: `write_temp_config_file`?

greenwoodcm: that works for me.  seems odd to me that a function would be named that but return a directory instead of a file, but whatever.

Keats: ah fair enough `create_config_load_path`?

components/site/src/sass/serde.rs
Keats: can we do a naive escape (eg just a .replace)? It will do something weird if the user has an escaped string but that seems even less likely...

components/site/src/sass/mod.rs
greenwoodcm: i like it

components/site/src/sass/serde.rs
greenwoodcm: yeah i like that.  i think what i'll do is naively escape double quotes and search for the characters that i know would need an escape, failing if i find any.  that seems like a good happy medium between allowing the might-happen thing and disallowing the shouldn't-happen thing with a helpful message.

Keats:
@greenwoodcm do you think you will have time to address the comments?

greenwoodcm:
oof, sorry this has been sitting forever, family life has me busy these days...i do plan on making these changes, thanks for the reminder

Keats:
No worries, I'm on paternity leave myself so not super active either

e792a8:
Hi, could anyone take a look at [my solution](https://github.com/e792a8/zola/commit/088ca62bc53e4f89ec0f2a9a8c111775927bbce8)?
It utilizes `grass_compiler::Options::add_custom_fn()` and then we just call in Sass `$cfg : zola-config();`. No need for any extra files.
If anyone is interested, I will open a new PR for that.

Keats:
Looks good! Are you sure you need to use the grass-compiler crate though?

e792a8:
Yes, as grass does not expose some types (Builtin, Value, etc.) required for add_custom_fn. [ref](https://docs.rs/grass/latest/grass/struct.Options.html#method.add_custom_fn)

Keats:
Interesting. I would maybe raise an issue on the grass repo since this exact usecase might be somewhat common

Keats:
It would probably be easier to template the Sass files.

PR #2263, opened 2023-07-29
template:feeds: add extra block

heitorPB:

Sanity check:

## Code changes

---

This PR adds an extra block on the feeds templates to allow users to customize the feeds with extra entries.

Forum post: https://zola.discourse.group/t/adding-tags-to-the-rss-atom-feeds/1820/4

I'm not happy with the names of the blocks, I think we can do better. What are your suggestions?


components/templates/src/builtins/atom.xml
```diff
@@ -48,6 +48,9 @@
         {% else %}
         <content type="html" xml:base="{{ page.permalink | escape_xml | safe }}">{{ page.content }}</content>
         {% endif %}
+
+        {%- block atom_entry_extra %}
+        {% endblock atom_entry_extra -%}
```
heitorPB: I still don't understand why, but these lines cause the tests to fail:

```
     Running tests/invalid.rs (/home/h/projects/zola/target/debug/deps/invalid-9ae7bc0e52bdd68b)

running 1 test
test errors_on_index_md_page_in_section ... FAILED

failures:

---- errors_on_index_md_page_in_section stdout ----
thread 'errors_on_index_md_page_in_section' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: Msg("Failed to parse '__zola_builtins/atom.xml'"), source: Some(Error { kind: Msg("  --> 52:9\n   |\n52 |         {% block atom_entry_extra %}\n   |         ^---\n   |\n   = unexpected tag; expected an `else` tag, an endfor tag (`{% endfor %}`), or some content"), source: None }) }', components/templates/src/lib.rs:28:6
```

The message is:

```bash
$ echo -e "  --> 52:9\n   |\n52 |         {%- block atom_entry_extra %}\n   |         ^---\n   |\n   = unexpected tag; expected an `else` tag, an endfor tag (`{% endfor %}`), or some content"
  --> 52:9
   |
52 |         {%- block atom_entry_extra %}
   |         ^---
   |
   = unexpected tag; expected an `else` tag, an endfor tag (`{% endfor %}`), or some content
```

We can't add a block to a builtin template? I am not sure what I am missing here.

heitorPB: I tried to add `<!-- Optional extra content -->` inside the block to have some content, but I got the same error.

I also tried adding these lines to `components/templates/src/lib.rs:19`:
```rust
        ("__zola_builtins/atom.xml", "{% extends \"__zola_builtins/atom_base.xml\" %}{% block atom_entry_extra %}<>{% endblock atom_entry_extra %}"),
        ("__zola_builtins/rss.xml",  "{% extends \"__zola_builtins/rss_base.xml\" %}{% block rss_item_extra %}<>{% endblock rss_item_extra %}"),
```

And no success.

I'm not familiar with details on Tera, but does this mean we can't have an empty `block` in the built-in templates?

Keats: It's because the block is in a for loop, which is not allowed in Tera. Maybe the extend approach doesn't work then.

heitorPB: Oh, that's a bummer.

I reverted the commits and updated the docs on how to get the templates and modify. Maybe that's worth of a merge.

Keats (approved):

Issue #2239, opened 2023-07-04
Exclude directories from zola check


thibaultamartin:
I have a website that has a blog section, and the blog has a lot of entries with a lot of external links. Running zola check takes several hours and makes it impossible to add to my CI.

Being able to exclude the blog directory from the links checker would be extremely useful for us!

Keats:
Have you seen the `skip_prefixes`config option  in https://www.getzola.org/documentation/getting-started/configuration/ ? It might not solve the issue straightaway but if you have 10k links to GitHub that you don't need to check for example it could alleviate the issue for now. 

That does seem ok to add as a feature though.

Any other issues/shortcomings with Zola for the matrix.org website?

thibaultamartin:
`skip_prefixes` doesn't really solve the problem for us since we can have many different addresses in the 21122 external links zola has to check.

> Any other issues/shortcomings with Zola for the matrix.org website?

Zola has generally been a joy to work with! It does the job and does it well, it takes a handful of seconds to build a website it took us several minutes to build before (though in all fairness it was an outdated Gatsby, and it had been severely abused).

The main limitation I've hit was [not being able to use get_page/get_section in shortcodes](https://www.getzola.org/documentation/content/shortcodes/), which is well documented.

> Shortcodes are rendered before the page's Markdown is parsed so they don't have access to the page's table of contents. Because of that, you also cannot use the [get_page](https://www.getzola.org/documentation/templates/overview/#get-page)/[get_section](https://www.getzola.org/documentation/templates/overview/#get-section)/[get_taxonomy](https://www.getzola.org/documentation/templates/overview/#get-taxonomy)/[get_taxonomy_term](https://www.getzola.org/documentation/templates/overview/#get-taxonomy-term) global functions. It might work while running zola serve because it has been loaded but it will fail during zola build.

A bit unfortunate, but we could work around it. Many thanks for all your work on Zola, it's a really good SSG :)

Raymi306:
@thibaultamartin I have a draft PR up, if you want to try building my branch and testing the new configuration option (note that I have not tested it myself beyond making sure it builds).
It goes in the link checker section of the config as `ignored_files` and takes a list of file globs

Keats:
It's been available for a while now

PR #2264, opened 2023-07-29
Allow ignoring files when link checking

Raymi306:
For #2239

Sanity check:

## Code changes

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -159,10 +159,7 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     for (file_path, links) in external_links {
         for link in links {
             if should_skip_by_prefix(link, &site.config.link_checker.skip_prefixes)
-                || should_skip_by_file(
-                    file_path,
-                    ignored_files_globset,
-                )
+                || should_skip_by_file(file_path, ignored_files_globset)
             {
                 skipped_link_count += 1;
             } else {
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -157,8 +157,12 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     // the ones that have invalid URLs
     for (file_path, links) in external_links {
         for link in links {
-            if should_skip_by_prefix(link, &site.config.link_checker.skip_prefixes) ||
-                should_skip_by_file(file_path, site.config.link_checker.ignored_files_globset.as_ref().unwrap()) {
+            if should_skip_by_prefix(link, &site.config.link_checker.skip_prefixes)
+                || should_skip_by_file(
+                    file_path,
+                    site.config.link_checker.ignored_files_globset.as_ref().unwrap(),
+                )
+            {
                 skipped_link_count += 1;
             } else {
                 match get_link_domain(link) {
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -155,12 +155,13 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     let mut invalid_url_links: u32 = 0;
     // First we look at all the external links, skip those the user wants to skip and record
     // the ones that have invalid URLs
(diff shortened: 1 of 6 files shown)


Raymi306:
I will address the test failures soon

Keats:
Hi @Raymi306 , are you still planning on addressing them?

Raymi306:
I apologize for the delay, I will look into this soon

Raymi306:
I did some updating of my fork and some cherrypicking to try and get this PR back in to shape.
I looked at 2 failing tests and I think that their checks are no longer relevant so I edited the tests to make them pass.
I am a little sick and it has been a long time since I worked on this, I need to write tests for the changes still.

Raymi306:
I note that components/site/link_checking has no tests and is rather monolithic, might be worth a "todo" comment to refactor the code to make it easier to test chunks of?

components/utils/src/globs.rs
```diff
@@ -0,0 +1,20 @@
...
+    // of the globber always returns false.
+    let mut glob_set_builder = GlobSetBuilder::new();
+    for pat in ignore {
+        let glob = match Glob::new(pat) {
+            Ok(g) => g,
+            Err(e) => bail!("Invalid ignored_{} glob pattern: {}, error = {}", name, pat, e),
+        };
+        glob_set_builder.add(glob);
+    }
+    Ok(glob_set_builder.build().expect(&format!("Bad ignored_{} in config file.", name)))
```
Keats: Don't use `expect`, we can convert that an error

src/cmd/serve.rs
```diff
@@ -759,6 +759,7 @@ pub fn serve(
 }
 
 fn is_ignored_file(ignored_content_globset: &Option<GlobSet>, path: &Path) -> bool {
+    // NOTE: globsets SHOULD be always Some unless config init fails, so we could unwrap
```
Keats: I don't think we need that note, it implies knowledge about other parts of the code. The function takes an `Option` so we should still check. Otherwise pass the GlobSet directly without the option

components/site/src/link_checking.rs
```diff
@@ -105,6 +106,10 @@ fn should_skip_by_prefix(link: &str, skip_prefixes: &[String]) -> bool {
     skip_prefixes.iter().any(|prefix| link.starts_with(prefix))
 }
 
+fn should_skip_by_file(file_path: &Path, glob_set: &GlobSet) -> bool {
```
Keats, line 109: not sure it's worth a fn if it's only used once in this file

components/site/src/link_checking.rs
```diff
@@ -152,7 +157,12 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     // the ones that have invalid URLs
     for (file_path, links) in external_links {
         for link in links {
-            if should_skip_by_prefix(link, &site.config.link_checker.skip_prefixes) {
+            if should_skip_by_prefix(link, &site.config.link_checker.skip_prefixes)
+                || should_skip_by_file(
+                    file_path,
+                    site.config.link_checker.ignored_files_globset.as_ref().unwrap(),
```
Keats: can you unwrap before the loop so it's easier to read?

Keats (changes requested):

components/site/src/link_checking.rs
Raymi306: should_skip_by_prefix was my point of reference here, it also is only used once in this file.

would you like me to remove both?

components/site/src/link_checking.rs
Raymi306: done

src/cmd/serve.rs
Raymi306: done, I removed the comment and left the match

components/utils/src/globs.rs
Raymi306: done

components/site/src/link_checking.rs
Keats: No that's fine then

Keats (approved):

Keats:
Thanks!

Issue #1787, opened 2022-03-03
Investigate tree-sitter to replace syntect


Keats:
Has anyone used it? The last time I looked at tree-sitter it didn't have many grammars but a quick look shows it's getting better. Our syntect syntaxes are stuck on _old_ versions of the grammars because of new features in the Sublime grammar format not supported by Syntect.
See https://github.com/nvim-treesitter/nvim-treesitter#supported-languages for a list of supported languages.

An alternative would be a basic textmate highlighter using VSCode syntaxes/themes since that's what everyone seems to be using these days.

GreatestOfAllTimee:
it's a nice move, Zola can't stay at the old versions of the grammars.

mwcz:
I haven't used tree-sitter as a library but it's _really_ nice in nvim.

jakelogemann:
👍 ts is the bee's knees. 

The official [tree-sitter-highlight](https://crates.io/crates/tree-sitter-highlight) crate has a few nice examples in the README... I haven't used it before, but I'm interested to try...

[ref](https://tree-sitter.github.io/tree-sitter/contributing#published-packages)

Keats:
If it's adopted (I don't know yet, I need to see the theming capabilities and just try it on various inputs), it would be its own package that can probably be on crates.io as well. I'd like to move all the lines numbers/highlights etc in it.

Keats:
List of parsers from neovim: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua

Jieiku:
sounds awesome.

Jieiku:
What would the tree-sitter output look like? would it list a ton of classes in the generated html like the current syntect solution does when you use css mode? or would it use classes that refer to css variables which we can then set to specific colors and styles, eg:
--z-1, --z-2, --z-3, etc... with a couple modifiers for bold, italic, and underline: --z-b, --z-i, --z-u

Keats:
Ideally it would the same kind as the current syntect output

Jieiku:
All the class definitions make the page source code much larger in size, but I can see how it would make things simpler as far as generating goes. If you simply used classes which refer to colors, then you would have to have some sort of lookup table per programming language. (because a bracket in one language might be colored, but in another language it might not be colored or colored differently)

I just wish there was a simple way to have much leaner generated html for syntax highlighting while using the css method.

Keats:
[...]

PR #2268, opened 2023-08-12
Add llvm syntax highlighting

### LINKED ISSUE #2267: Add syntax highlighting for llvm
# Bug Report

## Environment

Zola version:

## Expected Behavior
Code blocks with `ll` should enable syntax highlighting for llvm code. For example if you view the following block on github, you will see what the highlighting is supposed to look like:

```ll
define i32 @foo() {
entry:
  br i1 true, label %if.then, label %if.else
if.then:
  ret i32 10
if.else:
  ret i32 12
}
```

## Current Behavior
There is no syntax higlighting when `ll` is used.

michaelmaitland:
Closes #2267


Keats:
Thanks for the PR. I'm not going to add more built in syntaxes for now while we're looking at alternatives for highlighters.

michaelmaitland:
That’s unfortunate. Do you have any reason why we can’t add it today and remove it when we have an alternative syntax highlighting solution implemented? By that logic let’s remove all syntax highlighting for all languages since we’re looking into rehauling syntax highlighting that might get implemented one day.

Keats:
It's trivial to add syntaxes as an end user.

> By that logic let’s remove all syntax highlighting for all languages since we’re looking into rehauling syntax highlighting that might get implemented one day.

Let's not add more when the future is uncertain, no reason to break more stuff people rely on.

michaelmaitland:
> It's trivial to add syntaxes as an end user

can you link to this? I don’t know how to do this without maintaining my own downstream zola repo and build. 

> no reason to break more stuff people rely on

is there a reason why you think this change would break anything? It looks like zola has a clear track record of adding new languages without introducing any new problems. Why would this be any different?

Keats:
In https://www.getzola.org/documentation/content/syntax-highlighting/  search for `extra_syntaxes_and_themes`

> Why would this be any different?

We're looking at tree-sitter in https://github.com/getzola/zola/issues/1787 which supports fewer languages than Sublime, or sometimes they support it but the parser is 100MB and I'm not going to add it to Zola because it's too big.

michaelmaitland:
#1787 has been open for 1,5 years and the work you need from tree-sitter does not look like it is actively being worked on nor is there a timeline for its completion. 

The [link](https://www.getzola.org/documentation/content/syntax-highlighting/) you sent me to says to make a PR to add a syntax. 

> If the language you want to highlight is not on this list, please follow our [contribution guidelines](https://github.com/getzola/zola/blob/master/CONTRIBUTING.md) to add support for a new language.

It’s a small syntax I’m adding and there’s a very low chance it’s going to break anything based on the fact that we’re following an established method of adding syntax’s that hasn’t caused problems in the last. 

I’m asking for you to make the change in the interim while you figure out tree-sitter, based on the instructions in the documentation you linked me of how to add a new syntax.

Keats:
I've just updated the docs.

PR #2272, opened 2023-08-15
[docs] Add drafts to serve command in tutorial

ptxmac:
Sanity check:

  * Did a quick search for "draft"; didn't find any

If following the [tutorial](https://www.getzola.org/documentation/getting-started/overview/) as is won't work because Zola defaults to *NOT* render draft blog posts. 

This change just enables draft serving. Another solution would be to mark the posts as `draft = false`

It might also be a good idea to mention the draft concept in the guide

### DIFF
--- a/docs/content/documentation/getting-started/overview.md
+++ b/docs/content/documentation/getting-started/overview.md
@@ -70,7 +70,7 @@ Let's start the Zola development server within the newly created `myblog` direct
 
 ```bash
 $ cd myblog
-$ zola serve
+$ zola serve --draft
 Building site...
 Checking all internal links with anchors.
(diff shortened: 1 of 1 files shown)


Keats:
The tutorial doesn't seem to be using drafts?

ptxmac:
> The tutorial doesn't seem to be using drafts?

I think that's the problem - the default setup is that all posts are draft, so if you follow the tutorial you'll get a empty list of posts

ptxmac:
re-reading the docs it actually sounds that the intended behaviour is that `draft` should default to `false`. 

It appears I made a mistake locally - this is not my day; two PR's, one wrong, one already fixed 😂

PR #2284, opened 2023-08-24
docs: Update installation.md to use --locked

### LINKED ISSUE #2283: Unable to install nor build on Guix
There is no Zola package for Guix. Moreover, Guix's Rust version is at 1.68 atm. While Zola only builds with 1.70 and higher. Guix also doesn't support rustup, so there is no way to install Zola on Guix at the moment.

I would like to kindly request to somehow specify an MSRV (minimum supported Rust version) that ideally is more than 2 months old like 1.70 is right now. 
```
error: failed to compile `zola v0.17.2 (/home/steven/software/zola)`, intermediate artifacts can be found at `/var/cargo-target`

Caused by:
  package `clap_builder v4.4.0` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.68.2
  Either upgrade to rustc 1.70.0 or newer, or use
  cargo update -p clap_builder@4.4.0 --precise ver
  where `ver` is the latest version of `clap_builder` supporting rustc 1.68.2
```

stevenroose:
Fixes https://github.com/getzola/zola/issues/2283.

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -204,14 +204,14 @@ From a terminal, you can now run the following commands:
 ```sh
 $ git clone https://github.com/getzola/zola.git
 $ cd zola
-$ cargo install --path .
+$ cargo install --path . --locked
 $ zola --version
 ```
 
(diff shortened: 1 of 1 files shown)


Keats (approved):
Thanks!

PR #2286, opened 2023-08-28
fix broken link to Tera

alexrutar:
Sanity check:

## Details
The existing link to Tera in the documentation page is broken: update it to a working one. I'm not sure if this is the correct URL for the fix. I haven't created an issue for this since the fix is very minor, hopefully this is alright!


Keats (approved):
Thanks, I thought I updated all of it already :(

Issue #2298, opened 2023-09-10
Do not check the same url more than once if it is the same on different pages/languages


biodranik:
For sites with many languages, the same URLs/links are checked many times. That delays `zola check` and unnecessarily overloads servers.

Keats:
Weird, I thought we were doing that already.

biodranik:
@Keats no, I've checked the code and reproduced the issue on our [zola website](https://github.com/organicmaps/organicmaps.github.io/). Please CC me in the PR, I'm interested to learn how to do it properly in Rust :)

Keats:
I won't have time to do it myself right now but it should be easy enough to do if someone wants to try

PR #2301, opened 2023-09-12
fix: do not check the same url more than once

peterprototypes:
This is a fix for https://github.com/getzola/zola/issues/2298

Before:

DeepinScreenshot_select-area_20230912162708

After:

DeepinScreenshot_select-area_20230912162728

I've used the https://github.com/organicmaps/organicmaps.github.io/ repo for the before and after tests.

Sanity check:

## Code changes

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -179,17 +179,11 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     for link in checked_links.iter() {
         if dedup_links_by_domain.contains_key(link.domain.as_str()) {
             let deduped_links = dedup_links_by_domain.get_mut(link.domain.as_str()).unwrap();
-
-            if deduped_links.contains_key(link.external_link.as_str()) {
-                deduped_links.get_mut(link.external_link.as_str()).unwrap().push(link);
-            } else {
-                deduped_links.insert(&link.external_link, vec![link]);
-            }
+            deduped_links.entry(&link.external_link).or_insert_with(Vec::default).push(link);
         } else {
-            dedup_links_by_domain.insert(
-                link.domain.as_str(),
-                HashMap::from([(link.external_link.as_str(), vec![link])]),
-            );
+            dedup_links_by_domain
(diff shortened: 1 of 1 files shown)


components/site/src/link_checking.rs
```diff
@@ -170,53 +170,59 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
...
+                deduped_links.get_mut(link.external_link.as_str()).unwrap().push(link);
+            } else {
+                deduped_links.insert(&link.external_link, vec![link]);
+            }
         } else {
-            links_by_domain.insert(link.domain.as_str(), vec![link]);
+            dedup_links_by_domain.insert(
+                link.domain.as_str(),
+                HashMap::from([(link.external_link.as_str(), vec![link])]),
+            );
```
Keats: I think using the entry api would be nicer here

Keats (changes requested):

peterprototypes:
@Keats looks like me and @biodranik worked on the same thing. I think his solution is a bit easier to follow, as it avoids the complexity of a nested HashMap. The only difference in end result is that I changed the output to show the number of unique external links instead of the total number of links.

biodranik:
Printing the real number of _unique_ urls that are checked is a good idea. 

In my approach failures are also printed for each page where the same link was used, so you can see how often the failed link is used, and on which pages. Is it the same for your implementation?

peterprototypes:
> Printing the real number of _unique_ urls that are checked is a good idea.
> 
> In my approach failures are also printed for each page where the same link was used, so you can see how often the failed link is used, and on which pages. Is it the same for your implementation?

Yes, it's the same. I also clone the result from the first checked link to all messages.

biodranik:
Learning Rust in this way is a lot of fun! I've also updated my PR #2305, @Keats WDYT?

Keats:
I think #2305 approach is a bit neater

peterprototypes:
> I think #2305 approach is a bit neater

Agreed :+1:

PR #2305, opened 2023-09-12
Avoid unnecessary checking for already checked links

### LINKED ISSUE #2298: Do not check the same url more than once if it is the same on different pages/languages

biodranik:
Fixes #2298

For localized [Organic Maps zola site](https://github.com/organicmaps/organicmaps.github.io/) the time for `zola check` link checking has reduced from 3 minutes to less than 1 minute.

I'm new to Rust, so any ideas on how to write better code are welcome!

### DIFF
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -52,7 +52,7 @@ pub fn check_internal_links_with_anchors(site: &Site) -> Vec<String> {
             full_path.push(part);
         }
         // NOTE: This will also match _index.foobar.md where foobar is not a language
-        // as well as any other sring containing "_index." which is now referenced as
+        // as well as any other string containing "_index." which is now referenced as
         // unsupported page path in the docs.
         if md_path.contains("_index.") {
             let section = library.sections.get(&full_path).unwrap_or_else(|| {
--- a/components/site/src/link_checking.rs
+++ b/components/site/src/link_checking.rs
@@ -1,6 +1,6 @@
 use core::time;
-use std::path::Path;
-use std::{collections::HashMap, path::PathBuf, thread};
+use std::path::{Path, PathBuf};
+use std::{cmp, collections::HashMap, thread};
 
 use config::LinkCheckerLevel;
 use libs::rayon::prelude::*;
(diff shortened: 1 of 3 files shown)


components/site/src/link_checking.rs
```diff
@@ -171,7 +171,8 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
     }
 
     println!(
-        "Checking {} external link(s). Skipping {} external link(s).{}",
+        "Checking {} unique link(s) out of {} total external link(s) on all pages. Skipping {} external link(s).{}",
```
Keats: I don't know if we need the distinction between unique vs total. We can keep the current message but show the unique number

components/site/src/link_checking.rs
```diff
@@ -199,24 +200,41 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
...
+                        let mut checked_links: HashMap<&str, Option<link_checker::Result>> =
+                            HashMap::new();
                         links
                             .iter()
                             .filter_map(move |link_def| {
                                 num_links_left -= 1;
 
+                                // Avoid double-checking the same url (e.g. for translated pages).
+                                let external_link = link_def.external_link.as_str();
+                                if checked_links.contains_key(external_link) {
```
Keats: you can use `if let Some(...) = checked_links.get(external_link)` which would be cleaner

Keats (changes requested):

biodranik:
@Keats done, PTAL

Raymi306:
To fix the CI issue, you can run `cargo fmt`

biodranik:
Thanks! Also `brew install rustfmt` is needed first :) PTAL.

Keats (approved):

Keats:
Some of the code can be cleaned up a bit, I'll post a link to the commit after doing it.

biodranik:
@Keats Thanks! You can also leave suggestions if it's easier )

PR #2312, opened 2023-09-19
Simple console error message dialog in `zola serve` webpages

### LINKED ISSUE #2303: [Feature Request] Show `zola serve` errors in web browser.
I think it would be very useful to also serve a basic page with the build error instead of just print it to the CLI and stop the server. I think it would speed up zola users' workflow, and would not be *that* difficult to implement a basic error template.

I'd be interested in adding this feature, let me know if this is something that is wanted.

grantshandy:
Resolves #2303.

screenshot

This PR creates a mutable static error variable in `crate::cmd::serve`, and inserts a console-like error message dialog onto all outgoing HTML responses. The actual implementation is a little sub-optimal, but it's the best I could do considering two different threads have to share state between 4+ nested `FnMut()`s.

I tried to make the message dialog as neutral as possible, if there are any zola-specific styles it should have, let me know :)

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -28,9 +28,10 @@ use std::sync::mpsc::channel;
 use std::thread;
 use std::time::{Duration, Instant};
 
-use hyper::header;
+use hyper::http::HeaderValue;
 use hyper::server::Server;
 use hyper::service::{make_service_fn, service_fn};
+use hyper::{body, header};
 use hyper::{Body, Method, Request, Response, StatusCode};
 use mime_guess::from_path as mimetype_from_path;
 use time::macros::format_description;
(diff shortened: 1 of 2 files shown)


Keats (approved):
I like that a lot!

src/cmd/serve.rs
```diff
@@ -148,6 +165,61 @@ async fn handle_request(req: Request<Body>, mut root: PathBuf) -> Result<Respons
...
+        }
+
+        // Push the error message (wrapped in an HTML dialog box) to the end of the HTML body.
+        //
+        // The message will be outside of <html> and <body> but web browsers are flexible enough
+        // that they will move it to the end of <body> at page load.
+        let html_error = format!(
+            r#"<div style="all:revert;position:fixed;display:flex;align-items:center;justify-content:center;background-color:rgb(0,0,0,0.5);top:0;right:0;bottom:0;left:0;"><div style="background-color:white;padding:0.5rem;border-radius:0.375rem;filter:drop-shadow(0,25px,25px,rgb(0,0,0/0.15));overflow-x:auto;"><p style="font-weight:700;color:black;font-size:1.25rem;margin:0;margin-bottom:0.5rem;">Zola Build Error:</p><pre style="padding:0.5rem;margin:0;border-radius:0.375rem;background-color:#363636;color:#CE4A2F;font-weight:700;">{error_str}</pre></div></div>"#
+        );
+        bytes.extend(html_error.as_bytes());
```
Keats, line 215: Can we get some invalid HTML generated by inserting after `</html>`?

grantshandy: Yes, this way is technically invalid HTML, but it worked on all the browsers I tested it on. If that's too much of a problem I can try to remove `</html>` before appending the message box 🤔

Keats: If it works on all browsers it's fine, not the weirdest HTML they will see

Keats:
Thanks!

PR #2313, opened 2023-09-19
Fix the link to the author's site

dxvsh:
Hello! I recently found out about Zola and it seems like a great project to me! 
I was just browsing the main site and saw that the link to the author's site is broken. Its missing a `www` after `https://`. 

I'm not sure if I've updated the link at the right place but I could find no other occurrence of this address in the code.

Really sorry if I have done something wrong here. This is my first PR to a project and I'm not fully clear about the process yet :sweat_smile:.

Thank you!

Sanity check:

### DIFF
--- a/docs/templates/index.html
+++ b/docs/templates/index.html
@@ -99,7 +99,7 @@
             {% endblock content %}
         </div>
         <footer>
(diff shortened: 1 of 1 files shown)


Keats (approved):
Thanks!

Issue #1356, opened 2021-02-15
Don't render pages within a section


arthmis:
Hi I'm trying to take some markdown files that follow the same pattern and render them onto one page. I want to organize them within a section, however I don't want to render the actual markdown files, only `_index.md`. I have not found anyway to tell Zola to not render pages within a section. I'm essentially trying to take advantage of a for loop to render each page onto `index.html`.

Keats:
No way to do that right now. I don't think it's planned either.

arthmis:
Ok cool, so I'll have to remove the files after build process then and I guess from the sitemap too. I guess I can do that, there aren't too many pages. Should I keep this issue open?

Keats:
Sure

edwardwolcher:
Would adding a "render_pages" front-matter variable for sections and/or a "render" front-matter variable for pages be accepted as a pull request? 

I'd also love to be use unrendered markdown pages this way.

techware01:
Interested in this topic as well. Would be great to have a render front-matter for single pages as well. Any news on that?

Keats:
We could add `render` to pages and a `render_pages` to sections.
What would happen if `render_pages = false` and a page in it has `render = true` though? I would say ignore the page

Keats:
It's been released for a while

PR #2318, opened 2023-09-25
Prevent generating folder & index.html for a specific page when "rendes = false"

endesigner:
Adds front matter flag to prevent generation of page folder & index.html
[discussion](https://zola.discourse.group/t/do-not-render-page/584)

Sanity check:

## Code changes

---
#1356 may be related even though front matter flag that this feature provides must be applied to every page rather than whole section.


endesigner:
Any clue on the discrepancy of test runners?

Keats:
Not entirely sure

components/content/src/page.rs
```diff
@@ -115,6 +115,10 @@ impl Page {
         page.word_count = Some(word_count);
         page.reading_time = Some(reading_time);
 
+        if !page.meta.render {
```
Keats, line 118: Do we want to still process slugs etc even if we don't render things?

endesigner: Hi Keats,

No, my intention was to stop processing of slugs, paths and sitemap entries if `meta.render` is set. 

I assumed that, since `parse()` returns at this point, the job is done but your question makes me doubt that :) I'm not fluent in the codebase (or Rust for that matter) so if there are any other places where I should do the plugging please let me know.

Keats: It's mostly a question of whether we want a fully populated page, except it's not rendered which I think is better?

Keats: I think it's probably better to have the fully processed pages so if you query it in a template, you have all the info you need

Keats:
Something funky happened to the PR, can you rebase? Otherwise i'll take the commit directly on `next`

Keats:
Added in `next` branch

Issue #1319, opened 2021-01-20
Docker image: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown


src386:
It seems like it is currently not possible to build any Docker images `FROM balthek/zola`.

Dockerfile:
```
FROM balthek/zola:0.13.0 AS builder
COPY . .
RUN zola build

FROM nginx:latest
COPY --from=builder /workdir .
```

Build:
```bash
$ docker build -t mycustomimage .
Sending build context to Docker daemon  73.65MB
Step 1/5 : FROM balthek/zola:0.13.0 AS builder
 ---> 59e24c2db347
Step 2/5 : COPY . .
 ---> 1bc09f3100c4
Step 3/5 : RUN zola build
 ---> Running in 8b4ab29b3984
OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown
```

The `balthek/zola` image uses `FROM: scratch`as a base layer : https://github.com/getzola/zola/blob/9444b920d2e5a1bb3943edb49a71003c0020774c/Dockerfile#L19 so there is no shell, no binary, nothing except `zola`. It is possible to use `docker run` with volume mounts to execute `zola`, but it is currently not possible to build another image based on `balthek/zola`.

src386:
Workaround:
```
FROM debian:buster-slim
COPY --from=balthek/zola:0.13.0 /usr/bin/zola /usr/bin/zola
ADD . .
RUN zola build
```

Keats:
Isn't that the point of a Docker image for a specific tool? The leanest possible and you can still `COPY` if needed.

src386:
> Isn't that the point of a Docker image for a specific tool? The leanest possible and you can still `COPY` if needed.

IMO `scratch` based images are a pain because it makes troubleshoot harder when something does not work.

The workaround is good enough for me so I will close this issue.

oleastre:
For info, there is now an example how to use the zola image in a multi stage build without any work around.

PR #2327, opened 2023-10-11
Add multi-stage dockerfile build example

oleastre:
Sanity check:

This is a simple documentation improvement proposal to add an example how to use the zola docker image in a multi stage build. 

This should answers questions like #1319 and avoid workaround like creating derived docker images.

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -180,7 +180,8 @@ You can now browse http://localhost:8080.
 
 #### Multi-stage build
 
-Build current folder content and create a lightweight web server image using static-web-server:
+Since there is no shell in the Zola docker image, if you want to use it from inside a Dockerfile, you have to use the
+exec form of `RUN`, like:
 
 ```Dockerfile
 FROM ghcr.io/getzola/zola:v0.17.1 as zola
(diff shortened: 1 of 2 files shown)


docs/content/documentation/getting-started/installation.md
```diff
@@ -178,6 +178,22 @@ You can now browse http://localhost:8080.
...
+FROM ghcr.io/getzola/zola:v0.17.1 as zola
+
+COPY . /project
+WORKDIR /project
+RUN ["zola", "build"]
+
+FROM ghcr.io/static-web-server/static-web-server:2
+WORKDIR /
+COPY --from=zola /project/public /public
+```
```
Keats, line 192: That part should not be in the installation doc, it's running the output side in a docker container

Keats (changes requested):

docs/content/documentation/getting-started/installation.md
oleastre: Should I just remove that part ? 
Or do you think there is a better place in the documentation where to put the whole example ?
Cutting the Dockerfile just after the build is possible, but the resulting image is a bit useless in itself.

Keats: Maybe put it in the deploy section?

Keats (approved):
Thanks

Issue #1524, opened 2021-06-10
page.permalink doubles the base_url because the generated href is relative link


hbina:
# Bug Report

AFAIK, I am just following the structure provided in the tutorial.
I have something like this:

```html
{% extends "base.html" %} {% block content %}
<h1 class="title">{{ section.title }}</h1>
<p>Click <a href="althttpd.c">here</a> to see the source code</p>
<h2>Chapters</h2>
<ul>
  {% for page in section.pages | reverse %}
  <li><a href="{{ page.permalink }}">{{ page.title }}</a></li>
  {% endfor %}
</ul>
{% endblock content %}
```

But the generated HTML looks like this:

```html
<ul>
  <li><a href="hbina.github.io/althttpd/2/">Escape</a></li>
  <li><a href="hbina.github.io/althttpd/3/">tvms</a></li>
  <li><a href="hbina.github.io/althttpd/1/">Introduction</a></li>
</ul>
```

Which causes it to point to `hbina.github.io/hbina.github.io/<link>`

## Environment

Github Pages but locally it works fine.

Zola version:

```bash
hbina.github.io on  master [$] 
❯ zola --version
zola 0.13.0
```

## Expected Behavior
Tell us what should have happened.

It should generate absolute paths.

## Current Behavior
Tell us what happens instead of the expected behavior. If you are seeing an
error, please include the full error message and stack trace. You can get the
stacktrace of a panic by adding `RUST_BACKTRACE=1` when running a `zola` command.

## Step to reproduce
Please provide the steps to reproduce the issue.
If the issue is hard to reproduce, please provide a sample repository or sample
that triggers the bug.

Try to publish this repo at this commit https://github.com/hbina/hbina.github.io/tree/bdef05f33316a5edcbdfcf4652b6745cae724909

Keats:
The links are based on whatever is set in https://github.com/hbina/hbina.github.io/blob/bdef05f33316a5edcbdfcf4652b6745cae724909/config.toml#L2
It should be prefixed with http/https and we should probably have a warning/error about that.

hbina:
It works now. Thanks!

PR #2328, opened 2023-10-13
docs: show that `base_url` should contain protocol

rikhuijzer:
I was confused, like https://github.com/getzola/zola/issues/1524, by the example in the documentation for `base_url`. I also got double links on my website; just like in the issue. Showing the example with the protocol/scheme should probably make it a little bit more clear.


Keats (approved):
Thanks!

PR #2340, opened 2023-10-28
Change to updated version of checkout action

c-git:
I think we should update to the new version of checkout in the example. The current example is pinned to a very old version of checkout. I also just pinned it to version 4 instead of 4.0.0 so the documentation doesn't require updates as often to be up to date. That's also how they show it in the example in [their readme](https://github.com/actions/checkout#usage).

Small change, was easier to do than to explain what I wanted to do. Less time investment than creating an issue (to ensure in include all required context).

Sanity check:

### DIFF
--- a/docs/content/documentation/deployment/github-pages.md
+++ b/docs/content/documentation/deployment/github-pages.md
@@ -47,7 +47,7 @@ jobs:
     if: github.ref == 'refs/heads/main'
     steps:
       - name: checkout
-        uses: actions/checkout@v3.0.0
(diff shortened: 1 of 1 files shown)


Keats:
Can you do the PR on the next branch?

c-git:
Yes I can rebase it

Edit: Actually just realized that's not necessary just create a new PR.

c-git:
That's cool, I deleted the branch but there was a restore button so I could just reopen it.

PR #2344, opened 2023-10-29
Update version of checkout action in examples

c-git:
I think we should update to the new version of checkout in the example. The current example is pinned to a very old version of checkout. I also just pinned it to version 4 instead of 4.0.0 so the documentation doesn't require updates as often to be up to date. That's also how they show it in the example in [their readme](https://github.com/actions/checkout#usage).

Small change, was easier to do than to explain what I wanted to do. Less time investment than creating an issue (to ensure in include all required context).

Sanity check:

### DIFF
--- a/docs/content/documentation/deployment/github-pages.md
+++ b/docs/content/documentation/deployment/github-pages.md
@@ -46,7 +46,7 @@ jobs:
     if: github.ref == 'refs/heads/main'
     steps:
       - name: checkout
-        uses: actions/checkout@v3.0.0
+        uses: actions/checkout@v4
       - name: build_and_deploy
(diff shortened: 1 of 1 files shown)


Keats (approved):

c-git:
Hi did this one to replace [this one](https://github.com/getzola/zola/pull/2340) but aren't documentation normally done against master? That's the impression I got from the PR template.

I noticed that the version of the deployment action was updated on master but it hasn't been updated yet on this branch.

Keats:
Oh shit sorry, I thought for some reason it was the actual CI. Can you re-open https://github.com/getzola/zola/pull/2340?

c-git:
Actually I can't... I deleted the branch but I can just do a new one doesn't take long.

Issue #2341, opened 2023-10-28
Document RSS Feed Author


c-git:
# Documentation issue

## Summary
> What is the issue? Is the documentation unclear? Is it missing information?

I'm trying to set the author in the rss feed and the can't seem to find how to do it in the documentation. 

## Proposed solution

> A quick explanation of what you would like to see to solve the issue.
> If you want to add content, please explain what you were looking for 
> and what was your process while looking at the current documentation.

I'd like to see the steps required to set the author documented and easy to find.

I checked the page documenting [feeds](https://www.getzola.org/documentation/templates/feeds/) and didn't find it. Then I did a search on the documentation using the search at the top for the word "author" but only turned up results that appear to be for themes. Thus I came to the conclusion this is not documented yet.

At this point I'm mostly checking if that is the case before I try to create a PR to document it. Of corse if it does already exist please pardon my ignorance and point me in the right direction.

I am lead to believe that it does exist due to this [issue](https://github.com/getzola/zola/issues/2024) and this [PR](https://github.com/getzola/zola/pull/2092) and observing that the author is set to Unknown in the currently output feed and not empty.

I think if it goes in the extra section of the `config` and `front matter` then those would be the best places to document it by showing it with the default value and a comment explaining what it is (Like the rest of the documentation). And then a mention of it on the feed page. Possibly with a short summary, while the repetition would make it easier to find, it would cause more work to update if it changes in the future as more places would need to be updated, leading to possibly inconsistent documentation.

If it should be or is documented somewhere else please let me know.

Keats:
It should be documented yes. it looks ups in order: front-matter, config or default to unknown

c-git:
Thanks going to give it a try now. Cuz was just going to update my site.

c-git:
Turns out it is already documented for the [front matter](https://github.com/getzola/zola/blob/152eccc44751d070ae34b2805e7fa85a06c3667c/docs/content/documentation/content/page.md?plain=1#L131) but I guess this is not included in the search index I didn't see it in the results. Have to run now will finish later this evening hopefully.

c-git:
Documentation merged

PR #2347, opened 2023-10-30
Document author

c-git:

> The place to discuss new features is the forum: <https://zola.discourse.group/>
> If you want to add a new feature, please open a thread there first in the feature requests section.

Sanity check:

This adds author to the page that shows what fields are available in the configuration.

It also explains how the `author` from the configuration and `authors` from the front matter are used for feeds.

I'm on the fence about how to add author to the configuration as it is blank and not set to "Unknown" that happens in the feed template. However, the way I've done it causes the config not to work if you just copy and paste it, without editing it. I think that's probably bad but I patterned on the front matter documentation where the date and update are done the same way. I considered commenting out the line but then it's easier to miss. Not sure what is the best way to treat this.

### DIFF
--- a/docs/content/documentation/getting-started/configuration.md
+++ b/docs/content/documentation/getting-started/configuration.md
@@ -77,6 +77,9 @@ feed_filename = "atom.xml"
 # files are always copied, regardless of this setting.
 hard_link_static = false
 
+# The default author for pages
+author = 
+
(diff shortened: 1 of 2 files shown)


c-git:
I'm unable to link this PR to the related [issue](https://github.com/getzola/zola/issues/2341).

Keats (approved):

Keats:
Thanks!

PR #2372, opened 2023-12-13
Sorted subsections

bemyak:
This PR adds subsection sorting and introduces `lower` and `higher` fields to make sections more consistent with pages. As pointed [here](https://zola.discourse.group/t/next-previous-section-based-on-weight/528/2), these fields are supposed to be there.

An open question remains: how should we treat subsections of a transparent section? I raised this question [here](https://zola.discourse.group/t/should-transparant-sections-pass-their-subsections/1965), but so far there is no definite conclusion, so I'm leaving it out of the scope of this PR.

---

## Sanity check:

## Code changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.19.0 (unreleased)
+
+
 ## 0.18.0 (2023-12-18)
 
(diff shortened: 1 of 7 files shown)


Keats:
Done in the commit above. Better late than never!

PR #2386, opened 2023-12-20
fix: set the debian version used to build the Docker image

### LINKED ISSUE #2385: Docker image for 0.18.0 is broken
# Bug Report

## Environment

Zola version: 0.18.0

## Expected Behavior

The Docker image should build without failure.

## Current Behavior

The Docker image fails to build with "GLIBC not found" errors.

## Step to reproduce

My Dockerfile looks like this:

```
FROM ghcr.io/getzola/zola:v0.18.0 AS builder
WORKDIR /app
COPY . .
RUN ["zola", "build"]
```

When building it with:

```
docker build -t foo/foo .
```

It fails with:

```
 => ERROR [builder 4/4] RUN ["zola", "build"]                                                  0.2s
------
 > [builder 4/4] RUN ["zola", "build"]:
0.188 zola: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by zola)
0.188 zola: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by zola)
0.188 zola: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by zola)
```

oleastre:
This avoid conflicting GLibc version between the build step and the distroless cc one.

Resolves #2385

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:slim AS builder
+FROM rust:slim-bookworm AS builder
 
 RUN apt-get update -y && \
   apt-get install -y make g++ libssl-dev && \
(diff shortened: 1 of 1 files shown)


Keats (approved):
Thanks!

PR #2388, opened 2023-12-22
Allow not rendering pages

clarfonthey:
* [x] Have you created/updated the relevant documentation page(s)

Depends on #2407.

## `render = false`

First, this updates the meaning of `render = false`:
* Entries in the sitemap will not be included (no change)
* Files are not generated to disk (no change)
* Things are written to `SITE_CONTENT` instead of disk (the big change)

This means that `zola serve` will still render non-rendered content, although I consider this mostly a benefit to debugging, not an issue. Since `zola serve` is not intended for anything besides local development, I consider this a reasonable compromise.

The way I actually implemented this change was by simply updating `write_content` to allow writing to `SITE_CONTENT` even if it's run under `zola build`, so that non-rendered pages, sections, and taxonomies can still be accessed, but aren't actually written to disk. This helps ensure that things are more consistent (you don't have to break at the right point in a function) and hopefully things just work how people expect.

## Exposure to templates

This was another thing I added after realising that, since `render = false` doesn't affect creating permalinks at all (and probably shouldn't), it would be nice for templates to be able to detect whether a page/section is rendered so they can avoid linking things. So, templates now have access to the `render` variable, which is defaulted to true so you can just check `page.render`, etc.

## Explanation

Right now, the reasoning behind `render = false` for sections is so that people can use `get_section` or `get_page` internally without actually rendering pages for things. So, by this logic, ensuring that these rendered pages show up in `SITE_CONTENT` feels like the most straightforward way of doing things.


components/content/src/front_matter/page.rs
```diff
@@ -39,6 +39,11 @@ pub struct PageFrontMatter {
     pub datetime_tuple: Option<(i32, u8, u8)>,
     /// Whether this page is a draft
     pub draft: bool,
+    /// Whether to render the page or not. Defaults to `true`.
+    /// Useful when the page is only there to organize things but is not meant
```
Keats, line 43: what does it mean to have a page to organize things?

Keats:
Can you split the refactoring in another PR?

clarfonthey:
Just to clarify, which specific refactorings? I am fine with just merging the first two commits if that's what you mean, or I could modify the latter three commits to remove `render = false` from pages if that's what you mean.

Keats:
Basically split the render = false and the rest in 2 different PRs

clarfonthey:
Alright, I'll just start with the first two commits for now. Those are now #2407.

components/content/src/front_matter/page.rs
clarfonthey: I'm not quite sure the best way to word things. One example of something I use it for is my CV - I have jobs as different non-rendered pages in a non-rendered section and just output them directly onto the template for the parent section. This way, it's organised into different parts but this doesn't affect the final output.

clarfonthey:
Going to close this for now, will provide another PR when I come up with a better version.

PR #2395, opened 2023-12-30
Allow IPv6 serve, default `base_url` to listen interface instead of 127.0.0.1

clarfonthey:

This *could* work fine if you just replaced `SocketAddrV4` with `SocketAddr`, but I decided to tidy things up a bit more while I was looking at it. The main changes:

1. Instead of defaulting to 127.0.0.1, `zola serve` will make the `base_url` equal to the interface it's listening on by default (prevents you from doing `zola serve -i (thing) -u (same thing)`)
2. Parses `interface` explicitly as an IP address instead of just parsing `(interface):port` as a socket address. This means that you should provide IPv6 addresses without the brackets and that `base_url` will automatically add the brackets as needed.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -71,8 +71,8 @@ pub enum Command {
         force: bool,
 
         /// Changes the base_url
-        #[clap(short = 'u', long, default_value = "127.0.0.1")]
-        base_url: String,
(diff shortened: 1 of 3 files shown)


Keats (approved):
Seems ok, i'll try it a bit

PR #2397, opened 2024-01-01
Match `<!--more-->` in addition to `<!-- more -->`

clarfonthey:

I noticed this when trying to migrate a few WriteFreely posts to Zola. WriteFreely uses `<!--more-->` when added via the rich-text editor, whereas Zola was picky about the spaces around the tags. I replaced the check with a regex so that spaces and case don't matter any more.

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -665,7 +665,8 @@ mod tests {
         let bottom = "Here's the compelling conclusion.";
         let bottom_rendered = format!("<p>{bottom}</p>");
         // FIXME: would add a test that includes newlines, but due to the way pulldown-cmark parses HTML nodes, these are passed as separate HTML events. see: https://github.com/raphlinus/pulldown-cmark/issues/803
-        let mores = ["<!-- more -->", "<!--more-->", "<!-- MORE -->", "<!--MORE-->", "<!--\t MoRe \t-->"];
+        let mores =
+            ["<!-- more -->", "<!--more-->", "<!-- MORE -->", "<!--MORE-->", "<!--\t MoRe \t-->"];
         let config = Config::default();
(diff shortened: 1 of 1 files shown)


Keats:
Can you add a test?

clarfonthey:
Sorry for taking a bit to get to this-- a lot of my focus was on trying to get my particular static site setup working, and I was just submitting PRs for the changes I made that I thought would be reasonable to upstream.

I added some tests directly to the markdown parser, since it seems that most of the current summary-splitting tests are in the content itself. This way, it's a little closer to a unit test than an integration test.

Unfortunately, newline-split `<!-- more -->` would not work under this setup due to the way `pulldown-cmark` works, and I filed a bug to see if upstream would be interested in fixing this. But at least for now, allowing extra spaces between the more tags, and case-insensitivity, seems like a net improvement.

I also apologise that the test code is a little weird, adding in `trim`s instead of detecting newlines between paragraph tags, but I figured that this would be more resilient in case the parser were to change to remove those.

components/markdown/src/markdown.rs
```diff
@@ -485,7 +494,8 @@ pub fn markdown_to_html(
                     });
                 }
                 Event::Html(text) => {
-                    if text.contains("<!-- more -->") {
+                    dbg!(&text);
```
Keats: you can remove that

components/markdown/src/markdown.rs
```diff
@@ -485,7 +494,8 @@ pub fn markdown_to_html(
...
+                    if MORE_DIVIDER_RE.is_match(&text) {
```
Keats: you should check first if `has_summary` is already set to true, otherwise you might push multiple continue reading event

Keats (changes requested):

Keats (approved):

Issue #2378, opened 2023-12-18
"`feed_filename` for default language is specified twice" even though it's not


Yaroslav-95:
# Bug Report

## Environment

Zola version: 0.17.2 and master

## Expected Behavior
I have set feed_filename to a custom name "feed.xml" in my config.toml at the base or top section of the file, and haven't set it in my default language's section.

On Zola <= 0.17.1 this configuration works and generates a "feed.xml" file under each of the sections where I have `generate_feed = true`.

## Current Behavior
Zola fails to serve or build the site with the following message:
```
Warning: config.toml contains both default language specific information at base and under section `[languages.en]`, which may cause merge conflicts. Please use only one to specify language specific information
Error: Failed to serve the site
Error: `feed_filename` for default language is specified twice, as "feed.xml" and "atom.xml".
```

## Step to reproduce
Try to build a site with a custom `feed_filename` set at the top section with multiple languages with no `feed_filename` in the default language's section.

For an example of such a config.toml: https://git.yaroslavps.com/yaroslavps.com/tree/config.toml

Keats:
Duplicate of https://github.com/getzola/zola/issues/2360, the issue is the same afaik.

Yaroslav-95:
I saw that issue but I thought it wasn't related because that one talks about `generate_feed` and it only results in a warning. I also set generate_feed in my config.toml and there's no problems with it. However, feed_filename generates an error that prevents the site from building.
They do seem quite similar, so most probably you're right they are the same. I'll follow that issue, maybe even try to fix it myself if I find the time.

Keats:
Re-opening in the meantime until i look at the cause

SumDonkuS:
Hi,
I saw the [other issue](https://github.com/getzola/zola/issues/2360) was listed as a "good first issue".  I'd be excited to help out.  I'll just put together a PR [here](https://github.com/getzola/zola/pull/2400).  Please let me know if I am missing anything since this will be my first PR for this project.  Thank you again for the great project.

Keats:
Duplicate of #2360

Issue #2360, opened 2023-11-21
Unnecessary configuration or unnecessary warning with `generate_feed` in multilingual site


ogarcia:
# Bug Report

## Environment

Zola version: 0.17.2

## Current Behavior

Given this configuration.

<details>
<summary>config.toml</summary>

```toml
base_url = "https://example.com"
compile_sass = false
build_search_index = false
generate_feed = true
default_language = "es"

[markdown]
highlight_code = true

[languages.es]
generate_feed = true
title = "Título molón"
description = "Descripción molona"

[languages.en]
generate_feed = true
title = "Fancy title"
description = "Fancy description"
```

</details>

And this `content` files (optional but to generate `atom.xml`).

<details>
<summary>content/hello.md</summary>

```
+++
title = "Hola"
date = 2023-11-21
+++

Hola
```

</details>

<details>
<summary>content/hello.en.md</summary>

```
+++
title = "Hi"
date = 2023-11-21
+++

Hi
```

</details>

Running `zola serve` or `zola build` gives this warning:

> Warning: config.toml contains both default language specific information at base and under section `[languages.es]`, which may cause merge conflicts. Please use only one to specify language specific information

I know the culprit is defining `generate_feed` _globally_ outside the language blocks (`[languages.es]` and `[languages.en]`), but the problem is that if you don't define it there the feed is not generated.

## Expected Behavior

In my view, if the `generate_feed` entry is to be defined in the `[language.xx]` sections, it should not be necessary to define it globally as well.  Or at least, if it is necessary to do so then it should not give that warning.

Keats:
> I know the culprit is defining generate_feed globally outside the language blocks ([languages.es] and [languages.en]), but the problem is that if you don't define it there the feed is not generated.

Hmm that sounds like a bug for the default language.

Keats:
Yup we have a `if self.config.generate_feed` check in the code. Should be an easy fix

argon17:
I'd like to work on this as my first issue.
@Keats Should I take this, if it's not being fixed already by someone?

Keats:
Have a look at the next branch, I don't know off the top of my head. Have a look at the PR above: https://github.com/getzola/zola/pull/2400 as well which is outdated

argon17:
Hi @Keats, I spent some time understanding the codebase around this issue and have a couple of observations:

[...]

PR #2400, opened 2024-01-04
Issue 2378 lang level feed config

SumDonkuS:
Sanity check:

## Code changes

Looking at [issue 2378](https://github.com/getzola/zola/issues/2378) and the related [issue 2360](https://github.com/getzola/zola/issues/2360), found that there were a few places that needed to check the LangaugeOptions.  To handle the change for detecting the settings for `feed_filename`, the code has been switched to having this config be optional.  This also allowed for switching back to the first implementation of the `merge_field` macro. `generate_feed` checks both the section for the default language as well as the global area for its config.  If it exists in both places an error is thrown for the collision.

No documentation change for this bug fix.

This is my first PR for this project.  Please let me know if there is anything that I'm missing.  Thank you again for the great project.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -1046,7 +1046,7 @@ impl Site {
                 None => return Ok(()),
             };
 
-        for (feed, feed_filename) in feeds.into_iter().zip(self.config.feed_filenames.iter()) {
+        for (feed, feed_filename) in feeds.into_iter().zip(self.config.languages[lang].feed_filenames.iter()) {
             if let Some(base) = base_path {
                 let mut components = Vec::new();
                 for component in base.components() {
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -744,7 +744,6 @@ impl Site {
         start = log_time(start, "Rendered orphan pages");
         self.render_sitemap()?;
         start = log_time(start, "Rendered sitemap");
-
         self.render_all_feeds()?;
         self.render_themes_css()?;
         start = log_time(start, "Rendered themes css");
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -771,8 +771,10 @@ impl Site {
         for (code, language) in &self.config.languages {
             let is_default_language = code == &self.config.default_language;
 
-            let skip_default_language_feed_generation = is_default_language && !self.config.generate_feeds && !language.generate_feeds;
-            let skip_other_language_feed_generation = !is_default_language && !language.generate_feeds;
+            let skip_default_language_feed_generation =
+                is_default_language && !self.config.generate_feeds && !language.generate_feeds;
+            let skip_other_language_feed_generation =
+                !is_default_language && !language.generate_feeds;
             if skip_default_language_feed_generation || skip_other_language_feed_generation {
                 continue;
             }
@@ -784,11 +786,7 @@ impl Site {
             };
 
             let code_path = PathBuf::from(code);
-            let base_path = if is_default_language {
-                None
-            } else {
-                Some(&code_path)
-            };
+            let base_path = if is_default_language { None } else { Some(&code_path) };
 
             self.render_feeds(pages, base_path, code, |c| c)?;
             let debug_message = if is_default_language {
@@ -1063,7 +1061,9 @@ impl Site {
                 None => return Ok(()),
             };
 
-        for (feed, feed_filename) in feeds.into_iter().zip(self.config.languages[lang].feed_filenames.iter()) {
+        for (feed, feed_filename) in
+            feeds.into_iter().zip(self.config.languages[lang].feed_filenames.iter())
+        {
             if let Some(base) = base_path {
                 let mut components = Vec::new();
                 for component in base.components() {
@@ -1230,8 +1230,8 @@ fn log_time(start: Instant, message: &str) -> Instant {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use content::{FileInfo, Library, Page};
     use config::Config;
+    use content::{FileInfo, Library, Page};
     use tempfile::{tempdir, TempDir};
 
     fn create_page(title: &str, file_path: &str, lang: &str, config: &Config) -> Page {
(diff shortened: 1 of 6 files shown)


components/config/src/config/mod.rs
```diff
@@ -202,12 +203,18 @@ impl Config {
...
-                ""
-            } else {
-                "/"
-            };
+        let trailing_bit = if path.ends_with('/')
+            || self
+                .feed_filename
+                .as_ref()
+                .map(|ff| path.ends_with(ff))
+                .unwrap_or_else(|| path.ends_with(DEFAULT_FEED_FILENAME))
```
Keats: I would maybe put that as a variable above so the condition is easier to read.

components/site/src/lib.rs
```diff
@@ -1022,6 +1029,19 @@ impl Site {
         Ok(())
     }
 
+    pub fn language_feed_filename(&self, lang: &str) -> String {
```
Keats: Can we not return `&str`?

components/site/src/lib.rs
```diff
@@ -749,7 +749,13 @@ impl Site {
         start = log_time(start, "Rendered sitemap");
 
         let library = self.library.read().unwrap();
-        if self.config.generate_feed {
+        if self.config.generate_feed
+            || self
+                .config
+                .languages
+                .get(&self.config.default_language)
+                .map_or(false, |lang_opt| lang_opt.generate_feed)
```
Keats: looks like we do need better config querying based on a language like you did with `language_feed_filename`.
No need to change anything here, just thinking aloud

SumDonkuS:
Thank you for catching those things for me.  I'll keep a closer eye out for those in the future.  Let me know if the modification to querying `generate_feed` from the config looks like what you were thinking.  Thank you again.

components/config/src/config/mod.rs
SumDonkuS: Good call.  I have that one changed to keep the if statement more readable.

components/site/src/lib.rs
SumDonkuS: Let me know if the change I have is what you were thinking.

components/site/src/lib.rs
SumDonkuS: Good call.  Completely spaced that.  Thank you for the heads up.

components/config/src/config/languages.rs
```diff
@@ -18,7 +18,7 @@ pub struct LanguageOptions {
     pub generate_feed: bool,
     /// The filename to use for feeds. Used to find the template, too.
     /// Defaults to "atom.xml", with "rss.xml" also having a template provided out of the box.
```
Keats: need to change the docstring no?

components/config/src/config/mod.rs
```diff
@@ -66,7 +67,7 @@ pub struct Config {
     pub feed_limit: Option<usize>,
     /// The filename to use for feeds. Used to find the template, too.
     /// Defaults to "atom.xml", with "rss.xml" also having a template provided out of the box.
```
Keats: same here

Keats (commented):
I think it's ok. A bit iffy with Option<String> but maybe it's fine

SumDonkuS:
Thank you for spotting those docs.  They should now be modified to describe how the default values are handled.

Keats:
Note: that PR conflicts with https://github.com/getzola/zola/pull/2477, i'll have a look at rebasing it after 2477 is merged

Keats:
The other PR has been merged, do you have time to rebase it or should i do it?

Keats:
There seems to be more spots to fix than I thought. I'll release the next version of Zola first and get back to that after, otherwise I'm never going to do that release.

SumDonkuS:
Sorry for the delay.  Work pulled me away for a bit.

Keats:
I've put a tiny commit on the next branch: https://github.com/getzola/zola/pull/2547/commits/4c5cc5399b1b0f6590b6b7e964166107ed7298dd for some obvious issues.
Do you want to restart work on that PR? Probably better to start from scratch tbh

SumDonkuS:
Sorry.  Missed your comment about starting fresh after rebasing.  Let me know if you would like to start from scratch.

components/site/src/lib.rs
```diff
@@ -744,27 +744,7 @@ impl Site {
...
-
-        for (code, language) in &self.config.other_languages() {
-            if !language.generate_feeds {
-                continue;
-            }
-            let pages: Vec<_> = library.pages.values().filter(|p| &p.lang == code).collect();
-            self.render_feeds(pages, Some(&PathBuf::from(code)), code, |c| c)?;
-            start = log_time(start, "Generated feed in other language");
-        }
+        self.render_all_feeds()?;
```
Keats, line 747: This name is not accurate since there are also section feeds. `render_site_feeds`?

components/site/src/lib.rs
```diff
@@ -784,6 +764,41 @@ impl Site {
...
 
+    pub fn render_all_feeds(&self) -> Result<()> {
+        let mut start = Instant::now();
+        let library = self.library.read().unwrap();
+
+        for (code, language) in &self.config.languages {
+            let is_default_language = code == &self.config.default_language;
+
+            let skip_default_language_feed_generation =
+                is_default_language && !self.config.generate_feeds && !language.generate_feeds;
```
Keats, line 775: should that ambiguity be resolved when loading the config? Eg error if the value at the top level and in a default language level differs.
This way you can now just do something like `self.config.languages.iter().filter(|x| x.generate_feeds)` for the loop

components/site/src/lib.rs
```diff
@@ -1209,3 +1226,200 @@ fn log_time(start: Instant, message: &str) -> Instant {
     }
     now
 }
+
+#[cfg(test)]
+mod tests {
```
Keats, line 1231: Can those be moved in the tests folders? This file is already long enough without tests

Keats:
The PR is outdated. The issue is still open.

PR #2407, opened 2024-01-10
Refactor: create_directory responsibly

clarfonthey:
These are the first two commits of #2388 separated into their own PR. Specifically:

* At some point, `ensure_directory_exists` was made identical to `create_directory`. I removed the former and replaced all instances with the latter.
* A lot of methods were running `create_directory` iteratively on every component despite it using `create_dir_all` internally. I decided to update these to have fewer `create_directory` calls so they could be better tracked.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -638,7 +638,6 @@ impl Site {
         create_dirs: bool,
     ) -> Result<PathBuf> {
         let write_dirs = self.build_mode == BuildMode::Disk || create_dirs;
-        create_directory(&self.output_path)?;
 
         let mut site_path = RelativePathBuf::new();
         let mut current_path = self.output_path.to_path_buf();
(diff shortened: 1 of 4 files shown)


components/site/src/lib.rs
```diff
@@ -977,8 +970,6 @@ impl Site {
 
     /// What it says on the tin
     pub fn render_sitemap(&self) -> Result<()> {
-        ensure_directory_exists(&self.output_path)?;
```
Keats, line 980: I'd rather keep those around even if in theory it's already created elsewhere

Keats (changes requested):

components/site/src/lib.rs
clarfonthey: The reason why I removed it here is because `write_content` always creates directories when called, so, you can rely on that instead of creating them ahead of time.

This is why the only remaining instances where directories created are there, and when copying assets.

Keats: It's ok in the grand scheme of things but creates inter-dependencies in case we decide to not use write_content for some reason. Maybe it's fine.

clarfonthey: That's fair. My thought process was that we should delay creating the directory as long as possible to avoid making them unnecessarily, but I'm open to adding back in checks that are guarded on whether things will actually be written.

PR #2409, opened 2024-01-12
Add MIME type to `get_image_metadata`

clarfonthey:

This is required for OpenGraph under the `og:image:type` field, and probably other cases.

Also updated the documentation to clarify that `get_image_metadata` doesn't return the extension of the image, but the most common one for its format, so for example `image.jpeg` will show its format as `jpg`.


Keats (approved):

Keats:
Thanks!

Issue #2359, opened 2023-11-20
Fetaure Request: Add the possibility to pass the language to get_section


ogarcia:
For example, right now to get a section you call `get_section(path="blog/_index.md")` but that brings up the section with the default language. It would be very interesting if you could do `get_section(path="blog/_index.md", lang=lang)`.

I've been looking for ways to get around this but I can't think of any other. Because adding the language in a way similar to this will fail with the default language that has no code:
```
get_section(path=["blog/_index", lang, "md" | join(sep=".")])
```

SumDonkuS:
If this is of interest, I think I can put something together for the optional `lang` argument.  Looks like there was some conversation over [here](https://zola.discourse.group/t/lang-option-for-get-section/1921), but not sure if it was ever decided upon.

Keats:
You can do a PR, I don't think this is controversial

Keats:
Fixed in 0.23

PR #2410, opened 2024-01-13
Issue 2359 get section by lang

SumDonkuS:
Adding optional `lang` argument for `get_section` global function.  Language code will be added to the filename if it is not the default language.  Related issue [here](https://github.com/getzola/zola/issues/2359).

Sanity check:

## Code changes

### DIFF
--- a/components/templates/src/global_fns/content.rs
+++ b/components/templates/src/global_fns/content.rs
@@ -106,6 +106,21 @@ impl GetSection {
     pub fn new(base_path: PathBuf, library: Arc<RwLock<Library>>) -> Self {
         Self { base_path: base_path.join("content"), library }
     }
+
+    fn add_lang_to_path(path: &str, lang: &str) -> Result<String> {
+        match path.rfind('.') {
+            Some(period_offset) => {
+                let prefix = path.get(0..period_offset);
+                let suffix = path.get(period_offset..);
+                if prefix.is_none() || suffix.is_none() {
+                    Err(format!("Error adding language code to {}", path).into())
+                } else {
+                    Ok(format!("{}.{}{}", prefix.unwrap(), lang, suffix.unwrap()))
+                }
+            }
+            None => Ok(format!("{}.{}", path, lang)),
+        }
+    }
 }
 impl TeraFn for GetSection {
     fn call(&self, args: &HashMap<String, Value>) -> Result<Value> {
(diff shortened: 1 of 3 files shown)


Keats:
Thinking about it a bit more, we don't have it for `get_page` either... We should probably add it to both to be consistent and share the code when possible

components/templates/src/global_fns/content.rs
```diff
@@ -119,19 +140,39 @@ impl TeraFn for GetSection {
...
-        let full_path = self.base_path.join(&path);
-        let library = self.library.read().unwrap();
-
-        match library.sections.get(&full_path) {
-            Some(s) => {
-                if metadata_only {
-                    Ok(to_value(s.serialize_basic(&library)).unwrap())
-                } else {
-                    Ok(to_value(s.serialize(&library)).unwrap())
+        let lang =
```
Keats, line 196: We could also pass the list of available languages defined in config.toml and error if not in the list

SumDonkuS: Thank you.  Added check for configured languages.  Let me know what you think.

SumDonkuS:
Good call on adding it to `get_page` as well.  I have that added the code and tried to generalize as much code across to be used by both as possible.  Please let me know what you think.

components/templates/src/global_fns/content.rs
```diff
@@ -71,14 +72,60 @@ impl TeraFn for GetTaxonomyUrl {
...
+                Err(format!("Error adding language code to {}", path).into())
+            } else {
+                Ok(Cow::Owned(format!("{}.{}{}", prefix.unwrap(), lang, suffix.unwrap())))
+            }
+        }
+        None => Ok(Cow::Owned(format!("{}.{}", path, lang))),
+    }
+}
+
+fn calculate_path<'a>(
```
Keats: i'm not sure calculate_path is the right name. `get_path_with_lang`? Not sure

docs/content/documentation/templates/overview.md
```diff
@@ -154,6 +160,12 @@ If you only need the metadata of the section, you can pass `metadata_only=true`
 {% set section = get_section(path="blog/_index.md", metadata_only=true) %}
 ```
 
+If selecting a specific language for the section, you can pass `lang` with the language code to the function:
```
Keats, line 169: I would mention for pages and here that it's equivalent to `{% set section = get_section(path="blog/_index.fr.md") %}`

SumDonkuS:
Great.  Changed the function name to be more specific, like you suggested.  Also, added to the docs equivalencies based on if the language passed is the default language or not.

Keats (approved):

Keats:
Thanks!

PR #2419, opened 2024-01-22
Refactor: add BuildMode::Skip

clarfonthey:

**Note: this is rebased on top of #2418, and that should be merged first.**

This is an attempt at incorporating the refactoring done in #2388 without changing behaviour. It helps pave a path toward `render = false` for pages while making things a bit more consistent for the existing `render = false` for sections and taxonomies.

Effectively, `render = false` will still add things to the library, but their pages and aliases won't be rendered. We could potentially convert having `aliases` for something `render = false` to a hard error, but at least for now, it makes no sense to create links to something that isn't actually rendered.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -676,8 +676,19 @@ impl Site {
         Ok(current_path)
     }
 
-    fn copy_asset(&self, src: &Path, dest: &Path) -> Result<()> {
-        copy_file_if_needed(src, dest, self.config.hard_link_static)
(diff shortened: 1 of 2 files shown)


clarfonthey:
Actually, going to close this first since I realise it needs a bit more work. After looking more into how the code works, `get_page` and `get_section` don't even go through `SITE_CONTENT`, so, we probably shouldn't rely on using `BuildMode` and instead just not running the `write_content` method.

This was the original justification for adding `BuildMode::SKIP` (since it bypasses `SITE_CONTENT`) but after going through it more, that's probably not the right choice.

PR #2428, opened 2024-01-28
Refactor: remove even more unnecessary create_directory

clarfonthey:

These are some things that were missed even by #2407.

After doing some more investigation, even copying files will automatically create directories, so, we don't need these calls either. It simplifies the code a bit more and will also ensure that directories aren't created unless files are actually inside them.

Another reason for this change is it also makes calls to `write_content` and `write_section` a bit less confusing, since right now, if multiple calls to it are made, the final argument is set to `false` to avoid doing work, which can make following the code a bit more challenging since it's not always clear when this argument is applied.

----

One small change here: this adds the same create-parent logic to `create_file` as `copy_file` and `copy_file_if_needed`, so that these three methods are consistent. Part of the reason why these were necessary before was because directories had to be manually created sometimes, and this was a bit inconsistent, so, this PR fixes that.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -635,20 +635,13 @@ impl Site {
         components: &[&str],
         filename: &str,
         content: String,
-        create_dirs: bool,
     ) -> Result<PathBuf> {
-        let write_dirs = self.build_mode == BuildMode::Disk || create_dirs;
-
         let mut site_path = RelativePathBuf::new();
         let mut current_path = self.output_path.to_path_buf();
 
         for component in components {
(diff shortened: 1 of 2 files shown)


Keats:
Looks like legit CI failures?

clarfonthey:
They were, and sorry for taking so long to get back to this. Essentially, I thought that since these directories were always being made for copied files, they were also made for created files, which was not the case. I decided to go ahead and update things so that this is the case, but I understand if you'd rather not merge this PR because of that. While I personally think it makes the code simpler, that's not really an objective thing, just my opinion.

PR #2326, opened 2023-10-10
GitHub-Style Footnotes

DerDrodt:
This PR adds the option of rendering footnotes at the bottoms in GitHub-Style, based on the implementation of [notriddle](https://github.com/raphlinus/pulldown-cmark/blob/ad19879e6b4f411dbdd78ffa6a9cf429e35afe47/examples/footnote-rewrite.rs).

Closes #1285.

For instance, the following markdown

```md
### Footnotes

Footnote 1 link[^first].

Footnote 2 link[^second].

Duplicated footnote reference[^second].

[^first]: Footnote **can have markup**.

[^second]: Footnote text.
```
Gets rendered to:
```html
<h3 id="footnotes">Footnotes</h3>
<p>Footnote 1 link<sup class="footnote-reference" id="fr-first-1"><a href="#fn-first">[1]</a></sup>.</p>
<p>Footnote 2 link<sup class="footnote-reference" id="fr-second-1"><a href="#fn-second">[2]</a></sup>.</p>
<p>Duplicated footnote reference<sup class="footnote-reference" id="fr-second-2"><a href="#fn-second">[2]</a></sup>.</p>
<hr><ol class="footnotes-list"><li id="fn-first" class="footnote-definition">
<p>Footnote <strong>can have markup</strong>. <a href="#fr-first-1">↩</a></p>
</li>
<li id="fn-second" class="footnote-definition">
<p>Footnote text. <a href="#fr-second-1">↩</a> <a href="#fr-second-2">↩2</a></p>
</li>
</ol>
```

This style is disabled by default and can be turned off via the `markdown.bottom_footnotes` option in the `config.toml`.

PR #2432, opened 2024-02-06
Update pulldown_cmark dep to v0.10, and add pulldown_cmark_escape dep.

timonvo:
The pulldown_cmark escaping functionality is now shipped in a separate
pulldown_cmark_escape crate
(https://crates.io/crates/pulldown-cmark-escape), starting with v0.10.0.

The markdown.rs module has to be adapted to a few API changes in
pulldown_cmark, and we have to introduce explicit handling of  alt
text to ensure it continues to be properly escaped.

There are also a few other behavior changes that are caught by the
tests, but these actually seem to be desired, so I've updated the insta
snapshot files for those tests to incorporate those changes.

Sanity check:

Yes, but see the related https://github.com/getzola/zola/pull/2326#issuecomment-1929315804.

## Code changes

   --> N/A

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -4,19 +4,20 @@ use errors::bail;
 use libs::gh_emoji::Replacer as EmojiReplacer;
 use libs::once_cell::sync::Lazy;
 use libs::pulldown_cmark as cmark;
+use libs::pulldown_cmark_escape as cmark_escape;
 use libs::tera;
 use utils::net::is_external_link;
 
 use crate::context::RenderContext;
 use errors::{Context, Error, Result};
-use libs::pulldown_cmark::escape::escape_html;
+use libs::pulldown_cmark_escape::escape_html;
 use libs::regex::{Regex, RegexBuilder};
 use utils::site::resolve_internal_link;
 use utils::slugs::slugify_anchors;
 use utils::table_of_contents::{make_table_of_contents, Heading};
 use utils::types::InsertAnchor;
 
-use self::cmark::{Event, LinkType, Options, Parser, Tag};
+use self::cmark::{Event, LinkType, Options, Parser, Tag, TagEnd};
 use crate::codeblock::{CodeBlock, FenceSettings};
 use crate::shortcode::{Shortcode, SHORTCODE_PLACEHOLDER};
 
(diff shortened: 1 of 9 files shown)


components/markdown/src/markdown.rs
```diff
@@ -254,6 +255,7 @@ pub fn markdown_to_html(
     let mut error = None;
 
     let mut code_block: Option<CodeBlock> = None;
+    let mut alt_text: bool = false;
```
Keats: can you add a comment for what this is? No need for the type as well

components/markdown/tests/snapshots/markdown__all_markdown_features_integration.snap
```diff
@@ -83,7 +81,7 @@ line 1 of code
 line 2 of code
 line 3 of code
 </code></pre>
-<p>Block code &quot;fences&quot;</p>
+<p>Block code "fences"</p>
```
Keats, line 84: how come it's not escaping anymore here?

components/markdown/tests/snapshots/markdown__can_handle_heading_ids.snap
```diff
@@ -22,6 +20,6 @@ expression: body
 <h1 id="text-there">text <sup class="footnote-reference"><a href="#1">1</a></sup> there</h1>
 <div class="footnote-definition" id="1"><sup class="footnote-definition-label">1</sup>
 <p>footnote</p>
-<h1 id="classes" class="bold another">Classes</h1>
 </div>
+<h1 id="classes" class="bold another">Classes</h1>
```
Keats, line 24: so it was inserting the header in the footnote?

components/markdown/tests/snapshots/markdown__all_markdown_features_integration.snap
timonvo: I believe this is due to a behavior change in pulldown-cmark, rather than the changes I made in this pull request. I think it might be https://github.com/pulldown-cmark/pulldown-cmark/pull/830, specifically. It introduces a new `escape_html_body_text` which should be used for escaping text nodes and which only escapes '<', '>', and '&', as opposed to `escape_html` which should be used for escaping HTML attributes and which escapes single and double quotes as well.

I believe that pulldown-cmark's `push_html` will now convert Event::Text nodes to HTML using `escape_html_body_text` (see src/html.rs in that pull request), hence why we see these quotes not being escaped anymore. This is also why we need to handle the alt attribute's text manually now, to ensure it gets attribute-escaped rather than body-escaped (which is now the default).

And FWIW, all other existing uses of `escape_html` that I could find in Zola were for escaping HTML attributes, so I think none of them need to be switched to `escape_html_body_text`.

components/markdown/tests/snapshots/markdown__can_handle_heading_ids.snap
timonvo: Yeah, seems like it. pulldown-cmark 0.10 contains a bunch of footnotes-related fixes, which I'm guessing are responble for this change. Since the new version seems to be correct and the old version seemed wrong, I didn't investigate further to figure out exactly which commit fixed this.

components/markdown/src/markdown.rs
timonvo: Ah yes, good call. I've actually renamed the variable to `inside_attribute` as well, to be a little more descriptive.

timonvo:
Thanks for the review, I think this is ready for another round. Note that I also added a note to the CHANGELOG.md file to draw attention to the minor behavior changes people may see after this change.

Keats (approved):

Keats:
Thanks!

Issue #1840, opened 2022-04-28
Feature Request: Provide `transparent` option value to templates in the `section` object


ararunaufc:
# Bug Report

## Environment

Zola version: 0.15.3

## Expected Behavior
I described the problem I had in more detail [here](https://zola.discourse.group/t/how-to-find-if-a-section-is-set-to-transparent/1284).

In a nutshell, I'd like to have access to the value provided to the `transparent` option of a section when in a template.

## Current Behavior
As a workaround, I'm duplicating the option in the `[extra]`.

## Step to reproduce
_(Not applicable)_

PR #2462, opened 2024-03-19
Fix section.transparent serialization

wold5:
Fixes passing `section.transparent` for https://github.com/getzola/zola/commit/e8b04bb11a9464d515b47bd76bd732d7e83d6965 and https://github.com/getzola/zola/issues/1840

I cobbled a test together from the examples, but suspect it only checks for a section having `transparent` set internally, and not a template having access to `{{ section.transparent }}`... Since both rely on `meta.transparent` anyway, it's perhaps best to leave it at this.

If you have any remarks, let me know.

```
#[test]
fn check_transparancy() {
    let (mut site, _tmp_dir, _public) = build_site("test_site");
    site.load().unwrap();

    let mut path = env::current_dir().unwrap().parent().unwrap().parent().unwrap().to_path_buf();
    path.push("test_site");
    let posts_path = path.join("content").join("posts");
    let transparent_section_path = posts_path.join("2018");

    let library = site.library.read().unwrap();
    let section = library.sections.get(&transparent_section_path.join("_index.md")).unwrap();   
    assert_eq!(section.meta.transparent, true);
}
```

### DIFF
--- a/components/content/src/ser.rs
+++ b/components/content/src/ser.rs
@@ -161,6 +161,7 @@ pub struct SerializingSection<'a> {
     translations: Vec<TranslatedContent<'a>>,
     backlinks: Vec<BackLink<'a>>,
     generate_feed: bool,
+    transparent: bool,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


Keats (approved):
Thanks

Issue #2412, opened 2024-01-17
zola check: checking vs checked link number does not match


Raymi306:
# Bug Report
```
Checking 18 external link(s). Skipping 0 external link(s).
> Checked 19 external link(s): 0 error(s) found.
```

## Environment

Zola version: 0.18

## Expected Behavior

I'd think that the numbers ought to be the same.  
As an aside and reminder for myself to explore later, it also seems to take a bit longer than I would expect, I wrote an async link checker in Python that typically checked many more links in the time Zola took.

## Current Behavior

18 != 19

## Step to reproduce

I can provide my repo at a later date, I am actively overhauling things and just noticed this and thought I'd ask if it was expected

Keats:
That is weird.

yhoiseth:
Tried on my own with 0.18.0:

```
yngve@UT-C0D3E-027 zola-website % zola check
Checking site...
Checking all internal links with anchors.
> Successfully checked 15 internal link(s) with anchors.
Checking 102 external link(s). Skipping 0 external link(s).
> Checked 122 external link(s): 18 error(s) found.
Error: Failed to check the site
Error: Found 18 broken external link(s)
  1. Broken link in…
```

amtanq:
I can pick this up if it's still available

ta-nishq:
Raised a fix: https://github.com/getzola/zola/pull/2472

PR #2472, opened 2024-04-03
Fix link check report inconsistency

ta-nishq:
Fixes: https://github.com/getzola/zola/issues/2412

## Root Cause
In the first log we are removing duplicate links by using HashSet but in the second log we were counting all links

## Tests
Skipped tests cause this is a logging issue and does not affects the functionality


Keats (approved):
Thanks!

Issue #2083, opened 2023-01-21
Both RSS and Atom feeds?


orlp:
Is there a way to support having both an Atom and an RSS feed at the same time?

Keats:
Not right now no

dennisse:
You can create a new template to achieve this.

Just create a template for a feed, and use the `get_section`-function (https://www.getzola.org/documentation/templates/overview/#get-section) to get the root section and use it. Then create a new post/page under `content`, give it the correct path of your new rss-feed, and set its template to the new template.

Here is an example-template for a new atom-feed, `feed.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="{{ lang }}">
    <title>{{ config.title }}
        {%- if term %} - {{ term.name }}
    {%- elif section.title %} - {{ section.title }}
        {%- endif -%}
    </title>
    {%- if config.description %}
    <subtitle>{{ config.description }}</subtitle>
    {%- endif %}
    {%- if not feed_url is defined %}
        {%- set feed_url = page.permalink %}
    {% endif %}
    <link href="{{ feed_url | safe }}" rel="self" type="application/atom+xml"/>
  <link href="
      {%- if section -%}
        {{ section.permalink | escape_xml | safe }}
      {%- else -%}
        {{ config.base_url | escape_xml | safe }}
      {%- endif -%}
    "/>
    <generator uri="https://www.getzola.org/">Zola</generator>
    <updated>{{ last_updated | default(value=now()) | date(format="%+") }}</updated>
    <id>{{ feed_url | safe }}</id>
    {%- if not pages is defined %}
        {%- set getpages = get_section(path="_index.md") %}
        {%- set pages = getpages.pages %}
    {%- endif %}
    {%- for p in pages %}
    {%- if not p.relative_path is starting_with("pages/") %}
    <entry xml:lang="{{ p.lang }}">
        <title>{{ p.title }}</title>
        <published>{{ p.date | date(format="%+") }}</published>
        <updated>{{ p.updated | default(value=p.date) | date(format="%+") }}</updated>
        <link rel="alternate" href="{{ p.permalink | safe }}" type="text/html"/>
        <id>{{ p.permalink | safe }}</id>
        <content type="html">{{ p.content }}</content>
    </entry>
    {%- endif %}
    {%- endfor %}
</feed>
```
And then I create the page `content/pages/feed.xml.md` with this content:
```
+++
[...]

PR #2477, opened 2024-04-19
Added support for multiple feeds (i.e. generating both Atom and RSS)

LunarEclipse363:
## Introduction
As requested many times before, for example:
- https://zola.discourse.group/t/generate-both-rss-and-atom-feeds/441
- #2083

## Potential improvements
~~I don't know if/how Zola handles deprecating config options, if there's anything I should add to let the user know the feed related options changed please let me know how to do it.~~
Implemented backwards-compatibility as suggested by @selfisekai 

## Formalities
**Sanity check:**

**Code changes**

**If the change is a new feature or adding to/changing an existing one:**

### DIFF
--- a/components/config/src/config/languages.rs
+++ b/components/config/src/config/languages.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::config::search;
 use crate::config::taxonomies;
+use crate::config::might_be_single;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(default)]
@@ -15,9 +16,11 @@ pub struct LanguageOptions {
     /// Description of the site. Defaults to None
     pub description: Option<String>,
     /// Whether to generate feeds for that language, defaults to `false`
+    #[serde(alias = "generate_feed")]
     pub generate_feeds: bool,
     /// The filenames to use for feeds. Used to find the templates, too.
     /// Defaults to ["atom.xml"], with "rss.xml" also having a template provided out of the box.
+    #[serde(deserialize_with = "might_be_single")]
     pub feed_filenames: Vec<String>,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
     /// Whether to generate search index for that language, defaults to `false`
--- a/components/config/src/config/languages.rs
+++ b/components/config/src/config/languages.rs
@@ -4,9 +4,9 @@ use errors::{bail, Result};
 use libs::unic_langid::LanguageIdentifier;
 use serde::{Deserialize, Serialize};
 
+use crate::config::might_be_single;
 use crate::config::search;
 use crate::config::taxonomies;
-use crate::config::might_be_single;
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(default)]
@@ -20,7 +20,7 @@ pub struct LanguageOptions {
     pub generate_feeds: bool,
     /// The filenames to use for feeds. Used to find the templates, too.
     /// Defaults to ["atom.xml"], with "rss.xml" also having a template provided out of the box.
-    #[serde(deserialize_with = "might_be_single")]
+    #[serde(alias = "feed_filename", deserialize_with = "might_be_single")]
     pub feed_filenames: Vec<String>,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
     /// Whether to generate search index for that language, defaults to `false`
@@ -69,7 +69,8 @@ impl LanguageOptions {
         merge_field!(self.title, other.title, "title");
         merge_field!(self.description, other.description, "description");
         merge_field!(
-            self.feed_filenames.is_empty(),
+            self.feed_filenames.is_empty()
+                || self.feed_filenames == LanguageOptions::default().feed_filenames,
             self.feed_filenames,
             other.feed_filenames,
             "feed_filename"
--- a/components/config/src/config/languages.rs
+++ b/components/config/src/config/languages.rs
@@ -14,11 +14,11 @@ pub struct LanguageOptions {
     pub title: Option<String>,
     /// Description of the site. Defaults to None
     pub description: Option<String>,
-    /// Whether to generate a feed for that language, defaults to `false`
-    pub generate_feed: bool,
-    /// The filename to use for feeds. Used to find the template, too.
-    /// Defaults to "atom.xml", with "rss.xml" also having a template provided out of the box.
-    pub feed_filename: String,
+    /// Whether to generate feeds for that language, defaults to `false`
+    pub generate_feeds: bool,
+    /// The filenames to use for feeds. Used to find the templates, too.
+    /// Defaults to ["atom.xml"], with "rss.xml" also having a template provided out of the box.
+    pub feed_filenames: Vec<String>,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
     /// Whether to generate search index for that language, defaults to `false`
     pub build_search_index: bool,
@@ -66,9 +66,9 @@ impl LanguageOptions {
         merge_field!(self.title, other.title, "title");
         merge_field!(self.description, other.description, "description");
         merge_field!(
-            self.feed_filename == "atom.xml",
-            self.feed_filename,
-            other.feed_filename,
+            self.feed_filenames.is_empty(),
+            self.feed_filenames,
+            other.feed_filenames,
             "feed_filename"
         );
         merge_field!(self.taxonomies.is_empty(), self.taxonomies, other.taxonomies, "taxonomies");
@@ -79,7 +79,7 @@ impl LanguageOptions {
             "translations"
         );
 
-        self.generate_feed = self.generate_feed || other.generate_feed;
+        self.generate_feeds = self.generate_feeds || other.generate_feeds;
         self.build_search_index = self.build_search_index || other.build_search_index;
 
         if self.search == search::Search::default() {
@@ -101,8 +101,8 @@ impl Default for LanguageOptions {
         LanguageOptions {
             title: None,
             description: None,
-            generate_feed: false,
-            feed_filename: "atom.xml".to_string(),
+            generate_feeds: false,
+            feed_filenames: vec!["atom.xml".to_string()],
             taxonomies: vec![],
             build_search_index: false,
             search: search::Search::default(),
@@ -129,8 +129,8 @@ mod tests {
         let mut base_default_language_options = LanguageOptions {
             title: Some("Site's title".to_string()),
             description: None,
-            generate_feed: true,
-            feed_filename: "atom.xml".to_string(),
+            generate_feeds: true,
+            feed_filenames: vec!["atom.xml".to_string()],
             taxonomies: vec![],
             build_search_index: true,
             search: search::Search::default(),
@@ -140,8 +140,8 @@ mod tests {
         let section_default_language_options = LanguageOptions {
             title: None,
             description: Some("Site's description".to_string()),
-            generate_feed: false,
-            feed_filename: "rss.xml".to_string(),
+            generate_feeds: false,
+            feed_filenames: vec!["rss.xml".to_string()],
             taxonomies: vec![],
             build_search_index: true,
             search: search::Search::default(),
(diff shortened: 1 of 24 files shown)


components/config/src/config/languages.rs
```diff
@@ -14,11 +14,11 @@ pub struct LanguageOptions {
     pub title: Option<String>,
     /// Description of the site. Defaults to None
     pub description: Option<String>,
-    /// Whether to generate a feed for that language, defaults to `false`
-    pub generate_feed: bool,
-    /// The filename to use for feeds. Used to find the template, too.
-    /// Defaults to "atom.xml", with "rss.xml" also having a template provided out of the box.
-    pub feed_filename: String,
+    /// Whether to generate feeds for that language, defaults to `false`
+    pub generate_feeds: bool,
```
selfisekai, line 18: this will be a breaking change, imo should be marked `#[serde(alias = "generate_feed")]` for backwards compat

components/config/src/config/languages.rs
```diff
@@ -14,11 +14,11 @@ pub struct LanguageOptions {
...
+    /// The filenames to use for feeds. Used to find the templates, too.
+    /// Defaults to ["atom.xml"], with "rss.xml" also having a template provided out of the box.
+    pub feed_filenames: Vec<String>,
```
selfisekai, line 21: `feed_filename = "rss.xml"` should still be accepted too for compat. in another project i've solved this with a `serde(deserialize_with)` macro: https://codeberg.org/selfisekai/electron_tasje/src/commit/3d34593c358a1d68741f3c3d0554447f33d9ec16/src/config.rs#L74-L102

LunarEclipse363:
Ah I just realized I need to run `cargo test --all` for all the tests to run locally, gonna look into what's wrong.

LunarEclipse363:
Forgot rustfmt :sweat_smile:

LunarEclipse363:
I've tested this for my own website and generating two feeds seems to work without a problem, both at the root level of the website and for taxonomies.

It should be ready to merge now unless there's some issues I couldn't find.

jamwil:
> Forgot rustfmt 😅

@LunarEclipse363 Story of my life.

LunarEclipse363:
For reference, this PR is ready to merge, the force-push was just a rebase where the only conflict was in the changelog.

The CI failing now is unrelated to the PR, it looks like some dependency increased its MSRV:
https://dev.azure.com/getzola/zola/_build/results?buildId=3278&view=logs&jobId=bad3c584-618a-56e1-e8ff-fa893be575d7&j=bad3c584-618a-56e1-e8ff-fa893be575d7&t=15b64e0a-658a-5cf8-fb48-72caeb8060e1
```
error: package `clap_derive v4.5.4` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.71.1
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap_derive@4.5.4 --precise ver
where `ver` is the latest version of `clap_derive` supporting rustc 1.71.1
```

Keats:
Can you rebase on the next branch? I've fixed the CI except for Windows

components/config/src/config/languages.rs
```diff
@@ -14,11 +15,13 @@ pub struct LanguageOptions {
...
-    pub generate_feed: bool,
-    /// The filename to use for feeds. Used to find the template, too.
-    /// Defaults to "atom.xml", with "rss.xml" also having a template provided out of the box.
-    pub feed_filename: String,
+    /// Whether to generate feeds for that language, defaults to `false`
+    #[serde(alias = "generate_feed")]
+    pub generate_feeds: bool,
+    /// The filenames to use for feeds. Used to find the templates, too.
+    /// Defaults to ["atom.xml"], with "rss.xml" also having a template provided out of the box.
+    #[serde(alias = "feed_filename", deserialize_with = "might_be_single")]
```
Keats: the `might_be_single` name is not great

components/config/src/config/mod.rs
```diff
@@ -395,6 +400,48 @@ impl Default for Config {
...
+    T: DeserializeOwned,
+    D: Deserializer<'de>,
+{
+    let v = MightBeSingle::deserialize(deserializer)?;
+    Ok(v.into())
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub(crate) enum MightBeSingle<T> {
```
Keats: If we do support that, we need to add a deprecation message when there is only one value. I'm not entirely we need to though, it's a pretty simple change to do from the user POV.

As for name... `SingleOrVec`?

components/content/src/front_matter/section.rs
```diff
@@ -68,8 +68,8 @@ pub struct SectionFrontMatter {
     #[serde(skip_serializing)]
     pub aliases: Vec<String>,
     /// Whether to generate a feed for the current section
-    #[serde(skip_serializing)]
-    pub generate_feed: bool,
+    #[serde(skip_serializing, alias = "generate_feed")]
```
Keats: you don't need it here i think

components/site/src/feeds.rs
```diff
@@ -73,18 +73,23 @@ pub fn render_feed(
...
 
-    let feed_filename = &site.config.feed_filename;
-    let feed_url = if let Some(base) = base_path {
-        site.config.make_permalink(&base.join(feed_filename).to_string_lossy().replace('\\', "/"))
-    } else {
-        site.config.make_permalink(feed_filename)
-    };
+    let mut feeds = Vec::new();
+    for feed_filename in &site.config.feed_filenames {
+        let mut context = context.clone();
```
Keats, line 78: That's pretty wasteful.

Can you render them one by one and just updating `context.insert("feed_url", &feed_url);` in the loop? I can't remember what `additional_context_fn(context);` does though

Keats (changes requested):

components/config/src/config/mod.rs
LunarEclipse363: The problem I ran into while testing before adding the backwards-compatibility is that Zola doesn't error out if you leave the old value names in the config file, instead using default values, which is a very bad way to handle it. The user will think they have the configuration option set while it's not until they realize it's broken purely due to a single-letter rename.

I can change the name and add a deprecation message, sure.

components/site/src/feeds.rs
LunarEclipse363: `render_template()` takes ownership of the context so I need to clone it on every loop regardless

components/content/src/front_matter/section.rs
LunarEclipse363: Zola will ignore unknown fields here as well though which again ends up with the "suddenly an option stopped working without an error" problem

Which I'm not sure is solvable within serde

components/config/src/config/mod.rs
LunarEclipse363: Update: adding a meaningful deprecation message seems to be much harder than I expected due to the constraints of how serde works

Keats:
I would be tempted to just do a breaking change if there's no easy way to indicate that the single name is deprecated.

LunarEclipse363:
> I would be tempted to just do a breaking change if there's no easy way to indicate that the single name is deprecated.

I would recommend against that. Unless there is a way to make it a hard error to use the old syntax, it will mean that suddenly Zola quietly stops generating feeds for existing projects, which would be terrible UX.

I actually ran into this while testing - Zola just ignores the old configuration keys without an error if there is no compatibility mechanism.

Would adding `#[serde(deny_unknown_fields)]` to the config struct work? This would make it a breaking change, but also a clear error, so users would know action is required instead of Zola silently failing to render feeds.

I'm not sure if that would allow us to get a good error message specific to the deprecated fields though, I would need to look into what kind of error the deserializer returns and if it allows checking for this sort of information.

Keats:
> Would adding #[serde(deny_unknown_fields)] to the config struct work? 

It would be nice yes. The migration guide can be added to the changelog

LunarEclipse363:
This is more complicated than it needs to purely because there is one test that apparently requires that you can put arbitrary keys into the front-matter instead of only into an `[extra]` section.

Anyway, this only adds a special deprecation message for the front matter (I figured out how to do it in the end, it's not pretty) and relies on `#[serde(deny_unknown_fields)]` for the config file to avoid adding like 5 unnecessary types into the parsing code or sharing them between modules based on "this has the same error text" rather than "this is the same thing".

Very much a breaking change now, and I imagine `#[serde(deny_unknown_fields)]` might have unforeseen consequences for some users.

Anyway, everything is in the changelog so I guess it's fine.

Keats:
> This is more complicated than it needs to purely because there is one test that apparently requires that you can put arbitrary keys into the front-matter instead of only into an [extra] section.

Which test is that?

components/content/src/front_matter/section.rs
```diff
@@ -67,13 +68,23 @@ pub struct SectionFrontMatter {
     /// redirect to this
     #[serde(skip_serializing)]
     pub aliases: Vec<String>,
+    /// Deprecated
+    #[serde(skip_serializing)]
+    pub generate_feed: Deprecated<DeprecatedGenerateFeed>,
```
Keats: :+1:

Keats (approved):

LunarEclipse363:
> > This is more complicated than it needs to purely because there is one test that apparently requires that you can put arbitrary keys into the front-matter instead of only into an [extra] section.
> 
> Which test is that?

I don't remember off the top of my head but if you add `#[serde(deny_unknown_fields)]` to the section front matter struct, it will fail.

Keats:
> This is more complicated than it needs to purely because there is one test that apparently requires that you can put arbitrary keys into the front-matter instead of only into an [extra] section.

That test was invalid, you can just do:

```
-        f.write_all(b"+++\nslug=\"hey\"\n+++\n").unwrap();
+        f.write_all(b"+++\n+++\n").unwrap();
```

`#[serde(deny_unknown_fields)]` does find a few other issues so that's nice to have

Keats (approved):
Nice, thanks

apiraino:
@LunarEclipse363 while I appreciate the flexibility that this patch brings to Zola, I think that this might catch users by surprise:
1) A mention in the changelog does not make it very explicit that it is a breaking change
2) If a zola website does not have anything configured for `feed_filenames`, the new default (`["atom.xml"]`) will just break the site building (see #2561 and other commits issue linked to this patchset)
3) CIs could auto-update to the latest Zola version and website just stop building, because ...
4) ... some themes *might* use the old syntax to get the feed filename from the config

I think the correct way to handle this could have been:
1) make this retrocompatible (example: handle the old case of `feed_filenames` as a string and internally make it an array of 1 item)
2) announce the breaking change more prominently in the changelog and the github release so people with websites and themes authors are aware and can upgrade their configs without rush

To be clear: I appreciate the time you took to deliver this patch, just my .2 cents on what I think could be the fallout here

LunarEclipse363:
> I think the correct way to handle this could have been:
> 
>1. make this retrocompatible (example: handle the old case of `feed_filenames` as a string and internally make it an array of 1 item)
> 
>2. announce the breaking change more prominently in the changelog and the github release so people with websites and themes authors are aware and can upgrade their configs without rush

1. I originally took the time to use a custom deserializer and implement backwards-compatibility (with the help of @selfisekai), however @Keats asked me to rewrite this to be a breaking change. There was also a WIP version of this which had better error messages but the code was really ugly because `serde` isn't designed to accomodate deprecating a field. By ugly I mean like 20 lines of code repeated for every deprecated field, where the actual message was 1 line.

2. I have no influence over how Zola releases are handled but I agree

LunarEclipse363:
Regarding earlier versions that explored other solutions:

> I originally took the time to use a custom deserializer and implement backwards-compatibility 

see commit 2dae1d1d07cda99e7c654508e6a02387461fce95

> There was also a WIP version of this which had better error messages but the code was really ugly

see commit 56cb421080c9e759d0313bdc6e87c4a75c128035, more specifically [here](https://github.com/getzola/zola/commit/56cb421080c9e759d0313bdc6e87c4a75c128035#diff-cdc44264544e1413a2a74f7594b84d601a4c6c265ab515bdb46c441175d52007R25-R48) and [there](https://github.com/getzola/zola/commit/56cb421080c9e759d0313bdc6e87c4a75c128035#diff-4b7773a787675bbab913aaea8fe63da12f6f5e65055c2dc163527cff2aaef8a2R71-R87)

Keats:
I think in that case the breaking change was ok since old usage would error out.
I'll try to add back the breaking section for the next releases (like in https://github.com/getzola/zola/blob/master/CHANGELOG.md#breaking)

PR #2480, opened 2024-04-21
Implemented bottom footnotes with backreferences

### LINKED ISSUE #1285: Markdown footnotes enhancement
# Enhancement

## Environment

Zola version: 0.12.2 compiled from git

## Have footnotes link back to where they were footnoted

It would be great if there was a way to have a link-back to the location of the footnote. So if the site visitor is reading an article, they can click the footnote and then click the little return icon to go back to where the footnote was declared.

totikom:
Added option `bottom_footnotes`, which renders footnotes with backreferences.
It is basically a copy-paste from [this](https://github.com/pulldown-cmark/pulldown-cmark/blob/master/pulldown-cmark/examples/footnote-rewrite.rs) example from `pulldown-cmark`.
Fixes #1285 
Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change? (There are no open PR for this feature)

## Code changes


components/markdown/src/markdown.rs
```diff
@@ -239,6 +241,158 @@ fn get_heading_refs(events: &[Event]) -> Vec<HeadingRef> {
     heading_refs
 }
 
+fn fix_github_style_footnotes(old_events: &mut Vec<Event>) {
```
Keats: What do we need to fix?

Keats (changes requested):
This adds a complex function, it needs to have tests

components/markdown/src/markdown.rs
totikom: Renamed to `convert_footnotes_to_github_style()`.

components/markdown/src/snapshots/markdown__markdown__tests__def_before_use.snap
```diff
@@ -0,0 +1,55 @@
+---
```
Keats, line 1: Can we actually do integration testing? Rather than testing the AST events, it would be better to compare the output html I think?

totikom: Yep, sure. I can pass the vector of events to ` cmark::html::push_html()` and compare against generated htm.

I've changed `insta` tests to html for better readability, today I'll Add integration tests.

totikom: @Keats ready for review.

Keats: Sorry I've just realised: how does this interact with the new footnote feature from pulldown-cmark 0.10? It relies on it right?

totikom: > Sorry I've just realised: how does this interact with the new footnote feature from pulldown-cmark 0.10? It relies on it right?

Well, not really.  The new feature affects the syntax of the footnotes, while my PR changes the parsed event stream.

As far as I understand, the new feature of `pulldown-cmark` is not required for this PR to work.

Keats: Yep, it's just that I believe the new feature is turned on already but I guess it's just a change in the markdown that doesn't affect the events.

totikom: Yes, they explicitly [state](https://github.com/pulldown-cmark/pulldown-cmark/blob/1cfaf02e516d24fd8c93f39b422548610c541aae/pulldown-cmark/examples/footnote-rewrite.rs#L17), that old footnote syntax is also supported.

docs/content/documentation/getting-started/configuration.md
```diff
@@ -135,6 +135,9 @@ smart_punctuation = false
 # For example, `![xx](...)` is ok but `![*x*x](...)` isn’t ok
 lazy_async_image = false
 
+# Whether footnotes are rendered in the GitHub-style (at the bottom, with back references) or plain.
+bottom_footnotes = false
```
apiraino: This bit of documentation (or the flag name?) felt slightly ambiguous before using it. What is it meant by _plain_ here?

IIUC whether this flag is true or false, footnotes are _always_ rendered at the bottom, right? The only differences I notice is that the format (e.g. text[1] or text<sup>1</sup>) and if there a backlink or not.

In both cases it seems to work just great. Thanks @totikom !

components/markdown/src/markdown.rs
```diff
@@ -239,6 +241,158 @@ fn get_heading_refs(events: &[Event]) -> Vec<HeadingRef> {
...
+    let filtered_events = events.into_iter().filter_map(|event| {
+        match event {
+            // New footnote definition is pushed to the stack
+            Event::Start(Tag::FootnoteDefinition(_)) => {
+                footnote_bodies_stack.push(vec![event]);
+                None
+            }
+            // The topmost footnote definition is popped from the stack
+            Event::End(TagEnd::FootnoteDefinition) => {
+                // unwrap will newer fail, because Tag::FootnoteDefinition start always comes before
```
apiraino:
```suggestion
                // unwrap will never fail, because Tag::FootnoteDefinition start always comes before
```

docs/content/documentation/getting-started/configuration.md
totikom: Have you tried reordering footnotes and text?
```markdown
[^2]: this footnote should appear the 2nd

Some text[^1] referencing[^2] footnotes[^3].

Another text mentioning footnotes[^2].

[^3]: this footnote should appear 3th
[^1]: this footnote should appear 1st
```

apiraino: oh interesting, I didn't try that! So, _plain_ here means "render footnotes where they are in source Markdown file". Maybe this could be worded a bit more explicit ion the comment. Or maybe it's just me, in this case ignore this comment :slightly_smiling_face:

totikom: That's how it renders with my feature enabled:
image
And that's how it renders with my feature disabled:
image

totikom: BTW, paragraphs in footnotes looks like an issue, but it is out of scope for this PR.

Keats (approved):

Keats:
Thanks!

Issue #2448, opened 2024-03-03
Enhance test suite around `zola serve`


jamwil:
Setting this up to discuss/track the addition of more thorough endpoint testing around the `serve` command. Better test coverage will support some refactoring efforts, including #2446 along with a better url path mounting implementation as discussed in #2311.

jamwil:
Apologies for the lack of action on this so far. It's still on my radar—energy level be damned.

jamwil:
Just marking here that the unit tests introduced in #2484 seem to fail intermittently in Windows CI. Will leave this issue and try to identify what's causing that.

jamwil:
Hmm. So I finally got around to testing these on a Windows box. All green. Unfortunately, this is CI-specific.

Edit: One thing I can try is to copy the test site into a temporary directory on each test invocation. That should, in theory, fix any concurrency issue.

PR #2311, opened 2023-09-18
Fix --base-url improper path and protocol handling using `zola serve`

jamwil:
Correctly parse the protocol and path from the base url, and differentiate between the bound interface and the base URL when using `zola serve`.

Closes #2170.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [x] Have you created/updated the relevant documentation page(s)?

PR #2482, opened 2024-04-25
Restore trailing slash behaviour in serve command

jamwil:
This is a follow on to #2311 which restores the switch that will strip the trailing slash from the base url when calling `zola serve` if there is not trailing slash on the base url as defined in the site configuration.

I noticed this discrepancy as I was beginning to prepare tests for #2448. I believe this now aligns with the prior behaviour.

Sanity check:

## Code changes

* [x] Have you created/updated the relevant documentation page(s)? (N/A)

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -398,7 +398,7 @@ fn create_new_site(
 
     let mut constructed_base_url = construct_url(&base_url, no_port_append, interface_port);
 
-    if !site.config.base_url.ends_with("/") {
+    if !site.config.base_url.ends_with("/") && constructed_base_url != "/" {
         constructed_base_url.truncate(constructed_base_url.len() - 1);
     }
 
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -396,7 +396,11 @@ fn create_new_site(
         |u| u.to_string(),
     );
 
-    let constructed_base_url = construct_url(&base_url, no_port_append, interface_port);
+    let mut constructed_base_url = construct_url(&base_url, no_port_append, interface_port);
+
+    if !site.config.base_url.ends_with("/") {
+        constructed_base_url.truncate(constructed_base_url.len() - 1);
+    }
 
     site.enable_serve_mode();
     site.set_base_url(constructed_base_url.clone());


jamwil:
Howdy @Keats! I'm not sure why it doesn't let me request a review, but I think this is good to go.

src/cmd/serve.rs
```diff
@@ -396,7 +396,11 @@ fn create_new_site(
         |u| u.to_string(),
     );
 
-    let constructed_base_url = construct_url(&base_url, no_port_append, interface_port);
+    let mut constructed_base_url = construct_url(&base_url, no_port_append, interface_port);
+
+    if !site.config.base_url.ends_with("/") && constructed_base_url != "/" {
```
Keats, line 401: hm what does this do in practice? Do we want it?

jamwil: Haha, to be honest I was hoping you might know better than I.

It restores the logic here: <https://github.com/getzola/zola/commit/4bf67afd875b99d2bad6fe252d6fb6e44f52b313#diff-05643f021a24da1c9475eb37428b799095c0cfbd79f1b79e658c39615746315dL357-L361>

This never came up as I was testing #2311 as it doesn't actually seem to have any effect on the generated site. All the permalinks render as one might expect whether you strip that trailing slash or not. EXCEPT, in one case that I've been able to spot so far.

On my site, using the Anpu theme for a base, the `config.toml` calls for a few navigation links that are specified like so:

```toml
base_url = "https://jamwil.com"

# snip

[extra]
anpu_menu_links = [
    { url = "$BASE_URL/about/", name = "About" },
    { url = "$BASE_URL/categories/", name="Categories" },
    { url = "$BASE_URL/tags/", name = "Tags" },
]
```

On *these links*, and only these links it seems, if that logic does not exist to strip the trailing slash to match `base_url` in the site config, then the links render with an extra slash, e.g., `http://127.0.0.1:1111//about`.

***

To verify that the logic here matches the prior behaviour, I took the new unit tests that I've begun preparing in #2448 and grafted them onto the parent commit of 4bf67af, and the applicable tests pass.

I admit I find the logic around trailing slashes in zola to be a bit intractable. It seems we expect them throughout, but we *do not* expect it on the base URL as specified in `config.toml`, so this little snippet allows the `serve` logic to match the `build` logic.

I'm super curious what your thoughts are vis-à-vis trailing slashes in general and how they should be treated. Perhaps my understanding is incorrect, or perhaps this is more indicative of the theme constructing URLs in a non-idiomatic way? There may be some opportunity to clean things up beyond this PR, but this one is merely to match the behaviour on the `next` branch to that on `master` so there are no surprises in the next release.

Keats: In theory every URL in zola should have a trailing slash. I'm guessing we should probably remove one from the base_url when loading the config if there is one to ensure it is uniform and remove a few checks?

jamwil: That seems like a logical approach to me. Thanks for merging!

Issue #2421, opened 2024-01-22
JPEG XL support for image processing?


veluca93:
I admit I might be a bit partial in asking for this ;) 
Now that Safari supports decoding JPEG XL images, would it make sense to add support for producing JPEG XL images with `resize_image`?

Keats:
Is it supported in Chrome? I thought they removed it but i haven't looked at it in ages

veluca93:
Unfortunately not...
I personally hope that that will change in the not-so-far future, but who knows...
For now, I'd be happy with reaching Safari (i.e. all iOS devices) :-)

Keats:
But is there a poin adding a format that doesn't work on the most used browser for a static site generator?

veluca93:
> But is there a poin adding a format that doesn't work on the most used browser for a static site generator?

It can be used with `<picture>` tags to send JXL images to those browsers that do support it (i.e. ~all iOS users, which is a significant fraction of the mobile users), reducing bandwidth usage for them while still sending JPG/PNG images to browsers that don't.

At least, that's how I was planning to use it :-) 

If you think it's enough of a use case, I'd be willing to send a PR to add support - if not, since I *am* not quite impartial on this topic :P, I will not insist further ;)

Keats:
I think that's fine to add as long as the library works on linux/mac/windows

PR #2488, opened 2024-05-01
Add JPEG XL support to image processing.

veluca93:
As discussed in #2421.

Sanity check:

## Code changes


components/imageproc/src/processor.rs
```diff
@@ -69,6 +70,44 @@ impl ImageOp {
                 };
                 buffered_f.write_all(memory.as_bytes())?;
             }
+            Format::JXL(q) => {
+                let mut encoder = jpegxl_rs::encoder_builder();
+                if let Some(q) = q {
+                    if q == 100 {
+                        encoder.lossless(true);
+                    } else {
+                        // See `JxlEncoderDistanceFromQuality` for the formula.
```
Keats: maybe they can expose https://libjxl.readthedocs.io/en/latest/api_encoder.html#_CPPv429JxlEncoderDistanceFromQualityf and add a method to the builder? They can't expect everyone to copy paste the formula...

Keats (changes requested):
Can you add some tests? They are in imageproc/tests

components/imageproc/src/processor.rs
veluca93: I made a PR for that: https://github.com/inflation/jpegxl-rs/pull/48

I assume it will take some time for a new release. How do you think we should handle this in the meantime?

veluca93:
> Can you add some tests? They are in imageproc/tests

Absolutely - what do you think I should test? For now I just added a PNG to JXL test.

components/imageproc/src/processor.rs
Keats: Let's keep the formula and we'll see when it gets released

Keats:
Maybe a JXL output with a JXL input? The CI failures look legit though

veluca93:
> Maybe a JXL output with a JXL input? The CI failures look legit though

Ah, JXL *reading* is not supported :D do you think I should add that?

As for the CI... I believe I need to add the system dependency, but I have no idea how to configure that - any pointers?

Keats:
> Ah, JXL reading is not supported :D do you think I should add that?

It depends whether it can be added or not. If so then yes

> As for the CI... I believe I need to add the system dependency, but I have no idea how to configure that - any pointers?

It looks like there is a `vendored` feature for statically linking it

veluca93:
> > Ah, JXL reading is not supported :D do you think I should add that?
> 
> It depends whether it can be added or not. If so then yes
> 
> > As for the CI... I believe I need to add the system dependency, but I have no idea how to configure that - any pointers?
> 
> It looks like there is a `vendored` feature for statically linking it

Did both of those. Also, my pr to jpegxl-rs was accepted and a release should be coming soon, so perhaps waiting a day or two and switching to the next version of jpegxl-rs would be the ideal solution.

In case you are wondering why I did not use the "image" feature of jpegxl-rs: it depends on a different version of "image", and trying to upgrade the dependency for zola to 0.25 produced some dependency incompatibility errors.

Keats:
> it depends on a different version of "image", and trying to upgrade the dependency for zola to 0.25 produced some dependency incompatibility errors.

I will update to 0.25 soonish, but I haven't looked at the image feature of jpeg-xl-rs. If that's better, I can update it now and you rebase your PR on it?

veluca93:
> > it depends on a different version of "image", and trying to upgrade the dependency for zola to 0.25 produced some dependency incompatibility errors.
> 
> I will update to 0.25 soonish, but I haven't looked at the image feature of jpeg-xl-rs. If that's better, I can update it now and you rebase your PR on it?

If you can do that, it'd be great!

Looks like there are some build issues with jpegxl-rs, so it'll be a while before this PR can be merged anyway :-)

Keats:
`next` has been updated with image 0.25

veluca93:
Thanks! I updated the PR accordingly (also squashing the commits, since I was going to rebase anyway...)

Keats:
Can you rebase? CI is fixed minus windows which i'll look at later

veluca93:
> Can you rebase? CI is fixed minus windows which i'll look at later

I made a mess with rebasing, but it should be done now :-)

components/imageproc/src/format.rs
```diff
@@ -6,12 +6,14 @@ const DEFAULT_Q_JPG: u8 = 75;
...
-    /// JPEG, The `u8` argument is JPEG quality (in percent).
+    /// JPEG, The `u8` argument is JPEG quality (1..100).
     Jpeg(u8),
     /// PNG
     Png,
-    /// WebP, The `u8` argument is WebP quality (in percent), None meaning lossless.
+    /// WebP, The `u8` argument is WebP quality (1..100), None meaning lossless.
     WebP(Option<u8>),
+    /// JPEG XL, The `u8` argument is quality (1..100), None meaning lossless.
+    JXL(Option<u8>),
```
Keats: Let's call it `JpegXL`, it's going to look less weird than JXL and refer to `jpeg_xl` in the code.

Keats (changes requested):

components/imageproc/src/format.rs
veluca93: Done!

Keats (approved):

Keats:
CI failures seem legit

veluca93:
> CI failures seem legit

Yep, they are, I am hoping they will be resolved on the jpegxl_rs side (https://github.com/inflation/jpegxl-rs/pull/50) and jxl side (https://github.com/inflation/jpegxl-rs/issues/51) as soon as I can get to it :-)

Keats:
Closing until there is a library that is less complex to build on supported platforms.

PR #2269, opened 2023-08-13
vi serve change detection fix

Raymi306:
**IMPORTANT: Please do not create a Pull Request adding a new feature without discussing it first.**

The place to discuss new features is the forum: <https://zola.discourse.group/>
If you want to add a new feature, please open a thread there first in the feature requests section.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [x] Have you created/updated the relevant documentation page(s)?

---

Fixes #2266 

A little bit of code duplication going on.
I made sure to leave a helpful comment explaining the extra branch.

Side note:  I noticed that notify library is 2 major versions behind.  Would you be interested in PRs for bumping major versions of dependencies, or don't fix what isn't broken?

Issue #2077, opened 2023-01-17
Update notify dependency


Keats:
v5 has been released and it involves some changes.

The main one is that with the debounced version you do not get the type of event which is probably fine since we don't really care about that.

We do want to only react to a single file in the debounce time though: we should get all the changes to print them but then only do the minimum required. 
Eg if 3 Sass files have been changed, we should print those 3 files that have changed but only recompile Sass once.

If a Sass file and a Content file have changed, we can skip the Sass compilation since it will be done when rebuilding the content.

Kidsan:
I would like to have a go at this but I'm not sure what approach to take for collecting the received messages. My understanding is that the watcher will send a message for each individual file changed. Should we log the changed files and cache these received messages, and then evaluate that cache separately or is there an easier approach that I am not seeing

Keats:
Yeah I'm not entirely sure to be honest. Maybe we can write our own debouncer that handles that rather than use the notify mini debouncer?

0xpr03:
If you mean "report only once per interval": debouncer-mini does that by default (otherwise it's a bug). There is also debouncer-full if you need more features.

Keats:
Can we get a list of all the debounced events in the interval? We want to do some additional filtering after, eg if 10 different templates files have been modified, we want to reload the templates only once.

0xpr03:
It should report all timed-out events in one interval. So that sounds to me like what you want.
But that also means, if you have a timeout of 2s and events come in at 0,1,2,3s; you end up with events at delivered at 2s(event at 0s), 4s(1,2) and after 6s(3). Hope that helps.

Not sure if some "cooldown" window could be added / helps more.

iamorphen:
I played with this a bit, and the implementation gets messy if we try and be smart about handling a batch of events minimally. We're lucky if the batch of events requires a full site rebuild, because then we can let the rebuild cover all other events. If we have multiple events and none require a site rebuild, we have to add logic to make sure that the union of all changes required is applied correctly, and its this logic that gets messy due to a variety of checks and sensitive ordering (ex: check for full rebuild cases, then partial update cases, etc.).

I have a local implementation based on `notify-debouncer-full`. I have not read the code of `notify-debouncer-mini`, but if it is sensitive to the same events that `-full` is, including [opening and closing files](https://docs.rs/notify/6.1.1/notify/event/enum.AccessKind.html), we probably don't want `-mini` because we can't get at the event type and therefore filter those events out.

[...]

PR #2498, opened 2024-05-16
Fix hot reloading for `config.toml` in `zola serve` on Linux

### LINKED ISSUE #2266: zola serve not reloading when config.toml is changed
# Bug Report

`zola serve` does not detect changes to config.toml, but says it should

## Environment
Linux
Zola version: 0.17.2

## Expected Behavior

If config.toml gets written to, zola serve should reload

## Current Behavior
Tell us what happens instead of the expected behavior. If you are seeing an
error, please include the full error message and stack trace. You can get the
stacktrace of a panic by adding `RUST_BACKTRACE=1` when running a `zola` command.

Nothing happens

## Step to reproduce
Please provide the steps to reproduce the issue.
If the issue is hard to reproduce, please provide a sample repository or sample
that triggers the bug.

```
zola init
zola serve
vi config.toml
# edit a variable and save
```

iamorphen:
# Introduction

Fixes #2266 , alternative to #2269 which appears to be stalled and has a large diff against `next`. In particular, this fixes hot reloading for changes made to `config.toml` via the caching strategy used by editors such as vim on Linux. At the time of this writing, changes written to `config.toml` via at least vim and neovim on Linux are not noticed at all by `zola serve`.

## Code changes

1. Watch `config.toml`'s containing folder--not `config.toml` directly--for changes. See Details for why.
2. Each item to watch tracks whether to watch it recursively.
3. Ignore changes to files peer to `config.toml`, assuming that all other files we care about are nested more deeply than the config file. This is the case at the time of this writing, and the Zola docs suggest that `config.toml` is the only top-level file we watch in a zola project.
4. When announcing the files we're watching, replace the config file parent directory with the config file name to not confuse the end user, since we ignore all ordinary files peer to `config.toml`.

## Details

Editors such as vi, vim, and neovim cache changes to swap files. When a user wants to write to the original file, such editors may delete the original file and replace it with the swap file. This changes the inode of the "original" file. The file-watching strategy on Linux is inotify which watches for changes via inodes, not file names. So when an editor like vim deletes the file we're watching, we lose event information on that file after the final delete event. If we instead watch the owning directory, we can observe events on all file changes and with some filtering pretend like we're watching the same `config.toml` file all along.

The reason this bug is not observed on macOS is because macOS uses FSEvents to watch for file system changes. FSEvents monitors on the directory level. The change here makes behavior on Linux approximate that of FSEvents.

zola uses Notify v4, and `notify::watcher` returns a `RecommendedWatcher`. The recommended watcher on macOS [is `FsEventWatcher`](https://docs.rs/notify/4.0.17/x86_64-apple-darwin/notify/type.RecommendedWatcher.html) whereas it [is `INotifyWatcher`](https://docs.rs/notify/4.0.17/x86_64-unknown-linux-gnu/notify/type.RecommendedWatcher.html) on Linux.

## Testing

Due to where the ignore logic is done in `serve.rs:serve`, we can't really extend the tests in the same module to check this feature.

I have manually checked `zola serve` on Linux and macOS by editing `config.toml` and theme files and observing hot reloading. I also edited `theme.toml` at the top-level of a zola project, and such changes were correctly ignored.

I don't have a Windows platform to test on.

## UI

Due to the watch list remapping, `zola serve` still announces the following to the end user (note `config.toml` even though we watch the parent directory).

```
Listening for changes in /foo/bar/{config.toml,content,static,templates}
Press Ctrl+C to stop
```

### DIFF
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -96,6 +96,12 @@ pub fn filter_events(
             continue;
         }
 
+        // Ignore ordinary files peer to config.toml. This assumes all other files we care
+        // about are nested more deeply than config.toml or are directories peer to config.toml.
+        if path != config_path && path.is_file() && path.parent() == config_path.parent() {
+            continue;
+        }
+
         let (change_k, partial_p) = detect_change_kind(root_dir, &path, config_path);
         meaningful_events.insert(path, (partial_p, simple_kind.unwrap(), change_k));
     }
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -46,7 +46,6 @@ use notify_debouncer_full::{new_debouncer, notify::RecursiveMode, notify::Watche
 use ws::{Message, Sender, WebSocket};
 
 use errors::{anyhow, Context, Error, Result};
-use pathdiff::diff_paths;
 use site::sass::compile_sass;
 use site::{Site, SITE_CONTENT};
 use utils::fs::{clean_site_output_folder, copy_file};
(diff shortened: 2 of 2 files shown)


iamorphen:
CI shows a compilation error in `notify` on Windows. I can't personally debug that locally. It seems to be a `notify`-internal issue. I could try to bump the `notify` version zola uses and see if that fixes the problem. I can try that in a separate draft MR.

Keats:
For notify, there's https://github.com/getzola/zola/issues/2077 and it looks like it's on v6 now. We will need to upgrade before the next release anyway so if you have time that would be great

Keats (approved):

iamorphen:
My hacked up experiment with `notify-debouncer-full` in #2499 passed all CI. I'll see about making the implementation production-worthy.

iamorphen:
For posterity, the reason the Windows build issue popped up seems to be because `libc` in the CI run is 0.2.154 whereas other successful CI runs ([ex](https://dev.azure.com/getzola/zola/_build/results?buildId=3288&view=logs&j=d2980ad8-b87d-5294-4faf-0af1145b0486&t=4895aa0f-ffeb-520a-bc7b-cf61e25f8fa6)) use 0.2.151. The versions of `winapi` and `notify` otherwise seem identical across recent CI runs. I found `master` to use 0.2.151, but something in `next` moved us to `0.2.154`.

iamorphen:
I rebased and re-tested locally on Linux. I'm just one data point, but all's working on my end.

src/fs_utils.rs
```diff
@@ -96,6 +96,12 @@ pub fn filter_events(
             continue;
         }
 
+        // Ignore ordinary files peer to config.toml. This assumes all other files we care
+        // about are nested more deeply than config.toml or are directories peer to config.toml.
+        if path != config_path && path.is_file() && path.parent() == config_path.parent() {
```
iamorphen, line 101: I added an additional condition here, `path.is_file()`. Presumably we could be watching something like `sass/` peer to `config.toml`, and a user could remove the directory for some reason and then add it back later, all while `zola serve` is watching. We don't want to skip such changes, so I filter on regular files, not directories.

Keats:
Works on mac as well

Keats:
Thanks!

iamorphen:
Sorry, was away for a bit, just wanted to say thank you also for your time here!

PR #2499, opened 2024-05-17
experiment: hack in notify-debouncer-full to test Windows build

iamorphen:
# Introduction

This MR is not meant to be merged. The implementation is a hack to test whether upgrading `notify*` will fix the Windows build CI issue encountered in #2498 . If this fixes the CI issue, I will overhaul the implementation and open a different MR to land the `notify` version bump.

(Currently the implementation works locally on Linux but is too sensitive, often not refreshing the site correctly because we react to multiple events too quickly. Implementation overhaul should explore batching.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -773,6 +773,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.12"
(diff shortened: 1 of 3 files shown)


iamorphen:
Passed all CI and served its purpose, so will close.

PR #2503, opened 2024-05-20
feat: move from notify v4 to notify-debouncer-full v0.3

### LINKED ISSUE #2077: Update notify dependency

iamorphen:
# Introduction

Closes #2077 , unblocks #2498 .

This updates our `notify` dependency from v4 to `notify-debouncer-full` v0.3. The newer `notify` libraries behave significantly differently from the version zola currently uses, so we need to introduce more logic to generally preserve the current end-user experience behind the scenes. The most significant difference is that `serve` now needs to react to a [`Vec<DebouncedEvent>`](https://docs.rs/notify-debouncer-full/latest/notify_debouncer_full/type.DebounceEventResult.html) instead of a [single `DebouncedEvent`](https://docs.rs/notify/4.0.17/notify/fn.watcher.html).

# Code Changes

1. Use `notify-debouncer-full`, which adds debouncing functionality peripheral to the newer `notify` v6. This functionality was built directly into v4.
2. Map fine-grained `notify` file system event kinds into our own local abstraction for ease of use.
3. Introduce `get_relevant_event_kind` to map the vector of `notify` file system events to our abstraction and filter out events we aren't interested in. This moves previously inlined filtering logic into its own function. The new function lives in a new library, `components/fs_event_utils`.
4. Sort the incoming events by timestamp, descending. Add the events into a hash map where the key is the event path. This way we keep only the latest event for each unique path. This helps in cases where editors like Vim raise multiple events (which don't seem to be debounced) on a single path.
5. Munge the collection in (4) into a new map that bins changes by change type.
6. Iterate over the change map from (5) to process the group of events. This means we can make multiple changes to the site in rapid succession if the user changes a lot of files within the debounce period.
7. `reload_sass` and `reload_templates` now pass an absolute path to the `sass` and `templates` directories, respectively, to `rebuild_done_handling` to participate in livereload. To support this, the `Site` struct now stores the sass and template directories as fields.

# Testing

## Unit Tests

This MR introduces a new unit test to test the new internal support function listed above. Run with `cargo test --package fs_event_utils`.

## Integration Testing

I created the following script to rapidly modify `test_site`.

<details>
<summary>Expand for script.</summary>

```
# Config
echo " " >> test_site/config.toml
echo " " >> test_site/config.toml

# Content
echo " " >> test_site/content/root-page-1.md
echo " " >> test_site/content/posts/fixed-slug.md

# Templates
echo " " >> test_site/templates/index.html
echo " " >> test_site/templates/page.html

# Static files
echo " " >> test_site/static/site.css
echo " " >> test_site/static/scripts/hello.js

# SASS
echo " " >> test_site/sass/blog.scss
echo " " >> test_site/sass/scss.scss

# Themes
echo " " >> test_site/themes/sample/theme.toml
echo " " >> test_site/themes/sample/sass/sample.scss
```

</details>

Here is the result of running the test on `zola` built on `next`.

<details>
<summary>Expand for results on next.</summary>

```
Change detected @ 2024-05-19 20:15:53
-> Config changed. The browser needs to be refreshed to make the changes visible.
-> Creating 35 pages (1 orphan) and 12 sections
Done in 87ms.

Change detected @ 2024-05-19 20:15:54
-> Content changed /Users/orphen/code/zola/test_site/content/root-page-1.md
-> Creating 35 pages (1 orphan) and 12 sections
Done in 46ms.

Change detected @ 2024-05-19 20:15:54
-> Content changed /Users/orphen/code/zola/test_site/content/posts/fixed-slug.md
-> Creating 35 pages (1 orphan) and 12 sections
Done in 45ms.

Change detected @ 2024-05-19 20:15:54
-> Template changed /Users/orphen/code/zola/test_site/templates/index.html
Reloading only template
Done in 20ms.

Change detected @ 2024-05-19 20:15:54
-> Template changed /Users/orphen/code/zola/test_site/templates/page.html
Reloading only template
Done in 20ms.

Change detected @ 2024-05-19 20:15:54
-> Static file changed /Users/orphen/code/zola/test_site/static/site.css
Done in 0ms.

Change detected @ 2024-05-19 20:15:54
-> Static file changed /Users/orphen/code/zola/test_site/static/scripts/hello.js
Done in 0ms.

Change detected @ 2024-05-19 20:15:54
-> Sass file changed /Users/orphen/code/zola/test_site/sass/blog.scss
Done in 0ms.

Change detected @ 2024-05-19 20:15:54
-> Sass file changed /Users/orphen/code/zola/test_site/sass/scss.scss
Done in 0ms.

Change detected @ 2024-05-19 20:15:54
-> Themes changed.
-> Creating 35 pages (1 orphan) and 12 sections
Done in 44ms.

Change detected @ 2024-05-19 20:15:54
-> Themes changed.
-> Creating 35 pages (1 orphan) and 12 sections
Done in 45ms.
```

</details>

The config changes are rolled into a single update, and the content, template, static, sass, and theme changes are all processed individually (2 changes each).

Cumulative time to update is ~310 ms.

Here is the result of running the test on this feature branch.

<details>
<summary>Expand for test result on this branch.</summary>

```
Change detected @ 2024-05-20 09:10:48
-> Template file(s) changed /Users/orphen/code/zola/test_site/templates/index.html, /Users/orphen/code/zola/test_site/templates/page.html
Reloading only template
Done in 49ms.

Change detected @ 2024-05-20 09:10:48
-> Static file changed /Users/orphen/code/zola/test_site/static/scripts/hello.js
-> Static file changed /Users/orphen/code/zola/test_site/static/site.css
Done in 0ms.

Change detected @ 2024-05-20 09:10:48
-> Themes changed.
-> Creating 35 pages (1 orphan) and 12 sections
Done in 54ms.

Change detected @ 2024-05-20 09:10:48
-> Sass file(s) changed /Users/orphen/code/zola/test_site/sass/blog.scss, /Users/orphen/code/zola/test_site/sass/scss.scss
Done in 1ms.

Change detected @ 2024-05-20 09:10:48
-> Config changed. The browser needs to be refreshed to make the changes visible.
-> Creating 35 pages (1 orphan) and 12 sections
Done in 43ms.

Change detected @ 2024-05-20 09:10:48
-> Content changed /Users/orphen/code/zola/test_site/content/posts/fixed-slug.md
-> Creating 35 pages (1 orphan) and 12 sections
-> Content changed /Users/orphen/code/zola/test_site/content/root-page-1.md
-> Creating 35 pages (1 orphan) and 12 sections
Done in 87ms.
```

</details>

(The order of events is different, subject mostly to OS event scheduling and whatever the backing `notify` library is at the mercy of.) Template changes are batched into a single update, sass changes are batched into a single update, themes changes are batched into a single update, the config changes are still rolled into 1, and the static files and content are processed individually (looped over).

Cumulative time to update is ~235 ms. This single test is by no means a reliable profiling, but the results are not surprising, since we perform fewer operations overall relative to current behavior on `next`.

## Manual Tests

I modified my own personal site using zola built from this branch; in particular, I edited files with Vim. This is what revealed the need for `debounce_events`.

# Alternatives

It is possible we perform multiple site rebuilds in one debounce period, for example by modifying the config and modifying the theme. I investigated adding an optimization that would skip all remaining event processing in a debounce period if some change requires a site rebuild. While doable, this handling greatly complicates the logic in my opinion.

First, we can no longer simply `match` on `ChangeKind` because we would need to prioritize handling of particular events. This is mostly done by a branch-order-matters `if` that first checks for config changes, then template changes, then etc. In the case that we don't do a site rebuild in a debounce period, we have to make sure we correctly apply all required updates to account for each change.

If we do a site rebuild, we also need a clean way to end event processing early. We can do this with labeled `match` cases and `break` to those labels, but we'd also need to account for printing the time taken to apply site changes. Time-tracking and breaking at various points during the event handling makes understanding control flow more difficult.

To keep the user-facing experience similar with what's on `master`, we would also need to let the end-user know which files we observed changing, and the reporting of such becomes non-trivial (ex: if config changed, print usual config-changed message and then list all other files; but if templates changed, print usual templates-changed message and then list all other files; but if you don't rebuild the site and only do partial updates , then...).

I don't know whether end users are often applying many site changes in a single debounce window. I personally do not. I don't think pursuing this optimization is worth it from an ROI-to-end-user and maintainability perspective.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -22,6 +22,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use std::cell::Cell;
+use std::collections::HashMap;
 use std::fs::read_dir;
 use std::future::IntoFuture;
 use std::net::{IpAddr, SocketAddr, TcpListener};
@@ -44,7 +45,7 @@ use libs::globset::GlobSet;
 use libs::percent_encoding;
 use libs::relative_path::{RelativePath, RelativePathBuf};
 use libs::serde_json;
-use notify::{watcher, RecursiveMode, Watcher};
+use notify_debouncer_full::{new_debouncer, notify::RecursiveMode, notify::Watcher};
 use ws::{Message, Sender, WebSocket};
 
 use errors::{anyhow, Context, Error, Result};
@@ -53,10 +54,11 @@ use site::sass::compile_sass;
 use site::{Site, SITE_CONTENT};
 use utils::fs::{clean_site_output_folder, copy_file, is_temp_file};
 
+use crate::fs_event_utils::{get_relevant_event_kind, SimpleFSEventKind};
 use crate::messages;
 use std::ffi::OsStr;
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, Hash, PartialEq)]
 enum ChangeKind {
     Content,
     Templates,
@@ -65,6 +67,7 @@ enum ChangeKind {
     Sass,
     Config,
 }
+impl Eq for ChangeKind {}
 
 #[derive(Debug, PartialEq)]
 enum WatchMode {
@@ -485,7 +488,7 @@ pub fn serve(
 
     // Setup watchers
     let (tx, rx) = channel();
-    let mut watcher = watcher(tx, Duration::from_secs(1)).unwrap();
+    let mut debouncer = new_debouncer(Duration::from_secs(1), /*tick_rate=*/ None, tx).unwrap();
 
     // We watch for changes on the filesystem for every entry in watch_this
     // Will fail if either:
@@ -501,8 +504,8 @@ pub fn serve(
             WatchMode::Condition(b) => b && watch_path.exists(),
         };
         if should_watch {
-            watcher
-                .watch(root_dir.join(entry), RecursiveMode::Recursive)
+            debouncer.watcher()
+                .watch(&root_dir.join(entry), RecursiveMode::Recursive)
                 .with_context(|| format!("Can't watch `{}` for changes in folder `{}`. Does it exist, and do you have correct permissions?", entry, root_dir.display()))?;
             watchers.push(entry.to_string());
         }
@@ -606,24 +609,24 @@ pub fn serve(
     })
     .expect("Error setting Ctrl-C handler");
 
-    use notify::DebouncedEvent::*;
-
-    let reload_sass = |site: &Site, path: &Path, partial_path: &Path| {
-        let msg = if path.is_dir() {
-            format!("-> Directory in `sass` folder changed {}", path.display())
-        } else {
-            format!("-> Sass file changed {}", path.display())
-        };
+    let reload_sass = |site: &Site, paths: &Vec<&PathBuf>| {
+        let combined_paths =
+            paths.iter().map(|p| p.display().to_string()).collect::<Vec<String>>().join(", ");
+        let msg = format!("-> Sass file(s) changed {}", combined_paths);
         console::info(&msg);
         rebuild_done_handling(
             &broadcaster,
             compile_sass(&site.base_path, &site.output_path),
-            &partial_path.to_string_lossy(),
+            &site.sass_path.to_string_lossy(),
         );
     };
 
-    let reload_templates = |site: &mut Site, path: &Path| {
-        rebuild_done_handling(&broadcaster, site.reload_templates(), &path.to_string_lossy());
+    let reload_templates = |site: &mut Site| {
+        rebuild_done_handling(
+            &broadcaster,
+            site.reload_templates(),
+            &site.templates_path.to_string_lossy(),
+        );
     };
 
     let copy_static = |site: &Site, path: &Path, partial_path: &Path| {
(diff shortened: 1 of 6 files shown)


src/cmd/serve.rs
```diff
@@ -606,24 +620,24 @@ pub fn serve(
...
+    let reload_sass = |site: &Site, paths: &Vec<&PathBuf>| {
+        let combined_paths =
+            paths.iter().map(|p| p.display().to_string()).collect::<Vec<String>>().join(", ");
+        let msg = format!("-> Sass file(s) changed {}", combined_paths);
         console::info(&msg);
         rebuild_done_handling(
             &broadcaster,
             compile_sass(&site.base_path, &site.output_path),
-            &partial_path.to_string_lossy(),
+            &site.sass_path.to_string_lossy(),
```
iamorphen, line 620: @Keats , a question on this change and a similar one in `reload_templates`. Previously we passed partial paths here, but now I'm passing an absolute path to the root of the sass directory. I couldn't find (or resolve, for some reason) protocol docs for livereload v7. Is this change okay? Functionally it seems to work.

Keats: I have no idea. If it works it's probably fine, i'll try it locally.

src/cmd/serve.rs
```diff
@@ -690,52 +704,85 @@ pub fn serve(
...
-                    Rename(_, path) | Create(path) | Write(path) | Remove(path) | Chmod(path) => {
-                        if is_ignored_file(&site.config.ignored_content_globset, &path) {
-                            continue;
-                        }
+                    // We currently only handle notify events that report a single path per event.
+                    // Enforce that assumption to avoid surprises. See `get_relevant_event_kind`.
+                    assert!(
+                        event.event.paths.len() == 1,
+                        "Didn't expect notify event with multiple paths."
+                    );
```
Keats: Either `get_relevant_event_kind` should make sure of that or we should return an error rather than a panic

src/cmd/serve.rs
```diff
@@ -690,52 +704,85 @@ pub fn serve(
...
-                        match detect_change_kind(root_dir, &path, &config_path) {
-                            (ChangeKind::Content, _) => {
-                                console::info(&format!("-> Content changed {}", path.display()));
+                // Editors such as Vim will create numerous events when operating on a single file.
+                // These events are not combined above. Take the latest event for each unique path.
+                let filtered_events = debounce_events(&meaningful_events);
+
+                // Map of change_kind -> (partial_path, full_path, event_kind).
+                let changes: HashMap<ChangeKind, Vec<(PathBuf, &PathBuf, &SimpleFSEventKind)>> =
+                    map_events_to_change_map(&filtered_events, &root_dir, &config_path);
```
Keats: Can we extract the whole logic of finding and sorting/filtering events to a separate function so it's easy to test and doesn't take space in the actual loop?

src/cmd/serve.rs
```diff
@@ -763,55 +810,153 @@ pub fn serve(
...
-                            (ChangeKind::StaticFiles, p) => copy_static(&site, &path, &p),
-                            (ChangeKind::Sass, p) => reload_sass(&site, &path, &p),
-                            (ChangeKind::Themes, _) => {
-                                console::info("-> Themes changed.");
-
+                        }
+                        ChangeKind::Templates => {
+                            let partial_paths: Vec<&PathBuf> =
+                                change_group.iter().map(|(p, _, _)| p).collect();
+                            let full_paths: Vec<&PathBuf> =
```
Keats, line 820: when we do get partial and full paths?

src/cmd/serve.rs
```diff
@@ -763,55 +810,153 @@ pub fn serve(
...
+        }
+    });
+    filtered.dedup_by(|(p1, _, _), (p2, _, _)| p1 == p2);
+
+    filtered.into_iter().map(|(p, e, _)| (p, e)).collect()
+}
+
+/// Return a map storing all events of a like `ChangeKind` behind one key.
+/// Each element maps `change_kind -> (partial_path, full_path, event_kind)`.
+fn map_events_to_change_map<'a>(
```
Keats: let's move all that code to another file

src/cmd/serve.rs
```diff
@@ -763,55 +810,153 @@ pub fn serve(
...
+        },
+        _ => None,
+    }
+}
+
+/// Return a new vector with no duplicate `PathBuf` elements. For any duplicate, keep the `PathBuf`
+/// with the latest `Instant` value.
+fn debounce_events(
+    events: &Vec<(PathBuf, SimpleFSEventKind, Instant)>,
+) -> Vec<(PathBuf, SimpleFSEventKind)> {
```
Keats: Assuming the events are ordered (or that can will order them), we could just use a hashmap with the path as the key and insert them in order

Keats (changes requested):

src/cmd/serve.rs
iamorphen: Thanks for your time. Before I start, to confirm, are you asking about moving only the new code in this merge request or also existing code like `is_ignored_file`, `is_temp_file`, etc. participating in the filtering?

Keats: I think it might make sense to move everything to a new file? But no strong feeling about it

iamorphen: Thanks. In that case, I will focus on just my changes for now.

src/cmd/serve.rs
iamorphen: If I understand your question correctly, we get full paths from the incoming events from `notify`, and we get partial paths from `detect_change_kind`. In the implementation on `next`, check the partial paths to determine whether shortcodes were modified. We pass the full paths to `reload_templates`. I'm not sure if the latter is necessary, so just following what I see on `next`.

src/cmd/serve.rs
iamorphen: Fixed. It now prints an error and skips that event.

src/cmd/serve.rs
iamorphen: Done.

src/cmd/serve.rs
iamorphen: Okay, thanks. I took the liberty to use a hash map earlier on, and that allowed me to remove two of the functions I introduced. Upside is that the diff is smaller, downside is that we lost some test guarantees because some logic is inlined into `serve`.

src/cmd/serve.rs
iamorphen: I've introduced `components/fs_event_utils`. I took a guess at where to put code. If there's a better place, please advise.

iamorphen (commented):
I have updated the description to match the implementation of the latest update. I re-ran my manual test and confirmed the same behavior.

src/cmd/serve.rs
Keats: Let's put it in the root component, next to `cli.rs`

iamorphen: Done.

Keats:
LGTM. I'll make some changes to the code org later but it works fine locally. Thanks a lot!

iamorphen:
Thanks Keats!

Keats:
Here's the commit i've added: https://github.com/getzola/zola/pull/2382/commits/b693e62aa0e788bae6bfe1060d94765e59c14248
Just moving more things to that new file to try to keep the `serve.rs` reasonable

iamorphen:
Nice, thank you for the context and refactor.

PR #2507, opened 2024-05-24
Add support for Fuse.js search format

SIGSTACKFAULT:
Sanity check:
* [x] Have you opened a thread on the forum? *[yes](https://zola.discourse.group/t/fuse-js-search-format/2139) but nobody responded*

the Fuse.js format looks like this (I'm working on [Bevy's website](https://github.com/bevyengine/bevy-website/):

```
[
  {
    "title": "2D",
    "body": "\n  TODO\n  \nExplain how 2D graphics in Bevy is actually 3D with a weird camera projection\nDefine the relevant abstractions\n\n\n\n",
    "path": "/learn/book/graphics/2d/"
  },
  {
    "title": "Sprite Sheets",
    "body": "\n  TODO\n  \nDemonstrate how to import and use a sprite sheet for basic animations\n\n\n\n",
    "path": "/learn/book/graphics/2d/sprite-sheets/"
  },
  {
    "title": "Sprites",
    "body": "\n  TODO\n  \nDemonstrate how to import and use a sprite\n\n\n\n",
    "path": "/learn/book/graphics/2d/sprites/"
  }
]
```

it happens to also be compatible with tinysearch, and almost compatible with stork.


components/search/src/fuse.rs
```diff
@@ -0,0 +1,52 @@
+use config::Search;
+use content::Library;
+
+/// build index in Fuse.js format.
+pub fn build_index(lang: &str, library: &Library, config: &Search) -> super::Result<String> {
+    #[derive(serde::Serialize)]
+    struct Item {
+        title: Option<String>,
+        body: Option<String>,
+        path: String,
```
Keats: Do we really need owned values if we're just going to serialize it?

components/search/Cargo.toml
```diff
@@ -8,3 +8,5 @@ errors = { path = "../errors" }
 content = { path = "../content" }
 config = { path = "../config" }
 libs = { path = "../libs" }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
Keats: use `libs::serde_json` instead so we only have the lib defined in one crate

components/site/src/lib.rs
```diff
@@ -799,19 +799,30 @@ impl Site {
     }
 
     fn index_for_lang(&self, lang: &str) -> Result<()> {
-        let index_json = search::build_index(lang, &self.library.read().unwrap(), &self.config)?;
-        let (path, content) = match &self.config.search.index_format {
-            IndexFormat::ElasticlunrJson => {
-                let path = self.output_path.join(format!("search_index.{}.json", lang));
-                (path, index_json)
+        let extension = match &self.config.search.index_format {
```
Keats: that should be a method on the `IndexFormat` enum rather than putting it here

components/site/src/lib.rs
```diff
@@ -799,19 +799,30 @@ impl Site {
...
+            IndexFormat::ElasticlunrJavascript | IndexFormat::FuseJavascript => "js",
+            IndexFormat::ElasticlunrJson | IndexFormat::FuseJson => "json",
+        };
+        let path = &self.output_path.join(format!("search_index.{}.{}", lang, extension));
+        let library = self.library.read().unwrap();
+        let content = match &self.config.search.index_format {
+            IndexFormat::ElasticlunrJavascript | IndexFormat::ElasticlunrJson => {
+                search::build_index(lang, &library, &self.config)?
```
Keats: we should probably put elastic search into its own file to keep it symmetric

components/site/src/lib.rs
```diff
@@ -799,19 +799,30 @@ impl Site {
...
             }
         };
-        create_file(&path, &content)
+        drop(library); // no need to hold on to this guard while writing
+        create_file(
+            path,
+            match self.config.search.index_format {
+                IndexFormat::ElasticlunrJson | IndexFormat::FuseJson => content,
+                IndexFormat::ElasticlunrJavascript | IndexFormat::FuseJavascript => {
+                    format!("window.searchIndex = {}", content)
```
Keats, line 818: This can probably go on the enum as well?

Keats (changes requested):

Keats:
Sorry I've seen the message but didn't get to respond yet.

components/search/src/fuse.rs
SIGSTACKFAULT: I didn't want to mess with lifetimes and the elasticlunr logic also clones everything. But i'll change it.

components/search/Cargo.toml
SIGSTACKFAULT: done; is there a similar way to get Serde that i should use?

components/site/src/lib.rs
SIGSTACKFAULT: Moved the whole filename logic inside IndexFormat

components/site/src/lib.rs
SIGSTACKFAULT: done

components/site/src/lib.rs
SIGSTACKFAULT: of course it could, but I feel like it would be a weird thing to do? completely subjective.

SIGSTACKFAULT: like at that point i might as well put the entire thing inside IndexFormat

SIGSTACKFAULT:
bump @Keats because i replied to some of your reviews with followups

components/search/Cargo.toml
Keats: i don't think so because of the derive feature

Keats (approved):

docs/content/documentation/content/search.md
```diff
@@ -4,29 +4,46 @@ weight = 100
...
+index_format = "elasticlunr_javascript" # or "elasticlunr_json"
+```
+
+If you are using a language other than English, you will also need to include the corresponding JavaScript stemmer file.
+See <https://github.com/weixsong/lunr-languages#in-a-web-browser> for details.
+
+### Fuse
+
+Compatible with [fuse.js](https://www.fusejs.io/), [tinysearch](https://github.com/tinysearch/tinysearch), and almost
+compatible with [stork](https://stork-search.net/)
```
Keats: What does it mean "almost compatible"? Does it work with stork or not?

SIGSTACKFAULT: [stork's format](https://stork-search.net/docs/config-ref) is very similar but the keys are slightly different, it's toml, and it wants a bit of a extra metadata. One could conceivably convert fuse_json to stork with a shell script and a copy of [`jq`](https://github.com/jqlang/jq)

I should probably either write all that in the docs or remove the reference entirely

SIGSTACKFAULT: actually maybe i should have fuse_json have a `url` key instead of `path`, since fuse doesn't care and that will be more compatible with stork and tiny.

SIGSTACKFAULT: oooh wait i see why `include_path` exists

Keats (approved):

Keats:
Thanks!

Keats:
Note: I haven't tried it myself but I assume you did right?

SIGSTACKFAULT:
@Keats yeah i have a branch of bevy-website working using this. When will it ship?

Keats:
I need to have a look at all PRs and issues but i hope this week or the next one

Jieiku:
Just wanted to add that the fuse_json index format helped with implementing pagefind in abridge, here is the demo: https://abridge-pagefind.pages.dev/

I planned to update tinysearch in abridge to use fuse_json, but a while ago I ran into an issue that I have not been able to figure out: https://github.com/tinysearch/tinysearch/issues/175

PR #2516, opened 2024-06-09
Improve docs header style

FireIsGood:
Sanity check:

---

The current website style for the header is a bit jank and leaves a bit to be desired. The search bar and navigation are both left aligned which makes it feel a bit cluttered, though that's not too bad of an issue. The real issue I had with it was how the mobile styles looked. The spacing method causes the links to be shifted left a bit and the search bar to sit right against the text.

image

## Before

Desktop styles

| Home | Docs |
|--------|--------|
| home | docs |

Mobile styles

| Home | Docs |
|--------|--------|
| home mobile | docs mobile | 

## After

Desktop styles

| Home | Docs |
|--------|--------|
| home | docs |

Mobile styles

| Home | Docs |
|--------|--------|
| home mobile | docs mobile |

## Additional Information

A lot of the jank comes from how the nav was previously just using inline-block and margin-right based styles.

Also, there was a random `.container` style applied to only the home page's header which was reused on the contents of the body as well which made restyling a pain. I have made the decision to remove it from the header section entirely for consistency between the home and docs pages.

There are also a lot of other evil styles such as the `.selling-point` styles in general, though that would be more of a style rewrite which is more opinionated and likely better saved for a full website rewrite.


docs/templates/index.html
```diff
@@ -102,7 +103,7 @@ <h2>Augmented Markdown</h2>
             ©2017-{{ now() | date(format="%Y") }} — <a class="white" href="https://www.vincentprouillet.com">Vincent Prouillet</a> and <a class="white" href="https://github.com/getzola/zola/graphs/contributors">contributors</a>
         </footer>
 
-        <script type="text/javascript" src="{{ get_url(path="elasticlunr.min.js") }}"></script>
-        <script type="text/javascript" src="{{ get_url(path="search.js") }}"></script>
+        <script type="text/javascript" src="{{ get_url(path='elasticlunr.min.js') }}"></script>
+        <script type="text/javascript" src="{{ get_url(path='search.js') }}"></script>
```
Keats, line 107: I prefer `"` to `'`, any reasons to change those?

Keats (commented):
LGTM except the quote change

docs/templates/index.html
FireIsGood: If the editor lsp or formatter is set to regular HTML, it can get hung up on the fact that there are double quotes inside the strings and refuses to work. My reason for this case was that in JS you can use single quotes in double quotes and vice versa, so they should likely be the same in HTML.

I believe the specific reason in this case was that I was using Prettier which was angry that the tags weren't closed due to some weirdness there.

Keats: good point i didn't think of that

PR #2517, opened 2024-06-09
Fix theme images loading

FireIsGood:
Sanity check:

---

The theme images currently have heights equal to one of the images with the other fitting via `object-fit: cover`. This causes a lot of layout shift and inconsistent heights.

To fix this, I just added an `aspect-ratio: 16 / 9` as a progressive enhancement.

## Before

Testing with a slower connection to show layouts before the images load:

image

## After

Testing with a slower connection to show layouts before the images load:

image

## Additional Info

This project's docs site should probably try and find a formatter, as lots of the styles were inconsistent and formatting was most of the diff here. If the maintainer wishes to reduce the diff for this pull request, they may revert/drop the `chore: format themes file` commit.

### DIFF
--- a/docs/sass/_themes.scss
+++ b/docs/sass/_themes.scss
@@ -31,6 +31,7 @@
 
     img {
       width: 100%;
+      aspect-ratio: 16 / 9;
       height: 90%;
       object-fit: cover;
     }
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

PR #2519, opened 2024-06-10
fix ignore links to #top when checking anchors

### LINKED ISSUE #2495: Ignore links to #top when checking anchors
# Bug Report

## Environment

Zola version: 0.18.0

## Expected Behavior

[The HTML spec](https://html.spec.whatwg.org/#select-the-indicated-part) includes some special handling for a `#top` fragment added to the URL:

> 10. If *decodedFragment* is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string `top`, then return the [top of the document](https://html.spec.whatwg.org/#top-of-the-document).

This means that adding `[some text](#top)` to a page correctly creates a link to the top of the page, and Zola should treat the link as always being valid since the anchor is technically always there.

## Current Behavior

```
$ zola check
Checking site...
> Checked 1 internal link(s) with anchors: 1 target(s) missing.
Failed to check the site
Error: The anchor in the link `@/index.md#top` in /tmp/testsite/content/index.md does not exist.
```

## Steps to reproduce

Add a Markdown link to `#top` in any page, and run `zola check`.

## Workaround

Linking to `#` can work instead, as the HTML spec also defines it as an edge case:

> 3. If *fragment* is the empty string, then return the special value **top of the document**.

ta-nishq:
Fixes #2495

### DIFF
--- a/components/content/src/utils.rs
+++ b/components/content/src/utils.rs
@@ -8,13 +8,6 @@ use utils::fs::is_temp_file;
 use utils::table_of_contents::Heading;
 
 pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
-    // HTML specs state a special anchor: "top" (case-insensitive)
-    // which is used to refer to the start of the page
-    // https://html.spec.whatwg.org/#select-the-indicated-part
-    if anchor.eq_ignore_ascii_case("top") {
-        return true;
-    }
-
     for heading in headings {
         if heading.id == anchor {
             return true;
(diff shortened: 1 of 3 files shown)


components/content/src/utils.rs
```diff
@@ -8,6 +8,13 @@ use utils::fs::is_temp_file;
 use utils::table_of_contents::Heading;
 
 pub fn has_anchor(headings: &[Heading], anchor: &str) -> bool {
+    // HTML specs state a special anchor: "top" (case-insensitive)
+    // which is used to refer to the start of the page
+    // https://html.spec.whatwg.org/#select-the-indicated-part
+    if anchor.eq_ignore_ascii_case("top") {
```
Keats: Is it the right place to do it? This checks for the anchors in the heading, it feels like this check should be before calling that function.

ta-nishq: updated!

Keats (approved):

PR #2549, opened 2024-06-25
Fix YAML date parsing

### LINKED ISSUE #2538: Zola 0.19.0 doesn't accept all valid YAML dates
# Bug Report

With 0.19.0, Zola can now understand more date formats. However, I believe there is a bug in the implementation which makes the millisecond separator (but not the milliseconds) required

## Environment

Zola version: 0.19.0

## Expected Behavior
`2024-06-22 18:42:00 +02:00` should be a correct datetime

## Current Behavior
```
Error: Failed to build the site
Error: Error when parsing front matter of section `path/to/file.md`
Error: Reason: YAML deserialize error: Error("Unable to parse datetime", line: 2, column: 1)
```

If I add a single dot after the seconds (`2024-06-22 18:42:00. +02:00`), the issue goes away.

## Step to reproduce

1. Create a Markdown file with a date like `2024-06-22 18:42:00 +02:00`
2. Try to `zola build` or `zola serve`
3. Observe the error

## Possible solution

I believe there is an error in the regex:

https://github.com/getzola/zola/blob/v0.19.0/components/utils/src/de.rs#L12

As you can see, the millisecond (aka fraction) part of the string is optional, but the period is not. This contradicts the [YAML Working Draft on Timestamp](https://yaml.org/type/timestamp.html) which was mentioned in the original issue.

```diff
-    let re = Regex::new(r#"^"?([0-9]{4})-([0-9][0-9]?)-([0-9][0-9]?)([Tt]|[ \t]+)([0-9][0-9]?):([0-9]{2}):([0-9]{2})\.([0-9]*)?Z?([ \t]([-+][0-9][0-9]?)(:([0-9][0-9]?))?Z?|([-+][0-9]{2})?:([0-9]{2})?)?|([0-9]{4})-([0-9]{2})-([0-9]{2})"?$"#).unwrap();
+    let re = Regex::new(r#"^"?([0-9]{4})-([0-9][0-9]?)-([0-9][0-9]?)([Tt]|[ \t]+)([0-9][0-9]?):([0-9]{2}):([0-9]{2})(?:\.([0-9]*))?Z?([ \t]([-+][0-9][0-9]?)(:([0-9][0-9]?))?Z?|([-+][0-9]{2})?:([0-9]{2})?)?|([0-9]{4})-([0-9]{2})-([0-9]{2})"?$"#).unwrap();
```

## Additional info

Same error also happens if I omit the seconds altogether (`2024-06-22 18:42 +02:00`). I may be opinionated, but I think requiring those should be optional, too.

Relates to https://github.com/getzola/zola/issues/2071, https://github.com/getzola/zola/pull/2208

kytta:
This PR makes a few changes to how date is parsed in YAML preambles. Fixes #2538.

### New regex

Initially, I planned to address #2538 by adding the fraction separator (`.`) to the optional group so that it also becomes optional. However, I have made a few more changes:

- **named groups:** instead of indexed access, one can access date components via their name
- Parsing is now a bit **off-spec**
  - when one reads the [timestamp working draft](https://yaml.org/type/timestamp.html), one can see that they use two different parts for dates with and without time and tz
  - the former (just date) _does not_ allow for short months and days (i.e. without leading zero), and is thus compliant with ISO 8601
  - the latter (date, time, and optional offset) allow months, days, and hours without leading zeroes. This is _non-compliant_, but user-friendly
  - I find it weird that both variants are present and have made the choice to allow months and days without leading zeroes in both 'short' and 'long' forms
- as a side effect, there is **no group duplication** any more, and the regex is now shorter
- I also make use of **non-capturing groups** to split the regex more cleanly

You can explore the regex on regex101: [h5wyK4](https://regex101.com/r/h5wyK4/2); and RegExr: [824pi](https://regexr.com/824pi)

### Fixes offset parsing

As far as I can understand from code, the parsing of the offset was done incorrectly:

- to convert to nanosecond, seconds were multiplied by 100 million (instead of 1 billion)
- instead of a floating point number, an integer was multiplied
- the integer was parsed from the decimal part without the dot

I have revised the code and arrived at this solution:

- the regex allows up to 9 decimal points in the fraction part
- the fraction part get parsed as `f64` number. It is between 0 and 1 (exclusive)
- this floating point number represents a fraction of a second
- this floating point number gets multiplied by one billion (nanoseconds in a second) and applied to the offset

### New tests

I have re-organized some tests and have added some more tests. I've taken the new test cases from the [TOML test suite](https://github.com/toml-lang/toml-test), but have only taken those that make sense with the YAML spec

## Sanity check:

## Code changes

### DIFF
--- a/components/utils/src/de.rs
+++ b/components/utils/src/de.rs
@@ -23,18 +23,12 @@ pub fn parse_yaml_datetime(date_string: &str) -> Result<time::OffsetDateTime> {
         if let Some(minutes_) = captures.name("minute") { minutes_.as_str() } else { "0" };
     let seconds =
         if let Some(seconds_) = captures.name("second") { seconds_.as_str() } else { "0" };
-    let fractional_seconds_raw =
-        if let Some(fractionals) = captures.name("fraction") { fractionals.as_str() } else { "" };
-    let fractional_seconds_intermediate = fractional_seconds_raw.trim_end_matches("0");
+    let fraction_raw =
+        if let Some(fraction_) = captures.name("fraction") { fraction_.as_str() } else { "" };
+    let fraction_intermediate = fraction_raw.trim_end_matches("0");
     //
     // Prepare for eventual conversion into nanoseconds
-    let fractional_seconds = if fractional_seconds_intermediate.len() > 0
-        && fractional_seconds_intermediate.len() <= 9
-    {
-        fractional_seconds_intermediate
-    } else {
-        "0"
-    };
+    let fraction = if fraction_intermediate.len() > 0 { fraction_intermediate } else { "0" };
     let maybe_timezone_hour = captures.name("offset_hour");
     let maybe_timezone_minute = captures.name("offset_minute");
 
@@ -58,7 +52,7 @@ pub fn parse_yaml_datetime(date_string: &str) -> Result<time::OffsetDateTime> {
         .replace_hour(hours.parse().unwrap())?
         .replace_minute(minutes.parse().unwrap())?
         .replace_second(seconds.parse().unwrap())?
-        .replace_nanosecond(fractional_seconds.parse::<u32>().unwrap() * 100_000_000)?)
+        .replace_nanosecond((fraction.parse::<f64>().unwrap_or(0.0) * 1_000_000_000.0) as u32)?)
 }
 
 /// Used as an attribute when we want to convert from TOML to a string date
--- a/components/utils/src/de.rs
+++ b/components/utils/src/de.rs
@@ -147,23 +147,31 @@ mod tests {
     use time::macros::datetime;
 
     #[test]
-    fn yaml_spec_examples_pass() {
+    fn yaml_draft_timestamp_pass() {
+        // tests only the values from the YAML 1.1 Timestamp Draft
+        // See https://yaml.org/type/timestamp.html
         let canonical = "2001-12-15T02:59:43.1Z";
         let valid_iso8601 = "2001-12-14t21:59:43.10-05:00";
         let space_separated = "2001-12-14 21:59:43.10 -5";
         let no_time_zone = "2001-12-15 2:59:43.10";
         let date = "2002-12-14";
-        assert_eq!(parse_yaml_datetime(canonical).unwrap(), datetime!(2001-12-15 2:59:43.1 +0));
+        assert_eq!(
+            parse_yaml_datetime(canonical).unwrap(),
+            datetime!(2001-12-15 02:59:43.100 +00:00)
+        );
         assert_eq!(
             parse_yaml_datetime(valid_iso8601).unwrap(),
-            datetime!(2001-12-14 21:59:43.1 -5)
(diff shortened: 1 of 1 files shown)


kytta:
Ooops; I've hoped that `[skip ci]` would work for Azure 😅

components/utils/src/de.rs
```diff
@@ -9,23 +9,20 @@ use serde::{Deserialize, Deserializer};
 
 pub fn parse_yaml_datetime(date_string: &str) -> Result<time::OffsetDateTime> {
     // See https://github.com/getzola/zola/issues/2071#issuecomment-1530610650
-    let re = Regex::new(r#"^"?([0-9]{4})-([0-9][0-9]?)-([0-9][0-9]?)([Tt]|[ \t]+)([0-9][0-9]?):([0-9]{2}):([0-9]{2})\.([0-9]*)?Z?([ \t]([-+][0-9][0-9]?)(:([0-9][0-9]?))?Z?|([-+][0-9]{2})?:([0-9]{2})?)?|([0-9]{4})-([0-9]{2})-([0-9]{2})"?$"#).unwrap();
+    let regex = Regex::new(r#"(?m)^"?(?P<year>[0-9]{4})-(?P<month>[0-9][0-9]?)-(?P<day>[0-9][0-9]?)(?:(?:[Tt]|[ \t]+)(?P<hour>[0-9][0-9]?):(?P<minute>[0-9]{2}):(?P<second>[0-9]{2})(?:\.(?P<fraction>[0-9]*))?[ \t]*(?:(?P<utc>Z)|(?P<offset>(?P<offset_hour>[-+][0-9][0-9]?)(?::(?P<offset_minute>[0-9][0-9]))?))?)?"?$"#).unwrap();
```
kytta: This is the dumbest error ever 🫣
Rule to myself: never blindly commit untested code from browser, especially if copy-pasting 😅 

```suggestion
    let re = Regex::new(r#"(?m)^"?(?P<year>[0-9]{4})-(?P<month>[0-9][0-9]?)-(?P<day>[0-9][0-9]?)(?:(?:[Tt]|[ \t]+)(?P<hour>[0-9][0-9]?):(?P<minute>[0-9]{2}):(?P<second>[0-9]{2})(?:\.(?P<fraction>[0-9]*))?[ \t]*(?:(?P<utc>Z)|(?P<offset>(?P<offset_hour>[-+][0-9][0-9]?)(?::(?P<offset_minute>[0-9][0-9]))?))?)?"?$"#).unwrap();
```

components/utils/src/de.rs
kytta, line 1: **todo:** Add more tests

kytta (commented):
Don't mind me, it's easier for me to track what I need to do when I return to this

kytta:
@Keats I've fixed the fraction separator being required and have further optimized the regex. I've added some tests (test cases courtesy of TOML); if you don't mind them overlapping, I think it's a good test suite for this functionality 😄

Keats:
Neat, thanks

PR #2555, opened 2024-06-28
Exclude paginated pages in sitemap

### LINKED ISSUE #2527: Paginated pages should be absent from sitemap.xml
# Bug Report

I am using pagination to show blog posts. I have enabled sitemap.xml. I use ahrefs for SEO site audits. It shows me [this](https://help.ahrefs.com/en/articles/2652498-non-canonical-page-in-sitemap-error-in-site-audit) issue which arises because [my sitemap](https://bauva.com/sitemap.xml) has links for paginated pages which it shouldn't since the links for all blog posts and the link for the list of blog post is there.

## Environment

Zola version: 0.18.0

## Expected Behavior

The /blog/ URL should be a part of sitemap.xml along with all the blog articles but not /blog/page/1/.

## Current Behavior

The paginated pages, i.e. /blog/page/1/, are a part of sitemap.xml, which makes search engines' lives difficult, and it could possibly lead to issues in indexing.

## Step to reproduce

Switch on pagination and sitemap generation.

pranitbauva1997:
This fixes #2527. @Keats, Can you please have a look at this and suggest improvements?

Sanity check:

## Code changes

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -29,6 +29,13 @@ pub enum Mode {
     Check,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum ExcludePaginatedPagesInSitemap {
+    None,
+    All,
+}
+
 #[derive(Clone, Debug, Deserialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct Config {
@@ -102,6 +109,8 @@ pub struct Config {
     pub generate_sitemap: bool,
     /// Enables the generation of robots.txt
     pub generate_robots_txt: bool,
+    /// Whether to exclude paginated pages in sitemap; can take values "none", "all"
+    pub exclude_paginated_pages_in_sitemap: ExcludePaginatedPagesInSitemap,
 }
 
 #[derive(Serialize)]
@@ -123,6 +132,7 @@ pub struct SerializedConfig<'a> {
     search: search::SerializedSearch<'a>,
     generate_sitemap: bool,
     generate_robots_txt: bool,
+    exclude_paginated_pages_in_sitemap: ExcludePaginatedPagesInSitemap,
 }
 
 impl Config {
@@ -287,6 +297,10 @@ impl Config {
         self.mode == Mode::Check
     }
 
+    pub fn should_exclude_paginated_pages_in_sitemap(&self) -> bool {
+        self.exclude_paginated_pages_in_sitemap == ExcludePaginatedPagesInSitemap::All
+    }
+
     pub fn enable_serve_mode(&mut self) {
         self.mode = Mode::Serve;
     }
@@ -340,6 +354,7 @@ impl Config {
             search: self.search.serialize(),
             generate_sitemap: self.generate_sitemap,
             generate_robots_txt: self.generate_robots_txt,
+            exclude_paginated_pages_in_sitemap: self.exclude_paginated_pages_in_sitemap,
         }
     }
 }
@@ -405,6 +420,7 @@ impl Default for Config {
             extra: HashMap::new(),
             generate_sitemap: true,
             generate_robots_txt: true,
+            exclude_paginated_pages_in_sitemap: ExcludePaginatedPagesInSitemap::None,
         }
     }
 }
@@ -1066,4 +1082,6 @@ base_url = "example.com"
         let config = Config::parse(config).unwrap();
         assert!(config.generate_robots_txt);
     }
+
+    // TODO: add a test for excluding paginated pages
 }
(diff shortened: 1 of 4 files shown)


components/config/src/config/mod.rs
```diff
@@ -98,6 +98,9 @@ pub struct Config {
     pub markdown: markup::Markdown,
     /// All user params set in `[extra]` in the config
     pub extra: HashMap<String, Toml>,
+
+    /// Whether to exclude paginated pages in sitemap
+    pub exclude_paginated_pages_in_sitemap: bool,
```
Keats: Bikeshedding: i don't know whether it should go on another nested level in the config since we do regret top level fields everytime.. Maybe a seo struct?

I believe it should also be an enum:
- none
- first only
- all

since i did see some people indexing only the first page.

docs/content/documentation/templates/pagination.md
```diff
@@ -52,6 +52,13 @@ A paginated taxonomy gets two variables aside from the `paginator` variable:
 
 See the [taxonomies page](@/documentation/templates/taxonomies.md) for a detailed version of the types.
 
+## SEO
+
+It is preferable to not include paginated pages in sitemap since they are non-canonical pages as explained
```
Keats: I would avoid making a call like that since it seems there are a lot of opinions: https://www.reddit.com/r/TechSEO/comments/10gfiob/should_pagination_pages_be_included_in_your/
Just describe the setting behaviour

docs/content/documentation/templates/pagination.md
```diff
@@ -52,6 +52,13 @@ A paginated taxonomy gets two variables aside from the `paginator` variable:
...
+[here](https://help.ahrefs.com/en/articles/2652498-non-canonical-page-in-sitemap-error-in-site-audit) and
```
Keats: let's not include external links

Keats (changes requested):

docs/content/documentation/templates/pagination.md
pranitbauva1997: Understood

docs/content/documentation/templates/pagination.md
pranitbauva1997: Makes sense.

components/config/src/config/mod.rs
pranitbauva1997: Let's take an example of my site https://bauva.com since I have tested this PR on it. The [blog list page](https://bauva.com/blog/) is paginated and it contains the link to the next page. The current sitemap includes `/blog/` and `/blog/page/1/` and `/blog/page/2/`. After the current PR, the sitemap includes `/blog/` which can still work for search engines to discover the other paginated pages because the "next" and "prev" buttons. Also for the current behaviour `/blog/` and `/blog/page/1/` are the exact same pages which IMO should be avoided.

If we add another option like first only, the sitemap would contain `/blog/` and `/blog/page/1/` which will cause the same non-canonical page problem.

pranitbauva1997:
@Keats Can you please re-review this?

components/config/src/config/mod.rs
Keats: I think let's still make it a string because otherwise it will be a breaking change if we decide to have 3 behaviours. Maybe `paginated_pages_in_sitemap` as a name?

pranitbauva1997: Sure

pranitbauva1997:
@Keats Made the required changes of using a string value instead of boolean so that we can further expand the scope by including other options. Can you please re-review?

components/config/src/config/mod.rs
```diff
@@ -102,6 +102,8 @@ pub struct Config {
     pub generate_sitemap: bool,
     /// Enables the generation of robots.txt
     pub generate_robots_txt: bool,
+    /// Whether to exclude paginated pages in sitemap; can take values "none", "all"
+    pub exclude_paginated_pages_in_sitemap: String,
```
Keats: Sorry I meant something like `LinkCheckerLevel` where it's a String for the end user but we do turn it into a Rust enum: https://github.com/getzola/zola/blob/master/components/config/src/config/link_checker.rs#L7-L36

Keats (changes requested):

components/config/src/config/mod.rs
pranitbauva1997: @Keats, just to clarify, do you mean that the user behaviour of the PR in its current state is accurate but the internal data structures need to be changed to make it more extensible in future?

If that's the case, won't the refactor make sense when there are things to extend in the future? But if user behaviour needs to change, I should make that change in this PR.

Keats: You're correct. The main thing is that using an enum with serde serialization you will get error messages if you use invalid string values. Here they will just be ignored which isn't great.

pranitbauva1997: Got it, let me make the change right away

pranitbauva1997:
@Keats Made the changes and tested them on my website. Earlier wrong values weren't throwing errors. Now, they are throwing descriptive errors.

Keats (approved):
Nice, thanks!

pranitbauva1997:
@Keats thanks for approving the PR. This is my first contribution over here, so what would be the next steps? Please let me know if there is anything to be done from my side.

Keats:
Nothing, that's all good!

PR #2569, opened 2024-07-12
docs(installation): add instructions for openSUSE Leap and Tumbleweed

LeoniePhiline:
When installing `zola` from repositories, the `bash`, `zsh` and `fish` autocompletion packages are co-installed as recommended packages.

Package details for all distributions:
https://software.opensuse.org/package/zola?baseproject=ALL


Keats (approved):

Keats:
Thanks!

Issue #2551, opened 2024-06-26
"zola  serve" doesn't detect any change


arkydon:
# Bug Report

## Environment

Zola version: 0.19.1, 0.19.0
Windows 11 22H2, Build version: 22621.3737

## Expected Behavior
Using "zola serve" command, it should detect changes to a website and reflect it.

## Current Behavior
In both 0.19.0 and 0.19.1 it can't detect any changes anymore, so I have to rerun "zola serve" every time I make any changes in my website code. In the terminal it's always at the initial "Listening for changes".

## Step to reproduce
1. Run command "zola serve"
2. Make any change to templates or sass or content.

# Note
I tried it in Windows Terminal, Powershell and the old command prompt. I tried running as admin too.

Keats:
We've fixed it for Linux but for Windows I'd need someone with a Windows machine to check what's happening. See https://github.com/getzola/zola/pull/2542 for the linux PR

knzai:
I'm seeing this in Mac as well (no changes result in a reload, even to templates or pages or anything), but maybe the linux fix will also fix that.

Keats:
Have you tried 0.19.1? It has always worked for me on Mac

arkydon:
> Have you tried 0.19.1? It has always worked for me on Mac

Yes. Both 0.19.0 and 0.19.1 have this issue.

myfonj:
Tried [`zola 0.19.1`](https://github.com/getzola/zola/releases/tag/v0.19.1) on `Windows 10 Version 10.0.19045 Build 19045` (With latest Firefox and Edge) and watcher still does not seem to catch any change `:(`. 

According to devtools the websocket is being established (`ws://127.0.0.1:1024/livereload`) but never sends any message.

Most recent release with working live reload on my system seems to be [`0.18.0`](https://github.com/getzola/zola/releases/tag/v0.18.0).

knzai:
I was already on 0.19.1, but it started working with no changes, so idek. I've historically seen mac's do weird things with reloading dev server file detection before, so probably not anything on your end. Just following up to confirm I am *not* seeing this behavior on Mac, so ignore my earlier comment I guess.

Maybe a permissioning hiccup or something on one of my installs.

Snowdrama:
I also ran into this issue, I just installed it yesterday. Not that I'm 100% sure if it's related but it's showing the path and I'm curious if it's maybe related to the issue. 

I'm using 0.19.1
```
[...]

PR #2542, opened 2024-06-23
fix: treat any data change as a modification

lukehsiao:
fix: treat any data change as a modification

For unknown reasons, it seems some environments emit a `DataChange::Any` event,
rather than specifying content or size, when a file is edited. As an example:

    [src/fs_utils.rs:72:9] &event = DebouncedEvent {
        event: Event {
            kind: Modify(
                Data(
                    Any,
                ),
            ),
            paths: [
                "/home/redacted/content/blog/2024-06-23_example.md",
            ],
            attr:tracker: None,
            attr:flag: None,
            attr:info: None,
            attr:source: None,
        },
        time: Instant {
            tv_sec: 78544,
            tv_nsec: 936740729,
        },
    }

Consequently, because this isn't treated by Zola as a modification, the
site is not rebuilt, which regresses on previous behavior.

This patch fixes this particular case by treating any data modification
events as a modification.

Closes: https://github.com/getzola/zola/issues/2536

---

**IMPORTANT: Please do not create a Pull Request adding a new feature without discussing it first.**

The place to discuss new features is the forum: <https://zola.discourse.group/>
If you want to add a new feature, please open a thread there first in the feature requests section.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [x] Have you created/updated the relevant documentation page(s)?

PR #2575, opened 2024-07-15
fix: handle another fs event to fix change monitoring on Windows 10

iamorphen:
# Introduction

Toward #2551 . We needed to expand my narrower filtering in #2498 to handle some platforms I could not test for. The fix here is similar to that in #2542 .

## Code changes

1. Bin [`notify::event::ModifyKind::Any`](https://docs.rs/notify/6.1.1/notify/event/enum.ModifyKind.html) as a modification event. Previously we filtered it out.
2. Add test coverage for this event kind.

`ModifyKind::Any` is not a catch-all event kind but rather a "no detailed context" one. This seems to be the only granularity we are exposed to on Windows 10.

## Testing

I performed manual testing on Windows 10. I modified the config file and files in content, sass, static, templates, and themes and confirmed that site reload is fixed, at least on my runtime. I also tested modifications with editors like vim and simple Notepad. They raise a different set of events, but both contain the modify kind in this change, so this change covers both cases.

It would help to have someone else confirm this change if possible.

### DIFF
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -39,6 +39,8 @@ fn get_relevant_event_kind(event_kind: &EventKind) -> Option<SimpleFileSystemEve
             Some(SimpleFileSystemEventKind::Create)
         }
         EventKind::Modify(ModifyKind::Data(_))
+        // Windows 10 only reports modify events at the `Any` granularity.
+        | EventKind::Modify(ModifyKind::Any)
         // Intellij modifies file metadata on edit.
         // https://github.com/passcod/notify/issues/150#issuecomment-494912080
         | EventKind::Modify(ModifyKind::Metadata(MetadataKind::WriteTime))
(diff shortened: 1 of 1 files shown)


iamorphen:
In the event anyone else tests, on Windows, issuing ctrl+c to `zola` built on `next` as-is at the time of this writing (c666ee10e8839d0c3ceb109a1fe66e2daf8e3b80) results in

```
Encountered an error: uncategorized error
Enable a logger to see more information.
```

Usability is not ultimately affected.

Keats:
Hm that's a bummer i need to see if we don't get spammed by various events on Linux/Mac. Thanks for the PR!

Keats:
Looks fine for me on Mac

iamorphen:
Thanks for your review and testing! I just tested on Linux in the same manner I tested on Windows originally, just making modifications to the config, content, templates, themes, etc., one by one. I don't see any spamming on my end.

Keats:
Ok let's wait to see some more Windows people checking in but that looks good for me

Snowdrama:
Tested this on Windows 10, worked for me

Keavon:
I compiled this and tested it on my Windows 10 machine and it fixes the problem. I guess I'll be running this version from source until a hotfix release can be made, which I hope can be ASAP! Thanks for solving this pretty serious problem.

iamorphen:
Thanks for your time, folks!

bwaterford:
This fix works for me on Windows 7.

(I had to change some of the dependency versions in Cargo.lock to get it to compile running Rust 1.77.2)

PR #2580, opened 2024-07-19
Move sublime into config module

bemyak:
This is maybe a bit of controversial change, but I stumbled upon inability to build my crate (which depends on some parts of Zola) using Nix. The problem is that Nix builds every crate in an isolated environment, but since the `config` component depends on sublime files outside its crate root, the build fails.

I discussed the issue with a person who maintains the rust harness for Nix, and he suggested fixing it in Zola: https://github.com/ipetkov/crane/discussions/666#discussioncomment-10079661
There's a bit more info attached there.

Hope you will consider these changes :blush: 
It would also make publishing Zola's subcrates to crates.io a bit more feasible.

---
Sanity check:


Keats:
I think it's fine? I can't think of issues caused by that.

bemyak:
I don't think there are any issues caused by this change, it's more like a matter of aesthetics :slightly_smiling_face: 

If you don't object, maybe let's get it merged?

Keats:
Yep sorry forgot about it

bemyak:
Thank you! 🙏

PR #2581, opened 2024-07-21
Allow inline summary cutoff

### LINKED ISSUE #2562: `page.summary` only works when isolated on a newline.
# Bug Report

I've upgraded from Zola 0.18.0 to 0.19.1 today and found that my use of `page.summary` is broken. 

When writing the `<!-- more -->` tag inline into markdown content, the `page.summary` variable does not get populated. For example

```md
Hello hello <!-- more --> this is a test.
```

will not work. However, 

```md
Hello hello
<!-- more -->
this is a test
```

will work. Unfortunately, this will add a newline between `Hello hello` and `this is a test` when using `page.content`, which is a bummer because now I cant take part of a paragraph as a page summary anymore.

It would be cool to get back the functionality for `<!-- more -->` to be inlinable. I assume this might have something to do with the recent fix for footnotes?

## Environment

Zola version: 0.19.1

clarfonthey:

This worked previously, but was broken by a recent change. Essentially, because the `InlineHtml` event is not considered for the `<!-- more -->` marker, it has no meaning at the moment when not placed on its own line.

This alters the code to allow this again, but does so in a more reasonable way:

1. It properly closes tags that are left open in the summary. (Inline HTML is not considered; you're on your own if you do that.)
2. It runs a `summary-cutoff.html` template at the end of the summary if this happens, which acts depending on the summary before the cutoff and the current language. It defaults to showing an ellipsis if the summary doesn't end in punctuation.

I've also added tests to verify this behaviour works.

Fixes #2562.


Keats:
Honestly I'm not too sure about that. The original intent when it was added _was_ to put it on an empty line and putting it in the middle of some HTML  - I don't know...

clarfonthey:
That's extremely fair. In my case, I don't use it very often, but the current case where I do use it is for cutting sentences off as a kind of cliffhanger. Things like "So, I took a look at X, and \<cutoff\>". I just made a complete version that accounts for all kinds of markdown syntax because if I was going to do anything, I might as well do it right instead of just adding a `</p>` sometimes. (Which is what my code did before this was properly supported.)

And like, I guess that there might be cases where you might want to cut off in the middle of a blockquote or something, although those are probably much rarer.

If you do want it to remain on its own line, I can remove the summary-specific stuff and just keep it to fixing shortcodes for now. I do think that `<span>{{ code() }}</span>` not working and `{{ code() }}` working by itself is a little weird.

nyanpasu64:
Not sure about the technical details of this PR (I haven't looked into the code), but I found that even `<!-- more -->` in the middle or end of a regular Markdown paragraph (without HTML elements) is no longer detected as a marker. I use this heavily in my blog (https://nyanpasu64.gitlab.io/), including breaking summaries in the middle of a paragraph, and cannot update to 0.19.x until this is fixed. Is this related to this PR?

clarfonthey:
Short answer, yes,

Long answer, yes but it will change the code so that it actually emits the end of the paragraph and leave a cutoff marker in its place, so, you will need to modify your templates if they were manually adding the paragraph end marker back (alongside an ellipsis) in some capacity.

LunarEclipse363:
> 2. It leaves a `<span class="summary-cutoff"></span>` element before those tags are closed so it can be properly styled. This is slightly more portable than just using an ellipsis, since you can choose what character to use via CSS instead.

This could potentially be a built-in template the user can override like `anchor-link.html`.

The issue with putting the ellipsis in via CSS is that some user agents (Atom/RSS feed readers for example) might not load the stylesheet at all and then the ellipsis is gone entirely. I'm also not sure what implications would it have for accessibility.

clarfonthey:
> This could potentially be a built-in template the user can override like `anchor-link.html`.

Actually, yeah, that seems the most reasonable. Could also provide the template with the contents of the summary so you could, for example, only emit an ellipsis if there isn't punctuation at the end of the line. I'll make that the default.

clarfonthey:
Marking this as draft since I'm moving the shortcode fixes to a separate PR. Will rebase on top of that once it merges.

clarfonthey:
Rebased this on the other change and added a template version of the summary cutoff. By default, it ends the summary in an ellipsis if it was cut off inline.

components/markdown/src/markdown.rs
```diff
@@ -793,15 +796,50 @@ pub fn markdown_to_html(
             .position(|e| matches!(e, Event::Html(CowStr::Borrowed(CONTINUE_READING))))
             .unwrap_or(events.len());
 
+        // determine closing tags missing from summary
```
Keats, line 799: Could we error if the summary marker is _inside_ a tag? Eg `<i <-- more --> class='' />`

components/markdown/src/markdown.rs
```diff
@@ -793,15 +796,50 @@ pub fn markdown_to_html(
...
+                c.insert("summary", &summary_html);
+                c.insert("lang", &context.lang);
+                let summary_cutoff = utils::templates::render_template(
+                    SUMMARY_CUTOFF_TEMPLATE,
+                    &context.tera,
+                    c,
+                    &None,
+                )
+                .context("Failed to render summary cutoff template")?;
+                summary_html.push_str(&summary_cutoff);
```
Keats, line 836: Do we want a template or a string in the config? Template is more powerful but might be overkill?

clarfonthey: I feel like the template is ideal, although it's unlikely many people will change the template. The original idea was proposed as an easy option since even `anchor-link.html` has a simple template associated with it.

Without a template, you can't choose whether to emit an ellipsis based upon the closing punctuation, so, I'm inclined to say that the template is strictly better.

components/markdown/src/markdown.rs
clarfonthey: Since that's just straight-up invalid HTML, I'd be more inclined to just make the `<!-- more -->` regexp only match if it's by itself (i.e. not nested inside something like that). Will add a test for it.

Personally, I think that an error would be weird; we don't emit errors for invalid HTML otherwise, so, it's better to just ignore it and treat it like otherwise malformed HTML.

Keats: Fair enough

clarfonthey: So, I tried adding a test for this, but the markdown parser just isn't happy with it. If it's valid HTML, it parses, and if it's not, it just escapes it. So, in your example, it shows up as `&lt;i` at the end of the summary.

I'm inclined to just say that's okay, honestly. We could detect if it's inside any custom HTML tags, like `<i><!-- MORE --></i>`, but that goes against the principle that we aren't accounting for custom HTML tags at all and those just remain opened.

Keats: Sounds good

components/templates/src/builtins/summary-cutoff.html
```diff
@@ -0,0 +1 @@
+{% if summary is matching("\PP$") %}&hellip;{% endif %}
```
Keats: Can we default to an ellipsis without the test and mention that in the docs as an example?
`matching` will currently create a new regex everytime this template is called so it's not ideal perf wise as a default

Keats (changes requested):
One nit but good otherwise

components/templates/src/builtins/summary-cutoff.html
clarfonthey: Oh, I had no idea that was the case; I'm fine with that, then.

clarfonthey: FWIW, no idea where Tera 2 is currently, but I'd be willing to help figure out how to solve this problem there. Feels like it should be doable to let the functions know about the constant arguments and optimize themselves accordingly.

Keats: It's possible to fix in tera 1 I think yes, just have a cache for the compiled regexes

clarfonthey:
Small nudge @Keats; no rush, but wanted to see if there's anything else you'd like before merging this.

Keats (approved):
Sorry for the delay

PR #2591, opened 2024-07-31
fix(taxonomies): fix error messages

samueltardieu:
- Some error messages refer to `get_taxonomy_term_by_name` instead of `get_taxonomy_term`
- If a term cannot be found in a taxonomy, the error message erroneously pretends that the taxonomy itself is unknown instead of the term.

Sanity check:

## Code changes


Keats (approved):
Thanks!

PR #2592, opened 2024-07-31
chore(components/config): cleanup code

samueltardieu:
Cleanup code by using `Option::is_some_and()` and `Option::as_deref()` instead of recoding them.

Sanity check:

## Code changes

### DIFF
--- a/components/config/src/config/taxonomies.rs
+++ b/components/config/src/config/taxonomies.rs
@@ -32,18 +32,10 @@ impl Default for TaxonomyConfig {
 
 impl TaxonomyConfig {
     pub fn is_paginated(&self) -> bool {
-        if let Some(paginate_by) = self.paginate_by {
-            paginate_by > 0
-        } else {
-            false
-        }
+        self.paginate_by.is_some_and(|paginate_by| paginate_by > 0)
     }
 
     pub fn paginate_path(&self) -> &str {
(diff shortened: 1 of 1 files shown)


components/config/src/config/taxonomies.rs
```diff
@@ -32,18 +32,10 @@ impl Default for TaxonomyConfig {
 
 impl TaxonomyConfig {
     pub fn is_paginated(&self) -> bool {
-        if let Some(paginate_by) = self.paginate_by {
-            paginate_by > 0
-        } else {
-            false
-        }
+        self.paginate_by.is_some_and(|paginate_by| paginate_by > 0)
```
Keats, line 35: TIL about `is_some_and`

Keats (approved):
Thanks!

components/config/src/config/taxonomies.rs
samueltardieu: The method name is explicit but uncommon. It is often called `exists()` ([for example in Scala](https://www.scala-lang.org/api/2.13.3/scala/Option.html#exists(p:A=%3EBoolean):Boolean)) in other functional languages, because it can be applied to construct of more than one element, such as lists, with the same semantics.

I've also seen `self.paginate_by.into_iter().any(|paginate_by| paginate_by > 0)`  used before `is_some_and()` was introduced.

PR #2600, opened 2024-08-07
Clarify shortcode usage

rikhuijzer:
I was reading the [Writing a shortcode](https://www.getzola.org/documentation/content/shortcodes/#writing-a-shortcode) section and got confused about how exactly to use it. There is a nice example a few paragraphs below, but it's hard to find upon a first read.

I've looked into moving some paragraphs around, but I didn't see a better order. So instead this PR suggests to add a link to the example.

image


Keats (approved):

Keats:
Thanks

PR #2601, opened 2024-08-07
fix: shortcode in inline elements

ta-nishq:
### BTS
Step 1: Minimum reproducible example (to break shortcodes)
```markdown
<a href="{{ basic() }}">{{ basic() }}</a>
```

Step 2: Investigating code which references `@@ZOLA_SC_PLACEHOLDER@@`
- markdown.rs (renders markdown, point of interest)
- mod.rs
- parser.rs

Step 3: Investigating `shortcode.render` call in `markdown.rs`
- for samples which are rendered correctly the `shortcode.render` was invoked for each shortcode
- for samples which render `@@ZOLA_SC_PLACEHOLDER@@`, `shortcode.render` was **NOT** invoked
- source of shortcodes: `html_shortcodes` list (which had the shortcode references even if they were not rendered)

Step 4: Investigating missing shortcodes
- `!$range.contains(&shortcode.span.start)` skips shortcodes which are not in the range
- Understanding `range` and `shortcode.span`
- Understanding parsed content (cmark markdown parser)
  - `iterator(event, range)` where range -> range in the markdown source

Step 5: Investigating ranges
```plaintext
content: <pre>•  <a href="@@ZOLA_SC_PLACEHOLDER@@">@@ZOLA_SC_PLACEHOLDER@@</a>•</pre>••<a href="@@ZOLA_SC_PLACEHOLDER@@">@@ZOLA_SC_PLACEHOLDER@@</a>
note: • represents newline (replaced for easy offset calculation)
parse_tree:
  Start(HtmlBlock) 0..77
  Html(Borrowed(<pre>•)) 0..6
    Html(Borrowed(  <a href="@@ZOLA_SC_PLACEHOLDER@@">@@ZOLA_SC_PLACEHOLDER@@</a>•)) 6..70
  Html(Borrowed(</pre>•)) 70..77
  End(HtmlBlock) 0..77
  Start(Paragraph) 78..140
    InlineHtml(Borrowed(<a href="@@ZOLA_SC_PLACEHOLDER@@">)) 78..112
    Text(Borrowed(@@ZOLA_SC_PLACEHOLDER@@)) 112..135
    InlineHtml(Borrowed(</a>)) 135..139
  End(Paragraph) 78..140
shortcodes:
  17..40
  42..65
  87..110
  112..135
logs:
  Start(HtmlBlock) 0..77
  Html(Borrowed(<pre>•)) 0..6
[...]

PR #2606, opened 2024-08-09
Fix shortcode/continue-reading parsing with inline HTML (smaller version)

### LINKED ISSUE #2565: Broken shortcodes on 0.19
I had a shortcode called `email` which contained my email address for reasons, and used it as such

```
<a name="contact-me" href="mailto:{{ email() }}">{{ email() }}</a>
```

This used to work but I have since upgraded to 0.19.1 and it simply outputs @@ZOLA_SC_PLACEHOLDER@@

I had to move the entire tag into the shortcode for it to work again, but this smells like a bug and a regression to me.

clarfonthey:
Version of #2581 with the new summary cutoff behaviour removed. (It now cuts off mid-paragraph like before, but doesn't do anything special about it.)

Done on master branch so it can presumably be released as a new micro version, but can move to next branch if preferred.

Supersedes #2601, fixes #2565.

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -684,9 +684,7 @@ pub fn markdown_to_html(
                         event
                     });
                 }
-                Event::Html(text) | Event::InlineHtml(text)
-                    if !has_summary && MORE_DIVIDER_RE.is_match(text.as_ref()) =>
-                {
+                Event::Html(text) if !has_summary && MORE_DIVIDER_RE.is_match(text.as_ref()) => {
                     has_summary = true;
                     events.push(Event::Html(CONTINUE_READING.into()));
                 }
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -36,6 +36,10 @@ static MORE_DIVIDER_RE: Lazy<Regex> = Lazy::new(|| {
         .unwrap()
 });
 
+static FOOTNOTES_RE: Lazy<Regex> = Lazy::new(|| {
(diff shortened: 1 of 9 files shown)


components/markdown/tests/summary.rs
```diff
@@ -45,3 +42,32 @@ And some content after
...
+* Program <!-- more --> something
+* Eat
+* Sleep
+    "#,
+    );
+    insta::assert_snapshot!(body);
+}
+
+#[test]
+fn foontnotes_summary() {
```
ta-nishq: nit: maybe you meant this (similar change for snap file)
```suggestion
fn footnotes_summary() {
```

components/markdown/src/markdown.rs
```diff
@@ -679,17 +684,15 @@ pub fn markdown_to_html(
...
-                        has_summary = true;
-                        events.push(Event::Html(CONTINUE_READING.into()));
-                        continue;
-                    }
-                    if !contains_shortcode(text.as_ref()) {
-                        events.push(Event::Html(text));
-                        continue;
-                    }
-
+                Event::Html(text) | Event::InlineHtml(text)
```
Keats: I believe it might better to remove the InlineHtml from there for now so you don't get broken HTML

Keats (changes requested):
That should be on next still, there will be a 0.19.2 release after this gets merged.

components/markdown/src/markdown.rs
clarfonthey: That's fair; I'll make that change but leave in the test that shows that there's no summary when this happens.

Keats (approved):

Keats:
Nice thanks

Issue #2604, opened 2024-08-09
Zola panics if static directory does not exist (with compile_sass = false)


ta-nishq:
# Bug Report

## Environment
OS: macOS 14.5
Zola version: 0.19.1
Processor Class: arm

## Expected Behavior
If static folder is required then the error should point to this (also this should be consistent irrespective of compile_sass flag). This is not a big issue (for me), but documenting it.

## Current Behavior
Zola panics if static folder does not exist and compile_sass is set to false

## Step to reproduce
- create new zola site by `zola init test`
- ensure to disable sass compilation
- remove static directory `rmdir static`
- run `zola serve`

## Stacktrace
```bash
RUST_BACKTRACE=full zola serve

Building site...
Checking all internal links with anchors.
> Successfully checked 0 internal link(s) with anchors.
-> Creating 0 pages (0 orphan) and 0 sections
Done in 10ms.

thread 'main' panicked at src/cmd/serve.rs:507:59:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
stack backtrace:
   0:        0x102b9e304 - __mh_execute_header
   1:        0x102253b84 - __mh_execute_header
   2:        0x102b8d30c - __mh_execute_header
   3:        0x102b9e120 - __mh_execute_header
   4:        0x102b9a2d4 - __mh_execute_header
   5:        0x102b9b8d0 - __mh_execute_header
   6:        0x102b9e698 - __mh_execute_header
   7:        0x102b9e600 - __mh_execute_header
   8:        0x102b9a79c - __mh_execute_header
   9:        0x102d987fc - __mh_execute_header
  10:        0x102d98c58 - __mh_execute_header
  11:        0x1020f1008 - __mh_execute_header
  12:        0x1020fd024 - __mh_execute_header
  13:        0x1020aa8f0 - __mh_execute_header
  14:        0x102109664 - __mh_execute_header
```

Keats:
Ok this one is easy to fix, there's an unwrap at the line pointed in the snippet which we should handle.

PR #2608, opened 2024-08-10
fix: zola panic if no static dir

ta-nishq:
Fixes: #2604

### Root Cause:
- Application panicked at this code `std::fs::canonicalize(&output_path).unwrap()` when there is no static directory and compile_sass is set to false
- Reason for this is `output_path` (which in default case is public) does not get created
  - Folder gets created when there is a static folder or we have compile_sass set
- Added `create_directory` code to create the directory if it does not exist before canonicalize

### Alternate fix
Handle unwrap and show an error


Keats (approved):
Not entirely sure why it's causing an issue because of the static folder but lgtm

PR #2623, opened 2024-08-26
add optional "filter" type to image resize function

aaron404:

Sanity check:

## Code changes

- I have added a new `filter` parameter to the `resize_image` function. It defaults to `lanczos3` which is what was previously hardcoded. I had to add a mirror of the `Filter` type from the `image` crate to impl `Hash` for it. I'm not sure if there's a cleaner way to do this.
- I added a sprite from [pixelfrog](https://pixelfrog-assets.itch.io/tiny-swords) that has public domain usage rights to help demonstrate the differences of the filters in the updated docs.


aaron404:
I'm not sure if I have incorrectly configured something, but it seems that running `cargo test` from the repo root does not execute all the tests that are used by the automated builder. I have a few test cases to fix up

Keats:
`cargo test` will only run the root crate, you need to run `cargo test --all` to run tests for all the crates in a workspace

components/imageproc/src/ops.rs
```diff
@@ -46,12 +53,21 @@ impl ResizeOperation {
...
         };
 
+        let filter = match filter {
+            "lanczos3" => FilterType::Lanczos3,
+            "nearest" => FilterType::Nearest,
+            "triangle" => FilterType::Triangle,
+            "catmullrom" => FilterType::CatmullRom,
+            "gaussian" => FilterType::Gaussian,
+            _ => return Err(anyhow!("Invalid filter type: {}", filter)),
+        };
```
Keats, line 63: Can we convert directly to the filter type from the image crate and skip our own?

aaron404: I went with a custom type because FilterType from the image crate is not `Hash`, so it breaks downstream functionality when the processed image paths are generated.

Keats: I've just made a PR to impl it, let's see if it works

Keats: i think it's released in the most recent version

Keats:
@aaron404 are you interested in updating that PR?

aaron404:
@Keats I think I have missed something. You mentioned you made a PR to impl it (I assume impl `Hash` for `image`'s `FilterType`, but I wasn't able to find it. Can you point me to it?

Keats:
It's https://github.com/image-rs/image/commit/02a4928df9c14bdfa25594facd83391a820b6271
It should be available in recent `image` version

aaron404:
It looks like we also need FilterType to impl `Eq`. I have opened a [PR](https://github.com/image-rs/image/pull/2397) for that

aaron404:
Hmm.. `FilterType` also needs to be `Serialize` and `Deserialize` for use in `EnqueueResponse`. I'm not sure if it makes sense for `image` to impl these traits, but they do use a `serde` cfg feature, so it could be impl'd behind that check. I will submit a PR and see if they are open to it.

Keats:
Looks like both PRs have been merged?

Keats:
Replaced by https://github.com/getzola/zola/pull/3197

PR #2629, opened 2024-09-03
feat: build relases on gh-actions & support linux arm64

### LINKED ISSUE #2620: Docker image for arm64
Currently there seems to be only a Docker image available for the amd64 platform. I would have liked to run the Docker image on my server with arm64 architecture.

## Expected behaviour

Expected the following to work:

```
$ docker run --platform linux/arm64 -u "$(id -u):$(id -g)" -v $PWD:/app --workdir /app ghcr.io/getzola/zola:v0.19.1 build
```

## Actual behaviour

The above command fails with:

```
Unable to find image 'ghcr.io/getzola/zola:v0.19.1' locally
v0.19.1: Pulling from getzola/zola
Digest: sha256:6e4f78e3f1338b2f117e4a88106ddf01bc2727bf7dcd06a97a8ea36e0fe14edd
Status: Image is up to date for ghcr.io/getzola/zola:v0.19.1
docker: image with reference ghcr.io/getzola/zola:v0.19.1 was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64.
```

## Fix

As a workaround I manually compiled zola on the server and it worked like a charm. But adding arm64 releases to the Docker Hub would benefit other users as well. Could we do that?

I think adding something like `platforms: linux/amd64,linux/arm64/v8` to [cd-workflow.yml](https://github.com/getzola/zola/blob/38fddc692a99e7a5cc7a7dcd44ad605c976ccb17/.github/workflows/cd-workflow.yml#L44) might be enough already.

hegerdes:

Sanity check:

## Code changes

## What changed

Fixes #2620

This ports the azure-devops pipeline to GH-Actions to allow for easy docker arm build all in one platform.   
This adds a build & test job that that runs `cargo test --all`, `cargo fmt --check` and `cargo build --all`. All transferred from Azure Pipelines. This is run for the following targets:
 * x86_64-unknown-linux-gnu
 * aarch64-unknown-linux-gnu
 * x86_64-pc-windows-msvc
 * x86_64-apple-darwin
 * aarch64-apple-darwin

All target are naively compiled (also macos arm since GH has M1 runners) except for linux arm64. Due to GitHub limiting hosted arm runner to paid plans we need to cross-compile.

A test job can be found [here](https://github.com/hegerdes/zola/actions/runs/10690445901/job/29634833945)

The same setup is implemented for release which runs only on tags matching `v*`. All binaries are build uploaded as an artifact, singed via [GitHub attestations](https://github.blog/news-insights/product-news/introducing-artifact-attestations-now-in-public-beta/) and finally one release is created. See [demo workflow run](https://github.com/hegerdes/zola/actions/runs/10690464632)

The Docker images are build after that. There is an optional build arg `USE_GH_RELEASE` which defaults to `false`. If set to true it downloads the latest zola version (in the right ARCH) in the dockerfile instead of building it itself.  
This allows the (re)usage of already signed artifacts and avoids Docker QEMU emulation which would result in 1h+ build times. Users should be fine building it locally with the current source by running `docker build --tag zola .`

### To discuss:
1. In order to cross-compile I needed to create a file `.cargo/config.toml` and setting the liker to `aarch64-linux-gnu-gcc` for that arch. I have no real knowledge in rust and rust cross-compiling. Is there a better way, can we avoid that file? -> FIXED by using `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER` in GH Action only - this does not affect anyone build

3. Check the file names in the releases. I created a demo release on my fork (will delete later). Please double check: https://github.com/hegerdes/zola/releases

4. Is the approach with the optional arg in docker fine to download existing releases?


hegerdes:
Do you have runtime stats for azure-devops?

Just to compare:
image

**Note:** It does not take 45min. All jobs except for the last two run in parallel.

Keats:
The last release:

Screenshot 2024-09-05 at 10 29 11

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - x86_64-unknown-linux-gnu
+          - aarch64-unknown-linux-gnu
+          - x86_64-pc-windows-msvc
+          - x86_64-apple-darwin
+          - aarch64-apple-darwin
```
Keats, line 25: Do we need to run tests on different archs for the same OS?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,74 @@
...
+            echo "Compiling for ${{matrix.target}}"
+          fi
+        if: ${{ ! startsWith(matrix.os, 'windows') }}
+
+      - name: Install Rust (Windows)
+        run: |
+          curl -sSf -o rustup-init.exe https://win.rustup.rs
+          ./rustup-init.exe -y --default-toolchain ${{ matrix.rustup_toolchain }} --default-host x86_64-pc-windows-msvc
+          echo "##vso[task.setvariable variable=PATH;]%PATH%;%USERPROFILE%\.cargo\bin"
+        if: ${{ startsWith(matrix.os, 'windows') }}
```
Keats: Let's use https://github.com/dtolnay/rust-toolchain to install Rust, this was needed in Azure because there was no custom actions

.github/workflows/release.yml
```diff
@@ -0,0 +1,173 @@
...
+          else
+            echo "Compiling for ${{matrix.target}}"
+          fi
+
+      - name: Install Rust (Windows)
+        if: ${{ startsWith(matrix.os, 'windows') }}
+        run: |
+          curl -sSf -o rustup-init.exe https://win.rustup.rs
+          ./rustup-init.exe -y --default-toolchain ${{ matrix.rustup_toolchain }} --default-host ${{ matrix.target }}
+          echo "##vso[task.setvariable variable=PATH;]%PATH%;%USERPROFILE%\.cargo\bin"
```
Keats: same here, let's see if we can remove this custom code

Dockerfile
```diff
@@ -1,15 +1,21 @@
...
+RUN apt-get update -y && apt-get install -y make g++ libssl-dev curl jq tar gzip
 
 WORKDIR /app
 COPY . .
 
-RUN cargo build --release --target x86_64-unknown-linux-gnu
-
+RUN if [ "${USE_GH_RELEASE}" = "true" ]; then \
+    mkdir -p target/$(uname -m)-unknown-linux-gnu/release && \
+    export ZOLA_VERSION=$(curl -sL https://api.github.com/repos/getzola/zola/releases/latest | jq -r .name) && \
```
Keats: This is not reproducible, can you pass the version to the Dockerfile instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,173 @@
...
+
+            All artifacts are signed with this repos identity using Sigstore.
+            You can verify the signatures using the `GitHub` CLI.
+
+            ```shell
+            gh attestation verify --owner ${{ github.repository_owner }} <my-artifact>
+            ```
+          token: ${{ secrets.GITHUB_TOKEN }}
+          prerelease: ${{ contains(github.ref, '-pre') }}
+          files: artifacts/*
```
Keats, line 123: What's the reason behind splitting the Release-Build and Release jobs? Why not do it in 1 job?

.github/workflows/release.yml
```diff
@@ -0,0 +1,173 @@
...
+          generate_release_notes: true
+          fail_on_unmatched_files: true
+          body: |
+            Welcome to this new release of Zola ${{ github.ref_name }}!
+
+            All artifacts are signed with this repos identity using Sigstore.
+            You can verify the signatures using the `GitHub` CLI.
+
+            ```shell
+            gh attestation verify --owner ${{ github.repository_owner }} <my-artifact>
```
Keats, line 119: Right now Azure generates a list of commits since the last release (https://github.com/getzola/zola/releases/tag/v0.19.2), is it possible to keep it?

hegerdes: Yeah it should generate release notes with `generate_release_notes: true`. My demo release didn't have any real changes since it was the first release and I'm not working with PRs for active development and debugging.

I'm also using this action here: https://github.com/hegerdes/joplin-plugin-remote-note-pull/releases/tag/v1.3.0 and Pulumi also uses uses with some more changes than my small project, see https://github.com/pulumi/pulumi-awsx/releases

I think they might look a litte different to azures notes though.

Feel fee to add whatever you want to that release body you are the owner, it was just an example for now.

Dockerfile
hegerdes: Totally agree. Just wanted to check if this approach is fine before going all the way.  
I would pass an extra arg like `ZOLA_VERSION` with. I would set the default to latest and override in CI to the current tag. To enforce reproducibility we could also just fail if `USE_GH_RELEASE` is set but `ZOLA_VERSION` is not. Any takes on that?

.github/workflows/build-and-test.yml
hegerdes: GitHubs Action matrix is a little counter-intuitive. A job for each `target` (currently 5) and each `rustup_toolchain` (currently 1) is created. The `include` part sets the os. We use the native os for each target, except linux arm. I thought testing on the native arch is best when ever possible - but I know very litte about the rust build standards.

If you think testing with cross compiling is better or also needed required let me know the exact test matrix.

.github/workflows/release.yml
hegerdes: The build for each target run all on their own isolated runner. So the targets can be build native and in parallel. The runners don't share any storage so wee need to get all build assets on one runner. This is done via upload/download artifacts. The artifacts assets are temporary and will auto delete in one week.

When each runner creates a release we can have race conditions or partly failed release jobs. This ensures the release build for every target worked and only then one release is created. If one target fails -> no release.

The argument `prerelease` would allow you to create pre-releases when the tag looks like `v0.1.0-pre.0` It will not be marked as latest.

Dockerfile
Keats: Sounds good

Keats:
Nice, thanks!

hegerdes:
Thanks for merging this @Keats :) I really checked for errors but I would still recommend doing a pre-release once before switching completely over to the new release setup. A tag containing `-pre` should do the trick. 

If anything does not look right feel free to tag me anytime

hegerdes:
@Keats 
GitHub just released the Linux ARM runners to the public [yesterday](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/). So these could be used to avoid cross compiling for arm. It would also simplify CI a little (no extra args or target specific code).

If you are interested in it I can provide a follow-up PR. A demo CI-Run with build artefacts can be found here: https://github.com/hegerdes/zola/actions/runs/12817720665. I tested the release binaries on Linux (x86/arm) and Windows

Keats:
It can be added yes, thanks!

PR #2644, opened 2024-09-19
Remove brackets from `bottom_footnotes`

### LINKED ISSUE #2640: `bottom_footnotes` config adds unwanted square brackets to footnote references
Hello.

The square brackets around footnotes references applied when `bottom_footnotes` config is set is not documented anywhere, and honestly, looks ugly. They should either be removed or made as a config option.

The removal is a tiny change I made in my fork here [here](https://github.com/getzola/zola/commit/460e61a4b4bd7dc405459a7094cfb896535ca074). I did not want to create a pull request, however, without clearing that it is ok to do so.

The additional config option would be a bit more involved: Upgrade the `bottom_footnotes` config to an enum with `Default`, `Bottom`, `Github-style` etc.

asibahi:
closes #2640


Keats:
Can you do the PR on the next branch instead?

asibahi:
Changed the base branch from Github's interface but it seems to have added all of the main branch commits as part of this PR ?

PR #2645, opened 2024-09-19
remove square brackets

### LINKED ISSUE #2640: `bottom_footnotes` config adds unwanted square brackets to footnote references

asibahi:
closes #2640

redo of #2644

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -277,7 +277,7 @@ fn convert_footnotes_to_github_style(old_events: &mut Vec<Event>) {
                 // nr is a number of references to this footnote
                 let (n, nr) = footnote_numbers.entry(name.clone()).or_insert((n, 0usize));
                 *nr += 1;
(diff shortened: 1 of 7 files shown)


Keats:
Looks like we need to re-run the insta tests. Let me know if you need help with it or I can do it

asibahi:
Unfortunately CI and CI tests are voodoo to me. Why would the tests fail?

Keats:
We have integration tests testing the output of markdown rendering and they need to be updated. I'll handle it when I have a bit of time

Keats:
Pushed it to next but i fucked up the merge somehow so GH didn't recognise it :/

Issue #2473, opened 2024-04-13
No `paginator` in `get_section()`


charlesrocket:
There is no way to get paginator's page (`page/13`, etc) permalink due to standalone variable (#691).

Trying to work around this but `get_url()` refuses to error:

```
{% set foo = get_url(path="/posts/page/10142342340101263162411010112") %}
```

`foo` should fail in this case since the test site has only 10 posts with the section's pagination set to `2`.

Keats:
`get_url` doesn't check whether the path exists as some stuff gets added during the build process.

charlesrocket:
@Keats oh snap, that was my last idea. There are still sitemaps, but I have a grim feeling. All the work on the dynamic PWA cache just got cancelled because there's nowhere to pull pager permalinks from.

Keats:
Once we have tera2 it should be possible to revamp all the functions to get the paginator but it's far away

charlesrocket:
@Keats what about something like `validate_permalink()` since Tera 2 is not happening any time soon?

charlesrocket:
While there is no built-in solution, one could manually extract pagination settings and calculate required page numbers: charlesrocket/halve-z/pull/27

ZzMzaw:
Hello,

In order to set up a 'series' feature in a theme, I would need to know if pagination is reversed or not so that I can align series' pages indexes accordingly.
Everything works well in the series itself (which is a section) has I have both `section` and `paginator` variables.
Nevertheless, in the series' pages, I can just do a `get_section` to retrieve the series' section information and I don't have any way to retrieve pagination information.
Unfortunately, applying the workaround just above would not work in my case because my series' section has some content (and not only frontmatter).

Reversing the order of the pages listed in the section when `paginate_reversed` (or maybe something more meaningful such as `sort_reversed`) is set to `true` would work as well for me but I would expect it to have such a huge impact that it doesn't worth it in my opinion.

Even if adding the entire paginator seems not possible, do you think it would be possible to just add `paginate_by` and `paginate_reversed` in the serialized section returned by `get_section`?
https://github.com/getzola/zola/blob/051becaa1c681165c9f32d30e4e4bde3952765bc/components/content/src/ser.rs#L206-L229

If yes, I can try to tackle it and propose a PR in the weeks to come.
If you would prefer me to open a new issue instead of reusing this one, please tell me.

Thanks for your help and support.

Keats:
[...]

PR #2653, opened 2024-09-29
Add pagination info to sections

ZzMzaw:
Objective of this small update is to add some minimal information about pagination in the section itself for cases where the paginator is not available (e.g. when using `get_section`).

A small site for testing the feature is attached to this PR: [testsite.zip](https://github.com/user-attachments/files/17179446/testsite.zip).

This feature has been discussed in issue [2473](https://github.com/getzola/zola/issues/2473).

Sanity check:

## Code changes

### DIFF
--- a/components/content/src/ser.rs
+++ b/components/content/src/ser.rs
@@ -162,6 +162,9 @@ pub struct SerializingSection<'a> {
     backlinks: Vec<BackLink<'a>>,
     generate_feeds: bool,
     transparent: bool,
+    paginated: bool,
+    paginate_by: Option<usize>,
+    paginate_reversed: Option<bool>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


components/content/src/ser.rs
```diff
@@ -162,6 +162,9 @@ pub struct SerializingSection<'a> {
     backlinks: Vec<BackLink<'a>>,
     generate_feeds: bool,
     transparent: bool,
+    paginated: bool,
```
Keats: No need for that one, let's match what is in the frontmatter

Keats (changes requested):

components/content/src/ser.rs
ZzMzaw: I just removed the unecessary `paginated` attribute and retrieved raw values from the frontmatter.

Adapted testsite: [testsite.zip](https://github.com/user-attachments/files/17345078/testsite.zip)

Keats (approved):

Issue #2663, opened 2024-10-05
Bug: File watching during `zola serve` failing with Zola 19.2 on macOS 14.6.1


jamiedumont:
# Bug Report

## Environment

Zola version: 19.2
macOS: Sonoma 14.6.1

## Expected Behavior
File watching and live reload should detect file changes, rebuilding and reloading site whilst `zola serve` is running

## Current Behavior
`zola serve` behaves as expected  with a new Zola project but not with existing projects. This behaviour was noted in 19.1 but remained after upgrading to 19.2. It would appear that file changes aren't being detected as nothing appears in terminal running command.

## Step to reproduce
The watcher stops working after adding some files to a new Zola project. I've tried to pin point which files, tags, etc cause it to fail; but it's quite inconsistent. From a new project, adding something like a CSS file to `static` works, with changes detected. Adding an `index.html` to `templates` is detected in the first instance, but edits to that file aren't. Once the issue kicks in, changes in previously working files (like CSS in my example) aren't detected. The only way to return to a working state is to discard all changes with git.

I've been experiencing this for a number of weeks and working around it by restarting `zola serve` to view changes. I can't believe others will have been experiencing this without reporting it, so it must be specific to my environment and install.

I'll continue to test — perhaps removing Zola from my computer and reinstalling.

---

**EDIT** — Dropping back to `v18` and everything works as expected. Testing various versions I've been able to determine that — in my environment at least — this bug was introduced with `v19.0`. [This PR
](https://github.com/getzola/zola/pull/2503) seems the most likely change.

I'll pull down main and start pulling it apart to work out where the issue is on Monday.

jamiedumont:
Solved my own issue.

Worked out that my editor — Nova on macOS — is creating temporary files, which is failing the check `fn is_temp_file()` in file `/src/fs_utils.rs` at line 190. Sublime Text works correctly.

I'll dig into exactly which bit of `is_temp_file()` is an issue with Nova, and report back. 

Very niche bug as it's both dependent on Zola version — `> 18.0` — and Nova; although other editors may be an issue too.

jamiedumont:
Turns out that when Nova 11.10 is saving changes to an existing file, `notify_debouncer_full` is reporting the change as `EventKind::Modify(ModifyKind::Name(RenameMode::Any))` whereas other editors (Sublime in this case) create a `EventKind::Modify(ModifyKind::Data(DataChange::Content))`.

[...]

PR #2672, opened 2024-10-16
Catch editing events emitted by Nova on macOS

jamiedumont:
As per the bug report https://github.com/getzola/zola/issues/2663, I've added a branch to the match expression to catch `EventKind::Modify(ModifyKind::Name(RenameMode::Any))` events made by Nova on macOS when editing files.

It appears that this is the only event that makes it to `meaningful_events: HashMap<PathBuf, (PathBuf, SimpleFileSystemEventKind, ChangeKind)>` when using Nova.

In testing this change, I discovered that other editors including Helix 24.7 and Vim 9.0.2142 also emit this event, but go on to emit others.

This change didn't introduce problems when editing with:
* Helix 24.7
* Vim 9.0.2142
* Sublime Text Build 4180
on macOS 14.6.1.

Let me know if you want further testing (or tests) added.

Cheers, Jamie.


jamiedumont:
Just realised my branch is based on main and bringing in other commits. Apologies. I’ll close this and make a new PR based on next.

PR #2688, opened 2024-11-06
Improve accessibility by nesting bottom footnotes inside a footer element

ghost:
This adds semantic meaning to the footnotes organization and improves accessibility by aiding disabled users who rely on assistive devices such as screen readers which utilize semantic tags such as the footer element in order to navigate properly.

The hr element is semantically defined as representing a paragraph-level thematic break. Now that the footnotes are descendants of the footer element, the hr element originally preceding the footnotes list is no longer necessary (footnotes are no longer paragraph-level content) and thus replaced. However, the footer element is given the class "footnotes" to allow for styling, so the following CSS could be used to provide a stylistically equivalent visible border separating the footnotes from the content if so desired:

```css
    .footnotes {
      border-top: 2px groove gray;
    }
```

Zola Discourse link: https://zola.discourse.group/t/improve-accessibility-by-nesting-bottom-footnotes-inside-a-footer-element/2356

Sanity check:

## Code changes

This update does not require any interaction from the user to make use of it. However, if would be helpful, I could make a documentation change to mention the visible border CSS suggestion to users.

### DIFF
--- a/components/markdown/src/markdown.rs
+++ b/components/markdown/src/markdown.rs
@@ -304,7 +304,8 @@ fn convert_footnotes_to_github_style(old_events: &mut Vec<Event>) {
         return;
     }
 
-    old_events.push(Event::Html("<hr><ol class=\"footnotes-list\">\n".into()));
+    old_events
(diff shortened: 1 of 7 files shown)


Keats (approved):

Keats:
Thanks!

PR #2717, opened 2024-11-30
Add external links class option to Markdown configuration

welpo:

Sanity check:

## Code changes

---

Discussed in https://zola.discourse.group/t/style-internal-and-external-links-differently/1501/6.

This PR adds an optional `external_links_class` configuration option that allows adding a CSS class to all external links.

This makes it easier to style external links differently from internal ones, for example to add visual indicators for external links.

Example configuration:
```toml
[markdown]
external_links_class = "external"
```

with CSS:

```css
a.external::after {
    display: inline-block;
    vertical-align: middle;
    mask-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cg fill='none' stroke='currentColor' stroke-linecap='round' stroke-linejoin='round' stroke-width='2'%3E%3Cpath d='M11 5h-6v14h14v-6'/%3E%3Cpath d='M13 11l7 -7'/%3E%3Cpath d='M21 3h-6M21 3v6'/%3E%3C/g%3E%3C/svg%3E");
    margin-left: 0.1em;
    background-color: currentColor;
    width: 1em;
    height: 1em;
    content: "";
}
```

It looks like:

external_links_class

Note: for the docs, I wasn't sure whether to do (same as `author`):

```toml
external_links_class =
```

or (make it super obvious it's a string):

```toml
external_links_class = ""
```

Opted for `author` format, for consistency.

Happy to make any necessary changes!


Keats (approved):
LGTM

components/config/src/config/markup.rs
```diff
@@ -36,6 +36,8 @@ pub struct Markdown {
     pub highlight_themes_css: Vec<ThemeCss>,
     /// Whether to render emoji aliases (e.g.: :smile: => 😄) in the markdown files
     pub render_emoji: bool,
+    /// CSS class to add to external links
+    pub external_links_class: Option<String>,
```
Keats, line 40: ah maybe just one thing, should we error if the string contains `"`?

welpo: Setting `external_links_class = '"external"'` generates:

```html
<a external""="" href="https://www.getzola.org">Zola</a>
```

I could see someone trying to use quotes to assign two classes to external links, so validation there might help.

Since CSS classes can't include quotes (neither `'` nor `"`), should we check for both?

Would you prefer validation be done in `construct_external_link_tag`, when we try to use the class? Or early in `impl Markdown` (`markup.rs`):

```rust
impl Markdown {
    pub fn validate_config(&self) -> Result<()> {
        // Validate external link class doesn't contain quotes which would break HTML and aren't valid in CSS
        if let Some(class) = &self.external_links_class {
            if class.contains('"') || class.contains('\'') {
                bail!(
                    "External link class '{}' contains quotes which are not valid in CSS class names",
                    class
                )
            }
        }
        Ok(())
    }
// …
```

Keats: > Since CSS classes can't include quotes (neither ' nor "), should we check for both?

Yes. And definitely early in the config for the error checking

welpo: Done! I did not add a test to validate errors since (1) I didn't find other tests checking for specific error messages and (2) `external_links_class` validation happens in `Config::from_file()`, outside the test infrastructure.

components/markdown/tests/markdown.rs
```diff
@@ -133,6 +133,32 @@ fn can_use_smart_punctuation() {
...
+    // reset class, set target blank only
+    config.markdown.external_links_class = None;
+    config.markdown.external_links_target_blank = true;
+    let body = common::render_with_config("<https://google.com>", config.clone()).unwrap().body;
+    insta::assert_snapshot!(body);
+
+    // both class and target blank
+    config.markdown.external_links_class = Some("external".to_string());
+    let body = common::render_with_config("<https://google.com>", config).unwrap().body;
+    insta::assert_snapshot!(body);
```
Keats: is it possible to give a name or annotation to these snaps?

welpo: Done! I've taken the liberty to rename the snapshots from `can_use_external_links_options` for consistency and clarity.

Keats (approved):
Perfect thanks

mrec:
Sorry to necropost but just curious: what was the rationale for having a customizable `class` here, rather than using the standard `rel="external"` intended for exactly this purpose?

> Relevant to `<form>`, `<a>`, and `<area>`, it indicates the referenced document is not part of the current site. This can be used with attribute selectors to style external links in a way that indicates to the user that they will be leaving the current site.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel#external

welpo:
@mrec I did not know of that standard!

I believe that's a cleaner way to achieve the same goal.

Zola should probably add `rel="external"` by default to all external links, imo (or at least offer a `markdown`-section config).

mrec:
@welpo if nothing else I suspect it'd play better with accessibility tools - `rel="external"` has meaning to e.g. screenreaders in a way that an arbitrary CSS class doesn't. 

Obviously Zola couldn't drop support for the class since people are using it, but marking that as legacy and adding the standard attribute as a new config option sounds like a good idea. (I'm new here, but Zola overall seems to be very reluctant to start silently changing HTML output, possibly to avoid stepping on the toes of people who are making similar adjustments themselves via other routes.)

welpo:
@mrec that sounds good to me, but I'm just a contributor.

You could discuss the idea on a new issue and/or in discourse: https://zola.discourse.group/

Keats:
That would be fine to add yes, I forgot about that rel thing

ccjmne:
I'm also interested in generating `rel="external"` attributes for external link, either by default, or behind a configuration flag.

@Keats, would you be okay with a PR implementing that right away, or do you think it's better to have a thread on the Discourse forum first?

Keats:
by default is fine, no need for a thread

PR #2720, opened 2024-11-30
Added tags support for themes

harrymkt:

Sanity check:

## Code changes


docs/config.toml
```diff
@@ -5,6 +5,10 @@ description = "Everything you need to make a static site engine in one binary."
 compile_sass = true
 build_search_index = true
 
+taxonomies = [
+  { name = "tags"}
```
Keats: they should probably be named theme_tag no? In case we ever want other types of tags

Keats:
Can you remove the changes to the processed images?

harrymkt:
Ok, it's resolved now. I previously used the build command to test and that images got processed

docs/templates/themes.html
```diff
@@ -9,6 +9,7 @@
     <div class="themes-container">
         {% block theme_content %}
             <h1>Zola themes</h1>
+            <p><a href="{{ get_url(path = "theme-tags") }}">Search by tags</a></p>
```
Keats: I think I would wait until we actually have templates with tags before linking to it. No point linking to an empty page.

Keats (changes requested):

docs/templates/themes.html
harrymkt: Zluinav theme has that tags. Meanwhile I'll try finding another way out to see if the link can be invisible if there is no defined content with `theme-tags` taxonomy

Issue #2598, opened 2024-08-05
Undocumented or Faulty: linking colocated assets on a slug page through template


Yamavu:
# Documentation issue

## Summary
Listing all colocated assets of a page in a template `page.html` . The content `index.md` uses a **slug**

(tested on versions 0.18.0 and 0.19.1)

Files linked in markdown are unaffected.

file setup:
```txt
_index.md
page/
  image1.jpg
  image2.jpg
  index.jpg
```

From the documentation:
```html
<div>
{% for asset in page.assets -%}
  {%- if asset is matching("[.](jpg|png)$") -%}
    {% set image = resize_image(path=asset, width=240, height=180) %}
    <a href="{{ get_url(path=asset) }}" target="_blank">
      <img src="{{ image.url }}" />
    </a>
  {%- endif %}
{%- endfor %}
</div>
```

If the page also uses a custom **slug** variable, `{{ get_url(path=asset) }}` doesn't provide this correctly, 

Zola will only display the right output when the link is written like:
```
{{page.permalink | safe}}{{asset | split(pat='/') | last  | safe }}
```
Maybe I overlooked something, but linking all co-located assets seems to be quite essential for building almost any kind of art or photo gallery.

## Proposed solution
Either current documentation should be expanded or get_url() should include this functionality.

Raymi306:
I just ran into this as well.

Yamavu:
as far as I understood, _get_url ( path, lang="en", trailing_slash=false, cachebust=false )_ is also used internally which is why is simple and optimized. 

the function currently supposed to only work with MD files, not the page assets (even though the documentation suggests that). 

The function has a fallback option to also link to the static directory, and happens to work with un-sluggified content nodes too. It feels pretty alpha to me

Keats:
Basically it's for 2 things:

1. if it starts with `@/`, resolve the internal link
2. otherwise just make a permalink with what we've been given without checking anything

To handle 2 for assets with renamed slugs, it would require detecting it's a colocated asset, load the related page and generate the permalink from there (what you need to load exactly in the GetUrl fn is TBD, you don't want to add the whole site).
If someone wants to do a PR for that I would take it.

ZzMzaw:
I was wondering if adding permalinks of colocated assets alongside the one of the page would be an acceptable solution.

This would allow 1. to find assets through internal links which would solve the initial statement by just adding '@/' to the asset's path used in get_url.

[...]

PR #2726, opened 2024-12-06
Add internal resolution for assets in `get_url` function

ZzMzaw:
This PR solves issue #2598 through internal resolution for assets (i.e. using `@` in front of the asset path).

It allow to retrieve an assets permalink based on its associated page and according to the language.
Documentation regarding the galery shortcode has been updated accordingly.

Beware supports for resolving assets internally has only been implemented for the `get_url` function.
I am not sure if it should be generalized to links as well (as `get_url` can and should be used to feed the `href` of a link).

This PR could maybe provide an answer to the issue described by the PR #2146 (a not to-be-merged PR for demo purpose only) even if it doesn't fix it where the PR points out (not sure about this at all).

Sanity check:

## Code changes


ZzMzaw:
There is something wrong with serialization of assets in sub-directories on windows.
I'll look into it but as I don't have windows to reproduce the issue locally, I'll use this PR for testing.
I may force-push several time until I'm able to fix the issue.
I'll keep you posted once it is done.

ZzMzaw:
Tests are successful now, including on Windows.
It would be great if someone could test the galery shortcode on Windows.

Anyway, PR can be reviewed now.

components/content/src/file_info.rs
```diff
@@ -289,4 +301,26 @@ mod tests {
...
+                Path::new("/home/vincent/code/site/content/posts/tutorials/_index.fr.md"),
+                &PathBuf::new(),
+            ),
+        ];
+
+        for file in files {
+            assert!(file.colocated_path.is_some());
+            if let Some(colocated_path) = file.colocated_path {
+                assert!(!colocated_path.starts_with('/'));
+                assert!(colocated_path.ends_with('/'));
```
Keats: Can you assert what the actual path it?

components/content/src/utils.rs
```diff
@@ -58,6 +59,53 @@ pub fn find_related_assets(path: &Path, config: &Config, recursive: bool) -> Vec
...
+            asset_relative_path
+                .components()
+                .map(|component| component.as_os_str().to_string_lossy().to_string())
+                .collect::<Vec<String>>()
+                .join("/")
+        })
+        .map(|asset_relative_path_as_string| {
+            format!(
+                "/{}{}",
+                colocated_path.expect("Should have colocated path for assets"),
```
Keats: If it should have it, just pass &String for colocated_path instead of the Option

components/content/src/utils.rs
```diff
@@ -58,6 +59,53 @@ pub fn find_related_assets(path: &Path, config: &Config, recursive: bool) -> Vec
...
 
+/// Serializes assets source path for assets colocated with a section or a page
+pub fn serialize_assets(
+    assets: &Vec<PathBuf>,
+    parent_path: Option<&Path>,
+    colocated_path: Option<&String>,
+) -> Vec<String> {
+    assets
+        .iter()
+        .filter_map(|asset| asset.strip_prefix(parent_path.unwrap()).ok())
```
Keats: Pass parent_path: &Path rather than the option if all we're doing is unwrapping it

components/content/src/utils.rs
```diff
@@ -149,6 +197,120 @@ mod tests {
...
+            format!("{}{}", parent_permalink, "FFF.txt"),
+        );
+        expected_assets_permalinks.insert(
+            "posts/my-article/GRAPH.txt".to_string(),
+            format!("{}{}", parent_permalink, "GRAPH.txt"),
+        );
+        expected_assets_permalinks.insert(
+            "posts/my-article/subdir/GGG.txt".to_string(),
+            format!("{}{}", parent_permalink, "subdir/GGG.txt"),
+        );
```
Keats: Something like:

```
let m: HashMap<_, _> = HashMap::from_iter([("one", 1), ("two", 2), ("three", 3)]);
```

is cleaner

components/site/src/lib.rs
```diff
@@ -56,9 +56,12 @@ pub struct Site {
     pub static_path: PathBuf,
     pub templates_path: PathBuf,
     pub taxonomies: Vec<Taxonomy>,
-    /// A map of all .md files (section and pages) and their permalink
+    /// A map of all .md files (sections and pages) and their permalink
     /// We need that if there are relative links in the content that need to be resolved
     pub permalinks: HashMap<String, String>,
+    /// A map of all assets (non .md files colocated to sections and pages) and their permalink, distributed per lang
+    /// We need that if there are relative links in the content that need to be resolved
+    pub assets_permalinks: HashMap<String, HashMap<String, String>>,
```
Keats, line 64: what's the reason for separating by lang? Otherwise we could just add them to the permalinks maps
Are the assets copied for each language? I can't remember

components/content/src/file_info.rs
ZzMzaw: I just asserted what is really important for a colocated path which is not starting with a '/' and ending with a '/'.
I wanted to emphasize that because if you change the way you compute colocated path in the future but don't respect those two assertions, you'll break the current feature.

I definitely can split the check (one for the page and one for the section as they have different colocated path) in this test while adding a comment on what is really important to keep through time.
Would it be ok for you?

components/content/src/utils.rs
ZzMzaw: If `assets` is empty, I don't have any guarantee that `parent_path` will be Some.
I tried to reuse the same structure as the existing code but I can add an `if !assets.is_empty()` to unwrap this variable once instead of doing it in the iteration.
Doing it outside of `serialize_assets` to get rid of the `Option` in the signature would require to wrap the call of this function in the if statement mentionned above.
This would result in code duplication as I would have to do it twice, one for the pages and one for the sections.

What would you prefer me to implement?
1. Keep it as it is
2. Add the if statement in `serialize_assets` function to avoid unwrapping the option in the iteration
3. Add the if statement around the `serialize_assets`

components/content/src/utils.rs
ZzMzaw: If `assets` is empty, I don't have any guarantee that `colocated_path` will be Some.
I tried to reuse the same structure as the existing code but I can add an `if !assets.is_empty()` to unwrap this variable once instead of doing it in the iteration.
Doing it outside of `serialize_assets` to get rid of the `Option` in the signature would require to wrap the call of this function in the if statement mentionned above.
This would result in code duplication as I would have to do it twice, one for the pages and one for the sections.

What would you prefer me to implement?
1. Keep it as it is
2. Add the if statement in `serialize_assets` function to avoid unwrapping the option in the iteration
3. Add the if statement around the `serialize_assets`

components/content/src/utils.rs
ZzMzaw: Thanks for the head's up.
I'll fix it shortly.

components/site/src/lib.rs
ZzMzaw: Yes, assets are copied for each language.

This will allow to have different permalink depending on the lang for the same asset:
- For the default lang, `path/to/the/asset.jpg` would become `https://yourwebsite.com/slug/to/the/asset.jpg`
- For the non-default lang `lang`, `path/to/the/asset.jpg` would become `https://yourwebsite.com/lang/slug/to/the/asset.jpg`

I tested several things such as always defaulting to the default lang but I find this solution to be more clean and future proof.
When a lang is selected, your assets point to the same base URL than your page.
In addition, even if today it is not the case, if we introduce per-lang assets, it will just require minor adaptations.

Please let me know if you prefer to have something more simple.

components/content/src/utils.rs
ZzMzaw: OK, I had time to fix it right away in the end.

components/content/src/file_info.rs
Keats: Yes that would be fine

components/content/src/utils.rs
Keats: How much duplication are we talking about? I'm leaning towards 3

components/site/src/lib.rs
Keats: > Yes, assets are copied for each language.

Ah that's not ideal from content duplication/processing I guess. We can't even define a language specific asset and I don't think it's something that will be added.

I think it makes more sense for assets to be shared across languages and be located once in the folder of the main language page

components/content/src/utils.rs
ZzMzaw: We are just talking about a few lines (`if !assets.is_empty()` statement and of `parent_path` and `colocated_path` options unwrapping) duplicated twice, for pages and sections (where `serialize_assets` is currently used).

If we go for 3., we should apply the same logic to the `get_assets_permalinks` function by moving l.95-96 (`if !serialized_assets.is_empty()` statement and `colocated_path` option unwraping) around the function call.

Is it ok if I proceed that way?

components/site/src/lib.rs
ZzMzaw: > Ah that's not ideal from content duplication/processing I guess.

This is mandatory and may remain mandatory in the future from my understanding.
We'll break any relative links in non-default language pages otherwise.
Let's imagine we have a `test.file` colocated file and a link defined as follow: `<a href="test.file">`, it would work in default language (resolved as `https://yourwebsite.com/slug/to/the/test.file` by the browser) but would break in non-default ones in case content was not duplicated (resolved as `https://yourwebsite.com/lang/slug/to/the/test.file` by the browser).

> I think it makes more sense for assets to be shared across languages and be located once in the folder of the main language page

If it is still the case, I'll revert my changes to ignore the lang for assets in `get_url` function.
Is it ok for you?

components/content/src/utils.rs
Keats: I think it's fine. Eventually page/sections will converge probably so it's going to reduce the duplication

components/site/src/lib.rs
Keats: Keep like you have right now. In an ideal world, we would serve the assets from a single place and rewrite the links to assets to point to the right thing but that's not for now.

ZzMzaw:
Everything may have been fixed as per our discussion.
I let you check and resolve comments if it fits your expectations.
Feel free to let me know if there is still something to adapt.
Many thanks for your review.

components/content/src/page.rs
```diff
@@ -195,7 +197,26 @@ impl Page {
         if page.file.name == "index" {
             let parent_dir = path.parent().unwrap();
             page.assets = find_related_assets(parent_dir, config, true);
-            page.serialized_assets = page.serialize_assets(base_path);
+            if !page.assets.is_empty() {
```
Keats, line 200: We don't need that if right? It will be an empty vec if `page.assets` is empty. Same for the one below

components/content/src/utils.rs
```diff
@@ -58,6 +59,48 @@ pub fn find_related_assets(path: &Path, config: &Config, recursive: bool) -> Vec
...
+
+/// Create assets permalinks based on the permalin of the section or the page they are colocated with
+pub fn get_assets_permalinks(
+    serialized_assets: &Vec<String>,
+    parent_permalink: &str,
+    colocated_path: &str,
+) -> HashMap<String, String> {
+    let mut permalinks = HashMap::new();
+    if !serialized_assets.is_empty() {
+        for asset in serialized_assets {
```
Keats: If you make it into an iterator, you don't need that check

components/content/src/page.rs
ZzMzaw: I want to fail if `colocated_path` is none while `page.assets` is not empty because it should not happen (and this is covered by the `expect`) but no having a `colocated_path` is totally valid if there are no assets.
(similar reason for `serialized_assets` below).
In case we have no assets, `serialized_assets` and `assets_permalinks` should remain with default values.

I agree it would make more sense to regroup `serialize_assets` and `get_assets_permalinks` in the same block and retrieve `colocated_path` only once.
Would it be better if code was reworked as follow?
```
if !page.assets.is_empty() {
    let colocated_path = page.file.colocated_path.as_ref().expect("Should have a colocated path if we have assets");
    page.serialized_assets = serialize_assets(
        &page.assets,
        page.file.path.parent().unwrap(),
        colocated_path
    );
    page.assets_permalinks = get_assets_permalinks(
        &page.serialized_assets,
        &page.permalink,
        colocated_path
    );
}
```

I may have missed something so please don't hesitate to point it out.
Thanks for your feedback.

Keats: If it's an error that can happen it should be an error and not a panic. What you propose there is better yes

ZzMzaw: We should never reach the `expect` within the `if` so keeping the panic might be ok.
Anyway, user won't be able to circumvent it so it will have to be fixed in zola itself (and it means we may have done some incompatible changes).
All changes have been pushed.

Keats:
Closing in favour of https://github.com/getzola/zola/pull/3199

PR #2738, opened 2024-12-16
Let `extra_syntaxes_and_themes` reference builtin syntaxes

### LINKED ISSUE #1607: Using embedding in custom sublime syntaxes causes panic
# Bug Report

## Environment

Zola version: 0.14.1

## Expected Behavior
Zola builds the site without any errors.

## Current Behavior
I get the following panic : `thread '<unnamed>' panicked at 'Can only call resolve on linked references: ByScope { scope: <source.bash>, sub_context: None }', /build/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-4.6.0/src/parsing/syntax_definition.rs:191:18`

I did some digging, and my best guess as to what's causing the panic is that syntect can't see the existing bash syntax, or maybe it's not made aware of it. I'm not sure of how Zola is using syntect or how syntect works, so I don't really have a more precise idea, but I think that might be it. At the very least, I'm almost certain the embed in the sublime-syntax is triggering the issue here.

## Step to reproduce
- Run `zola init` to make a new basic project (and enable code highlighting)
- Create `content/_index.md` and inside create a code block with the `crontab` language (it's important to have the script part because that's where the bash embed is used)
```
    ```crontab
    0 0  *   *   *     /some/script
    ```
```
- Edit `config.toml` to add `extra_syntaxes = ["syntaxes"]` under the `[markdown]` section
- Create `syntaxes/Crontab.sublime-syntax` with the contents of [this file](https://github.com/michaelblyons/SublimeSyntax-Crontab/blob/54f1fa7ff0c9d18aea3790555dba6e533ce3749b/Crontab.sublime-syntax)
- In `syntaxes/Crontab.sublime-syntax`, replace `source.shell` with `source.bash` on lines 201 and 210
- Run `zola build`

### LINKED ISSUE #1678: Extra syntaxes embedding built-in syntaxes fail
# Bug Report

## Environment

Zola version: 0.41.1

## Expected Behavior

I would expect custom syntaxes defined using the `extra_syntaxes` system would be able to embed built-in syntaxes as Sublime can.

## Current Behavior

Extra syntaxes which embed built-in syntaxes cause Syntect, and in turn Zola, to panic.

```
Building site...
thread '<unnamed>' panicked at 'Can only call resolve on linked references: ByScope { scope: <source.js>, sub_context: None }', /build/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-4.6.0/src/parsing/syntax_definition.rs:191:18
```

## Step to reproduce

A minimal reproducing example site can be found [here](https://github.com/ajyoon/---zola-extra-syntax-bug-demonstration). The key offending lines are the embed references [here](https://github.com/ajyoon/---zola-extra-syntax-bug-demonstration/blob/main/syntaxes/foo.sublime-syntax#L14). Clone this repo and `zola serve` to panic.

## Discussion

The panic is set off upstream in Syntect [here](https://github.com/trishume/syntect/blob/aefae2f74dfaf6b193d09ad2b3842db6955889ee/src/parsing/syntax_definition.rs#L193), caused when `markup.rs` [loads the custom syntaxes](https://github.com/getzola/zola/blob/d8aebbd220121300952073da77b66870de2d5190/components/config/src/config/markup.rs#L56).

My understanding is this *could* be fixed fairly easily, but it may cost a performance hit as the [`highlighting.rs`'s `SYNTAX_SET` is precomputed](https://github.com/getzola/zola/blob/d8aebbd220121300952073da77b66870de2d5190/components/config/src/highlighting.rs#L10) and fixing this would require modifying that approach with one of a few possibilities like:
- Simply have [`resolve_syntax_and_theme`](https://github.com/getzola/zola/blob/d8aebbd220121300952073da77b66870de2d5190/components/config/src/highlighting.rs#L41) convert `SYNTAX_SET` back into a builder and load the extra syntaxes there
- Modify `SYNTAX_SET` and its precomputation scheme so it only precomputes a [`SyntaxSetBuilder`](https://docs.rs/syntect/latest/syntect/parsing/struct.SyntaxSetBuilder.html), then in `resolve_syntax_and_theme` add the extra syntaxes to it and then build at runtime.

It's easy to imagine this could be a costly fix for a fairly niche use-case.

I was able to find a workaround, which is to simply copy the referred-to `.sublime-syntax` files from Zola or Sublime's source into my site's custom syntax directories. If this workaround is considered acceptable in the long-term, I think a fine fix would to be simply document this limitation and workaround in [the relevant docs](https://www.getzola.org/documentation/content/syntax-highlighting/).

9999years:
Closes #1607
Closes #1678

This lets syntax definitions in `extra_syntaxes_and_themes` reference builtin syntaxes (by `scope:` or `include:` or similar).

Unfortunately, re-building the `SyntaxSet` is rather expensive. On my (small) personal website, this makes Zola render my site in 469ms rather than 63ms — a slowdown of nearly 8x! 

We could potentially mitigate this by serializing the computed `SyntaxSet` to a file, for example with a [`cached::DiskCache`](https://docs.rs/cached/latest/cached/stores/struct.DiskCache.html), but I don't see existing caching infrastructure so I'm not sure if that's a good tradeoff.

Is this wanted? The performance cost is not worth it with the current implementation, but maybe we can optimize it?

---

## Code changes


Keats:
That's a huge slowdown. Let's see if someone from syntect can weight in because that's too much IMO

9999years:
@Keats Yeah, agreed. I opened an issue upstream, we'll see what they think: https://github.com/trishume/syntect/issues/562

components/config/src/config/markup.rs
```diff
@@ -99,7 +101,7 @@ impl Markdown {
             return Ok((None, None));
         }
 
-        let mut ss = SyntaxSetBuilder::new();
+        let mut ss = SYNTAX_SET.clone().into_builder();
```
jyn514, line 104: i don't think this change is correct. `load_extra_syntaxes_and_highlight_themes` ends up in `extra_syntax_set`, which is used in `resolve_syntax_and_theme` here: https://github.com/getzola/zola/blob/434da458a38dd3e011ad147d410b61dac6e7570c/components/config/src/highlighting.rs#L44
and so this causes *all* syntaxes to be treated as Extra, not just the ones from `extra_syntaxes_and_themes`.

Issue #2741, opened 2024-12-19
Feature Request: Allow users to specify additional paths to watch for changes


rparrett:
Motivated by some friction experienced by Bevy maintainers, discussed here: https://github.com/bevyengine/bevy/discussions/16431#discussioncomment-11442950

As well as a [project of mine](https://github.com/rparrett/zola-test) where I am including content that is necessarily in another directory, and would love it if Zola would reload when it changes.

edit: Oops, it looks like this was previously suggested [in the forum](url), but never received a response.

cart:
To elaborate:

1. We have a "news" markdown page that calls a `release_notes` shortcode
2. The `release_notes` shortcode calls `load_data()` to load a `_release-notes.toml` file, which is an index that contains many "release note" entries containing things like a title and a link to the release note entry markdown.
3. We iterate over all of the "release note" entries and call `load_data(release_note.file_name)`  to retrieve the markdown and render it.

The problem is that if we change one of the individual release note markdown files, the final rendered page is not updated.

Here is a link to the shortcode: https://github.com/bevyengine/bevy-website/blob/main/templates/shortcodes/release_notes.md?plain=1

And here is a link to one of the news articles that doesnt hot-reload when a release note entry changes: https://github.com/bevyengine/bevy-website/blob/main/content/news/2024-11-29-bevy-0.15/index.md

And here is a link to one of the release note entries that connects to that article: https://github.com/bevyengine/bevy-website/blob/main/release-content/0.15/release-notes/14791_Required_Components.md

Keats:
I think that's fine to add as an argument to `zola serve`. I'm assuming we're fine with just completely reloading the site when something changes right?

rparrett:
It would be nice to be able to configure this with an array in `config.toml` so that contributors don't need to do anything extra beyond `zola serve`.

I have this working in a [branch](https://github.com/rparrett/zola/commit/660b57e1fb5b2bc99b3db0dce48b8f903ab08cc6), and can open a PR if that sounds okay. There may be some small implementation details to discuss.

Keats:
I would prefer that to be cli arguments, there are no `serve` specific things in config.toml right now. You can always write the command required in a makefile or docs

Keats:
It was done back in 0.20

PR #2745, opened 2024-12-20
Add optional arg to `serve` for extra paths to watch for changes

rparrett:
This PR implements a feature discussed in #2741.

Sanity check:

## Code changes

  There does seem to be some existing documentation related to the args for `serve`, but it is not comprehensive, and I suspect this isn't important enough to include there.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -89,6 +89,10 @@ pub enum Command {
         /// Default append port to the base url.
         #[clap(long)]
         no_port_append: bool,
+
+        /// Extra path to watch for changes, relative to the project root.
+        #[clap(long)]
+        extra_watch_path: Vec<String>,
     },
 
     /// Try to build the project without rendering it. Checks links
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -816,8 +816,19 @@ pub fn serve(
                                 site = s;
                             }
                         }
-                        ChangeKind::Unknown => {
-                            // Probably a user-provided watch location. We can't know exactly what to update.
+                        ChangeKind::ExtraPath => {
+                            let full_paths: Vec<&PathBuf> =
+                                change_group.iter().map(|(_, p, _)| p).collect();
+                            let combined_paths = full_paths
+                                .iter()
+                                .map(|p| p.display().to_string())
+                                .collect::<Vec<String>>()
+                                .join(", ");
+                            console::info(&format!(
+                                "-> {combined_paths} changed. Recreating whole site."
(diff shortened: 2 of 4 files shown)


src/cli.rs
```diff
@@ -89,6 +89,10 @@ pub enum Command {
         /// Default append port to the base url.
         #[clap(long)]
         no_port_append: bool,
+
+        /// Extra path to watch for changes, relative to the project root.
+        #[clap(long)]
+        extra_watch_path: Vec<String>,
```
Keats, line 95: Should it be `extra_watch_paths` since it's a vec?

src/fs_utils.rs
```diff
@@ -17,6 +17,7 @@ pub enum ChangeKind {
     StaticFiles,
     Sass,
     Config,
+    Unknown,
```
Keats: let's call it AdditionalDirectory or something similar, it's not strictly unknown

src/cmd/serve.rs
```diff
@@ -810,6 +816,12 @@ pub fn serve(
                                 site = s;
                             }
                         }
+                        ChangeKind::Unknown => {
+                            // Probably a user-provided watch location. We can't know exactly what to update.
```
Keats: can we add an info message like (text can change obviously):

```
console::info("-> {full_path} changed. Recreating whole site.");
```

Keats (changes requested):
Just some nits

src/cli.rs
rparrett: It's a vec, but the user would multiple paths by repeating the argument.
```bash
zola serve --extra_watch_path="a" --extra_watch_path="b"
```
So I think the singular makes the most sense. There's some precedent for this, e.g.
```bash
curl --header "Content-Type: application/json" --header "Authorization: Bearer token" https://api.example.com
grep --regexp="pattern1" --regexp="pattern2" file.txt
```
Happy to change it if you prefer though.

src/fs_utils.rs
rparrett: I went with `ExtraPath` to match other names used in this feature and added a doc comment to `ChangeKind`. Let me know if you'd prefer something else though.

src/cli.rs
Keats: That's fine, I thought it was `--extra_watch_paths=something,something_else` with clap being smart enough to split on the `,`

Keats (approved):

rparrett:
Thanks!

PR #2748, opened 2024-12-24
Allow nested shortcodes

ARitz-Cracker:
Fixes https://github.com/getzola/zola/issues/515

Hello, and Happy Christmas 🎄

This PR iterates on #2630 with the given feedback of _not_ requiring the use of the `markdown` filter to execute nested shortcodes. Additionally, unlike the previous PR, nested nth invocation increments follows a more intuitive pattern.

### Before:
- 1
- 2
  - 5
  - 6
    - 9
    - 10
  - 7
  - 8
- 3
- 4

### After:
- 1
- 2
  - 3
  - 4
    - 5
    - 6
  - 7
  - 8
- 9
- 10

Sanity check:

  - This PR replaces #2630

## Code changes
- The shortcode parser has been changed to stop at the appropriate {% end %}, instead of the first-encountered {% end %}. (Thanks to https://github.com/getzola/zola/pull/1475)
- A ~~clonable shared~~ `ShortcodeInvocationCounter` has been added
- The `markdown::shortcode::parser::Shortcode` struct now has `inner: Vec<Shortcode>`
- The `markdown::shortcode::parser::parse_for_shortcodes` function now also parses shortcode bodies
- Added a `flatten` method to `markdown::shortcode::parser::Shortcode` which renders the inner-shortcodes into the shortcode body
- The `render` method on `markdown::shortcode::parser::Shortcode` now calls `flatten`

  - As far as I'm aware, the documentation does not explicitly prohibit nested short-codes.

Let me know what you think!

### DIFF
--- a/components/markdown/src/shortcode/parser.rs
+++ b/components/markdown/src/shortcode/parser.rs
@@ -1,11 +1,11 @@
-use std::ops::Range;
+use std::{collections::HashMap, ops::Range};
 
 use errors::{bail, Context as ErrorContext, Result};
 use libs::tera::{to_value, Context, Map, Tera, Value};
 use pest::iterators::Pair;
 use pest::Parser;
 use pest_derive::Parser;
-use utils::templates::{ShortcodeFileType, ShortcodeInvocationCounter};
+use utils::templates::{ShortcodeDefinition, ShortcodeFileType, ShortcodeInvocationCounter};
 
 pub const SHORTCODE_PLACEHOLDER: &str = "@@ZOLA_SC_PLACEHOLDER@@";
 
@@ -13,14 +13,34 @@ pub const SHORTCODE_PLACEHOLDER: &str = "@@ZOLA_SC_PLACEHOLDER@@";
 pub struct Shortcode {
     pub(crate) name: String,
     pub(crate) args: Value,
+    // In practice, span.len() is always equal to SHORTCODE_PLACEHOLDER.len()
     pub(crate) span: Range<usize>,
     pub(crate) body: Option<String>,
     pub(crate) nth: usize,
+    pub(crate) inner: Vec<Shortcode>,
     // set later down the line, for quick access without needing the definitions
     pub(crate) tera_name: String,
 }
 
 impl Shortcode {
+    /// Attempts to fill the `tera_name` field from the provided definitions for self and all of self.inner.
+    ///
+    /// This returns an error if the definitions do not have this shortcode.
+    pub fn fill_tera_name(
+        &mut self,
+        definitions: &HashMap<String, ShortcodeDefinition>,
+    ) -> Result<()> {
+        if let Some(def) = definitions.get(&self.name) {
+            self.tera_name = def.tera_name.clone();
+        } else {
+            return Err(errors::anyhow!("Found usage of a shortcode named `{}` but we do not know about. Make sure it's not a typo and that a field name `{}.{{html,md}}` exists in the `templates/shortcodes` directory.", self.name, self.name));
+        }
+        for inner_sc in self.inner.iter_mut() {
+            inner_sc.fill_tera_name(definitions)?;
+        }
+        Ok(())
+    }
+
     pub fn file_type(&self) -> ShortcodeFileType {
         if self.tera_name.ends_with("md") {
             ShortcodeFileType::Markdown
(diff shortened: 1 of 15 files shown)


components/markdown/src/shortcode/parser.rs
```diff
@@ -1,27 +1,46 @@
...
     pub(crate) span: Range<usize>,
     pub(crate) body: Option<String>,
     pub(crate) nth: usize,
+    pub(crate) inner: Vec<Shortcode>,
     // set later down the line, for quick access without needing the definitions
     pub(crate) tera_name: String,
 }
 
 impl Shortcode {
+    /// Attempts to fill the `tera_name` field from the provided definitions for self and all of self.inner.
```
Keats, line 26: What's `tera_name`?

components/markdown/src/shortcode/parser.rs
```diff
@@ -30,7 +49,34 @@ impl Shortcode {
         }
     }
 
-    pub fn render(self, tera: &Tera, context: &Context) -> Result<String> {
+    /// Expands all inner-shortcodes and leaves self.inner empty.
+    ///
+    /// This function has no effect with shortcodes without bodies.
+    pub fn flatten(&mut self, tera: &Tera, context: &Context) -> Result<()> {
```
Keats: let's just call that `expand_inner_shortcodes` or `render_inner_shortcodes`

components/utils/src/templates.rs
```diff
@@ -34,6 +37,26 @@ impl ShortcodeDefinition {
     }
 }
 
+#[derive(Debug, Default, Clone)]
+pub struct ShortcodeInvocationCounter {
+    amounts: Arc<RwLock<HashMap<String, usize>>>,
```
Keats: Do we still need Arc<RwLock<..>>?

components/markdown/src/shortcode/parser.rs
ARitz-Cracker: `shortcodes/xxx.yy`. This function re-implements what was originally [here](https://github.com/getzola/zola/pull/2748/files/4b39a97bd715a76326a6fca883272d7263b70472#diff-82dc1d519d7c352b36650805f58639cec80b5aac264cc565b3919b1d78556814L19-R20) so it may be done recursively with inner shortcodes

components/markdown/src/shortcode/parser.rs
ARitz-Cracker: I'll go with `render_inner_shortcodes` in the next commit 👍

components/utils/src/templates.rs
ARitz-Cracker: Basically, yes, but only if it's gonna be a part of `markdown::context::RenderContext`.

This thing has interior mutability mostly due to the requirements of the previous PR, and when I started this PR, I wasn't 100% sure where `ShortcodeInvocationCounter` would actually be _needed_, but now that everything's settled, it looks like only `extract_shortcodes` uses this which calls itself so this thing can be refactored to get the interior mutability removed if we keep its existence to the recursive `extract_shortcodes` calls. Let me know if you'd like me to go through with this.

ARitz-Cracker:
Oh, another thing, I noticed that when it comes to MD shortcodes calling HTML shortcodes, the HTML shortcode must be multi-line otherwise its result will get wrapped in a `<p>` tag. Where would be the appropriate place to document this?

components/utils/src/templates.rs
Keats: Yeah let's keep things simple

Keats:
> Oh, another thing, I noticed that when it comes to MD shortcodes calling HTML shortcodes, the HTML shortcode must be multi-line otherwise its result will get wrapped in a <p> tag. Where would be the appropriate place to document this?

Not sure honestly. Sounds like a bug maybe?

ARitz-Cracker:
> > Oh, another thing, I noticed that when it comes to MD shortcodes calling HTML shortcodes, the HTML shortcode must be multi-line otherwise its result will get wrapped in a  tag. Where would be the appropriate place to document this?
> 
> Not sure honestly. Sounds like a bug maybe?

I... Actually don't think so, iirc if you have a single line with a `<span>` tag in an MD document, it being a part of a paragraph is expected behaviour, and after skimming through `pulldown-cmark`'s source, it doesn't seem to have has inherent knowledge on whether or not an HTML element is an inline element (like a span) or a block element (like a div) resolving such issues seems outside the scope of this PR.

ARitz-Cracker:
Applied the feedback given, lmk what you think!

Keats (approved):
Thanks!

PR #2750, opened 2024-12-27
Implement zola serve --store-html

### LINKED ISSUE #2377: `zola serve`: option to store html files on disk.
I'm using TailwindCSS as a CSS framework. It generates CSS classes definitions from the CSS classes present in the HTML.

If I configure TailwindCSS to look after the `templates` folder, it works fine. I use `tailwindcss --watch`, which generated its CSS to the `static` folder, and `zola serve`, which put assets in the `public` folder.

But when the CSS class names are dynamically generated by Zola macros, they're not present in the `templates` folder: only in the `public` folder, where generated HTML files live.

Problem is: only `zola build` actually write those files, `zola serve` store them in-memory, thus making them inaccessible to `tailwindcss`. It would be super useful to store to have an option like `--html-on-disk` to the `serve` subcommand for those kinds of workflows.

ppom0:
Fixes #2377


components/site/src/lib.rs
```diff
@@ -114,10 +116,10 @@ impl Site {
     }
 
     /// Enable some `zola serve` related options
-    pub fn enable_serve_mode(&mut self) {
+    pub fn enable_serve_mode(&mut self, both: bool) {
```
Keats: there's an enum for it, use it in the params instead of a bool

src/cli.rs
```diff
@@ -82,6 +82,10 @@ pub enum Command {
         #[clap(short = 'O', long)]
         open: bool,
 
+        /// Also store HTML in the public/ folder (by default HTML is only stored in-memory)
+        #[clap(short = 'H', long)]
```
Keats: I think only having long here is fine

Keats (changes requested):

components/site/src/lib.rs
ppom0: Does the new commit correspond to your idea?

I can also squash it onto the previous commit.

Keats (approved):
Yep perfect

Keats:
Looks like legit tests failures

ppom0:
Yes indeed!
Are you ok with the tests just testing serve without the `--store-html` option?

Keats:
Yes that's fine, thanks!

PR #2752, opened 2024-12-29
Remove COPR recommendation for installation

Gerrit0:
The package hasn't been updated since Zola 0.12, which led to me being quite confused when the syntax highlighting instructions in the docs didn't work.


Keats:
Thanks

PR #2776, opened 2025-01-17
Adds minify option to build command

### LINKED ISSUE #1924: Support minification only for the build
Adding `minify_html = true` enables minification globally, for `zola serve` as well as the `build` option.

I would like to propose an additional flag (e.g. `minify_html = build_only`) so that it happens only during the build.

ogarcia:
Now the build command supports minification and you can run `zola build --minify` to get the minified result only at build time. Closes #1924.

Now if the pull request is correct. Sorry for the previous one that I made a mix of branches.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -175,6 +175,10 @@ impl Site {
         self.output_path = path.as_ref().to_path_buf();
     }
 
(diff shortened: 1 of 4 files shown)


Keats (approved):
Thanks!

PR #2777, opened 2025-01-17
native arm compilation in ci for arm

hegerdes:

Sanity check:

## Code changes
This is a followup to #2629.  
GitHub added native arm runners for public use. [GitHub Changelog](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/)

Cross compile for arm is not needed anymore, removes extra ci steops. This also allows to test the final binary on each arch by rinning ist and showing its version. 

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,10 +4,6 @@ on:
   push:
     branches: ["*"]
 
-env:
-  # Cross-compilation for aarch64 requires a different linker
-  CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-gcc
-
 permissions:
   contents: read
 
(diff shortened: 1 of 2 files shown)


Keats (approved):
Nice!

hegerdes:
Not really nice when builds fail...

Im sincerely sorry, I tested it but seems like I copied the wrong path.  
Better self decide if the test call to zola should be removed or if the path should be fixed:
```yaml
      - name: Show version
        # Is this the same for every target wen running cargo build?
        run: ./target/debug/zola --version

```

Keats:
Ah yes let's just remove the zola --version call i think

hegerdes:
Should I add another PR or would you be okay with removing the `Show version` step?

Keats:
You can remove the show version step in another PR

Issue #2690, opened 2024-11-07
word_count and reading_time count the footnotes


liavhibe:
# Bug Report

## Environment

Zola version: 0.19.2

## Expected Behavior
word_count and reading_time should not count footnotes.

## Current Behavior
word_count and reading_time count the footnotes.

## Step to reproduce
create a page then add and remove some footnotes, you'll see the change in the output for both functions.

ghost:
https://github.com/getzola/zola/blob/a91ae35872639c34f32e26d48d3078ddb8101c81/components/content/src/page.rs#L110

Currently, the reading analytics are derived from `page.raw_content` which is all content (including footnotes). In order to get the word count for just the main content, the footnotes will need to be differentiated from the rest of the contents in `raw_content`.

Keats:
I would say it's already an approximation so I'm closing that as wontfix since fixing it will not change much in practice.

PR #2778, opened 2025-01-20
Improve get_reading_analytics()

soutade:
Remove Markdown footnotes & HTML comments before counting

For Bug #2690

### DIFF
--- a/components/content/src/utils.rs
+++ b/components/content/src/utils.rs
@@ -1,3 +1,4 @@
+use libs::regex;
 use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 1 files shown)


Keats:
Does it make a significant difference for an average article?

soutade:
> Does it make a significant difference for an average article?

I think NO.
Footnotes are not so much used and HTML comments very rare.
It was just to address ticket #2690

Keats:
Let's close it as wontfix then, no need to complicate something that is already an approximation. Thanks for the PR though

Issue #2676, opened 2024-10-23
`redirect_to` adds trailing slash - prohibiting the usage of links to id's


trembel:
# Bug Report
When using `redirect_to` on a section, e.g.
```
+++
redirect_to = "/about/#me"
+++
```
a trailing slash is appended to the redirect link, which leads to a non-working redirect (e.g. not to the desired element)

## Environment

Zola version: 0.19-2
## Expected Behavior
Redirect to the desired element in the html page

## Current Behavior
Redirect to the top of the html page

Keats:
We should also allow using a Zola link (eg `@/` pointing a md file)

Keats:
It;s been fixed for a while now

PR #2779, opened 2025-01-21
Handle permalink with anchor (doesn't add trailing slash)

soutade:
Handle permalink with anchor (doesn't add trailing slash)

For ticket #2676


Keats (approved):
Thanks!

PR #2782, opened 2025-01-27
Fix test can_build_site_with_pagination_for_taxonomy()

soutade:
Take in account not rendered pages for pagers computation.
As pages iterator has not always the same references order, numbers of "A" pagers can be different if it contains or not some not rendered pages


Keats (approved):
Thanks!

PR #2708, opened 2024-11-18
Add optional math support for markdown

recmo:
**IMPORTANT: Please do not create a Pull Request adding a new feature without discussing it first.**

The place to discuss new features is the forum: <https://zola.discourse.group/>
If you want to add a new feature, please open a thread there first in the feature requests section.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [x] Have you created/updated the relevant documentation page(s)?

This fixes #2019 without adding substantial complexity or dependencies. It optionally enables a common-mark extension available in `pulldown-cmark`. 

This will correctly parse delimited math, and render html like `<span class="math math-inline">\sum_i x_i<\span>`.

To render the latex using Katex, include something like this in the header:

```html
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/katex.min.css"
        integrity="sha384-nB0miv6/jRmo5UMMR1wu3Gz6NLsoTkbqJghGIsx//Rlm+ZU03BU6SQNC66uf4l5+"
        crossorigin="anonymous">
    <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/katex.min.js"
        integrity="sha384-7zkQWkzuo3B5mTepMUcHkMB5jZaolc2xDwL6VFqjFALcbeS9Ggm/Yr2r3Dy4lfFg"
        crossorigin="anonymous"></script>
    <script defer>
        document.addEventListener("DOMContentLoaded", function () {
            const macros = {};
            for (let el of document.getElementsByClassName('math')) {
              katex.render(el.textContent, el, {
                output: 'html',
                throwOnError: false,
                displayMode: el.classList.contains('math-display'),
                macros,
              });
            }
        });
    </script>
```

PR #2791, opened 2025-02-07
feat(markdown): add support for math rendering with typst and katex

cestef:

Sanity check:

	- https://github.com/getzola/zola/pull/2708 stalled

## Code changes


cestef:
I don't really know how to get rid of these `Cargo.lock` updates, but anyways

cestef:
This PR adds math rendering on the "server-side" with [typst](https://typst.app), while keeping the katex option open for latex.

A `GenericCache` struct has also been added for integrations like these to store the compilation results to save time on subsequent builds. Ideally, each integration would have its cache in the the global `<Site>.caches` attribute. Cache is stored as binary-encoded data at `.cache/<CACHE_NAME>`. Typst remote packages are also downloaded to this path. 

The output SVGs generated by typst can also be minified via the [svgo](https://svgo.dev/) CLI. This is kind of a hacky implementation because it relies on the CLI but it does the job.

Newly added configuration parameters:

```toml
math = "typst"                       # Defaults to "none", one of "typst", "katex"
math_svgo = true                     # Optimize the SVGs generated by the math renderer
math_svgo_config = "svgo.config.mjs" # Optional, Path to the svgo config
math_css = "styles/typst-embed.css"  # path to the CSS file to inject inside the SVGs generated 
```

CI is failing because `image-webp@0.2.1` (used by `typst` and `image`) and `twox-hash@2.1.0` (used for caching keys) require rustc >= 1.81, should we upgrade rustc or downgrade these dependencies ?

The size of the binary also drastically increases because of typst's ecosystem (`36.4 MiB` -> `62.0 MiB`)

claytonwramsey:
This is great work! One possible place to improve: the current implementation outputs SVG for equations. It might be better to use [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) to make sites smaller and more accessible.

cestef:
Thanks a lot :D Typst doesn't yet support HTML or MathML output (see https://github.com/typst/typst/issues/5512)

The basic katex implementation however already outputs HTML I think. I don't use LaTeX that much so if anyone is up to help with this, I'd be glad.

claytonwramsey:
I think it might be possible to write a backend visitor to output MathML (cf https://github.com/wcshds/typst-math-to-mathml-converter). This would be a lot easier than what the Typst team is building, since we don't have to handle everything else, but the downside is that we're then implementing a possibly buggy different spec from what Typst does. 

As a stopgap for accessibility, we might also just include the math source as alt text for the generated SVGs.

cestef:
This indeed sounds a bit risky, including the original source in the alt should be enough for now. I will tinker a bit with the library you linked to try out mathml rendering, I also stumbled accross this one:

https://github.com/xushengfeng/xmmath

components/config/src/config/markup.rs
```diff
@@ -65,6 +74,19 @@ pub struct Markdown {
...
     /// The default template can be overridden by creating a `anchor-link.html` in the `templates` directory
     pub insert_anchor_links: InsertAnchor,
+    /// Whether to enable math rendering in markdown files
+    pub math: MathRendering,
+    /// Whether to optimize generated math SVGs with svgo
+    pub math_svgo: bool,
+    /// Svgo configuration file path
+    pub math_svgo_config: Option<String>,
+    /// Injected CSS path for math rendering
+    pub math_css: Option<String>,
```
soqb: surely we should make this an object? it feels a bit flat so something (loosely) like:
```rust
// parses:
// `false` -> `Disabled`
// `true` -> `Enabled(None)`
// `"my_path"` -> `Enabled(Some("my_path"))`
#[derive(Clone, Debug, Default)]
enum MathSvgo {
    #[default]
    Disabled,
    Enabled(Option<String>),
}

// .. and maybe even allow
// `math = "typst"` as well as `math = { engine = "typst" }`..?
#[derive(Clone, Debug, Serialize, Deserialize, Default)]
struct MathOptions {
    engine: MathRendering,
    #[serde(default)]
    svgo: MathSvgo,
    #[serde(default)]
    css: Option<String>,
}
```

this way, this:
```toml
[markdown]
math = "typst"
math_svgo = true
math_css = "static/math.css"
math_svgo_config = "svgo.whatever"
```

becomes:
```toml
[markdown]
math = {
    engine = "typst"
    svgo = "svgo.whatever"
    css = "static/math.css"
}
```

and anyway, if you want to be flat you can write `[markdown.math]`.

components/content/src/page.rs
```diff
@@ -329,6 +338,7 @@ Hello world"#;
             &config,
             InsertAnchor::None,
             &HashMap::new(),
+            Arc::new(Caches::default()),
```
soqb: applies elsewhere but [`Arc` implements `Default`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#impl-Default-for-Arc%3CT%3E) for this case:

```suggestion
            Arc::default(),
```

components/markdown/src/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Generic cache using DashMap, storing data in a binary file
+#[derive(Debug)]
+pub struct GenericCache<K, V>
+where
+    K: Eq + Hash + Serialize + for<'de> Deserialize<'de>,
+    V: Serialize + for<'de> Deserialize<'de>,
+{
+    cache_file: PathBuf,
+    cache: DashMap<K, V>,
```
soqb, line 19: i'm not totally sure whether it matters at the kind of scale/concurrency zola operates on, but you might find [papaya](https://docs.rs/papaya) a better fit for a concurrent hash map. [this (biased) blogpost by the author](https://ibraheem.ca/posts/designing-papaya/) totally convinced me.

components/markdown/src/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// Get the directory where the cache is stored
+    pub fn dir(&self) -> &Path {
+        self.cache_file.parent().unwrap()
+    }
+
+    /// Create a new cache for a specific type
+    pub fn new(base_cache_dir: &Path, filename: &str) -> crate::Result<Self> {
+        // Create the base cache directory
+        fs::create_dir_all(base_cache_dir)?;
```
soqb: do we really want to create the dir/file even when the cache is never touched?

components/markdown/src/markdown.rs
```diff
@@ -436,6 +446,7 @@ pub fn markdown_to_html(
     opts.insert(Options::ENABLE_STRIKETHROUGH);
     opts.insert(Options::ENABLE_TASKLISTS);
     opts.insert(Options::ENABLE_HEADING_ATTRIBUTES);
+    opts.insert(Options::ENABLE_MATH);
```
soqb: i'd probably make this contingent on whether math is enabled. i think it's more graceful to have the math included raw rather than hide it completely. it's also (marginally) more practical for the pure commonmark folk.

components/markdown/src/markdown.rs
```diff
@@ -564,7 +597,7 @@ pub fn markdown_to_html(
                         cmark::CodeBlockKind::Fenced(fence_info) => FenceSettings::new(fence_info),
                         _ => FenceSettings::new(""),
                     };
-                    let (block, begin) = match CodeBlock::new(fence, context.config, path) {
+                    let (block, begin) = match CodeBlock::new(&fence, context.config, path) {
```
soqb: performance wise, though it's somewhat (but not totally) moot, you might want to clone `fence.language` beforehand so you don't have to pay the larger cost of cloning inside `CodeBlock::new`.

components/markdown/src/markdown.rs
```diff
@@ -573,17 +606,47 @@ pub fn markdown_to_html(
...
+                    code_block_language = fence.language.map(|s| s.to_string());
                 }
                 Event::End(TagEnd::CodeBlock { .. }) => {
                     if let Some(ref mut code_block) = code_block {
-                        let html = code_block.highlight(&accumulated_block);
-                        events.push(Event::Html(html.into()));
-                        accumulated_block.clear();
+                        let inner = &accumulated_block;
+                        match code_block_language.as_deref() {
+                            Some("typ") => {
```
soqb: is there a reason katex isn't supported here too?

components/markdown/src/markdown.rs
```diff
@@ -846,6 +962,10 @@ pub fn markdown_to_html(
             summary = Some(summary_html);
         }
 
+        if let Some(render_cache) = typst.render_cache {
+            render_cache.write()?;
+        }
```
soqb: why not `Compiler::write_cache`? and what about `KatexCompiler`?

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,15 @@ pub struct ThemeCss {
     pub filename: String,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum MathRendering {
+    #[default]
+    None,
+    Typst,
+    KaTeX,
```
soqb: nit: `Katex` would be more consistent.

components/markdown/src/math/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+pub enum ShouldMinify<'a> {
+    Yes(Option<&'a str>), // config
+    No,
+}
+
+pub trait Compiler<M: Copy + Eq + Hash, R: for<'de> Deserialize<'de> + Serialize> {
+    fn compile(&self, input: &str, mode: M, minify: &ShouldMinify) -> Result<R, String>;
+    fn set_cache(&mut self, cache: Arc<GenericCache<String, R>>);
+    fn write_cache(&self) -> Result<(), String>;
+}
```
soqb: love this :eyes:

cestef:
I have tried out a bit the library you linked, here's the result I get:

image

I am using `katex`'s font for the rendered mathml. There are still a lot of inconsitencies (the spaces between $a$ and $x$, the spacing feeling weird), but overall it seems like a viable solution if we were to tweak things a bit.

For reference, here's the result with SVG rendering:

image

soqb (commented):
this looks great! i've left a few comments but i've been toying locally for a few days and i'm very happy.

+1 for svg+mathml too. it's always what i gravitate to and it always works well. one day soon mathml will be widespread and complete enough that it will become a viable primary target, but for now, svg rendering is the only consistent option.

on that note, did you ever consider [mathjax](https://docs.rs/mathjax)? the latex support is a little better than katex these days.

components/markdown/src/markdown.rs
cestef: I don't use latex/katex that much, but I think there's no concept of "raw" rendering in katex. From what I understood, it's only made for math rendering, whereas typst can render whole documents. the "raw" mode was initially intended for rendering graphs/tables. Let me know if katex supports such things!

cestef:
Update:

- As per @soqb suggested, math-related configuration is now grouped into the `math` attribute:
```toml
[markdown.math]
engine = "typst" # typst, katex, or none
svgo = true # true, false, or "path"           
css = "styles/typst-embed.css" # optional
addon = "helpers.typ" # optional
```

We can also simplfy specify:

```toml
[markdown]
math = "typst"
```
Which is equivalent to 

```toml
[markdown]
math = {
	engine = "typst"
}
```

- A new `cache` configuration parameter has been added to control caching from typst or katex (and possibly more integrations in the future, thus, I put it in `markdown` instead of `markdown.math`):

```toml
[markdown]
cache = true # true, false or "path"
```

By default the caches are now stored in `~/.cache/zola/<CACHE_NAME>`.

- CodeBlocks are now wrapped inside a `CodeBlockType` enum. This is due to the introduction of "renderable" code blocks:
~~~md
## Wonderful test

This is markdown content.

```typ
= Hello, World!
This is a raw typst document: $y^2 = x^3 + a x + b$
```
~~~

image

- To allow for future additions (e.g. mathjax), a `MathCompiler` trait has been added.
	If raw rendering, is supported, code fence languages that it renders should be returned at `raw_extensions`.
	_Only the specified rendering engine is now created and boxed, this means that rendering both typst and katex in the same document is not possible_

- Added `addons` for math which allow users to add code snippets to all compiled expressions. This is useful when you need to provide helper functions, import default packages or override defaults.

PS: Thank you so much for your feedback soqdb 😄

Keats:
I wouldn't bother with the `math = "typst"` form in the config

cestef:
@soqb after benchmarking papaya and dashmap, it looks like the difference for our scale is _very_ _very_ small...
Benchmarks were ran on a site with around 500 elements in cache:

Hot run (read heavy):
zola-hot
Cold run (write heavy):
zola-cold

lilymonad:
I tried this PR, and used `katex`, but I had to include the `katex.css` file myself in my base html template. We need this CSS file in order to hide the mathml output. Wasn't it the job of the `css` attribute ?

cestef:
css attribute was for css injected _inside_ the svg (typst). For KaTeX, there is nothing injected. Maybe I should update the documentation for this.

lukasjuhrich:
Hey folks, great work! I'm quite interested in this PR.
Has there been preceding feature discussion I'm unaware of, or is this intended to be a POC that remains open?

I'll gladly check out this branch and use it for now, but I was wondering if there are any plans from the maintainer's side to upstream this, and if yes what the blockers to this PR are (inb4 maintainer capacity).

Cheers!

cestef:
This was intially just a PoC to fit my own needs for a blog, and I posted a PR here just in case anyone was interested. I'd gladly help you set up a clean PR to integrate it into zola!

p-r-o-g:
I can try to recreate the changes without merge conflict if that would help? Is someone actively working on getting this done?

cestef:
Merge conflicts were resolved

cestef:
Hey @keats! Any thoughts on this math rendering PR? Would love to get your feedback to move this forward 🙂

cappuccinocosmico:
Hey, just a random new user passing through. This PR looks great! For me getting the ability to render typst would be really helpful for some of my use cases. Mainly because typst has so much more functionality for rendering complex mathematics compared to regular katex:
image
Most of the diagram generation seems to work within math mode, and based on my reading of the PR you should be able to make diagrams with something like this
```md
# Markdown

Here *is* some **example** markdown code:
$#diagram(
	spacing: 8pt,
	cell-size: (8mm, 10mm),
	edge-stroke: 1pt,
	edge-corner-radius: 5pt,
	mark-scale: 70%,

	blob((0,1), [Add & Norm], tint: yellow, shape: hexagon),
	edge(),
	blob((0,2), [Multi-Head\ Attention], tint: orange),
	blob((0,4), [Input], shape: house.with(angle: 30deg),
		width: auto, tint: red),

	for x in (-.3, -.1, +.1, +.3) {
		edge((0,2.8), (x,2.8), (x,2), "-|>")
	},
	edge((0,2.8), (0,4)),

	edge((0,3), "l,uu,r", "--|>"),
	edge((0,1), (0, 0.35), "r", (1,3), "r,u", "-|>"),
	edge((1,2), "d,rr,uu,l", "--|>"),

	blob((2,0), [Softmax], tint: green),
	edge("<|-"),
	blob((2,1), [Add & Norm], tint: yellow, shape: hexagon),
	edge(),
	blob((2,2), [Feed\ Forward], tint: blue),
)$
```
Rendering these would require some modifications since you would have to include some packages from typst universe, but that is something I can totally volunteer to work on as soon as this gets merged.

cestef:
Hey, thank you so much for taking interest in this PR ! It is already possible to include packages by default via the `addon` field in the config:

```toml
[markdown.math]
engine = "typst"
svgo = "svgo.config.mjs"
css = "styles/typst-embed.css"
addon = "helpers.typ"
```

This will include the `helpers.typ` file for all math blocks. There, you can import your packages, change the default layout, define helper functions, etc.

What you're trying to achieve can also be done via a `typ` code block, which will be rendered as "raw" typst:

~~~md
```typ

= Hello, World!

This is a normal typst file 

$ a + b = 42 $
```
~~~

Keats:
I think that looks amazing. It's pretty big, I'll set up some time to properly review it

claytonwramsey:
Thanks again for implementing this!!!!

I noticed that Typst block rendering for figures in the latest revision (`b93c762`) doesn't seem to work with inline code. For instance, the following implementation doesn't work:

````md
```typ
= Hello, World!
```
````

But the following does:

````md
```typ,include=hello.typ
```
````

(provided that `hello.typ` contains `= Hello, World!`). It looks like this is a regression, since `d41fffc` doesn't have this issue.

When creating an inline feature, it looks like something wonky happens to the SVG. The text content of the code block is inserted after the SVG, and the generated SVG contains very little content. Below is a sample output from the inline declaration.

```html
<img src="data:image/svg+xml,%3Csvg%20class%3D%22typst-doc%22%20viewBox%3D%220%200%200%200%22%20width%3D%220pt%22%20height%3D%2210pt%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20xmlns%3Axlink%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%22%20xmlns%3Ah5%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxhtml%22%3E%0A%20%20%20%20%3Cg%2F%3E%0A%3C%2Fsvg%3E%0A" class="typst-display typst-doc" style=" width: 0em" loading="lazy" decoding="async" alt="">= Hello, World!
```

cestef:
@claytonwramsey Thanks for the feedback ! after taking a look at the code, it was a complete mess, I don't even know how it was supposed to work before lol, should be fixed in [`12fceaa` (#2791)](https://github.com/getzola/zola/pull/2791/commits/12fceaa4dc3baabf8424a277384cb74fad3f74b2)

components/libs/Cargo.toml
```diff
@@ -9,7 +9,22 @@ ammonia = "4"
 atty = "0.2.11"
 base64 = "0.22"
 csv = "1"
-elasticlunr-rs = { version = "3.0.2", features = ["da", "no", "de", "du", "es", "fi", "fr", "hu", "it", "pt", "ro", "ru", "sv", "tr", "ko"] }
+elasticlunr-rs = { version = "3.0.2", features = [
```
p-r-o-g: Korean "ko" seemingly went missing

Keats:
Can you rebase? I'll review it after

cestef:
This should be it, had to fight a bit with git rebase, sorry for the force-pushes

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,142 @@ pub struct ThemeCss {
     pub filename: String,
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum MathRenderingEngine {
+    #[default]
+    None,
+    Typst,
+    Katex,
```
Keats, line 32: Do we need both typst and katex?

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,142 @@ pub struct ThemeCss {
...
+    fn visit_string<E>(self, value: String) -> Result<BoolWithPath, E>
+    where
+        E: de::Error,
+    {
+        Ok(BoolWithPath::True(Some(value)))
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Default, Hash)]
+pub enum BoolWithPath {
```
Keats, line 75: What's that? Can you add a comment?

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,142 @@ pub struct ThemeCss {
...
+            BoolWithPath::True(None) => serializer.serialize_bool(true),
+            BoolWithPath::True(Some(s)) => serializer.serialize_str(s),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Default)]
+pub struct MathRenderer {
+    pub engine: MathRenderingEngine,
+    pub svgo: BoolWithPath,
```
Keats, line 106: is that a path to the binary?

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,142 @@ pub struct ThemeCss {
...
+    pub css: Option<String>,
+    pub addon: Option<String>,
```
Keats, line 108: Only one addon supported? What are addons?

components/config/src/config/markup.rs
```diff
@@ -23,6 +23,142 @@ pub struct ThemeCss {
...
+            #[serde(default)]
+            format: ImageFormat,
+        }
+
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum MathRendererConfig {
+            // String case: math = "engine"
+            Engine(MathRenderingEngine),
+            // Table case with full configuration
```
Keats, line 135: let's just only have the full table so we don't have to support that kind of things

components/config/src/config/markup.rs
```diff
@@ -67,6 +203,12 @@ pub struct Markdown {
     pub insert_anchor_links: InsertAnchor,
     /// Whether to enable GitHub-style alerts
     pub github_alerts: bool,
+    /// Whether to enable math rendering in markdown files
+    pub math: MathRenderer,
+    /// Whether to cache the rendered math
+    pub cache: BoolWithPath,
```
Keats, line 209: should that be in MathRenderer?

components/config/src/config/markup.rs
```diff
@@ -67,6 +203,12 @@ pub struct Markdown {
     pub insert_anchor_links: InsertAnchor,
     /// Whether to enable GitHub-style alerts
     pub github_alerts: bool,
+    /// Whether to enable math rendering in markdown files
+    pub math: MathRenderer,
```
Keats, line 207: How about making it an Option<MathRenderer>?

components/markdown/src/cache.rs
```diff
@@ -0,0 +1,106 @@
...
+use bincode;
+use dashmap::DashMap;
+use errors::Context;
+use serde::{Deserialize, Serialize};
+use std::fs::{self, File, OpenOptions};
+use std::hash::Hash;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+
+/// Generic cache using DashMap, storing data in a binary file
```
Keats, line 11: Can we do like we do for image processing where we generate the output filename based on the inputs so we can just check whether a file exists as kind of a cache.
In practice how often would the same math text be present in many pages?

components/markdown/src/context.rs
```diff
@@ -57,6 +88,23 @@ impl<'a> RenderContext<'a> {
         self.current_page_path = Some(path);
     }
 
+    /// Same as above
```
Keats, line 91: The comment above has been removed

components/markdown/src/markdown.rs
```diff
@@ -436,6 +448,9 @@ pub fn markdown_to_html(
     opts.insert(Options::ENABLE_STRIKETHROUGH);
     opts.insert(Options::ENABLE_TASKLISTS);
     opts.insert(Options::ENABLE_HEADING_ATTRIBUTES);
+    if context.config.markdown.math.engine != config::MathRenderingEngine::None {
+        opts.insert(Options::ENABLE_MATH);
+    }
```
Keats, line 453: so we still need the pulldown_cmark math option even with typst?

components/markdown/src/markdown.rs
```diff
@@ -451,6 +466,57 @@ pub fn markdown_to_html(
...
     let mut next_shortcode = html_shortcodes.pop();
     let contains_shortcode = |txt: &str| -> bool { txt.contains(SHORTCODE_PLACEHOLDER) };
+    let addon = context
+        .config
+        .markdown
+        .math
+        .addon
+        .as_ref()
+        .map(|path| read_file(Path::new(&path)).ok())
+        .flatten();
```
Keats, line 476: what's an addon

components/markdown/src/markdown.rs
```diff
@@ -451,6 +466,57 @@ pub fn markdown_to_html(
...
+
+    let styles = context
+        .config
+        .markdown
+        .math
+        .css
+        .as_ref()
+        .map(|path| read_file(Path::new(&path)).ok())
+        .flatten();
```
Keats, line 485: Do we need that during rendering or should users have the css in their main CSS files?

components/markdown/src/markdown.rs
```diff
@@ -451,6 +466,57 @@ pub fn markdown_to_html(
...
+        }
+    }
+
+    if matches!(context.config.markdown.math.svgo, BoolWithPath::True(_)) {
+        Svgo::default().check_bin().map_err(|e| {
+            Error::msg(format!(
+                "Error checking svgo installation, make sure it's installed and in your PATH: {}",
+                e
+            ))
+        })?;
```
Keats, line 518: How much difference do we get by running svgo on the generated svg? 
Zola doesn't want to call other tools during the process, it's probably easier to set up some pipelines outside of Zola (or add the concepts of post-build pipeline in Zola).

components/markdown/src/markdown.rs
```diff
@@ -451,6 +466,57 @@ pub fn markdown_to_html(
...
+    let styles = context
+        .config
+        .markdown
+        .math
+        .css
+        .as_ref()
+        .map(|path| read_file(Path::new(&path)).ok())
+        .flatten();
+
+    let (mut compiler, cache): (
```
Keats, line 487: let's be clear with naming, i got confused seeing compiler later down the file. Prefix with math_ ?

components/markdown/src/markdown.rs
```diff
@@ -698,6 +810,33 @@ pub fn markdown_to_html(
                         event
                     });
                 }
+
+                Event::InlineMath(ref content) | Event::DisplayMath(ref content) => {
```
Keats, line 814: So we have the math option for the inline math and codeblocks for bigger chunks?

components/markdown/src/math/svgo.rs
```diff
@@ -0,0 +1,50 @@
...
+            .map_err(|e| e.to_string())?;
+        child.stdin.as_mut().unwrap().write_all(svg.as_bytes()).unwrap();
+        let output = child.wait_with_output().map_err(|e| e.to_string())?;
+        if !output.status.success() {
+            return Err(String::from_utf8_lossy(&output.stderr).to_string());
+        }
+
+        Ok(String::from_utf8_lossy(&output.stdout).to_string())
+    }
+}
```
Keats, line 50: That's probably something that should not be in the PR

components/markdown/src/math/typst/format.rs
```diff
@@ -0,0 +1,55 @@
...
+    styles: Option<&str>,
+) -> String {
+    let height =
+        HEIGHT_RE.captures(svg).and_then(|caps| caps[1].parse::<f64>().ok()).unwrap_or(0.0);
+
+    let width = WIDTH_RE.captures(svg).and_then(|caps| caps[1].parse::<f64>().ok()).unwrap_or(0.0);
+    let mut svg = svg.to_string();
+
+    if render_mode == MathRenderMode::Raw {
+        // Add 10pt to the height to account for the padding
```
Keats, line 26: shouldn't be that done via CSS?

components/markdown/src/math/typst/format.rs
```diff
@@ -0,0 +1,55 @@
...
+            &format!("height=\"{}pt\"", height + 10.0),
+            1,
+        );
+    }
+
+    let shift = align.map(|align| height - align);
+    let shift_em = shift.map(|shift| shift / EM_PER_PT);
+
+    if let Some(styles) = styles {
+        svg = svg.replacen(">", &format!("><style>{}</style>", styles), 1);
```
Keats, line 38: it feels like being able to add classes would be better than style no?

components/markdown/src/math/typst/format.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    let url_encoded = urlencoding::encode(&svg);
+    format!(
+            "<img src=\"data:image/svg+xml,{url_encoded}\" class=\"{} typst-doc\" style=\"{} width: {}em\" loading=\"lazy\" decoding=\"async\" alt=\"\" />",
```
Keats, line 43: why not raw svg?

components/markdown/src/math/typst/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+            (0..face_count).map(move |face| {
+                Font::new(buffer.clone(), face).expect("failed to load font from typst-assets")
+            })
+        })
+        .collect()
+}
+
+/// Fake file
+///
+/// This is a fake file which wrap the real content takes from the md math block
```
Keats, line 47: takes -> taken?

components/markdown/src/math/typst/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        if path.exists() {
+            return Ok(path);
+        }
+
+        // Download the package
+        let package_url = format!(
+            "https://packages.typst.org/{}/{}-{}.tar.gz",
+            package.namespace, package.name, package.version
+        );
```
Keats, line 126: I feel like package declarations/installation should not be done while rendering markdown but upstream, eg defined in the config and a command to download things

components/markdown/src/math/typst/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+        // Prepare source based on mode
+        let source = match mode {
+            MathRenderMode::Display => templates::display_math(&source, self.addon.as_deref()),
+            MathRenderMode::Inline => templates::inline_math(&source, self.addon.as_deref()),
+            MathRenderMode::Raw => templates::raw(&source, self.addon.as_deref()),
+        };
+
+        // Generate cache key
+        let key = {
+            let mut hasher = XxHash64::with_seed(42);
```
Keats, line 242: Is the speed difference worth adding that crate? Image proc still uses the std DefaultHasher because in practice it's not really noticeable

components/markdown/src/math/typst/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+                    }
+                    MathRenderMode::Raw | MathRenderMode::Display => None,
+                };
+
+                format_svg(&minified, align, mode, self.styles.as_deref())
+            }
+            ImageFormat::Webp => {
+                // let _pixmap = typst_render::render(page, 300.0);
+                // TODO: svg2webp
+                unimplemented!("WebP is not supported yet")
```
Keats, line 307: Is it not supported in typst?

components/markdown/src/math/typst/templates.rs
```diff
@@ -0,0 +1,52 @@
...
+    let addon = addon.unwrap_or("");
+    format!(
+        r#"
+#set page(height: auto, width: auto, margin: 0pt, fill: none)
+#set text(16pt)
+{addon}
+{code}
+"#,
+    )
+}
```
Keats, line 52: Do we want people to be able to use their own templates?

components/site/src/lib.rs
```diff
@@ -95,6 +98,17 @@ impl Site {
         let imageproc = imageproc::Processor::new(path.to_path_buf(), &config);
         let output_path = path.join(config.output_dir.clone());
 
+        let caches = match config.markdown.cache {
+            config::BoolWithPath::True(ref maybe_path) => Some(Arc::new(match maybe_path {
```
Keats, line 102: When would the user want to specify their own cache directory?

docs/content/documentation/content/math.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+```toml
+[markdown]
+cache = "path/to/cache" # true, false, or "path/to/cache"
+```
+
+## SVG Optimization
+
+Enable `svgo = true` to optimize the generated SVG files using [SVGO](https://svgo.dev). This can significantly reduce the file size of complex mathematical expressions.
```
Keats, line 86: so the reason we do that is because it's inlined as an image right? So there's no SVG to optimize in a build pipeline. I'm surprised there's not a library like svgo in Rust but I would leave that part out for now

Keats (changes requested):

components/config/src/config/markup.rs
cestef: The idea behind the generic `MathCompiler` trait was to add support for multiple backends, many people still use LaTeX in my opinion.

components/config/src/config/markup.rs
cestef: the idea was to put it here in case we needed cache for other stuff than math rendering in the future, the comment above is erroneous

components/markdown/src/cache.rs
cestef: creating files can be quite expensive, especially when you have such high volume of rendering going on. As a reference, my blog with about 10 posts, each with moderate math content has about a 1000 maths elements (inline + block). Small variables such as $x$ or $y$ are very common and re-rendering them would be a loss of time. I posted a benchmark above for the average compilation time with 500 elements (cold run is about 30s, while hot is 0.8s)

components/markdown/src/markdown.rs
cestef: Yes, `ENABLE_MATH` in pulldown_cmark enables the **markdown** parsing of the math elements, which then fires the appropriate events when we iterate later on.

components/markdown/src/math/typst/format.rs
cestef: typst rendered SVGs tend to have a lot of children elements, which can cause serious performance issues if you write math-heavy content. Putting them inside an image drastically reduces the DOM elements count. https://developer.chrome.com/docs/lighthouse/performance/dom-size

components/markdown/src/math/typst/mod.rs
cestef: Packages are part of the typst compiler which resolves them for us at compile time.

components/markdown/src/math/typst/mod.rs
cestef: typst supports png output, but not webp yet. https://typst.app/docs/reference/png/

components/markdown/src/math/typst/templates.rs
cestef: I don't see much use, people can already configure templates further with the addon option, which includes additional code in the templates. inline and block templates only consist of base page settings and a math block. For further customization people can use the raw rendering via code blocks.

components/site/src/lib.rs
cestef: CI/CD pipelines or network shared cache, I can remove this if you feel it's overkill

Keats: Yeah i think it's better to keep things simple

components/config/src/config/markup.rs
Keats: But for the Zola crowd, do we need multiple renderer?

claytonwramsey: Chiming in as a Zola user who's been using cestef's fork: I personally am using the Typst math rendering right now, but I had to manually port all the math on my blog from LaTeX to do so. I think a lot of people using this feature might be moving over from a previous math implementation, so they will likely use LaTeX, but my experience is that writing new posts with Typst is way easier. In the ideal world, it would be possible to specify math on a per-page or per-block basis, but that might be a little too complex.

Keats: > my experience is that writing new posts with Typst is way easier

That's what I've seen online as well. I don't use any math/LaTeX whatsoever myself so I don't have an opinion myself on that other than keeping things simple

cappuccinocosmico: Imo, its about supporting legacy use cases, 95% of all math/physics papers are still written in latex, and its also the standard format for math in markdown: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions. I think the main reason everyone on this thread is using typst is just the attraction to rust. But if no one on this thread actually wants support for Katex, it might be worth it to remove it for now and only add it if someone requests over typst.

components/config/src/config/markup.rs
```diff
@@ -67,6 +203,12 @@ pub struct Markdown {
...
+    /// Whether to cache the rendered math
+    pub cache: BoolWithPath,
+    /// Whether to enable GitHub-style alerts
+    pub github_alerts: bool,
```
cappuccinocosmico, line 211: Github alerts seems to be duplicated in this enum, if you try to compile the program you get an error around this 
image

components/config/src/config/markup.rs
```diff
@@ -243,6 +385,9 @@ impl Default for Markdown {
             lazy_async_image: false,
             insert_anchor_links: InsertAnchor::None,
             github_alerts: false,
+            math: MathRenderer::default(),
+            cache: BoolWithPath::True(None),
+            github_alerts: false,
```
cappuccinocosmico, line 390: Also errors out on this declaration as well

components/config/src/config/markup.rs
p-r-o-g: >  I think the main reason everyone on this thread is using typst is just the attraction to rust

This is not the case for everyone probably, example is me, I am rather a Typst user & enthusiast (rust as well but not as much experience using it) and actually would like to use it for a documentation website involving lots of mathematics code. Typst is gaining functionality and convenience very quickly and left LaTeX already behind in most aspects (my opinion).

lilymonad: I personaly still use LaTeX because it is what I used untill now, and also for tikz figures and graphs (not with zola, just in my other documents). If Typst has the same math rendering quality as Katex and maintaining multiple engines is a pain, it would bother me to switch to Typst but I would do it anyways. I won't complain on this decision, I'm already grateful people made math rendering possible.

I can only see one downside of having only Typst: if people write things both in their blog and in research papers, they would need to write things in two languages each time. But I guess it's rare.

components/markdown/src/math/typst/mod.rs
nicolandu: Would it be easier to go the PNG=>WebP route?

components/config/src/config/markup.rs
lukasjuhrich: > I think the main reason everyone on this thread is using typst is just the attraction to rust

_Every_ rust rewrite has a portion of supporters purely based on the fact that it is new, fancy, and written in rust. Existence of such supporters is a terrible indicator to judge a project's actual quality, soundness of architectural decisions and overall adoption trajectory.

When typst first got out, I was very critical because it was quite young and it had a few rough edges, but I quickly flipped because it made me _way_ faster. I wrote my 140p thesis in typst and will never go back. After only one year of learning I was able to get my hands dirty with custom commands and document introspection in a way that after 7 years of latex I was never able to come close to.
Typst does numerous things a _ton_ better in a way that latex will never be able to: To name one, TeX is a macro processing language, and this will never change. That makes it impossible to give good syntactical error messages or editor completion (heuristics are always possible, but as soon as you have custom commands you are out the window), and you will not be able to change that. To name another, the packaging ecosystem is easy to use and growing rapidly. And we don't have to talk about compilation speed (although to be fair the static web-rendering use case is different from getting a live preview while typing, where one sees the most benefit).

All this is to say two things.
_First_, I guarantee you there will be a growing portion of users who will want to use typst and feel very strongly about its adoption in up-and-coming projects such as `zola`.
_Second_, LaTeX is still a requirement for most people, and will continue to be. A portion of such people are likely to cast out `zola` if they want to typeset math but don't want the burden of learning a „fancy new typesetting language that nobody uses and will probably die in a few years, anyway“. Especially in academia things are slowly moving (rightfully so), and the change in adoption will probably come with a generational change; some communities will never reach it. Something similar is happening with automated theorem proving, and something similar happened in math with the introduction of category theoretical methods. We're talking on the scale of decades here.

Sure, much of this is anecdotal, and I might be wrong about my projections, but I like to think spending 6+ months doing nothing but writing typst and having had extensive LaTeX experience beforehand counts for something other than „irrelevant because opinion-based and not measurable“.

I am however not sure what you should do with this information, because I can't seriously tell someone to “just support both” without being the one who has to carry the maintenance burden.

components/markdown/src/markdown.rs
leiserfg: The Katex module has a js engine embedded, isn't it possible to use it to run svgo (so calling the external tool is not needed?)

Dimev:
I've also stumbled upon https://docs.rs/math-core/latest/math_core/, as well as previously used https://docs.rs/latex2mathml/latest/latex2mathml/ myself. Both are rust projects, but I'm not aware of how complete they are.

Anyway, thank you for working on this! Typst is also great to work with.

OJFord:
Are you still working on this @cestef? I'd love to be able to use Zola with Typst, and your own blog using it looks great, but I'm not sure about using a fork with uncertain future for it even if 'production' is only a personal blog!

Thanks for your work on it so far & so glad to see Zola maintainers seem open to it.

I can potentially help, but if you're still willing to work on it perhaps the best way I can is just to test/use it and see how it goes?

cestef:
hi, thanks a lot for your interest in this pr! I kind of abandonned it in a somewhat "working" state, there is still quite a lot of refactoring needed, merging with upstream should be fine. the biggest issue is that integrating with the typst library directly makes the binary size explode. maybe a plugin system would make sense here ? it seems like a lot of work anyways. I will try to work a bit on this when I get home (I'm currently at FOSDEM), feel free to tinker with it in the meantime :D

OJFord:
> the biggest issue is that integrating with the typst library directly makes the binary size explode. maybe a plugin system would make sense here ?

Seems there's some attempt in that direction in #3007, but [not clear](https://github.com/getzola/zola/pull/3007#issuecomment-3509533752) to me if that's progressing/likely to happen.

It would seem to make sense for 'renderer' to be pluggable (built-in markdown) default, or some sort of hook system such that typst-ts could be used as [they describe](https://myriad-dreamin.github.io/typst.ts/cookery/direction/responsive.html). (They provide an implementation for [hexo](https://hexo.io) since that does support plugins, for example.)

Keats:
So I've been rewriting a lot of the internals and shortcodes no longer exist in the tera2 branch. Markdown rendering is also much simpler so it will be much easier to add for the codeblock part. As for replacing markdown entirely and adding typst as a full 2nd language for the content, I'm not 100% sure. I would want to make sure we can use the same code highlighting engine for example (typst uses syntect, zola giallo) and also see how many people actually want the whole thing in practice.

laloe74:
I have no need to use Typst and don't like volume bloat. But I need the native KaTeX conversion MathML, which is the last bit of Zola I need.

zlfn:
Hello, is there any progress with this feature? I think the problem of binary growing can be solved by feature flag. (like `indexing-ja`, `indexing-zh`)

Keats:
I would be fine with typst being included by default.

cadenkriese:
Typst 0.15 supports MathML in its HTML export mode, see https://typst.app/blog/2026/typst-0.15#math-ml. The browser support is getting better. The WebKit team, for example, spent significant time on bug fixes for MathML in Safari 27.

Keats:
Now that 0.23 is released we can have a look at typst/math again. Can we work on an issue to discuss exactly what we want/need to support. I'd like to do RFCs ideally but I need time to create the first one x)

thomasweitzel:
Is `pulldown-cmark`'s `ENABLE_MATH` the right abstraction if Zola wants both TeX and Typst as first-class math languages?

`pulldown-cmark` is [explicitly TeX-oriented](https://pulldown-cmark.github.io/pulldown-cmark/specs/math.html): `$...$` produces `InlineMath`, while `$$...$$` produces `DisplayMath`. Typst natively uses `$...$` for inline and `$ ... $` for block math. This PR therefore puts Typst formulas inside TeX-style Markdown delimiters and maps `$$...$$` to Typst's `$ ... $` internally.

The resulting `InlineMath` / `DisplayMath` events are renderer-neutral, but the syntax producing them isn't. Should Zola establish the TeX-style `$`/`$$` convention for all renderers, or should the parser support the native syntax of the selected math language?

Keats:
The first RFC is https://github.com/getzola/zola/pull/3258
We can use that as a template for the math support one. It's the same template as the Rust lang RFCs

PR #2793, opened 2025-02-09
feat: add zola version variable

### LINKED ISSUE #2749: Feature Request: Zola version variable
Jekyll has a variable for printing the current version number.

For example, with Jekyll version 4.3.3, ```{{ jekyll.version }}``` will print ```4.3.3```.

I use this in an existing Jekyll site to set the [generator metadata name](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name).

```html
<meta name="generator" content="Jekyll 4.3.3" />
```

This can be useful when debugging to be certain that a page was generated with a specific version of the software.

I would like to request a similar variable for Zola.

extua:
Simple PR to expose the Zola version number as a Tera variable. See issue #2749 outlining the feature request.

The only way to export the cargo version number from the top-level cargofile down to the workspace was to set

```toml
[workspace.package]
version = "0.19.2"
```

following [the Rust reference doc example](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table).
This means this version number also needs to be set here as well as at the top of the file.

Sanity check:

## Code changes

_This change is a (small) new feature and I can update the documentation after it is merged._


Cargo.toml
```diff
@@ -56,6 +56,8 @@ indexing-ja = ["libs/indexing-ja"]
 
 [workspace]
 members = ["components/*"]
+[workspace.package]
+version = "0.19.2"
```
Keats, line 60: No way to avoid specifying it twice?

extua: I don't think it's possible to de-duplicate these within the root cargofile. Another option would be to set the version number as a const in the root package, and then import it into the workspace for `components/content`. I'm not sure how to do that though.

salim-b: > No way to avoid specifying it twice?

As long as Cargo supports the full TOML spec (which I suppose), you could rewrite the second key in the [dotted key](https://toml.io/en/v1.0.0#keys) form:

```toml
[workspace]
members = ["components/*"]
package.version = "0.19.2"
```

cestef: To avoid specifying the version twice, we could retrieve `env!("CARGO_PKG_VERSION")` in the cli, and pass it to `Site::new()` ?

Keats: that's a good idea

extua: > To avoid specifying the version twice, we could retrieve `env!("CARGO_PKG_VERSION")` in the cli, and pass it to `Site::new()` ?

Sorry for the delay responding, I will try this out but am not super familiar with the way everything is structured so I need some time to work it out and create a minimal working example.

Keats:
I've added the commit to the next branch

PR #2797, opened 2025-02-12
Add `include` attribute for code fences

cestef:

https://zola.discourse.group/t/include-and-copy-attributes-for-code-fences/

Sanity check:

## Code changes

This PR adds support for two fence settings:

<code>```include=<filename></code>

- Allows including a local file as the content of this codeblock, useful for large files

<code>```copy</code>

- Adds a `data-copy` attribute to the code-block, useful if you want to add copy buttons only to some blocks.

### DIFF
--- a/components/markdown/src/codeblock/mod.rs
+++ b/components/markdown/src/codeblock/mod.rs
@@ -2,9 +2,11 @@ mod fence;
 mod highlight;
 
 use std::ops::RangeInclusive;
+use std::path::PathBuf;
 
 use errors::{bail, Result};
 use libs::syntect::util::LinesWithEndings;
+use utils::fs::read_file;
 
 use crate::codeblock::highlight::SyntaxHighlighter;
 use config::highlighting::{resolve_syntax_and_theme, HighlightSource};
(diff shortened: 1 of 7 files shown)


components/markdown/src/codeblock/mod.rs
```diff
@@ -112,19 +119,28 @@ impl<'config> CodeBlock<'config> {
...
             },
             html_start,
         ))
     }
 
+    pub fn include(&self, base: Option<&PathBuf>) -> Option<String> {
+        let path = base?.join(self.include.as_ref()?);
+        let res = read_file(&path);
+
+        res.ok()
```
ppom0: It may be better to add a warning when the path doesn't exist, instead of silently failing?
Would help debugging missing paths.

Keats:
Can you cherry pick the commits on the new next branch?

cestef:
@Keats this should do it

components/markdown/src/codeblock/mod.rs
```diff
@@ -112,19 +119,33 @@ impl<'config> CodeBlock<'config> {
...
         ))
     }
 
+    pub fn include(&self, base: Option<&PathBuf>) -> Option<String> {
+        let path = base?.join(self.include.as_ref()?);
+        let res = read_file(&path);
+        match res {
+            Ok(content) => Some(content),
+            Err(e) => {
+                eprintln!("Warning: Failed to include {:?}: {}", path, e);
```
Keats: I think that should be an error

Keats (changes requested):
Can you split the include and copy feature in different PRs? The copy one is easy but the include is a bit more involved (error handling, maybe caching etc)

kKdH:
Hey guys, I created a [post in the forum](https://zola.discourse.group/t/include-content-by-tagged-regions/2654) and asked for a similar feature to this one. Maybe you are interested to take a look.

Keats:
I think since the content is now templated, you can just use `load_data` function in the codeblock and that should work?

ppom0:
@Keats, do you mean that we can now use built-in functions in the .md content files?

I was not aware of that, that would be cool!

Keats:
Not right now but in the next version (release this week or the next), every .md content file is also a template just like the rest of your site.

PR #2832, opened 2025-03-27
Add support for building for x86_64-unknown-linux-musl

### LINKED ISSUE #2365: Hope provide Linux musl pre-builds , even if it's slow.
My Linux server is Centos8.

GLIBC version is 2.28. So can't work when use [zola-v0.17.2-x86_64-unknown-linux-gnu.tar.gz](https://github.com/getzola/zola/releases/download/v0.17.2/zola-v0.17.2-x86_64-unknown-linux-gnu.tar.gz)

I try build with source code. But my server configuration is too low. 1 core and 1GB. My server will crash when build . So I can't build succeed.

Please and so thanks!

dstaley:
This PR adds support for compiling for `x86_64-unknown-linux-musl`, which produces a statically-linked binary that does not depend on a specific version of glibc. This is useful in environments that might be using an older version of glibc than the one that Zola was built with when compiling on `ubuntu-22.04`. For example, Amazon Linux 2023 currently uses glibc 2.34, whereas `ubuntu-22.04` results in a binary depending on glibc 2.35.

Originally, I intended to accomplish this using `musl-tools`, but was unable to accomplish this due to the `webp` and `ring` packages not compiling correctly. `cargo-zigbuild` uses Zig as the compiler and linker, which seemed like an easier way than trying to figure out the magic combination of flags.

I've tested the produced binary on Amazon Linux 2023 and it runs successfully. Tools like `objdump` and `ldd` confirm it's a statically linked binary.

Fixes #2365 

Sanity check:

## Code changes

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-musl
           - aarch64-unknown-linux-gnu
           - x86_64-pc-windows-msvc
           - x86_64-apple-darwin
@@ -25,6 +26,8 @@ jobs:
             target: x86_64-pc-windows-msvc
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
+          - os: ubuntu-22.04
+            target: x86_64-unknown-linux-musl
           - os: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
           - os: macos-13
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-musl
           - aarch64-unknown-linux-gnu
           - x86_64-pc-windows-msvc
           - x86_64-apple-darwin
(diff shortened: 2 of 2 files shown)


Keats (approved):
Nice, thanks!

Have you tried running the musl binary and a normal cargo build --release to see the speed difference?

dstaley:
I wrote this PR via Docker running on an Apple Silicon Mac, which was several layers of virtualization and emulation, so I’ll give it a shot on one of my native Linux machines and see if there’s a noticeable difference! Do you happen to know of a decently sized Zola site that I can use to benchmark build times?

Keats:
You can generate some in https://github.com/getzola/zola/blob/master/components/site/benches/gen.py but I haven't run those in a long time so they might not be valid zola sites anymore

dstaley:
On the `small-blog`, glibc is 1.27 ± 0.11 times faster than musl. On `huge-blog`, glibc is 2.49 ± 0.10 times faster than musl.

mattaw:
Question: It looks like the CI `build-and-test.yml` is not building the `-musl` version, but the `-gnu` version - I am not experienced with rust, does this matter? 

[Reasoning `build-and-test.yml` currently is missing `--target ${{ matrix.target }}` from the cargo build commands, and the cargo-zigbuild step, and it failed to compile when I tried adding the musl target to the toolchain install and the cargo build statements.]

PR #2850, opened 2025-04-09
Allow generating empty feeds

### LINKED ISSUE #2837: No feed generated, resulting in a 404, if no page in section
# Bug Report

## Environment

Zola version: 0.20.0

## Expected Behavior

When I use `generate_feeds = true` in a section with no pages, I expect an Atom feed to be generated to that people can subscribe to it even if there’s no entry yet.

## Current Behavior

When I use `generate_feeds = true` in a section with no pages, no Atom feed is generated, and if I link to it Zola doesn’t warn me even though it leads to a 404.

## Step to reproduce

Create an empty section with `generate_feeds = true` and link to the Atom feed.

senekor:
closes #2837


Keats:
Thanks!

PR #2853, opened 2025-04-15
Allow arbitrary file extensions for templates

### LINKED ISSUE #2772: Generating feeds with files not ending in xml/html
# Bug Report

## Environment

Zola version: zola 0.19.2

## Expected Behavior

I can put any file extension in the templates folder and it be counted as a template, or failing that entering it as a `feed_filename` should convert it to a template for rendering feeds

## Current Behavior

```
sean@little-my ~/G/G/R/website (master)> zola serve
Building site...
-> Creating 6 pages (0 orphan) and 3 sections
Error: Failed to serve the site
Error: Tried to render `cal.ics` but the template wasn't found
sean@little-my ~/G/G/R/website (master) [1]> ls -l templates/cal.ics
-rw-r--r-- 1 sean users 716 Jan 13 13:37 templates/cal.ics
```

## Step to reproduce
1. change feed filenames to `feed_filenames = [ "atom.xml", "cal.ics"]`
2. create a file in templates named `cal.ics`
3. `zola serve` will fail with the above error

Optionally clone and `zola serve` https://gitlab.com/rust-manchester/website

Additionally I forked zola to see if what I was attempting was possible and got a working version using
https://github.com/seam345/zola/commit/0cb8148e30a0be2a9fcb802cd31408dab5568aa1
This was very hacked together though and only works for .ics extensions I imagine there are many more plain text formats that could be used to generate feeds in the same way

Did work though :D https://rustmanchester.co.uk/events/cal.ics

senekor:
There is no benefit to restricting file extensions of templates. Allowing it will enable things like custom JSON feeds. Users can also choose to use `.tera` as the extension for their templates, enabling some text editors to handle syntax highlighting better.

closes #2772

### DIFF
--- a/components/templates/src/lib.rs
+++ b/components/templates/src/lib.rs
@@ -44,8 +44,7 @@ pub fn render_redirect_template(url: &str, tera: &Tera) -> Result<String> {
 }
 
 pub fn load_tera(path: &Path, config: &Config) -> Result<Tera> {
-    let tpl_glob =
-        format!("{}/{}", path.to_string_lossy().replace('\\', "/"), "templates/**/*.{*ml,md,txt}");
+    let tpl_glob = format!("{}/{}", path.to_string_lossy().replace('\\', "/"), "templates/**/*");
 
     // Only parsing as we might be extending templates from themes and that would error
     // as we haven't loaded them yet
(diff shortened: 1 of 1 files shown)


Keats:
That's going to cause some issues potentially for example with shortcodes where we rely on the extension to distinguish their types. It's also potentially an issue in terms of consistency since now to extend a theme template it's always `filename.html` (or .xml). With that it might be  `.tera`, `.jinja2` etc.

senekor:
Can we at least allow `.json` templates? while ours is non-standard, there is a standard json feed formt: https://www.jsonfeed.org/

Keats:
We can add .json and .ics at least

senekor:
superseded by #2856

PR #2855, opened 2025-04-16
Sort pages by permalink if sort_by is unspecified

senekor:

I think this is consistent with the fact that the permalink is already used to break ties if two pages have the same `date` or `weight`.


components/content/src/sorting.rs
```diff
@@ -17,7 +17,7 @@ pub fn sort_pages(pages: &[&Page], sort_by: SortBy) -> (Vec<PathBuf>, Vec<PathBu
             SortBy::Title | SortBy::TitleBytes => page.meta.title.is_some(),
             SortBy::Weight => page.meta.weight.is_some(),
             SortBy::Slug => true,
-            SortBy::None => unreachable!(),
+            SortBy::None => true, // fallback to permalink sorting
```
Keats, line 20: I would rather add `SortBy::Permalink` and keep `SortBy::None` so we don't sort a section that doesn't need to be sorted

senekor: Okay will do

senekor:
superseded by #2857

PR #2780, opened 2025-01-22
Add AVIF support for resize_image() function

iWangJiaxiang:
The work for adding AVIF support for `resize_image()` has been finished, with pure rust crate.

With this feature, user may convert their images into avif format and the size will be reduced significantly.

However, the speed is slow in the developing env of GITHUB Code Space. I will take further tests on my machine.

Sanity check:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes
(Delete or ignore this section for documentation changes)

* [x] Are you doing the PR on the `next` branch?

If the change is a new feature or adding to/changing an existing one:

* [x] Have you created/updated the relevant documentation page(s)?

PR #2862, opened 2025-04-20
Improved AVIF support

lpulley:
This follows up #2780 with the ability to control the AVIF encoder speed (1-10) and fixes handling of non-RGB8/RGBA8 color modes when encoding AVIF.

---

Regarding non-RGB color support, Zola currently blows up if an input image for AVIF resizing is not RGB `u8`, I think because `image` hits asserts regarding the buffer size. :(

```
➜ zola build                    
Building site...
-> Creating 23 pages (0 orphan) and 0 sections
thread '<unnamed>' panicked at components/imageproc/src/processor.rs:83:25:
assertion `left == right` failed: Invalid buffer length: expected 507000 got 169000 for 500x338 image
  left: 507000
 right: 169000
thread '<unnamed>' panicked at components/imageproc/src/processor.rs:83:25:
assertion `left == right` failed: Invalid buffer length: expected 7482240 got 2494080 for 1920x1299 image
  left: 7482240
 right: 2494080
```

This fixes that by supporting all the (known) values for `image::ColorType`.

### DIFF
--- a/components/imageproc/src/format.rs
+++ b/components/imageproc/src/format.rs
@@ -17,14 +17,10 @@ const DEFAULT_SPEED_AVIF: u8 = 5;
 /// Thumbnail image format
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Format {
-    /// JPEG, The `u8` argument is JPEG quality (in percent).
-    Jpeg(u8),
-    /// PNG
+    Jpeg { quality: u8 },
     Png,
-    /// WebP, The `u8` argument is WebP quality (in percent), None meaning lossless.
-    WebP(Option<u8>),
-    /// AVIF, The first `u8` argument is AVIF quality (in percent). The second `u8` argument is AVIF speed.
-    Avif(u8, u8),
+    WebP { quality: Option<u8> }, // 'None' means lossless
+    Avif { quality: u8, speed: u8 },
 }
 
 impl Format {
@@ -42,7 +38,9 @@ impl Format {
         };
         match format_from_auto {
             "jpeg" | "jpg" => match quality.unwrap_or(DEFAULT_QUALITY_JPEG) {
-                valid_quality @ QUALITY_MIN_JPEG..=QUALITY_MAX_JPEG => Ok(Jpeg(valid_quality)),
+                valid_quality @ QUALITY_MIN_JPEG..=QUALITY_MAX_JPEG => {
+                    Ok(Jpeg { quality: valid_quality })
+                }
                 invalid_quality => Err(anyhow!(
                     "Quality for JPEG must be between {} and {} (inclusive); {} is not valid",
                     QUALITY_MIN_JPEG,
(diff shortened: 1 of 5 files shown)


components/imageproc/src/format.rs
```diff
@@ -2,6 +2,8 @@ use errors::{anyhow, Result};
 use std::hash::{Hash, Hasher};
 
 const DEFAULT_Q_JPG: u8 = 75;
+const DEFAULT_Q_AVIF: u8 = 70;
+const DEFAULT_S_AVIF: u8 = 10;
```
Keats: We should probably stop using single letter, I only understood S was for speed by looking a comment below

components/imageproc/src/format.rs
```diff
@@ -12,16 +14,24 @@ pub enum Format {
     Png,
     /// WebP, The `u8` argument is WebP quality (in percent), None meaning lossless.
     WebP(Option<u8>),
-    /// AVIF, The `u8` argument is AVIF quality (in percent), None meaning lossless.
-    Avif(Option<u8>),
+    /// AVIF, The first `u8` argument is AVIF quality (in percent). The second `u8` argument is AVIF speed.
```
Keats: No more lossless?

components/imageproc/src/format.rs
lpulley: I agree. I matched what was there, but I will change them to "QUALITY" and "SPEED".

components/imageproc/src/format.rs
lpulley: There never was. The default quality 70 was just applied further down :/

components/imageproc/src/format.rs
lpulley: `ravif` (the crate behind `image`'s AVIF implementation) uses different defaults (quality 80, speed 5); any interest in switching the defaults in Zola?

https://github.com/kornelski/cavif-rs/blob/ed676dd1a3d9726b740ad679843bad55d3a84ebd/ravif/src/av1encoder.rs#L82-L84

"Fast but crappy" doesn't strike me as a great default speed, and (anecdotally) quality 70 is disappointing. Feel free to play around with the options on this branch, though, and see what you think.

lpulley:
While working on this, it occurred to me that it'll be awkward if `resize_image` ever wants to support a format with a different range of valid values for quality or speed. It got me thinking that, instead of `format="avif", quality=90, speed=7`, maybe we should do `format="avif;quality=90;speed=7"`, or something like that. I thought it might make the documentation a bit easier to understand, too. But it's not immediately clear to me how the parsing would have to change to allow that.

components/imageproc/src/format.rs
Keats: We can switch to their default

lpulley:
Here; I tried making it so that the quality and speed ranges are checked separately for each format (and allowed quality 0 for WebP), with tests for the minimum and maximum qualities for each format.

The "min speed" AVIF tests are pretty slow. I might add a very small PNG (something like 30px) for those tests, to avoid wasting time on them.

Keats (approved):
Looks good thanks!

lpulley:
Thanks!

Upon review, I think quality 0 WebP hashes the same as lossless WebP... probably we need to change the `None` case for WebP hash to something like 255?

PR #2871, opened 2025-04-30
Fix hashes in `can_resize_image`

lpulley:
#2862 (and then #2872) changed the hashing procedure for `Format`, so these have changed.

https://github.com/getzola/zola/actions/workflows/build-and-test.yml has been failing on `next` since 36c07474282332f055ea5688ace5151b855610ee; this should fix that.


lpulley:
Looking good now: https://github.com/getzola/zola/actions/runs/14802063351

PR #2873, opened 2025-04-30
Avoid creating empty/incomplete processed image files

lpulley:
`ImageOp` creates the output file before writing to it (of course), but if writing the image into the file fails with some error, the file is left around (despite being empty or corrupted). This means a subsequent `zola build` raises no errors, and it "looks like" it worked on the second attempt, but the file(s) are empty or corrupt.

I think a good solution here is to write to a `tempfile` and then move it to the final destination.

---

The following events now occur when running a `zola build` that needs to create one new processed image (in this case `example.1345f5935a4925f7.avif`):

```
static/processed_images/ CREATE .tmpE0FV7i
static/processed_images/ OPEN .tmpE0FV7i
static/processed_images/ MODIFY .tmpE0FV7i
static/processed_images/ ATTRIB .tmpE0FV7i
static/processed_images/ MOVED_FROM .tmpE0FV7i
static/processed_images/ MOVED_TO example.1345f5935a4925f7.avif
static/processed_images/ CLOSE_WRITE,CLOSE example.1345f5935a4925f7.avif
```

### DIFF
--- a/components/imageproc/src/processor.rs
+++ b/components/imageproc/src/processor.rs
@@ -61,13 +61,15 @@ impl ImageOp {
         let f = File::create(&self.output_path)?;
         let mut buffered_f = BufWriter::new(f);
 
-        match self.format {
+        let write_result: Result<()> = match self.format {
             Format::Png => {
                 img.write_to(&mut buffered_f, ImageFormat::Png)?;
+                Ok(())
             }
             Format::Jpeg { quality } => {
                 let mut encoder = JpegEncoder::new_with_quality(&mut buffered_f, quality);
                 encoder.encode_image(&img)?;
+                Ok(())
             }
             Format::WebP { quality } => {
                 let encoder = webp::Encoder::from_image(&img)
@@ -77,6 +79,7 @@ impl ImageOp {
                     None => encoder.encode_lossless(),
                 };
                 buffered_f.write_all(memory.as_bytes())?;
+                Ok(())
             }
             Format::Avif { quality, speed } => {
                 let mut avif: Vec<u8> = Vec::new();
@@ -88,10 +91,14 @@ impl ImageOp {
                     img.color().into(),
                 )?;
                 buffered_f.write_all(&avif.as_bytes())?;
+                Ok(())
             }
-        }
+        };
 
-        Ok(())
+        if write_result.is_err() {
+            fs::remove_file(&self.output_path)?;
+        }
+        write_result
     }
 }
 
--- a/components/imageproc/src/processor.rs
+++ b/components/imageproc/src/processor.rs
@@ -41,6 +41,7 @@ impl ImageOp {
             return Ok(());
         }
 
+        let input_permissions = fs::metadata(&self.input_path)?.permissions();
         let reader =
             ImageReader::open(&self.input_path).and_then(ImageReader::with_guessed_format)?;
         let mut decoder = reader.into_decoder()?;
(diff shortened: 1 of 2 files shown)


Keats:
looks ok to me!

lpulley:
I'm not sure how to write a good test for this; ideally, writing to the output never fails 😅

I also changed to a `.tmp`-file approach because that guards against SIGINTs et al. If we write to a `.tmp` file and move it to the real output path only if writing to it succeeds, then even if `zola build` crashes, we shouldn't leave any corrupt/empty image files around. (I played with this by hand by just issuing a SIGINT to `zola build` while it's writing images.)

components/imageproc/src/processor.rs
```diff
@@ -21,6 +22,7 @@ use crate::helpers::get_processed_filename;
 use crate::{fix_orientation, ImageMeta, ResizeInstructions, ResizeOperation};
 
 pub const RESIZED_SUBDIR: &str = "processed_images";
+const TMP_OUTPUT_SUFFIX: &str = ".tmp";
```
Keats: can we use the tempdir (or tempfile can't remember) crate so the folders get cleaned up automatically?

lpulley: I vanished for a while; sorry! I can refactor this to use [tempfile] if you prefer, yeah.

[tempfile]: https://docs.rs/tempfile/latest/tempfile/

lpulley: Updated! See also: the updated `inotifywait` events in the PR description.

lpulley: The resulting files have kinda odd permissions (`0600`); not sure if Zola cares to change that.

lpulley: Updated to apply permissions of input file to the temporary output file before moving it.

I think my pushes don't trigger CI, but it would be nice to see this run on CI.

Keats: yes go for it, we use it already

woops i had the tab open too long and didn't see all the comments

Keats (approved):

Keats:
hm not sure why the CI is not starting... try to rebase from next and force push to try to make it do something?

lpulley:
I've rebased and forced a few times as I've worked, but it doesn't run. I assumed it's because I'm not a maintainer and it wants to protect access to repository secrets, as a security policy.

lpulley:
Maybe if you push an empty commit (`--allow-empty`) it'll go?

lpulley:
Ah, but it's a fork. That might be why (and you can't push, I think).

Keats:
Yeah I was just checking that, I don't think I can push

lpulley:
Seems like it doesn't work for several other contributors I checked, either. I seem to recall it's a security setting or something like that. (You don't want precious secrets being accessible to whatever job someone cooks up in their fork. Or precious CI time being consumed.)

Keats:
We don't have any secrets in the GH config and usually there's a button to click only for first time contributors to start CI :thinking:

Keats:
can you try closing that PR and opening another one?

Keats:
Can you rebase by any chance?

lpulley:
Ah, I see. Yeah, I'll do that.

Keats (approved):

Keats:
Thanks!

lpulley:
Happy to help! Thanks for getting this merged

PR #2886, opened 2025-05-10
Hash input file contents for imageproc output filenames

lpulley:
It doesn't seem quite right to me that `resize_image`'s output filename hash reads the *path* of the input file; shouldn't it read the *contents* instead? That way:

- if an input file is modified in place, Zola will not reuse the now-stale processed file
- if an input file is moved but retains the original filename, Zola will be able to reuse the existing processed file
  - (This one begs the question: should the original filename be included in the processed output filename at all? Should the processed output filename just be a hash digest and an extension? It would minimize the impact of renames and maybe even allow multiple identical input files to reuse a single processed output file.)

(Also, `GetImageMetadata::call` can just use the full path to the file as its cache key, so the return signature of `search_for_file` can be simplified.)

Since #2862/#2872 have changed the hash behavior on `next`, I figure this is maybe a good time to consider this, too, as it also affects the imageproc hash behavior.

### DIFF
--- a/components/imageproc/src/helpers.rs
+++ b/components/imageproc/src/helpers.rs
@@ -1,7 +1,10 @@
 use std::borrow::Cow;
 use std::collections::hash_map::DefaultHasher;
+use std::fs;
 use std::hash::{Hash, Hasher};
 use std::path::Path;
+
+use errors::{Context, Result};
 
 use crate::ResizeOperation;
 use crate::format::Format;
@@ -46,17 +49,16 @@ fn get_orientation(raw_metadata: Option<Vec<u8>>) -> Option<u32> {
     }
 }
 
-/// We only use the input_path to get the file stem.
-/// Hashing the resolved `input_path` would include the absolute path to the image
-/// with all filesystem components.
 pub fn get_processed_filename(
     input_path: &Path,
-    input_src: &str,
     op: &ResizeOperation,
     format: &Format,
-) -> String {
+) -> Result<String> {
     let mut hasher = DefaultHasher::new();
-    hasher.write(input_src.as_ref());
+    hasher.write(
+        &fs::read(input_path)
+            .with_context(|| format!("Failed to read file {}", input_path.display()))?,
+    );
     op.hash(&mut hasher);
     format.hash(&mut hasher);
     let hash = hasher.finish();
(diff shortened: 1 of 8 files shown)


lpulley:
@Keats do I have a good idea here? Bumping because of:

> Since #2862/#2872 have changed the hash behavior on `next`, I figure this is maybe a good time to consider this, too, as it also affects the imageproc hash behavior.

lpulley:
Hm... this is vulnerable to the input contents changing between enqueueing and image processing. I suppose the queue would have to hold the input contents instead of the input path?

Keats:
I think that makes sense

components/imageproc/src/helpers.rs
```diff
@@ -32,17 +35,16 @@ pub fn fix_orientation(img: &DynamicImage, path: &Path) -> Option<DynamicImage>
...
 pub fn get_processed_filename(
     input_path: &Path,
-    input_src: &str,
     op: &ResizeOperation,
     format: &Format,
-) -> String {
+) -> Result<String> {
     let mut hasher = DefaultHasher::new();
-    hasher.write(input_src.as_ref());
+    hasher.write(
```
Keats, line 58: does that change perf significantly? I can imagine reading a whole site of images is going to be much more computationally expensive than filenames

Keats: Also it doesn't make sense to read the file multiple times, once for the filename and once to actually operate on it

Keats (commented):
I like the idea but the performance changes seem to be a net negative even when you take into account regenerating the image if we change the filename

components/imageproc/src/helpers.rs
lpulley: Yeah, I've been thinking about how to approach this. We'd essentially want to read each input once (at most), and for each operation to hold a reference or key to the in-memory contents of the input for processing.

I'll push something for that if I figure out a good solution.

Keats:
This approach won't work perf-wise since we're loading the files twice. I'm not entirely sure how to fix that though.

lpulley:
Would you be interested in this as an enhancement if I can do it without loading the files twice (once to hash and once to process)? I suppose it would require holding them in memory, which might be infeasible. Or we'd have to decide the output path (and do the hashing) only after the image has been loaded for processing?

Keats:
> I suppose it would require holding them in memory, which might be infeasible

I don't think it's acceptable for sites with lots of images? I'm not sure. If there is a way to do it without holding everything then sure

Issue #2817, opened 2025-03-07
Implement GitHub flavor message boxes


rezamahdi:
In pulldown-cmark docs, it is noted that the crate is able to parse message boxes like warning, note and so on 

But it seems that Zola doesn't support them.

Is this feature implemented?

For example expect this:

> [!NOTE]
> This is a note

But is rendered as:

> [! NOTE]
> This is a note

Keats:
See https://github.com/getzola/zola/issues/2719

ctrlok:
@Keats, it is pretty different.

I'm experimenting with a custom build of zola with 

```rust
    opts.insert(Options::ENABLE_GFM);
    opts.insert(Options::ENABLE_SUBSCRIPT);
    opts.insert(Options::ENABLE_SUPERSCRIPT);
    opts.insert(Options::ENABLE_WIKILINKS);

```

And it is pretty fun. 

Enabling GFM allows you to write markdowns like this. 

```markdown
> [!CAUTION]
> Markdown's simplicity doesn't mean it lacks power. It's a perfect example of how elegant solutions often come from
> simple designs.
```

Which is transforms to 

```html
<blockquote class="markdown-alert-caution">
<p>Markdown's simplicity doesn't mean it lacks power. It's a perfect example of how elegant solutions often come from
simple designs.</p>
</blockquote>
```

As a theme builder, I can create styles for those blocks and allow potential users to use them. Here is an example of a theme I'm working on that supports all those blocks. 

![Image](https://github.com/user-attachments/assets/636422c2-bc54-40e3-aaf4-42f5d4139b11)

This GFM style blockquote is widely accepted by community: [hugo](https://gohugo.io/render-hooks/blockquotes/#basic-syntax) as an example, and it is supported by github 

> [!CAUTION]
> Markdown's simplicity doesn't mean it lacks power. It's a perfect example of how elegant solutions often come from
> simple designs.

I can implement it via shortcodes (and I did it in the past), but still as user, even I forget the names and syntax of those shortcodes, while this markdown syntax is standard and a lot of users know it already.  

Also, it didn't break almost any existing blockquote behavior and can be easily implemented by turning this on. 


wdyt?

Keats:
I mean, we can enable it but I am not entirely sure how the headers would work for i18n vs a shortcode

banditopazzo:
+1 for this feature. this would be very useful, I agree with @ctrlok that the github alerts are widely accepted and familiar to users.

[...]

PR #2893, opened 2025-05-16
Add github alerts

banditopazzo:
This PR wants to add github flavor alerts.

## Code changes
I followed your guide in the issue adding the feature with the proposed name `github_alerts` and default to `false`.
Added also the test, not sure if it's the proper way, this is my first contribution to the project.

Ref: #2817


Keats:
Thanks!

PR #2897, opened 2025-05-23
Update dependency onig

senekor:
This fixes a build issue on platforms using GCC 15.


Keats:
That should be picked up already automatically no?

senekor:
If zola is used as a library, yes. But when installing with `--locked`, then no.

picsel2:
> That should be picked up already automatically no?

Unfortunately not. For me it picked up the broken version when trying to simply build with `cargo build` on a fresh cloned repo using the `master` branch.

picsel2 (approved):
I get equivalent changes when running `cargo update onig`.

lgtm, it fixes the build

Issue #2889, opened 2025-05-15
[Search / fuse.js] Omit filtered fields from search index rather than adding as null


jlizen:
Currently my search config is like so:
```
[search]
index_format = "fuse_json"
include_path = true
include_title = true
include_content = false
include_description = false
include_date = false
```

IE, all I care to pass my index is title and path. My search handling only fuzzy matches on title, since that is what is relevant to my users for page discovery.

Meanwhile my search engine index is like so:
```
[{"url":"https://my-url", title":"my-title","description":null,"body":null,"path":"/my-path0/"}],
```

I have roughly 200k pages on my site, so all of these extra null fields add up. The index is currently around 30mb for context.

Clearly `fuse.js`'s javascript works fine with these fields set to null. Would it be ok to add a [skip_serializing_if](https://serde.rs/attr-skip-serializing.html) directive to the optional fields to avoid bloating the output? It would [go here](https://github.com/getzola/zola/blob/master/components/search/src/fuse.rs#L13).

I was going to just post-process my own index, but it'd be nicer to change the behavior for others too. Glad to cut a quick PR.

PS - The elasticlunr story is more complicated, since it is modeled by its own crate. Anyway it is a [heavierweight structure](https://github.com/mattico/elasticlunr-rs/blob/4db7fac70fa4d6281bf527d9fae07f5a2169f252/src/lib.rs#L197) in general so probably people looking for the super lightweight use case will use fuse.js like I did.

jlizen:
@Keats would you accept a PR for this? Ie, update the fuse.js struct to skip serializing optional fields and thus avoid outputing a bunch of null keys?

Keats:
yep that would be great

PR #2899, opened 2025-05-23
fix(search/fuse.js): skip serializing null keys when filtering them via zola config

jlizen:
Closes:
https://github.com/getzola/zola/issues/2889

Currently, even if you filter fields out of your `fuse.js` index:
```
[search]
index_format = "fuse_json"
include_path = true
include_title = true
include_content = false
include_description = false
include_date = false
```

They show up in the index as null fields, which bloats the index size for large sites:
```
[{"url":"https://my-url", title":"my-title","description":null,"body":null,"path":"/my-path0/"}]
```

This change adds a  `#[serde(skip_serializing_if = "Option::is_none")]` to the optional fields, so that they are instead just left out of the index.

## Testing

The new result for the equivalent example above is:
`[{"url":"https://my-url", title":"my-title","path":"/my-path0/"}]`

I have tested that `fuse.js`'s javascript bundle still handle this properly with the null fields missing. I did so by locally staging my site with `zola serve`, as well as manually inspect the index in console to make sure it was reflecting changes.

Everything worked fine. This cut 10mb off my index size.

Sanity check:

## Code changes

n/a

### DIFF
--- a/components/search/src/fuse.rs
+++ b/components/search/src/fuse.rs
@@ -10,9 +10,13 @@ pub fn build_index(lang: &str, library: &Library, config: &Search) -> Result<Str
     #[derive(serde::Serialize)]
     struct Item<'a> {
         url: &'a str,
+        #[serde(skip_serializing_if = "Option::is_none")]
         title: Option<&'a str>,
+        #[serde(skip_serializing_if = "Option::is_none")]
         description: Option<&'a str>,
+        #[serde(skip_serializing_if = "Option::is_none")]
         body: Option<String>, // AMMONIA.clean has to allocate anyway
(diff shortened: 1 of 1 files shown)


jlizen:
@Keats - this is my first time contributing to this repo - let me know if I missed anything. Thanks in advance!

I didn't bother documenting this behavior explicitly since it seems more like a bugfix than a meaningful API change for end users.

Keats:
This is potentially a breaking change, i'll need to add it to the Changelog but good otherwise

jlizen:
Thanks @Keats ! Any sense of when the next release is planned?

PR #2904, opened 2025-05-31
README: adds logo; DOCS: logo in header

indiscipline:
1. Adds the new logo to the README.md header.
2. Replaces documentation template header text with an image, adds an appropriate styling (not tested!).

It also would be nice to update the @getzola account logo with one of the square ones.


Keats:
Thanks! I'll update the GH logo

PR #2922, opened 2025-07-08
Add Lysator to examples

sermuns:
We are extremely happy with Zola and Tera as templating engine! Handcrafted templates for this website!


Keats:
Nice thanks!

PR #2936, opened 2025-07-26
Reword taxonomy page

ccjmne:
Improved literacy and formatting. I again took the liberty of formatting it to 80-characters wide, but that single commit can be dropped if necessary.


Keats:
I would prefer to drop the 80 chars limit commit yeah, I don't follow that myself so it introduces unncessary changes

ccjmne:
Alright, I'm happy either way! I'll drop that commit.

Keats:
Thanks!

PR #2938, opened 2025-07-26
Configurable file watcher debounce duration in live-reload mode

### LINKED ISSUE #2937: [Feature Request] Configurable debounce timer for file watcher
It’d be nice to add the ability for the user to configure a difference debounce value for the files watcher.

One full second seems quite excessive to me sometimes, when I could just edit my blog post and see it some handful of milliseconds afterwards on my second monitor.

Forum thread: https://zola.discourse.group/t/watch-debounce-configuration/2666

ccjmne:
- Introduce `-d` `--debounce <VALUE>` flag for `serve`
- Document `serve`'s `--debounce` flag in `cli-usage.md`

Let the file watcher use a configurable debounce value.

This commit preserves the (until now hard-coded) 1-second debounce value
as the default, but adds a new flag for the `serve` command, which lets
the user configure a debounce duration for consecutive rebuilds in watch
mode.

Closes #2937.

### DIFF
--- a/docs/content/documentation/getting-started/cli-usage.md
+++ b/docs/content/documentation/getting-started/cli-usage.md
@@ -102,8 +102,10 @@ fail to see your change or get an error, try restarting `zola serve`.
 
 By default, the live reload will be debounced by one full second so as to more
 gracefully handle multiple changes to your input files in rapid succession. You
-have control over that debouncing duration with the debounce flag. You may use
-`-d0` to disable it altogether.
+have control over that debouncing duration with the `--debounce <duration_ms>`
+flag.  
+You may use `-d1` to (virtually) disable it altogether: for technical reasons
+(and keeping things simple), a "debounce" of 0 is not supported.
 
 You can also point to a config file other than `config.toml` like so (note that the position of the `config` option is important):
 
--- a/docs/content/documentation/getting-started/cli-usage.md
+++ b/docs/content/documentation/getting-started/cli-usage.md
@@ -100,6 +100,11 @@ a hard refresh if possible. If you are using WSL2 on Windows, make sure to store
 Some changes cannot be handled automatically and thus live reload may not always work. If you
 fail to see your change or get an error, try restarting `zola serve`.
 
+By default, the live reload will be debounced by one full second so as to more
+gracefully handle multiple changes to your input files in rapid succession. You
+have control over that debouncing duration with the debounce flag. You may use
+`-d0` to disable it altogether.
+
 You can also point to a config file other than `config.toml` like so (note that the position of the `config` option is important):
 
 ```bash
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -102,8 +102,8 @@ pub enum Command {
         #[clap(long)]
         extra_watch_path: Vec<String>,
 
-        /// Debounce time in milliseconds for the file watcher
-        #[clap(short = 'd', long, default_value_t = 1000)]
+        /// Debounce time in milliseconds for the file watcher (at least 1ms)
+        #[clap(short = 'd', long, default_value_t = 1000, value_parser = clap::value_parser!(u64).range(1..))]
         debounce: u64,
     },
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -101,6 +101,10 @@ pub enum Command {
         /// Extra path to watch for changes, relative to the project root.
         #[clap(long)]
         extra_watch_path: Vec<String>,
+
+        /// Debounce time in milliseconds for the file watcher
+        #[clap(short = 'd', long, default_value_t = 1000)]
+        debounce: u64,
     },
 
     /// Try to build the project without rendering it. Checks links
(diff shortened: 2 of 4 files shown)


ccjmne:
Something's not quite right: you shoudn't use `-d0`, that'd get your CPU burning up. One millisecond is fine, though.
What do you reckon is best? Setting a minimum value for the `debounce` flag to `1`, or sneakily passing `1` along to the watcher-debouncer when the user chooses `0`?

nicolandu:
My advice would be to error, or at least warn, in the case of `-d0`. Silently changing configuration from what the user expects is probably not a good idea.

ccjmne:
I like that as well, I'll get right on it.

nicolandu:
Also, it would be good to document the flag (`-d`/`--debounce`) more thoroughly, such as by specifying its unit (milliseconds) in the docs.

ccjmne:
Done, and done.

I looked into the reason for the excessive CPU load with `--debounce 0`, and it seems to be a limitation of the debouncing library: it's not made to work efficiently with `0` there—which is reasonable.

The "ideal" resolution may have been to somehow bypass using the debouncer altogether if the user passes `--debounce 0`, but frankly, having another dependency (the debouncing library doesn't expose its underlying "watcher)" and the extra code to handle that special case seems like too much of a maintenance burden to pay for this little of a nitpick.

My puny mammalian brain can't really feel the difference between 1ms and 0ms anyways ;)

Keats:
The code looks fine, i'm just a bit hesitant about adding more flags to the CLI

Keats:
In practice though, is it actually worth it to have a debounce value that small? Zola is going to rebuild the thing every few (or even just every) keystrokes which seems a bit wasteful

ccjmne:
> In practice though, is it actually worth it to have a debounce value that small? Zola is going to rebuild the thing every few (or even just every) keystrokes which seems a bit wasteful.

Hang on a second: does Zola watch as you *type*, not as the files are changed on disk?

Surely that can't be, without some curious integration with your text editor. Or are you actually reflecting the changes back to your filesystem as you type?
On my end, using Vim, the live-rebuild watcher definitely only reacts when I commit my buffer to the disk, though I might be missing something huge and obvious: sorry if that's the case :)

TL;DR: What I'm observing is:
1. `zola serve`
2. type-type-type: nothing happens
3. `<Ctrl-S>` (or `:w`, you know what I mean)
4. one second goes by, *then*
5. zola rebuilds and reloads

With `zola serve -d1`, nothing changes but the removal of step 4; but I may be missing something.

Keats:
> Surely that can't be, without some curious integration with your text editor. Or are you actually reflecting the changes back to your filesystem as you type?

No but with auto-saving editors it might happen frequently enough

ccjmne:
> No but with auto-saving editors it might happen frequently enough

Oh alright. I would presume then that the debouncing happens at *that* level, where the editor wouldn't save your file to disk after each keystroke.

For myself, waiting one full second after I've virtually instructed Zola to rebuild my site is a bit of an irritation.
It's probably vastly exacerbated by the fact I'm still in the "build up the theme" phase, where I tweak styles and want to iterate rapidly: I can see how authoring *content* needs much less reactivity :)

nicolandu:
I just had a thought: what if we lower the debounce from 1000 ms to a lower value, say 50 or 100 ms, and ditch the complexity of an extra CLI flag?

Keats:
I've just tried in rustrover with a 50ms debounce and I do often get multiple builds in the same second (mostly when using backspace) so that would be a lot of wasted cpu for my own usecase :/

ccjmne:
@Keats, I think we'd all—but you primarily—benefit from understanding what is going on.

It's quite possible that you've configured something to commit changes, as you type, back to your filesystem (exceedinly?) often, apparently with some throttle or debounce: naively, I'd say that's an undesirable configuration.

Regardless, I truly don't think that your specific, undiagnosed set-up being quite curious should be the sole blocker that prevents a 3ms rebuild to take anything less than 1003ms.

Zola truly is a pretty neat tool, thank you for that!

Now if you still think that, in principle, waiting around is what is ultimately desirable for the project, that's acceptable. If you truly suspect that *any* (let alone a very large) debouncing is required, I'm not entirely convinced, and can't quite arrive to the same conclusion through my (admittedly modest) test.

ccjmne:
@Keats,

My previous message sounded harsh. If you really do want to figure out what I presume is writing to your disk surprisingly (unnecessarily?) often, I'd be genuinely curious and willing to help and find out.

If you want another parir of investigative eyes, I'd be happy to try and help!

yurivish:
I've also pined for instant "preview-on-save" and would appreciate any new development that makes this possible. 

It does sound like Keats's issue is potentially related to an interaction with a possible autosave functionality in their editor rather than being a direct consequence of the feature proposed in this PR.

In any case, thanks for Zola – I've been using it for my blog for over a year now and it's great!

Keats:
Sorry I forgot to reply. I am using stock RustRover, nothing custom. I think it's ok to have as an option

PR #2948, opened 2025-08-06
Fix `now()` to return same time on each call

### LINKED ISSUE #2947: `now()` returns a different time on each page
# Bug Report

## Environment

Zola version: 0.21.0

## Expected Behavior
The template function `now()` should return the same time for each page on a given site build.

## Current Behavior
The template function `now()` returns a different time on each invocation.

## Step to reproduce
Invoke `{{ now() }}` on multiple pages on the same site, then `zola serve` or `zola build`.

nicolandu:
This PR implements the same functionality as Tera, but with a single call to `chrono::Local::now()` at build time, instead of once per page. This change is non-breaking and 100% compatible with existing sites. No extra documentation should be needed if Zola's `now()` matches Tera's.
Fixes #2947.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2293,6 +2293,7 @@ dependencies = [
  "atty",
  "avif-parse",
  "base64",
+ "chrono",
  "csv",
  "elasticlunr-rs",
  "filetime",
--- a/components/libs/Cargo.toml
+++ b/components/libs/Cargo.toml
@@ -7,7 +7,9 @@ edition = "2021"
 ahash = "0.8"
 ammonia = "4"
 atty = "0.2.11"
+avif-parse = "1.3.2"
 base64 = "0.22"
+chrono = {version = "0.4.27", default-features = false, features = ["std", "clock"]}
 csv = "1"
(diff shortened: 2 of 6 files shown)


nicolandu:
@Keats, would it be simpler to dependency inject a clock provider into Tera instead than to patch `now()` in Zola?
If not, this should be ready to merge.

Keats:
> would it be simpler to dependency inject a clock provider into Tera instead than to patch now() in Zola?

How would that look?

nicolandu:
Bonjour @Keats!

The dependency injection approach would require rewriting Tera's `now` function to accept a custom `chrono::DateTime<Local>` provider (for example, accepting an `impl Fn() -> chrono::DateTime<Local>`), which could be `chrono::Local::now` by default. This could then be overridden in Zola to be a value obtained by a single call to `chrono::Local::now()` at the beginning of the website build process.

However, I think the easiest choice for now would be to go ahead with this present PR, since it is ready to merge. An issue could be opened in Tera in the near future for the above-mentioned changes, and I could potentially work on that in the future. If that lands, Zola could be easily modified to work with that later.

Merci!
Nicolas

nicolandu:
Fixed a silly `Cargo.toml` error, now it builds: an `optional=true` had slipped in when refactoring...

nicolandu:
Ready to merge!

Keats:
Thanks!

PR #2958, opened 2025-08-22
Fix link checking timeout

### LINKED ISSUE #2884: Link checking panics on timeout
# Bug Report

## Environment

Fedora Linux 42
Zola version: 0.20.0

## Expected Behavior

This is running against a site with 633 external links, the majority of which are broken.

## Current Behavior

Zola check (with RUST_BACKTRACE=1) results in the following output:

```
Checking site...
Checking 633 external link(s). Skipping 0 external link(s).
thread '<unnamed>' panicked at components/link_checker/src/lib.rs:62:44:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Decode, source: TimedOut }
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

It fails on this line.

https://github.com/getzola/zola/blob/2cc452cf56244fc602fe79452a54b3bf77a73dbc/components/link_checker/src/lib.rs#L62

When running with RUST_BACKTRACE=full, it takes a long time but does successfully check the site (and return the broken links).

## Step to reproduce

This check runs fine on Github Actions, but crashes when on my laptop on library WiFi. I'm certain it's a connection speed / timeout problem.

nicolandu:
Fixes #2884.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -44,6 +44,9 @@ jobs:
         with:
           toolchain: ${{ matrix.rustup_toolchain }}
 
+      - name: Cargo fmt
+        run: cargo fmt --check
+
       - name: Cargo build (Native TLS)
         run: |
           cargo build --all --no-default-features --features=native-tls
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

PR #2962, opened 2025-08-25
Correctly remove footnote-reference from summary

### LINKED ISSUE #2961: Footnotes are rendered in `page`s' `summary`-es
# Bug Report

Footnotes aren't removed in `summary` since the addition of back-references, when `bottom_footnotes` is `true` in the `config.toml`.

## Expected Behavior
A `page`'s `summary` should be rendered without its `<sup class="footnote-reference">` elements.

## Current Behavior
The `page`'s `summary`-es contain their `<sup class="footnote-reference">` elements.

## Diagnostic

It looks like this expected behaviour was introduced in:

https://github.com/getzola/zola/issues/1282

ccjmne:
This commit updates and fixes the regular expression to find and delete footnote references in the pages' summaries.

fecc3cf introduces the desired behaviour.

6a2b890 broke the suppression of footnote-reference in the summary by adding an id parameter to the <sup> elements for back-reference.

Fixes #2961.

Ref: https://zola.discourse.group/t/drop-footnote-references-from-summary-if-bottom-footnotes-true/2841/2

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -531,7 +531,7 @@ Hello world
 
     #[test]
     fn strips_footnotes_in_summary() {
-        let config = Config::default_for_test();
+        let mut config = Config::default_for_test();
         let content = r#"
 +++
 +++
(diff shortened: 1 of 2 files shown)


ccjmne:
Sorry for the many force-pushes.

I also modified the corresponding test so that it runs both with and without `bottom_footnotes` in the config—the latter which would have been failing until the fix provided by this PR.

ccjmne:
@Keats, I cannot post on https://zola.discourse.group anymore.
My posts were all flagged by "community" as `spam` some hour ago.

I only posted legitimate, constructive content and references.
Is there any recourse, please? :)

Keats:
I've approved them. Looks like the system didn't like external links for new accounts, I've added github as an exception I think it should be ok now?

ccjmne:
Excellent, it's all okay now again.
Thanks!

Keats:
Thanks!

components/content/src/page.rs
SomeoneSerge, line 1: I suppose this should be conditional on whether there's a `footnote-definition` above `<!-- more -->`?

PR #2969, opened 2025-09-02
Add option to attach `rel=external` to external links 

ccjmne:
Discussed in #2717.

It may be worth considering having it default to `true`... but I figured I would ask here, and let it be `false` at first, to avoid changing the current behaviour.

Though I would argue that this seems like a sensible default, and since we're gonna ship quite a few changes with the next release from `next`, it may be okay to introduce this "breaking" change. Just let me know!

Includes testing and documentation.

## Code changes

### DIFF
--- a/components/config/src/config/markup.rs
+++ b/components/config/src/config/markup.rs
@@ -240,7 +240,7 @@ impl Default for Markdown {
             external_links_target_blank: false,
             external_links_no_follow: false,
             external_links_no_referrer: false,
-            external_links_external: false,
+            external_links_external: true,
             smart_punctuation: false,
             definition_list: false,
             bottom_footnotes: false,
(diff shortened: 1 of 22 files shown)


components/config/src/config/markup.rs
```diff
@@ -234,6 +240,7 @@ impl Default for Markdown {
             external_links_target_blank: false,
             external_links_no_follow: false,
             external_links_no_referrer: false,
+            external_links_external: false,
```
Keats: I think it's fine to default to true. It's not going to break anything

Keats (changes requested):

components/config/src/config/markup.rs
ccjmne: Excellent!

ccjmne:
Oops, I forgot to update all the tests (the snapshots in particular).
I'll push one last commit that addresses this.

Keats (approved):

PR #2971, opened 2025-09-08
Update Vercel guide

ysfchn:

---

Updated Vercel guide to recommend using a custom command for obtaining musl binary of Zola, since Vercel's own Zola preset always gets glibc version even when it couldn't run on its own due to lacking glibc version in the system image.

```bash
echo "${ZOLA_VERSION:-"latest"}" | sed '/^latest$/!s/\(.*\)/tags\/v\1/' | \
    xargs -I% curl -fsSL "https://api.github.com/repos/getzola/zola/releases/%" | \
    grep -oP "\"browser_download_url\": ?\"\\K(.+linux-${ZOLA_LIBC:-"musl"}\\.tar\\.gz)" | \
    xargs curl -fsSL | tar -xz
```

Feels hacky than ever but I couldn't get to simplify than this.

Just for the reference, Vercel has [acknowledged the issue and said working on a fix](https://community.vercel.com/t/cant-build-site-with-zola-0-21-0/20782/15) to use musl binaries onwards but I'd prefer having the command on the Zola docs anyway.

### DIFF
--- a/docs/content/documentation/deployment/vercel.md
+++ b/docs/content/documentation/deployment/vercel.md
@@ -7,7 +7,7 @@ Vercel (previously Zeit) is similar to Netlify, making the deployment of your si
 The sites are hosted by Vercel and automatically deployed whenever we push a commit to our
 selected production branch (e.g, master).
 
-If you don't have an account with Vercel, you can sign up [here](https://vercel.com/signup).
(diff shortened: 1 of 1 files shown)


Keats:
Can you revert the formatting changes?

ysfchn:
Sure, should be done now :+1:

PR #2972, opened 2025-09-12
remove zola-theme-papermod (archived repo)

cydave:
I'm no longer maintaining the papermod theme port, as such I'd love to give other users the oppertunity to publish their version of a papermod port.

Not sure if removing the theme from your docs dir is sufficient or if you'd like to handle the archived theme differently, let me know.


Keats:
Done in https://github.com/getzola/themes/pull/189

PR #2994, opened 2025-10-06
add taxonomy_root configuration option

EvanCarroll:
Taxonomy functionality generates pages entirely for you. It makes sense to be able to home these in specific directory outside of the rest of the content. Here we do that with the new option `taxonomy_root`.

This is lead up to the functionality requested in here. https://zola.discourse.group/t/custom-path-for-taxonomy-pages/82

----

Sanity check:

* [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* [ x] Are you doing the PR on the `next` branch?

* [x ] Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -66,6 +66,9 @@ pub struct Config {
     /// If set, files from static/ will be hardlinked instead of copied to the output dir.
     pub hard_link_static: bool,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
+    /// Optional base path for all taxonomies. If set, all taxonomy paths will be relative to this path.
+    /// For example, if taxonomy_root is "blog" and taxonomy is "tags", the path will be /blog/tags/
+    pub taxonomy_root: Option<String>,
     /// The default author for pages.
     pub author: Option<String>,
 
@@ -410,6 +413,7 @@ impl Default for Config {
             feed_filenames: vec!["atom.xml".to_string()],
             hard_link_static: false,
             taxonomies: Vec::new(),
+            taxonomy_root: None,
             author: None,
             compile_sass: false,
             minify_html: false,
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -952,6 +952,10 @@ impl Site {
             components.push(taxonomy.lang.as_ref());
         }
 
+        if let Some(ref taxonomy_root) = self.config.taxonomy_root {
+            components.push(taxonomy_root.as_ref());
+        }
+
         components.push(taxonomy.slug.as_ref());
 
         let list_output =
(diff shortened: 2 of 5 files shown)


EvanCarroll:
If this is acceptable. I would like to take on the topic of adding taxonomy_path, which would be relative to taxonomy_root and allow you to set a custom path. This would be useful for example, if you're listing multiple different types of things like venues, events, etc and you want to rely on the taxonomy functionality to do that.

components/config/src/config/mod.rs
```diff
@@ -66,6 +66,9 @@ pub struct Config {
     /// If set, files from static/ will be hardlinked instead of copied to the output dir.
     pub hard_link_static: bool,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
+    /// Optional base path for all taxonomies. If set, all taxonomy paths will be relative to this path.
+    /// For example, if taxonomy_root is "blog" and taxonomy is "tags", the path will be /blog/tags/
+    pub taxonomy_root: Option<String>,
```
Keats, line 71: Do we want that to be a per-taxonomy config? I can maybe see people wanting that? Not sure though maybe we leave global for now and can add per-taxonomy override later.

components/site/src/lib.rs
```diff
@@ -987,9 +991,26 @@ impl Site {
...
+                            PathBuf::from(format!(
+                                "{}/{}/{}/{}",
+                                taxonomy.lang, taxonomy_root, taxonomy.slug, item.slug
+                            ))
+                        } else {
+                            PathBuf::from(format!(
+                                "{}/{}/{}",
+                                taxonomy.lang, taxonomy.slug, item.slug
+                            ))
+                        }
```
Keats, line 1013: some repetition but I'm not sure how to fix that

docs/content/documentation/getting-started/configuration.md
```diff
@@ -99,6 +99,10 @@ author =
 #
 taxonomies = []
 
+# Optional base path for all taxonomies. If set, all taxonomy paths will be relative to this path.
+# For example, if taxonomy_root is "blog" and taxonomy is "tags", the path will be /blog/tags/
+# taxonomy_root = "blog"
```
Keats, line 104: let's just add that it's not set as default so it will be /tags/

Keats (changes requested):

components/config/src/config/mod.rs
EvanCarroll: I'll do the work for per-taxonomy, path-specificity later. (Start on it immediately after this lands). But this alone is nice to just disambiguate this functionality from content stuff. If for no other reason then to tell all users,

1. Stuff in `/t/` is _taxonomy_ and automatically generated.
2. *ALL* other content has a markdown file in ./content.

docs/content/documentation/getting-started/configuration.md
EvanCarroll: Good idea!

Keats (approved):

Keats:
Thanks!

PR #3001, opened 2025-10-17
fix(markdown): prevent content duplication when using footnotes befor…

micahkepe:
## Sanity check:

## Code changes

- [x] Are you doing the PR on the `next` branch?

- Have you created/updated the relevant documentation page(s)? &rarr; _N/A_ (bug fix)

---

## Overview

This PR fixes a bug where content containing footnotes before the `<!-- more -->`
summary marker was being duplicated in the rendered HTML output.

## Problem

When a markdown file contains:

1. Content with footnote references before `<!-- more -->`
2. The `<!-- more -->` summary divider
3. Content and footnote definitions after the divider

The content before `<!-- more -->` was appearing twice in the rendered output - once as
intended before the continue-reading marker, and again duplicated after it.

**Example:**

```markdown
Hello world[^1].

<!-- more -->

Good bye.

[^1]: "World" is a placeholder.
```

This would render with "Hello world" appearing both in the summary section AND
duplicated in the body section.

## Root Cause

In `components/markdown/src/markdown.rs` at line 857, when rendering the body HTML, the
code was using ALL parsed markdown events:

```rust
cmark::html::push_html(&mut html, events.into_iter());
```

This meant that even though the summary had already been rendered from events before the
`continue_reading marker`, those same events were being rendered again for the body.

## Solution

Modified the body rendering logic (lines 856-862) to conditionally skip summary events
when `has_summary` is true:

```rust
// emit everything (or everything after summary if there is one)
if has_summary {
  // Only emit events after the summary cutoff to avoid duplication
  cmark::html::push_html(&mut html, events.into_iter().skip(continue_reading));
  } else {
    cmark::html::push_html(&mut html, events.into_iter());
}
```

Now when a summary exists, only the events AFTER the `continue_reading` marker are
rendered for the body.

## Testing

1. Updated existing test `test_summary_split` to expect the corrected behavior
2. Added new test `footnotes_in_summary_no_duplication` specifically for this bug scenario
3. All existing tests pass
4. Locally tested with debug build on actual blog content containing footnotes

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -567,19 +567,19 @@ And here's another. [^3]
             page.summary,
             Some("<p>This page use <sup>1.5</sup> and has footnotes, here\'s one. </p>\n<p>Here's another. </p>".to_string())
         );
+        // After fixing the duplication bug, page.content should start with continue-reading
+        // and NOT include the summary content again
         assert_eq!(
             page.content,
-            r##"<p>This page use <sup>1.5</sup> and has footnotes, here's one. <sup class="footnote-reference"><a href="#1">1</a></sup></p>
-<p>Here's another. <sup class="footnote-reference"><a href="#2">2</a></sup></p>
-<span id="continue-reading"></span>
-<p>And here's another. <sup class="footnote-reference"><a href="#3">3</a></sup></p>
-<div class="footnote-definition" id="1"><sup class="footnote-definition-label">1</sup>
+            r##"<span id="continue-reading"></span>
+<p>And here's another. <sup class="footnote-reference"><a href="#3">1</a></sup></p>
+<div class="footnote-definition" id="1"><sup class="footnote-definition-label">2</sup>
 <p>This is the first footnote.</p>
 </div>
-<div class="footnote-definition" id="2"><sup class="footnote-definition-label">2</sup>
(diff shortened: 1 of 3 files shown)


components/content/src/page.rs
```diff
@@ -567,19 +567,19 @@ And here's another. [^3]
             page.summary,
             Some("<p>This page use <sup>1.5</sup> and has footnotes, here\'s one. </p>\n<p>Here's another. </p>".to_string())
         );
+        // After fixing the duplication bug, page.content should start with continue-reading
+        // and NOT include the summary content again
```
Keats, line 571: That's not correct. `page.summary` has the beginning of the content to display on list pages but you want it part of the content because on the page itself you are not doing `{{page.summary}}{{page.content}}`, you will just do `{{page.content}}` and you expect the full text to be there

Keats (changes requested):
That would be a huge breaking change and is not something that we want to do.

micahkepe:
Thank you for your review. I agree we can close this PR for now. My reasoning is that the case of footnotes before the  summary marker is quite uncommon in real-world sites, so rather than making a breaking change to the core engine, it seems more appropriate to handle via theme logic if needed.

If more users bump into this scenario, I’m happy to revisit a more robust fix in the future.

Let me know if you’d prefer to keep the PR open a bit longer for discussion.

Issue #3006, opened 2025-10-27
Move minification into rendering and out of write_content (w/ patch)


EvanCarroll:
Background, currently

* `write_content` handles minification.
* zola follows the pretty url pattern of generating `foo/index.html` for `foo`. So a file generated for `/vcards/wikidata/Q136086504.vcf` ends up creating `/vcards/wikidata/Q136086504.vcf/index.html` via `write_content`

At that that point that `write_content` is called it only knows the destination it's generating. It doesn't know anything about the input. So in the above example, it sees it is generating an `index.html`. This means something like `.txt` and `ical` generation is all a hack based on the above..

Ideally, the rendering step would handle minifcation, because it knows what template _extension_ is being used. If the template is `.html` we know to minifiy.

Attached is a patch that proposes we change this, putting magnification into the rendering step. This isolates `write_content` the IO step.

I will supply a patch to handle this. This will allow you to generate anything with `minifiy_html` set to true, so long as the page's template is `.txt` or anything `*ml`. This may have even been the original design intent of the minification patch.

Issue #2993, opened 2025-10-04
Is there any way to create private post?


serephus:
This is **not** a feature request.

Anyway what I want is: a post protected with a question/password. (like descrypt after given the correct password in frontend)

mailq:
This is **not** an answer. "Private posts" is not the responsibility of a static website generator. This is a task for the webserver.

See e.g. <https://httpd.apache.org/docs/current/howto/auth.html> for Apache HTTP or <https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/> for NGINX. In case you use another webserver or service, have a lookout for "Basic Authentication" or "Access control".

Either organize your private pages into a sub directory with access control, or configure your server to protect individual paths with a common pattern.

Keats:
It's not possible with Zola itself since we just output HTML. @mailq provided a good answer for how to actually do that but it's a step after the zola build that depends on how/where you host it.

serephus:
sorry to open this again, essentially what I am looking for is something like this [hexo plugin](https://github.com/D0n9X1n/hexo-blog-encrypt), perhaps I should say encrypted posts. Anyway, is there anyway to achive this? I understand hexo is SSG too, so it should be possible, could a zola theme implement this?

mailq:
Sorry to close this again. Zola has no plugin mechanism where you can write a plugin to do what you want. The core is not responsible to add dynamic features for encryption and decryption. That is the job of your HTTP server. If you need this functionality anyway, then either do it the right way (HTTP Authorization), or use another tool (hexo, Wordpress, DIY)

EvanCarroll:
Creating a patch right now that will do this.

EvanCarroll:
@dearfl do you need this at the page level, or the the section level or site?

serephus:
> [@dearfl](https://github.com/dearfl) do you need this at the page level, or the the section level or site?

Thank you for doing this, personally I want this at page level. What I want to achive is that some posts can only be accessed by people that knows certain things about me, like my brithday or something.

EvanCarroll:
@dearfl If I built this, can I provide you a fork where this works and rely on you test it, because I think it's a *great* idea, but I don't expect it to be accepted into Zola until you prove it out. I built all the plumbing to do it though.

Do you want this to be..

* You give it a password and it fake-secures the page.
* It generates a cryptographically secure key which when provided decrypts the page.

serephus:
[...]

Issue #1684, opened 2021-12-07
Support gzip compression in `zola serve`


ajyoon:
While building sites I like to keep an eye on the transfer sizes of pages. This helps me keep sizes under control as I develop. However, because `zola serve` doesn't use any compression, these metrics in local development do not reflect production. This is especially true with text-based assets like HTML, JS, and CSS, which often compress very well.

If `zola serve` supported automatic gzip encryption, I would more easily be able to optimize for size where needed.

This could be done a few ways:
- Just enable it unconditionally in `zola serve` - I can't think of a use-case where this would be problematic, but I could be wrong
- Enable it by default, allowing an opt-out flag like `zola serve --no-compression`
- Implement it behind an opt-in flag like `zola serve --gzip`

Based on a glance through https://github.com/getzola/zola/blob/master/src/cmd/serve.rs any of therse approaches seem fairly straightforward to implement and I'd be happy to take a shot at it.

Keats:
How do you compare what zola serve would use and what you would use in the end? Do you control the compression level?

ajyoon:
For my use-case, I don't think an exact match of compression settings between local and production is necessary. I mostly just want a close-enough approximation of payload sizes given any production-realistic compression. For my purposes, actually, it wouldn't even matter if `zola serve` used brotli instead of gzip. (In that sense, I see this issue relates to #1520).

A low-config or no-config approach to this would also help prevent bloat, as I understand `zola serve` is meant to be fairly minimalist.

Keats:
I agree it would be nice to have compression done automatically on the output and on zola serve. As you say it would probably be a no config, just `compression = none | gzip | brotli`.

ajyoon:
I understand #1520 is concerned with built output as well as served results, but personally I think compressing build artifacts requires more nuance. For instance, would it transform filenames? Would it output both compressed and uncompressed versions? Would it apply to compression-friendly static assets like fonts? Would it use maximum compression ratio efficiency by default, as opposed to servers which often lean toward speed for on-the-fly compression? In my opinion, there are too many factors here to justify including it in a static site generator when it could be implemented by a fairly simple post-processing [`find` command](https://github.com/getzola/zola/issues/1520#issuecomment-868852530), especially considering that most static site hosting services already perform compression transparently.

I'd be happy to take a crack at implementing `zola serve` support for compression, but probably not file output as-is.

Keats:
[...]

Issue #2930, opened 2025-07-19
Really need to support HTML content files.


JetForMe:
I just ran into the limitation that content files must be markdown. This makes it impossible to put non-trivial HTML in the file. I had hoped Zola would be a good alternative to Jekyll, but this (and not using Tera in the content files) and the required workarounds (make a template for each content file you want like this) is pretty clunky.

Issue #2420 covers this, but was closed with the comment that this won't happen. Seems a bit opinionated, no?

BProg:
The way to do this right now is to create you html file under `templates/<yourfile>.html` and the in the `content/<yourpage>.md` use the `template = "<yourfile>.html"`

Inside your html

```
{% extends "base.html" %}

{% block content %}
<!-- YOUR HTML HERE -->
{% endblock content %}
```

JetForMe:
Yeah, that's kind of a hack, and what I did, but c'mon.

TheZoq2:
Also throwing in my support for this. I realized I can do some cool stuff by writing my content in typst. Since it outputs HTML it would be very nice to just be able to put its output in `content`

Keats:
If the content is full HTML, you can just create a .md file and put your HTML content preprended by `+++\n+++` and that should work? Assuming it's meant to go inside the page.html template

TheZoq2:
True, with the way html export in typst works I may need to do some postprocessing on it to make it work, but it is worth a shot!

lpcvoid:
Yeah, I'd also really like this. Currently using the workaround as well, it works, but it feels kind of hacky.

PR #3007, opened 2025-10-27
Implement a rendering pipeline with middleware (adds support for Brotli and Gzip, cleans up minification)

EvanCarroll:
I've updated this architecture to support a rendering pipeline which everything goes through.

* Minification is now a middleware.
* Livereload is now a middleware.

At the end of every pipeline we call `write()`. It's no longer called as IO call at the end of a function (which only know it was writing an `index.html`, and thus would minify `.txt`). This works and passes all tests,

* livereload knows if it's in a pipeline generated from an HTML
* minification knows if its in a pipeline generating HTML

With this infrastructure, this patch

* adds compression closing #1684
* and encryption #2993

AI summary of patch..

  Core Architecture

```
  ┌─────────────┐
  │  Renderable │ (trait)
  └──────┬──────┘
         │
         ├─ Page
         ├─ Section
         ├─ Taxonomy
         ├─ Pager
         ├─ Feed
         ├─ Sitemap
         └─ ... (11 types)
         │
         ▼
  ┌──────────────────┐
  │    Pipeline      │
  ├──────────────────┤
  │ 1. Render        │◄── Tera templates
  │ 2. Metadata      │
  │ 3. Middleware    │◄── Pluggable chain
  │ 4. Process       │
  │ 5. Write         │
  └──────────────────┘
         │
         ├─ LiveReloadMiddleware
         ├─ MinifyMiddleware
         └─ [future: compression, encryption]
         │
         ▼
  ┌──────────────────┐
  │  ProcessedContent│
  └──────────────────┘
         │
         ▼
  ┌──────────────────┐
  │  ContentWriter   │
  ├──────────────────┤
  │ BuildMode::Disk  │─── File system
  │ BuildMode::Memory│─── In-memory cache
  │ BuildMode::Both  │─── Both
  └──────────────────┘
```

  Key Components

  1. Renderable Trait (renderable.rs)

  - Purpose: Unified interface for all content types
  - Methods:
    - render() - produces HTML/XML/JSON from templates
    - metadata() - provides routing & content type info
  - Implementations: 11 types (Page, Section, Taxonomy, Feed, Sitemap, robots.txt, 404, Pager, Redirect, etc.)

  2. Pipeline (pipeline.rs)

  - Purpose: Orchestrates the 5-stage rendering process
  - Stages:
    a. Render: Template → String (via Tera)
    b. Metadata: Extract path, content type, permalink
    c. Context: Wrap in MiddlewareContext
    d. Middleware Chain: Sequential transformations
    e. Return: ProcessedContent ready to write

  3. Middleware Trait (middleware.rs)

  - Purpose: Extensible content transformation
  - Interface: process(&self, ctx: &mut MiddlewareContext)
  - Context includes:
    - content: String - the HTML/XML/JSON
    - metadata: ContentMetadata - path, template, content type
    - config: Arc<Config> - site configuration

  4. Concrete Middleware

  - LiveReloadMiddleware (middleware/livereload.rs):
    - Injects <script src="/livereload.js"> before </body>
    - Only processes ContentType::Html
  - MinifyMiddleware (middleware/minify.rs):
    - Removes whitespace from HTML
    - Conditional on config.minify_html
    - Only processes ContentType::Html

  5. ContentWriter (writer.rs)

  - Purpose: Handles output destination
  - BuildMode:
    - Disk: Write to filesystem (public/ directory)
    - Memory: Store in SITE_CONTENT cache (for live reload)
    - Both: Do both operations

  Design Benefits

  1. Separation of Concerns: Rendering, processing, and writing are isolated
  2. Type Safety: ContentType enum ensures correct middleware application
  3. Extensibility: Add new middleware without touching core logic
  4. Testability: Each component has unit tests
  5. Future-Ready: Designed for compression & encryption (per commit message)
  6. Unified Flow: All 11 content types go through the same pipeline

  Before vs After

  Before: Rendering logic scattered, minification/livereload handled per-type

  After: Single pipeline with pluggable middleware - every rendered item flows through the same chain

  This architecture mirrors common web frameworks (Express.js middleware, ASP.NET pipeline) and provides a clean foundation for adding post-processing features like
  compression or encryption.
  
 

----

* `write_content` isn't smart enough to do what we need.
* this is a cleaner architecture.
* Now pages with txt templates are not minified.
* Add test for text template whitespace preservation with minification
  * Test verifies that non-HTML templates (e.g., .txt) preserve whitespace
  * Created text_content.txt template with multiple spaces, tabs, and newlines
  * Created test page using the text template
  * Test passes with the minification fix (ed36c1e6) and fails without it
  * Updated page count assertions due to new test page

Addresses #3006


components/site/src/lib.rs
```diff
@@ -889,7 +897,11 @@ impl Site {
             None => "index.html",
         };
         let content = render_redirect_template(permalink, &self.tera)?;
-        self.write_content(&split, page_name, content)?;
+
+        // Aliases are always HTML redirects - minify if enabled
+        let final_content = self.maybe_minify(content)?;
```
Keats: Hm that's not great, we're just having to remember to call that everywhere we're writing HTML instead of once. We should instead be able to see whether the output is HTML or not and not generate an index.html for those like we do for sitemaps.

Keats (changes requested):

components/site/src/lib.rs
EvanCarroll: That specific part of the file you've highlighted is how the code currently works. `maybe_minify` does 

```
    fn maybe_minify(&self, content: String) -> Result<String> {
        if self.config.minify_html { minify::html(content) } else { Ok(content) }
    }
```

That is to say, it's just a conditional to check whether or not the config option is present at all. We can roll that into the routine itself if you find it confusing.

----

> Hm that's not great, we're just having to remember to call that everywhere we're writing HTML instead of once. We should instead be able to see whether the output is HTML or not and not generate an index.html for those like we do for sitemaps.

Fixing that would require re-architecting this. We currently have,

* The template,
* The filename you're writing.

And

* `write_content`, is the one that that writes files. it has no idea of the template.
* A lot of `render_` where each one call `write_content` with the content it's writing.

----

Problem, 

* If anything can `write_content`
* And, `write_content` isn't template aware and only knows of content and a destination file location,
* How can I tell whether or not minification is needed?

Currently you have to do that at rendering step.

----

So ultimately to do this without having to have the logic in everything that can produce html, we would need to create something between the renderer and the IO layer.

If you want a more drastic rewrite, I'll create you a middleware rendering pipeline that can solve this problem. Where anywhere middleware can answer the call at the end. But this requires blocking access to `write_content` and having the render's return the mime/type that that they're rendering to the next step in the pipeline.

This could also be used to solve, #1684

EvanCarroll: I'm going to close this because I rewrote all this and a lot more.

EvanCarroll:
@Keats if you determine this architecture isn't something you want to move towards, please tell me so I can make adjustments or quit closing out issues.

ccjmne:
Seems really cool, but the current MR is **a lot** for a human to digest.
Do you think you could keep the encryption middleware in another PR for the time being? Though maybe @Keats can review it like that.

EvanCarroll:
> Seems really cool, but the current MR is **a lot** for a human to digest. Do you think you could keep the encryption middleware in another PR for the time being? Though maybe @Keats can review it like that.

Fair and certainly doable. Because the architecture isn't cleanly in one patch it makes this patch even more complex. I tweak it a few times. If this is needed. what I can do is rewrite the the patches entirely to sink the commits separately. It's a waste of time though if that's not needed.

Actually if the middlewares themselves had their own config parser, they would be entirely self-contained and this would be trivial..

The next patch that I want to do pulls out markdown rendering putting that into the middleware too. And closes out #2930, and #2420 and #2790. This would make adding first-class Tyrst the same way trivial.

Keats:
It is indeed way too big to review. A few other things:

- The issue for secret content is closed so for now it shouldn't be implemented
- Same-ish for pre-compression, I don't think it's super important
- I think it's too complex overall? All we need to have is a way to differentiate HTML and other content and do different operations when wrtiting the content.

EvanCarroll:
I'll get back this and rebase the patch to make it easier to process. I got one more project I have to get off my plate. The architectural changes, should make all the middle wares one file (or relatively small) so you can pick whatever middleware you want.

Keats:
The `next` branch now has some kind of queue processing.

PR #3017, opened 2025-11-09
Introduce `before` and `after` filters

### LINKED ISSUE #2619: Add a `before` and `after` filter
Add a filter that removes all items with a `.date` field less than the provided date and a filter that does the opposite.

e.g.
```
sections.pages | before("2020-03-10")
```

```
sections.pages | after("2020-03-10")
```

## Use Case

I want to render talks I gave before a certain date in a different list than ones after.
I'm currently doing this by adding an extra frontmatter key into my pages but I'd rather do this by date and this would make that much easier.

## Naming

I think `before` and `after` would be good names for this. 

However, if these collide with existing names for anything, some prefix could be added to make them more unique like `dates_before` and `dates_after`.

ccjmne:

Fixes #2619.

- The "before" filter is exclusive, and the "after" one is inclusive; it
  may be better thought of as a "since" than a strict "after".
- Pages with no dates are never included.
- Dates can be 8601 default,       like: 2023-12-01T12:00:00Z
  or only have the date component, like: 2023-12-01

### DIFF
--- a/components/templates/src/filters.rs
+++ b/components/templates/src/filters.rs
@@ -5,13 +5,13 @@ use std::sync::{Arc, Mutex};
 
 use config::Config;
 
-use libs::base64::engine::{general_purpose::STANDARD as standard_b64, Engine};
+use libs::base64::engine::{Engine, general_purpose::STANDARD as standard_b64};
 use libs::regex::Regex;
 use libs::tera::{
-    to_value, try_get_value, Error as TeraError, Filter as TeraFilter, Result as TeraResult, Tera,
-    Value,
+    Error as TeraError, Filter as TeraFilter, Result as TeraResult, Tera, Value, to_value,
+    try_get_value,
 };
-use markdown::{render_content, RenderContext};
+use markdown::{RenderContext, render_content};
 
 #[derive(Debug)]
 pub struct MarkdownFilter {
@@ -92,6 +92,12 @@ impl RegexReplaceFilter {
     }
 }
 
+impl Default for RegexReplaceFilter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl TeraFilter for RegexReplaceFilter {
     fn filter(&self, value: &Value, args: &HashMap<String, Value>) -> TeraResult<Value> {
         let text = try_get_value!("regex_replace", "value", String, value);
@@ -100,13 +106,13 @@ impl TeraFilter for RegexReplaceFilter {
             None => {
                 return Err(TeraError::msg(
                     "Filter `regex_replace` expected an arg called `pattern`",
-                ))
+                ));
             }
         };
         let rep = match args.get("rep") {
             Some(val) => try_get_value!("regex_replace", "rep", String, val),
             None => {
-                return Err(TeraError::msg("Filter `regex_replace` expected an arg called `rep`"))
+                return Err(TeraError::msg("Filter `regex_replace` expected an arg called `rep`"));
             }
         };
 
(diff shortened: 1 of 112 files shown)


ccjmne:
With `before` being exclusive and `after` being inclusive, we have effectively access to the `>=` and `<` comparisons.

I figured it would be somewhat useful when honing in on a particular month or particular year:

- this month's articles: `section.pages | after(date="2025-11-01")`
- last year's articles: `section.pages | after(date="2024-01-01") | before(date="2025-01-01")`

Keats:
How about adding `inclusive` as param to both, defaulting to false so people can choose?

docs/content/documentation/templates/overview.md
```diff
@@ -111,6 +111,24 @@ To format a number for a specific locale, you can use the `locale` argument and
...
+{% for page in section.pages | before(date="2020-03-10") %}
+  <h2>{{ page.title }}</h2>
+{% endfor %}
+```
+
+### after
+Filters an array of pages/sections to keep only items with a `date` field on or after the specified date.  Items without a `date` field are excluded.
+
+```jinja2
+{% for page in section.pages | after(date="2020-03-10") %}
```
Keats: looking at the usage, I'm wondering if it would be better to have the before/after work on a date rather than Page[]. It would be usable in more places

ccjmne:
> How about adding `inclusive` as param to both, defaulting to false so people can choose?

Ha! Yeah, seems obvious in retrospect. I'll do just that (tonight).

docs/content/documentation/templates/overview.md
ccjmne: Yeah, it's a bit peculiar to have that filter work on an array of items that have a `date` property, that's pretty arbitrary.

To clarify you suggest we have some `after` and `before` filters that yield either `true` or `false` when they're used on a date, that we'd use like:

```jinja2
{% for page in section.pages %}
    {% if page.date | before(date="2025-01-01") %} {% continue %} {% endif %}
    {{ page.content | markdown | safe }}
{% endfor %}
```

Correct?
If so, that makes sense as well, I can do that :)

ccjmne: I don't suppose it'd make sense to have it run on BOTH OF `Page[]`/`Section[]` and `date`, right? It'd be somewhat neat to have a filter that directly eliminates stuff from an array, I quite like the `section.pages | after(xxx)` construct.

But supporting both `if page.date | before(xxx)` and `for page in section.page | before(xxx)` is semantically overloaded, I believe.
It's very, *very* easy to implement and maintain, however ;)

I'll go ahead and assume you'd rather want only the:
```jinjia2
{% if page.date | before(date="2025-01-01") %}
```

We can always revert that commit later anyway.

Keats: I think it's better to work on a date rather than on a page/section. I think for that it will be better to design an improved `filter` filter for tera2 so you can apply a filter on every elements of a collection

ccjmne: Oh hell yeah, a `filter` filter would be awesome :)

In the meantime, I went ahead and made the adjustments you requested. This approach will definitely be perfect once the `filter` filter is available, and it's already still quite useable.

Let me know how you want to proceed from here!

ccjmne:
Hey @Keats, I addressed your comments, do you think it makes sense to merge this as it now is?

components/templates/src/filters.rs
```diff
@@ -164,6 +164,57 @@ impl TeraFilter for NumFormatFilter {
     }
 }
 
+pub fn before<S: BuildHasher>(
```
Keats: why do you specify the build hasher?

ccjmne: These were the awkward product of some careless copy-pasting of the `base64_encode` or `base64_decode` filters in that same file... Oops.
Thanks for being vigilant.

Keats:
I've just realised, these shouldn't be filters but tests

ccjmne:
Ah, like `xxx is defined`? Is `defined` one such "test"?
I could look into having them be that, that does indeed make more sense :)

Keats:
Yes exactly, it's basically a shorthand for a function returning a bool

ccjmne:
@Keats, can such `test`s be created in this repository, or do they have to belong upstream to Tera/Tera2?

I can't quite figure out where to add them here otherwise :/

Keats:
It's possible but might as well wait for https://github.com/Keats/tera2 integration. I'll try to do that in January and that will change things so no point wasting time now.

Keats:
See https://github.com/Keats/tera2/pull/95 that should work once tera2 integration is done.

PR #3024, opened 2025-11-17
Provide original markdown title in TOC

### LINKED ISSUE #3023: Provide access to original markdown content in table of contents
I publish a technical blog, and every now and then, I really would want to have my headers like... well, all of these:

```
 ## The `DOM` markup logic
 ### The `<summary>` in the `<details>`
 ### The `[open]` attribute and the `[name]` linking
 ### File descriptors in `/proc`
 ## A fourth parent with `2.51`
 ## Miscellaneous `/proc` resources
 ## Interesting `/proc/<pid>` information
 ## Files descriptors `/proc/<pid>/fd`
 ## Using `column`
 ## Using `paste`
 ## Under the hood: an `iterable iterator` {#under-the-hood}
 ### The `iterator` protocol
 ### The `iterable` interface
 ## Not just `for..of`
 ## More friendliness with `moreutils`
 ## The anatomy of a `sed` command
 ### Some `GNU` `sed`-specific pseudo-addresses
 ## The legitimate use of `cat`
 ## `prefers-color-scheme`
 ## `prefers-reduced-motion`
 ## `prefers-contrast`
 ## `prefers-reduced-transparency`
 ## `prefers-reduced-data`
 ## The `media` HTML attribute
```

Yeah, that's a lot of `<code>` blocks... but I assure you, it makes sense ;)

In any case, these are a lot more easily visually "grepped" when they use the `<code>` tag that I intend them to leverage.
I'm thinking of having another property on the table of content's entries, alongside `title` (which strips everything that's not text), that we could choose to further process with `| markdown`.

If there's no objection, I'm planning on taking a look at the code and providing a PR tonight or tomorrow :)

Cheers!

ccjmne:
Provide original markdown title in TOC.

This commit introduces the "title_raw" property to the Heading struct,
so that the user may get adequately formatted entries in their table of
contents when the original title contains, for example, code highlight
content.

For example:

    ### The `<summary>` in the `<details>`

Would offer, in heading.title:

    The <summary> in the <details>

And in the newly introduced heading.title_raw:

    The <code>&lt;summary&gt;</code> in the <code>&lt;details&gt;</code>

Resolves #3023.

---

Sanity check:

* [x] Not already addressed by another ongoing PR
* [x] PR made against `next`?
* [x] Documentation updated?

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -44,6 +44,9 @@ jobs:
         with:
           toolchain: ${{ matrix.rustup_toolchain }}
 
+      - name: Cargo fmt
+        run: cargo fmt --check
+
       - name: Cargo build (Native TLS)
         run: |
           cargo build --all --no-default-features --features=native-tls
@@ -54,6 +57,3 @@ jobs:
 
       - name: Cargo test
         run: cargo test --all
-
-      - name: Cargo fmt
-        run: cargo fmt --check
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Install Zig and cargo-zigbuild
         run: |
           curl -sSfL https://ziglang.org/download/0.14.0/zig-linux-x86_64-0.14.0.tar.xz | tar -xJf -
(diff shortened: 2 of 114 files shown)


Keats:
Going to close that for now and see if more people want this based on the issue.

Issue #2919, opened 2025-07-01
Config file name conflict as `config.toml` is too generic


davlgd:
# Bug Report

## Environment

Zola version: 0.20

## Expected Behavior
As Zola uses `config.toml` as default file name, there is conflicts with other Rust projects doing the same, such as [Static Web Server](https://github.com/static-web-server). Both try to load the same file as default. 

A good change could be to use another default file name, such as `zola_config.toml`. It would also help to detect Zola as the static website generator to use for tools supporting different SSG. 

## Current Behavior
If I configure a project also using `config.toml` as default, it will try to load Zola's. I can point to another file, but it could be great to not have to do so. 

## Step to reproduce
Get a Zola project, host it with static-website-server:
```
cargo install static-web-server
static-web-server --root /path/to/the_folder_to_serve
```

As SWS will try to load Zola config file as its own, you'll see logs such as :

```
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "base_url" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "build_search_index" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "compile_sass" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "extra" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "ignored_content" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "markdown" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "slugify" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "taxonomies" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "theme" or unsupported
2025-07-01T11:55:22.319Z: Warning: unused configuration manifest key "title" or unsupported
```

I've recommended SWS to make the same change, but I think both projects needs to avoid such generic config file name as default.

mrec:
I would have expected that you'd point SWS at `/public` as the folder to serve, rather than the root of your Zola project. Everything produced by `zola build` goes in there. If you put your SWS `config.toml` in `/static` it should get copied over to `/public` where SWS can see it. Do you still get problems with that  setup?

davlgd:
[...]

PR #3028, opened 2025-11-23
feat: change default config file name to zola.toml

cscnk52:

Sanity check:

## Code changes

<hr/>

Related Zola forum discussion:

- https://zola.discourse.group/t/adding-schema-file-for-zola-config-file/2875
- https://zola.discourse.group/t/rename-the-default-configuration-file/2880

Now this PR only contain change default file name to `zola.toml`, also fallback to `config.toml` when default file not file. This is import for language server auto find and using schema files, like [Hugo change default config name from config.toml to hugo.toml](https://github.com/gohugoio/hugo/issues/8979).

And now the schema file have been deployed on [my GitHub pages](https://cscnk52.github.io/json-schemas/), next step I will try to add it to [www.schemastore.org](https://www.schemastore.org/), So that language server can auto recognize and use schema file.

Also see:

- https://github.com/SchemaStore/schemastore/blob/c3565e40e454c18619aa742d8c061b49febe56a6/CONTRIBUTING.md#avoid-generic-filematch-patterns

Related to

- https://github.com/SchemaStore/schemastore/pull/5170
- https://github.com/getzola/zola/issues/2919

### DIFF
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -66,6 +66,9 @@ pub struct Config {
     /// If set, files from static/ will be hardlinked instead of copied to the output dir.
     pub hard_link_static: bool,
     pub taxonomies: Vec<taxonomies::TaxonomyConfig>,
+    /// Optional base path for all taxonomies. If set, all taxonomy paths will be relative to this path.
+    /// For example, if taxonomy_root is "blog" and taxonomy is "tags", the path will be /blog/tags/
+    pub taxonomy_root: Option<String>,
     /// The default author for pages.
     pub author: Option<String>,
 
@@ -410,6 +413,7 @@ impl Default for Config {
             feed_filenames: vec!["atom.xml".to_string()],
             hard_link_static: false,
             taxonomies: Vec::new(),
+            taxonomy_root: None,
             author: None,
             compile_sass: false,
             minify_html: false,
--- a/components/config/src/config/mod.rs
+++ b/components/config/src/config/mod.rs
@@ -13,7 +13,7 @@ use libs::toml::Value as Toml;
 use serde::{Deserialize, Serialize};
 
 use crate::theme::Theme;
-use errors::{anyhow, bail, Result};
+use errors::{Result, anyhow, bail};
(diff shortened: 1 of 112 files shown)


Keats:
I don't really want to add that dependency or a command just for that. I think it's something that should stay in a fork of the person maintaining the schema

src/cli.rs
```diff
@@ -12,7 +12,7 @@ pub struct Cli {
     pub root: PathBuf,
 
     /// Path to a config file other than config.toml in the root of project
-    #[clap(short = 'c', long, default_value = "config.toml")]
+    #[clap(short = 'c', long, default_value = "zola.toml")]
```
Keats, line 15: Now that we have multiple possible values there we should make that optional so get_config_file_path can search for both config.toml/zola.toml when the user hasn't provided a path.

Also to become the default it needs to replace config.toml everywhere in the init command and possibly a few other places?

Keats (changes requested):

components/config/src/config/mod.rs
```diff
@@ -149,7 +149,7 @@ impl Config {
         };
 
         if config.base_url.is_empty() || config.base_url == DEFAULT_BASE_URL {
-            bail!("A base URL is required in config.toml with key `base_url`");
+            bail!("A base URL is required in zola.toml with key `base_url`");
```
Keats, line 152: let's say `config file` in all messages so it covers both config.toml and zola.toml

src/cmd/init.rs
```diff
@@ -121,7 +121,7 @@ fn populate(path: &Path, compile_sass: bool, config: &str) -> Result<()> {
     if !path.exists() {
         create_dir(path)?;
     }
-    create_file(&path.join("config.toml"), config)?;
+    create_file(&path.join("zola.toml"), config)?;
```
Keats, line 124: I can do it later but if you have time, can you extract it in a static var to avoid hardcoding it everywhere?

Keats:
I've pushed a patch the to the `next` branch.

PR #3033, opened 2025-11-28
build: Generate man pages at build time

### LINKED ISSUE #3032: Feature request: Add man pages for the CLI
Many commands that run on Unix-like operating systems come with man pages. Hugo can generate man pages for the Hugo CLI by running `hugo gen man`.[^1]

I think it would be useful to have man pages for the Zola CLI.

We can also write man pages directly using roff, but this requires updating both `src/cli.rs` and man pages, which can be a bit tedious. Using the [`clap_mangen`](https://crates.io/crates/clap_mangen) crate, we can generate man pages from the CLI definitions in `src/cli.rs`. In this case, updating `src/cli.rs` will also update man pages.

If this feature request is approved, I'll add `zola generate-man output_dir` (where `output_dir` is the directory to write man pages) to the CLI.

[^1]: <https://gohugo.io/commands/hugo_gen_man/>

sorairolake:

Sanity check:

## Code changes

---

Generates man pages for the Zola CLI at the package compile time. These are generated when the target is `unix` and the profile is the `release`. man page is documentation for Unix-like operating systems, so the man pages are not generated when the target is `windows`.

Closes #3032

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,8 @@ jobs:
         run: |
           mkdir -p artifacts
           cp -av target/${{ matrix.target }}/release/zola .
-          tar -czf ${{ github.event.repository.name }}-${{ github.ref_name }}-${{ matrix.target }}.tar.gz zola
+          cp -av "$(find target -path '*/zola-*/out' -type d)"/* artifacts
+          tar -czf ${{ github.event.repository.name }}-${{ github.ref_name }}-${{ matrix.target }}.tar.gz zola artifacts
         if: ${{ ! startsWith(matrix.os, 'windows') }}
 
       - name: Archive (Windows)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,8 @@ jobs:
         run: |
           mkdir -p artifacts
           cp -av target/${{ matrix.target }}/release/zola .
-          cp -av "$(find target -path '*/zola-*/out' -type d)"/* artifacts
+          OUTPUT_DIR="$(find target -path '*/zola-*/out' -type d)"
+          cp -av "$OUTPUT_DIR"/* artifacts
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -13,13 +13,24 @@ fn generate_pe_header() {
...
 }
 
+include!("src/cli.rs");
+
+fn generate_man_pages() {
+    use clap::CommandFactory;
+
+    let out_dir = std::env::var("OUT_DIR").unwrap();
+    let cmd = Cli::command();
+    clap_mangen::generate_to(cmd, out_dir).unwrap();
```
Keats, line 23: Will the file generated be picked up by the release process?

sorairolake: When you run `cargo build --release` on Unix-like operating systems, the following files are generated in `target/release/build/zola-*/out` directory:

- `zola-build.1`
- `zola-check.1`
- `zola-completion.1`
- `zola-init.1`
- `zola-serve.1`
- `zola.1`

Keats: We'd need to modify https://github.com/getzola/zola/blob/master/.github/workflows/release.yml#L72-L84 in linux to include those for the releases

.github/workflows/release.yml
```diff
@@ -73,7 +73,8 @@ jobs:
         run: |
           mkdir -p artifacts
           cp -av target/${{ matrix.target }}/release/zola .
-          tar -czf ${{ github.event.repository.name }}-${{ github.ref_name }}-${{ matrix.target }}.tar.gz zola
+          cp -av "$(find target -path '*/zola-*/out' -type d)"/* artifacts
```
Keats: there's going to be only one hash there for the zola-* right?

Keats: can we just run the find command above before as well so we have some visibility?

sorairolake: > there's going to be only one hash there for the zola-* right?

There is only one directory named `zola-*` that has a subdirectory named `out`.

> can we just run the find command above before as well so we have some visibility?

Yes, I think it's possible.

PR #3038, opened 2025-12-07
[fix] Try to preserve ICC color profiles when processing images

kleinesfilmroellchen:
As a photographer who frequently publishes their photos on a site generated with Zola, I have been annoyed since forever over Zola’s disregard of color profiles. I take most pictures in high-gamut AdobeRGB, and many people will also be using Rec2020 nowadays. Therefore, it is important that the processed images appear exactly as the originals do. Discarding color space information is *incorrect* and *not debatable*, it simply reinterprets color data in a different, incorrect color space (usually Rec.709 or sRGB), leading to wrong results.

Therefore, with this PR the image processing tries as much as possible to preserve ICC (color space) data, and prints a warning if it cannot do that. Therefore, this change is non-breaking in the sense that no existing page will fail to build, but processed images may appear different (*better*) than before.

Before:

grafik

After (note the green and red colors):

grafik

Caveats:
- Some formats *may* store color profiles in a different way. ICC is the industry standard, supported in all of our output formats, and as far as I can tell it’s even used for all the common color profiles like sRGB and AdobeRGB
- AVIF writer doesn’t support ICC. We print a warning in this case.
- Only the `image`-builtin WebP writer supports ICC, but it can only compress losslessly. Therefore, we only embed ICC if the compression is lossless (using a different writer from what we normally do). We print a warning otherwise.

Also adds tests with a few common color profiles and some of the existing images (which all don’t have ICC data).

---

Sanity check:

## Code changes


components/imageproc/src/processor.rs
```diff
@@ -68,33 +71,51 @@ impl ImageOp {
...
-                    img.as_bytes(),
-                    img.dimensions().0,
-                    img.dimensions().1,
-                    img.color().into(),
-                )?;
-                tmp_output_writer.write_all(avif.as_bytes())?;
+                let mut encoder =
+                    AvifEncoder::new_with_speed_quality(&mut tmp_output_writer, speed, quality);
+                add_color_profile(&mut encoder);
+                img.write_with_encoder(encoder)?;
```
kleinesfilmroellchen, line 120: This change does two more things beyond adding the profile:
- it also bypasses the encode to memory
- it removes the img.as_bytes() call, which makes a copy of the whole image

Those two should make the encode faster by 1x output data memcpy + 1x image data memcpy. Since it matches all the other standard encoders, the tests pass and my site still works, I’m confident this is correct.

components/imageproc/tests/test_imgs/adobergb.avif
kleinesfilmroellchen, line 1: I added this file to test AVIF ICC decoding support (see tests), but since AVIF decoding support itself hasn’t been added yet, it’s unused for now and the tests commented out. I decided to leave the file for now but can remove it if desired (it’s quite small anyways as you can see, and almost half its size is the AdobeRGB ICC profile).

docs/content/documentation/content/image-processing/index.md
```diff
@@ -70,6 +70,23 @@ orig_width: u32,
...
+- Only ICC-based color data is preserved.
+- For WebP, only lossless encoding supports ICC profiles.
+- AVIF encoding does not currently support ICC profiles.
+
+If the color space were to be lost, a warning is printed.
+The image is converted regardless, without proper color conversion.
+Therefore, if the source color space and the default color space of the target format (usually sRGB or Rec.709) do not match, colors will appear incorrect.
+
+Zola performs image processing on the raw pixel data, without taking the color space into account.
+This is not usually an issue, but non-linear color spaces (including sRGB, Rec.709, AdobeRGB, Display P3, and non-linear Rec2020) may exhibit imprecise color blending, especially when scaling up low-resolution images.
```
kleinesfilmroellchen, line 88: From what I can gather after digging through `image`’s source code, they do all the resampling in the default color space, which is really all they could do on their side — image has no knowledge of color spaces, only color data models, where there is no difference between linear RGB and logarithmic sRGB. Handling this properly is probably beyond the scope of zola, as it requires introspecting the color space for the used gamma, doing a nonlossy color space transform (especially tricky for 8-bit channels), scaling and transforming back, multiple nontrivial steps. The current state is fine for all but extreme cases where you actually deal with a lot of color blending and linear color blending should be applied. Therefore I don’t think this ever has to be supported by zola.

components/imageproc/src/processor.rs
```diff
@@ -68,33 +71,51 @@ impl ImageOp {
         }?;
         let mut tmp_output_writer = BufWriter::new(&tmp_output_file);
 
+        let has_color_profile = color_profile.is_some();
+        let add_color_profile = |encoder: &mut dyn ImageEncoder| {
+            if let Some(color_profile) = color_profile {
+                let _ = encoder.set_icc_profile(color_profile).inspect_err(|_| println!("processing {}: Warning: Image encoder for {} does not support color profiles, colors may be incorrect.", self.input_path.display(), self.format.extension()));
```
Keats: I don't think we want printlns here, especially since it's not actionable. I  want  Zola to start using log/env_logger so that would be a good place to start using it

components/imageproc/src/processor.rs
```diff
@@ -68,33 +71,51 @@ impl ImageOp {
...
             }
             Format::WebP { quality } => {
-                let encoder = webp::Encoder::from_image(&img)
-                    .map_err(|_| anyhow!("Unable to load this kind of image with webp"))?;
-                let memory = match quality {
-                    Some(q) => encoder.encode(q as f32),
-                    None => encoder.encode_lossless(),
-                };
-                tmp_output_writer.write_all(memory.as_bytes())?;
+                // use the `image` builtin encoder for lossless, as it supports color profiles
```
Keats, line 95: do we need the webp::Encoder if image supports it?

components/imageproc/tests/test_imgs/adobergb.avif
Keats: let's remove it if it's unused we can add a TODO in the test saying we are missing AVIF decoding

Keats (changes requested):

components/imageproc/src/processor.rs
kleinesfilmroellchen: I’m all for proper logging. Do you want me to add a basic `env_logger` or just the call to `log` in this PR? (I’d rather leave out the logging config to you since I don’t quite know what you want.)

components/imageproc/src/processor.rs
kleinesfilmroellchen: Yes, because it’s the only lossy WebP encoding method that exists; image’s encoder specifically mentions that it can only do lossless (and this is why the `new` method has this special name).

components/imageproc/src/processor.rs
kleinesfilmroellchen: Quite a few of our dependencies already use `log` anyways, so this doesn’t even increase the crate graph (apart from the actual logging frontend)

kleinesfilmroellchen: I’ve now added `log` itself in an earlier commit and switched all of these over.

Keats: I can handle setting up env_logger later

kleinesfilmroellchen:
@Keats Addressed all of your comments, let me know how you want to proceed in regards to `log`.

Keats (approved):

Keats:
Thanks!

Keats:
@kleinesfilmroellchen there's https://github.com/getzola/zola/issues/3077 and https://github.com/getzola/zola/issues/3081 do you think it's linked to that change?

kleinesfilmroellchen:
DQT doesn't look like it's related to ICC?

kleinesfilmroellchen:
As for the second one, we'd need the image itself to check

Overall, I am fairly confident that I didn't change *how* existing formats are read in. So everything should work/not work as before. (It's just that we ignored ICC data before and now look at it).

I don't remember if there's a fallback which discards ICC data if it fails to read?

Keats:
Yeah I'm not sure but that's the only change around image processing in 0.22 i think. I can try to bisect later if i find the time.

PR #3042, opened 2025-12-11
Port logging to `log`

kleinesfilmroellchen:
As per discussion in <https://github.com/getzola/zola/pull/3038#discussion_r2611103892>, we want to move to using [log](https://crates.io/crates/log) for all of our logging needs. This PR does both the backend work, and also adds an env_logger configured with:

- default level `zola=info`, so that only Zola’s output is visible by default (and no debugging)
- respecting terminal color configuration (CLICOLOR/NO_COLOR/terminal detection etc.)
- simple output format `<Level> <Message>` matching `mdBook`, except for `INFO` which only prints the message

Also migrate to `anstyle` which is more flexible than `termcolor` and also provides a utility for color mode auto-detection.

Caveats:
- Some commands have slightly changed output (e.g.
  removed “warning” text for log::warn), but a lot
  of them might still contain formatting symbols
  like “->” and “>”.
- Init command (and its tooling) is unchanged, as
  it should be a nice interactive command.
- There is no `log::success`, so the success-info
  distinction has been lost in some places.
- Separate prints close to each other have been
  transformed into one log call each. They might
  look less related than they did before.
- No new logging is added. Many components might
  benefit from extra logging within them.

---

Sanity check:

## Code changes


components/site/src/lib.rs
```diff
@@ -1254,7 +1255,7 @@ fn log_time(start: Instant, message: &str) -> Instant {
     let do_print = std::env::var("ZOLA_PERF_LOG").is_ok();
     let now = Instant::now();
     if do_print {
-        println!("{} took {}ms", message, now.duration_since(start).as_millis());
+        log::info!("{} took {}ms", message, now.duration_since(start).as_millis());
```
Keats: we meed to remove the custom logging with ZOLA_PERF_LOG env var all around

components/site/src/link_checking.rs
```diff
@@ -274,7 +275,7 @@ pub fn check_external_links(site: &Site) -> Vec<String> {
                     .collect::<Vec<_>>()
             });
 
-            println!(
+            log::info!(
```
Keats, line 278: i think all that functions should stay println? it's stuff we always want the end user to see, logs should be reserved to things like debugging or warnings

src/cmd/serve.rs
```diff
@@ -534,14 +535,14 @@ pub fn serve(
 
                 let local_addr = listener.local_addr().unwrap();
 
-                println!(
+                log::info!(
```
Keats: anything in `src` should stay println/eprintln

Keats (changes requested):
I think this is a bit too much. A lot of it that we always want to show should stay as println. I think logs is going to be used mostly for perf/what's happening at the info level and warnings like in your PR. I don't have a clear view of what it would look like yet but I'm fairly sure everything we're printing during a normal run (such as link checking, timing etc) should stay println

components/site/src/link_checking.rs
kleinesfilmroellchen: As long as the logging framework is configured correctly, it’s very easy to always show these messages. Possibly using a special `target` for log.

kleinesfilmroellchen:
I take issue with the current printing approach because it is sometimes hard to guess which subsystem is doing what unless the messages are very clear, and whether certain messages are even related. mdbook, which has a very similar CLI to zola, also uses `log` for printing its regular CLI output, using only a very basic style that includes the warning level and the message. (Technically, they use `tracing`, which is basically just a more advanced `log`. See e.g. [here](https://github.com/rust-lang/mdBook/blob/7b29f8a7174fa4b7b31536b84ee62e50a786658b/crates/mdbook-driver/src/mdbook.rs).)

Keats:
>  it is sometimes hard to guess which subsystem is doing what unless the messages are very clear

I don't want the classic `[2017-11-09T02:12:24Z DEBUG main]` for the main output of zola. We can define a basic formatter with env_logger (let's say just the message and changing colour for the level warning/error). We do lose the possibility of putting different colour inside a message but i guess that's fine. Can you add the formatting you had in mind in this PR as well?

kleinesfilmroellchen:
> > it is sometimes hard to guess which subsystem is doing what unless the messages are very clear
> 
> I don't want the classic `[2017-11-09T02:12:24Z DEBUG main]` for the main output of zola.

Me neither!

> We can define a basic formatter with env_logger (let's say just the message and changing colour for the level warning/error). We do lose the possibility of putting different colour inside a message but i guess that's fine.

I’m fairly certain you can still inject ANSI codes into the log message, AFAIK there’s quite a bit of support around that (color macros, detecting terminal color support, etc.). I’m not sure where we do that right now though, didn’t see it when going through the print statements. If I missed something please point it out and I’ll definitely add it back.

> Can you add the formatting you had in mind in this PR as well?

Okay, so I will add an `env_logger` with default config (so RUST_LOG works as above) and mdbook-like formatting, no coloring of the message, and colored log level. I won’t add an env_logger to the testing infra or any other binaries that may be floating around, just the main CLI.

Cargo.toml
```diff
@@ -36,6 +36,8 @@ open = "5"
 mime_guess = "2.0"
 # For essence_str() function, see https://github.com/getzola/zola/issues/1845
 mime = "0.3.16"
+env_logger = "0.11"
+anstyle = "1"
```
Keats: we are already using `termcolor` in components/console
It's fine to replace it but let's not have both

Keats:
I'll have a look at the output tomorrow

Cargo.toml
kleinesfilmroellchen: Yes we can remove the anstyle dependency, at least when disabling the color feature of env_logger. (That might require us to detect terminals manually tho, I will see.)

Keats:
Looking at the output: 

1. we should default to showing at least warn when env var not set
2. no need for INFO label. Ideally no need for WARN label as well but coloring the text is a bit trickier...

Cargo.toml
kleinesfilmroellchen: After looking into this more, we should use anstyle instead of termcolor. It also has a utility crate (`anstream`) which provides auto-color-detection, pretty much matching the manual logic in `console` but more accurate (and compatible with old Windows terminals). So I now also ported `console` to `anstyle` which was simple enough.

kleinesfilmroellchen:
> Looking at the output:
> 
>     1. we should default to showing at least warn when env var not set

I noticed this too earlier when testing, so now we fall back to `zola=info` as a logging configuration. This prevents other crates from adding their own confusing `INFO` prints by default (though I have not observed this so far).

>     2. no need for INFO label. Ideally no need for WARN label as well but coloring the text is a bit trickier...

I personally disagree, but will do

Keats:
Sorry can you rebase? The giallo branch merge created a bunch of conflicts

EvanCarroll:
This doesn't make sense, if you're using Axum and Tokio why aren't you tracing for logging?

Keats:
We wouldn't use anything from tracing other than the logs so it wouldn't make sense to use it

EvanCarroll:
Tracing would allow you to run debugging information on a span covering thinks like shortcode expansion or calls to cmark, and get comprehensive reporting -- when desired -- about what took time in generating the site. It would also allow you to get comprehensive timings for requests and it easily flows with axum/tokio (it's the support mechanism of doing this). Not to mention, I think it's probably already in the req stack somewhere by way of `hyper-util` and `axum`.  So it's basically more power and free.

If you want a patch to show it off and see, accept my axum patch and I'll provide an alternative for you to A-B.

Keats:
We dont care about the dev server at all for logging. I don't care
particularly about spans or instrumenting or multiple sinks either, zola is
a basic CLI we can easily handle it with basic logging.

EvanCarroll:
I'm not trying to be standoffish and it's not _just_ about the dev server either (though I don't know why you wouldn't want the things you don't care about). 

* It's no more difficult to use the free macros in tracing and.
* The framework is already in use by our deps.
* It will allow you to benchmark across async tasks and provide debugging about compilation speed.

It's literally switching `log::info!` to `tracing::info!`. There is even a note about migration to tracing from `log` just to ease the difficulty since everyone is moving that direction https://docs.rs/tracing/latest/tracing/#for-log-users

Keats:
I'll well aware of tracing, but since we won't use any of it _right now_, there's no point adding it. As you said, it's easy to swap later if we want to

components/site/src/link_checking.rs
EvanCarroll: I agree with @kleinesfilmroellchen here, generally you don't want to use `println!` and `eprintln!` when you have logging because you want to control context and highlighting with the same framework. This allows you to enable/disable coloring, filter the messages you want, and to reserve `println!` for things you want the user to see, like an IO prompt.

EvanCarroll:
I've rebased this onto next, I'll have a new copy up with migration to tracing and improvements soon.

Keats:
@kleinesfilmroellchen what were the questions?

kleinesfilmroellchen:
> @kleinesfilmroellchen what were the questions?

I forgot that the issue here was only rebasing. I’m already on this and it should be done in a few minutes.

Keats:
Nice, thanks!

PR #3043, opened 2025-12-12
Add optional AVIF decoding support

### LINKED ISSUE #1202: Feature request: Add AVIF support.
Adding this as a feature request now as image-rs seem to be well advanced with their avif support: https://github.com/image-rs/image/issues/1152
The original forum thread for reference: https://zola.discourse.group/t/feature-request-add-support-for-avif-images-to-resize-image/538

I don't have much development time at the moment due to real life, but happy to take a crack at this at a later date if no one else has claimed it.

kleinesfilmroellchen:
As per <https://github.com/getzola/zola/issues/1202#issuecomment-3621105845>, this adds AVIF decoding, but as an optional feature (similar to native TLS) which is off by default, keeping the official release build entirely self-contained. Distribution packagers are *strongly* recommended to enable the feature and make their `zola` package depend on their `libdav1d` package.

This doesn’t add tests for AVIF loading. It would be possible to merge this somehow with the feature, but it seems like wasted effort: even more changes to CI, and we generally expect `image` to work according to their documentation. (We don’t add any special code of our own to load AVIF files, after all.)

Should close #1202, unless there is a native Rust AVIF loader that we could use instead (In which case, close this PR, but I have looked and not found any).

---

Sanity check:

## Code changes

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - x86_64-unknown-linux-gnu-ubuntu-24
           - x86_64-unknown-linux-musl
           - aarch64-unknown-linux-gnu
           - x86_64-pc-windows-msvc
@@ -27,6 +28,8 @@ jobs:
             target: x86_64-pc-windows-msvc
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
+          - os: ubuntu-24.04
+            target: x86_64-unknown-linux-gnu-ubuntu-24
           - os: ubuntu-22.04
             target: x86_64-unknown-linux-musl
           - os: ubuntu-22.04-arm
@@ -48,8 +51,17 @@ jobs:
         run: cargo fmt --check
 
       - name: Cargo build (Native TLS)
+        if: ${{ matrix.os != 'ubuntu-24.04' }}
         run: |
           cargo build --all --no-default-features --features=native-tls
+          cargo clean
+
+      - name: Cargo build (AVIF Loading Support)
+        if: ${{ matrix.target == 'x86_64-unknown-linux-gnu-ubuntu-24' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y dav1d libdav1d-dev pkg-config
+          cargo build --all --no-default-features --features=load-avif
           cargo clean
 
       - name: Cargo build (Rust TLS)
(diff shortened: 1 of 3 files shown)


kleinesfilmroellchen:
Okay so it turns out that Ubuntu 22 has too old of a libdav1d version for this. What else is new, Canonical. I’ll add Ubuntu 24 CI which only builds with AVIF support to cut down the build times as much as possible.

kleinesfilmroellchen:
Welp, I broke the CI queue. The current use of the `matrix` feature in CI is kind of broken; Rust also doesn’t even pick up the target so the musl target isn’t even doing anything.

kleinesfilmroellchen:
@Keats are there any issues with this change? I only noticed today that this hasn’t been addressed yet at all.

Keats:
As mentioned in the comment, the feature needs to work for all supported platforms (windows/mac/linux) by default

kleinesfilmroellchen:
> As mentioned in the comment, the feature needs to work for all supported platforms (windows/mac/linux) by default

The feature is optional. Is it still a requirement? As stated, Ubuntu 22 simply has too old of a libdav1d to ever work. I can’t personally test macOS or Windows outside of CI.

kleinesfilmroellchen:
The whole point here is to give distributors the ability to enable AVIF support, with whatever mechanisms and extra work they need on their platform. As of now, no platform can enable this support even if they want to.

Keats:
> The feature is optional. Is it still a requirement? 

Yeah. I want someone to be able to pull a repo on Windows/Mac/Linux and be able to run the site. Dictionaries for search are an exception as features because of the size but they are not platform dependent.

kleinesfilmroellchen:
I’m interested in getting some help here, then. I can do some limited testing on macOS and Windows, but I don’t have a proper machine for either of those.

For Linux there is also the issue that this makes zola incompatible with distros which ship ancient versions of libdav1d, notably Ubuntu 22. I’m not sure what other major (and currently supported) distro is affected, though.

kleinesfilmroellchen:
So from the little research I have done:
- [rav1d](https://github.com/memorysafety/rav1d) exists, which is an ISRG project and a pure-Rust translation of libdav1d which seems to have feature-parity and supports all of zola’s platforms.
- It only has a C API, but that should be libdav1d-compatible.
- The `dav1d` Rust bindings (used by `image`) can be pointed to a specific build of libdav1d, so one could point them to a build of rav1d.

If someone wants to extend my PR to implement all of this in a build or package script, please go ahead.

PR #3044, opened 2025-12-14
Use giallo instead of syntect

Keats:
This won't work until you have giallo checked out in a sibling folder to zola and after running `just generate-dump`. If you try it on your site, make sure you include the default CSS for giallo:

```css
.giallo-l {
  display: block;
}
.giallo-ln {
  display: inline-block;
  user-select: none;
  white-space: pre;
  margin-right: 0.4em;
  padding: 0.4em;
  min-width: 3ch;
  text-align: right;
  opacity: 0.8;
}
```

I'll replace the path when rust-onig is released and therefore giallo can be on crates.io. I'll also remove the submodules and update docs etc at that point.


docs/config.toml
```diff
@@ -13,9 +13,11 @@ taxonomies = [
 index_format = "elasticlunr_json"
 
 [markdown]
-highlight_code = true
-highlight_theme = "kronuz"
 external_links_class = "external"
+
+[markdown.highlighting]
+theme = "github-dark"
+
```
Keats: The config has changed to allow more options. Setting a theme (or 2 for light/dark dual theming) will enable highlighting.

asibahi:
Hello.

I am wondering how this works with custom syntax. On my website I have a couple of articles with languages not supported natively, and I’m using [separate syntax files](https://git.sr.ht/~asibahi/site/tree/main/item/syntax). I’m wondering what the path for transferring those would be.

Keats:
Odin: https://github.com/DanielGavin/ols/blob/master/editors/vscode/syntaxes/odin.tmLanguage.json
smt: https://github.com/soaibsafi/smt-z3-vscode/blob/main/syntaxes/smt.tmLanguage.json

you should be able to download those in your site, add the paths to them in the config

```
[markdown.highlighting]
extra_grammars = [..., ...]
```
and it should work

Dimev:
Nice work! 
I currently style my code via CSS directly, so it may be nice if `theme = "...` is not required if `style = "class"` is present.

Would it also be possible for the `.giallo-l` spans to be block elements by default, without having to include CSS for this?

Dimev:
Also, is there still a way to generate the css for a given theme?

Keats:
> I currently style my code via CSS directly, so it may be nice if theme = "... is not required if style = "class" is present.

That's not possible, you need a theme to match the actual grammar scopes to theme scopes in textmate and I'm not going to expose that in Zola. How do you style via CSS?

> Would it also be possible for the .giallo-l spans to be block elements by default, without having to include CSS for this?

I'd say no, in theory only `span` are allowed in `<code>` blocks and I don't want to make that part configurable for the HTML renderer

Keats:
> Also, is there still a way to generate the css for a given theme?

Yes, I've just pushed the updated docs if you want to have a look. I won't have the time to do migration guide today though but it's pretty easy in theory.

Dimev:
> That's not possible, you need a theme to match the actual grammar scopes to theme scopes in textmate and I'm not going to expose that in Zola. How do you style via CSS?

Oh, didn't know that. I took a theme I liked and manually modified the colors in the css.

I also see the docs now, thanks!

asibahi:
> Odin: https://github.com/DanielGavin/ols/blob/master/editors/vscode/syntaxes/odin.tmLanguage.json smt: https://github.com/soaibsafi/smt-z3-vscode/blob/main/syntaxes/smt.tmLanguage.json
> 
> you should be able to download those in your site, add the paths to them in the config
> 
> ```
> [markdown.highlighting]
> extra_grammars = [..., ...]
> ```
> 
> and it should work

Thanks for the links. Saves me the search effort later. I was mostly pointing out that it is a breaking change that needs addressing (if you don't intend to keep syntect behind as a fallback).

Complete tangent question: Is it possible somehow to annotate inline elements with a language for coloring or not? In other words, something like `let x = 4;` and I can annotate the inline block being Rust somehow and get Rust coloring?

Keats:
> Saves me the search effort later. I was mostly pointing out that it is a breaking change that needs addressing (if you don't intend to keep syntect behind as a fallback).

I've updated the docs and changelog now so that should be clearer.

> Complete tangent question: Is it possible somehow to annotate inline elements with a language for coloring or not? In other words, something like let x = 4; and I can annotate the inline block being Rust somehow and get Rust coloring?

I don't think so? You can't put which language it is in an inline code. Maybe with shortcodes?

asibahi:
Hey I have been trying this and it doesn't seem to work. I have this line in my `config.toml` and it is still generating errors with `error_on_missing_language`

```toml
extra_grammars = ["./syntax/odin.tmLanguage.json", "./syntax/smt.tmLanguage.json"]
```

One more issue: empty lines seem to be deleted from code blocks. 

This is the new binary 
image

This is the old one (change of theme notwithstanding)

image

Keats:
> One more issue: empty lines seem to be deleted from code blocks.

Ah yeah that's a bug in the HTML renderer I think. I'll fix that.

> error_on_missing_language

is it complaining about odin/smt? It does look highlighted which wouldn't happen with the default plain grammar

asibahi:
> is it complaining about odin/smt? It does look highlighted which wouldn't happen with the default plain grammar

Yes. The highlighted image is Rust :D Odin and smt code created the error and refused to build with `error_on_missing_language = true`. It builds fine with it commented out, but they are not highlighted.

Keats:
I can't reproduce the empty lines being skipped but for the odin issue, it's actually called Odin in the grammar, with capital O. I will make giallo case insensitive for grammar/theme lookups to fix it properly.

asibahi:
Thanks. editing the names worked. I expected it to check `fileTypes` so I added those in as well, so I edited those, but did not think to change the name.

re: blank lines, [my site's repo with the changes is here](https://git.sr.ht/~asibahi/site/log/zola21). (I called the branch `zola21` despite next being `22`, .. anyway). The screenshot I took is from the beginning of the `z3` article. Let me know if I can minimize it in some way.

Keats:
Ok the issue was in the CSS, add `min-height: 1lh;` to giallo-l. I will update giallo and the docs

It will also show an empty line at the end, i'll fix that in giallo directly

izissise:
Hello, I recompiled zola with this new version including this change and I fail to understand how to use it.

```
[markdown.highlighting]
error_on_missing_language = true
style = "class"
theme = "github"
```

I put this in config.toml

but no matter the value I put for theme it keep saying `ERROR Theme `github` does not exist`

I couldn't not find where I should get theme nor where to put them at

Keats:
Themes are listed in https://github.com/getzola/giallo?tab=readme-ov-file#themes

izissise:
Thank you, I figured it out eventually

In the end, I didn't need to load `giallo.css`  in the page because the classes on the code elements didn't change, my previous css theme still work.
Maybe  choosing a theme is not necessary  when `style = "class"` ?

Keats:
> In the end, I didn't need to load giallo.css in the page because the classes on the code elements didn't change, my previous css theme still work.
Maybe choosing a theme is not necessary when style = "class" ?

Are you sure you are using 0.22? A theme is required to be set otherwise I get an error when loading the site. Also there's no way the classes would not change since the highlighting engine is completely different.

izissise:
> Are you sure you are using 0.22? A theme is required to be set otherwise I get an error when loading the site. Also there's no way the classes would not change since the highlighting engine is completely different.

I do have a `giallo.css` being generated in the static folder with something like this

```
/*
 * theme "github-light" generated by giallo
 */

.z-code {
  color: #24292E;
  background-color: #FFFFFF;
}

.z-hl {
  background-color: #F6F8FA;
}

.giallo-ln {
  color: #1B1F234D;
}

.z-comment { color: #6A737D; }
.z-constant { color: #005CC5; }
.z-entity { color: #6F42C1; }
.z-keyword { color: #D73A49; }
.z-storage { color: #D73A49; }
.z-string { color: #032F62; }
.z-support { color: #005CC5; }
.z-variable { color: #E36209; }
...
```

My css code look like this even before 0.22
```

// Unstyled Text Color, and background color
.z-code {
  color: var(--h1);//#ccc
  background-color: var(--h0);
}
// This is a comment
.z-comment {//, .z-punctuation.z-definition.z-comment
  color: var(--h2);//#888
  font-style: italic;
}
// a  (int a;)
.z-variable {
  color: var(--h1);
}
// "This is a string"
.z-string {//, .z-constant.z-other.z-symbol
  color: var(--h6);
}
// 10  (a = 10;)
.z-constant.z-numeric {
  color: var(--h4);
}
// /n
.z-constant.z-character.z-escape {//, .z-string.z-regexp
  color: var(--h7);
}
// false, true, etc...
.z-constant.z-language {
  color: var(--h4);
}
// BUF  (const BUF: usize = 10 * 1024;)
.z-constant.z-other {//, .z-constant.z-character
  color: var(--h4);
}
...
```

I did add the recommended giallo css

```
.giallo-l {
    display: inline-block;
    min-height: 1lh;
    width: 100%;
}
.giallo-ln {
    display: inline-block;
    user-select: none;
    margin-right: 0.4em;
    padding: 0.4em;
    min-width: 3ch;
    text-align: right;
    opacity: 0.8;
}
```

And did some other minors change as some languages name changed

There is an examples here: https://blog.izissise.net/posts/raconn/

Keats:
So it did pick up the github-light and it's not defined anywhere in your config.toml? As for the class names it will depend on the theme itself so the basic one should be the same (not many ways to call a variable or a constant eh) but I wouldn't rely on that.

izissise:
config.toml
```
[markdown.highlighting]
error_on_missing_language = true
extra_grammars = ["codehighlightsyn/gitignore.tmLanguage.json"]
style = "class"
theme = "github-light"
```

I do have the github-light theme.

Ok I understand now, every theme have they own class name, so I should probably just include giallo.css

Thanks

PR #3056, opened 2025-12-29
Logging rewrite to use tracing.

EvanCarroll:
This builds off of the work done in #3042 except uses the tracing framework.

In addition to all the functionality in #3042 (afaik) it also adds the ability to trace `zola serve`. While I find it very useful to know what requests `zola serve` got to easily eliminate issues with browser caching, I added option to silence this if you don't want it `--log-serve=false` or `--no-log-serve`. This is done with `tower_http`, a free benefit to having now fully adopted axum.

This patch also gives the user a better interface into tracing with `-v` and `-vv`.

While this patch is slightly longer, it adds the ability to filter messages you want and gives the logging framework information about the process with the tracing spans.


EvanCarroll:
If this is good, I can squash.

kleinesfilmroellchen:
> If this is good, I can squash.

Please retain my commits as-is. I am still waiting for @Keats to decide on the remaining open questions from the original PR.

Keats:
Have you asked @kleinesfilmroellchen before doing that PR? I also mentioned there was no need to add tracing before so I'm not sure why you're adding it.

EvanCarroll:
I told you I was going to provide an alternative. I don't feel this is any more complex, and I do believe you're getting a much better feature set out of it. It doesn't make sense to me to use a library made for debugging synchronous sequential operations in `zola`.

PR #3058, opened 2025-12-31
Fix components/utils build

SilensAngelusNex:
Running `cargo build` from `components/utils` fails because `components/utils/src/de.rs` uses `time::format_description`. This module doesn't exist unless `time`'s `formatting` feature is enabled.

```
error[E0433]: failed to resolve: could not find `format_description` in `time`
   --> components/utils/src/de.rs:7:11
    |
  7 | use time::format_description::well_known::Rfc3339;
    |           ^^^^^^^^^^^^^^^^^^ could not find `format_description` in `time`
    |
note: found an item that was configured out
   --> /home/wcarv/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/time-0.3.44/src/lib.rs:101:9
    |
100 | #[cfg(any(feature = "formatting", feature = "parsing"))]
    |          --------------------------------------------- the item is gated here
101 | pub mod format_description;
    |         ^^^^^^^^^^^^^^^^^^

error[E0599]: no method named `format` found for struct `OffsetDateTime` in the current scope
  --> components/utils/src/de.rs:84:27
   |
84 |                 Ok(Some(d.format(&Rfc3339).unwrap()))
   |                           ^^^^^^ method not found in `OffsetDateTime`

Some errors have detailed explanations: E0433, E0599.
```

## Sanity check:

* ✔️  Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* ✔️ Are you doing the PR on the `next` branch?
* ➖ Have you created/updated the relevant documentation page(s)?

### DIFF
--- a/components/utils/Cargo.toml
+++ b/components/utils/Cargo.toml
@@ -13,7 +13,7 @@ percent-encoding = { workspace = true }
 regex = { workspace = true }
 slug = { workspace = true }
 tera = { workspace = true }
-time = { workspace = true }
+time = { workspace = true, features = ["formatting"] }
 toml = { workspace = true }
 walkdir = { workspace = true }
 
@@ -21,4 +21,4 @@ errors = { workspace = true }
 
 [dev-dependencies]
 tempfile = "3"
-time = { workspace = true, features = ["macros"] }
+time = { workspace = true, features = ["formatting", "macros"] }


components/utils/Cargo.toml
```diff
@@ -13,12 +13,12 @@ percent-encoding = { workspace = true }
 regex = { workspace = true }
 slug = { workspace = true }
 tera = { workspace = true }
-time = { workspace = true }
+time = { workspace = true, features = ["formatting"] }
```
Keats, line 16: hu i thought it would automatically take the features defined in the main Cargo.toml

SilensAngelusNex: I think it is if you build from the workspace root. I noticed it trying to run `cargo test` from `components/imageproc`.

Keats: ah ok let's not do that then otherwise we're going to have features defined in random places rather than a single one

SilensAngelusNex: It does fail the same way when running `cargo build --package utils` from the root directory. I think building the whole project just currently happens to work because other crates in the workspace (namely the `zola` binary) do enable the feature, and all the workspace crates are using the same `time`.

AFAICT, there's two options to fix this:

1. 
    ```
    # Cargo.toml
    [workspace.dependencies]
    time = { version = "0.3", features = ["formatting" ]}
    ```
2. Adding `features = ["formatting" ]` to the specific crates that need the feature, like I'm doing here.

Looking at `cargo` discussion around workplace dependencies, I think the second option is the way they intend. See https://github.com/rust-lang/cargo/issues/12162#issuecomment-2010322151 and https://github.com/rust-lang/cargo/issues/12162#issuecomment-2020707188. That way you can tell which features each workspace crate actually depends on instead of just knowing that each workspace crate depends on some subset of the features specified in the workspace `Cargo.toml`.

SilensAngelusNex: https://github.com/rust-lang/cargo/issues/11551#issuecomment-1374535828 and https://github.com/rust-lang/cargo/issues/4463 are relevant too.

Keats: eh ok let's do that then, I thought it would get unified either way but i guess some packages can be published independently...

PR #3059, opened 2025-12-31
Add fields to get_image_metadata result

### LINKED ISSUE #3053: Support reading more image metadata fields
I'm running into a couple issues that I'd like to use my images' EXIF metadata to solve[^1]. However,  the only current way to read metadata ([`get_image_metadata`](https://www.getzola.org/documentation/templates/overview/#get-image-metadata)) only reads a few specific types (width, height, format, and mime), which don't include the ones relevant to my use cases.

The easiest thing to do would probably be to add more fields to `imageproc::ImageMetaResponse`, but that would force existing `get_image_metadata` calls to copy fields that they definitely won't use. Alternatively, we could add something like `get_image_metadata_by_tag`, which takes a list[^2] of tags[^3].

Something similar was mentioned in https://github.com/getzola/zola/issues/838#issuecomment-549910823.

[^1]: Populating mouse-over text using `ImageDescription` and sorting pictures by the date they were taken using `DateTimeOriginal`.

[^2]: Getting one tag at a time would cause all the metadata to be parsed multiple times, see the docs for [`Exif`](https://docs.rs/kamadak-exif/latest/exif/struct.Exif.html), the struct that Zola uses to retrieve individual metadata fields.

[^3]: Not really sure of best way to represent tags. [`exif::Tag`](https://docs.rs/kamadak-exif/latest/exif/struct.Tag.html#impl-Tag-1) has a list of constants for well-known tags, so a
    ```
    match tag_name {
        "KnownTagName" => exif::Tag::KnownTagName,
        ...
    }
    ```
    could work. Or maybe there's a way to expose the constants through Tera? Neither of those approaches let users define their own tags, but I think that's fine?

SilensAngelusNex:
Add the `description` and `created` fields to `ImageMetaResponse` so that they will be accessible to callers of `get_image_metadata`.

Fixes #3053

## Sanity check:

* ✔️ Have you checked to ensure there aren't other open [Pull Requests](https://github.com/getzola/zola/pulls) for the same update/change?

## Code changes

* ✔️ Are you doing the PR on the `next` branch?
* ✔️ Have you created/updated the relevant documentation page(s)?


SilensAngelusNex:
I made the type for `created` be `String`, but I'd really like to make it something more structured, like `exif::DateTime` (which can handle timezones and has sub-second resolution). I'm not really sure what the best way to define would be so that Tera can use it. Ideally, we'd still be able to compare it with `<` et al, but it seems like Tera doesn't support that so we'd have to provide comparison functions?

LMK if you think there's a better approach here.

Keats:
>  I'm not really sure what the best way to define would be so that Tera can use it. Ideally, we'd still be able to compare it with < et al, but it seems like Tera doesn't support that so we'd have to provide comparison functions?

Nope it's not possible, we would need to expose a specific type in the Value enum in Tera for datetimes, I considered it for tera2 but ended up not doing it because it adds lots of complexity to handle all the tz/various libraries. I will add some tests though, eg `my_date is after(d=other_date)`, `my_date is before(d=other_date)`. Not as nice as `<, >` but good enough

components/imageproc/src/helpers.rs
```diff
@@ -28,22 +29,44 @@ pub fn fix_orientation(img: &DynamicImage, raw_metadata: Option<Vec<u8>>) -> Opt
...
+    let metadata = metadata?;
+    let field = metadata.get_field(tag, ifd_num)?;
+    let val = &field.value;
+    let exif::Value::Ascii(bytes) = val else {
+        eprintln!("Found an unexpected field value for {:?}: {:?}", tag, val);
+        return None;
+    };
+
+    if bytes.len() == 0 {
+        eprintln!("String value for {:?} has empty Ascii vec.", tag);
```
Keats: can we bubble those as errors so the user will know which image it's for?

Keats: and is an empty value an error?

Keats (changes requested):

components/imageproc/src/helpers.rs
SilensAngelusNex: Fixed

components/imageproc/src/helpers.rs
```diff
@@ -23,27 +67,78 @@ pub fn fix_orientation(img: &DynamicImage, raw_metadata: Option<Vec<u8>>) -> Opt
...
+    let mut s = String::new();
+    s.push_str(&String::from_utf8_lossy(&*bytes[0]));
+    Ok(s)
+}
+
+pub fn get_description(metadata: Option<&Exif>) -> Result<String, ExifError<'_>> {
+    get_string_field(metadata, exif::Tag::ImageDescription, exif::In::PRIMARY)
+}
+
+pub fn get_created_datetime(metadata: Option<&Exif>) -> Result<String, ExifError<'_>> {
```
Keats: do all photos have a description/created? We don't really want to error if it's just missing for some reasons

Keats (commented):
LGTM, just a quick question

bobrocke:
I'm looking forward to this. Is it possible we could see this in the next Zola release?

Keats:
Yes, in 0.23. It's a huge release though so it's going to take some time

components/imageproc/src/helpers.rs
SilensAngelusNex: Good point. I think it'd be weird to not have a CreatedDatetime, just like it'd be weird to not have an Orientation, but I think it's fairly normal to be missing a Description. Moved things around so it's easier to omit error logs for fields that it's normal to be missing and so that we don't log multiple times if the metadata is completely missing or can't be parsed.

Keats: > I think it'd be weird to not have a CreatedDatetime,

Some places will strip all metadata of images, I don't know if they even keep a datetime?

SilensAngelusNex: So do you think none of CreatedDatetime Orientation or Description should print an error if they're missing? Or keep the print for Orientation, but skip it for the other two?

Keats: Everything should be optional

SilensAngelusNex: Done

Keats:
Thanks!

PR #3079, opened 2026-01-15
Fix: Include sub-pages of redirect sections in fuse search index.

schilkp:
The fuse search index exporter previously did not include sub-pages of redirect sections.

Have the fuse exporter apply the same logic for deciding whether a page should be included in the search index or not as the elasticlunr exporter, by updating it to also include such sub pages of redirect sections.

----

Sanity check:

## Code changes

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -11,9 +11,9 @@ pub struct Cli {
     #[clap(short = 'r', long, default_value = ".")]
     pub root: PathBuf,
 
-    /// Path to a config file other than config.toml in the root of project
-    #[clap(short = 'c', long, default_value = "config.toml")]
-    pub config: PathBuf,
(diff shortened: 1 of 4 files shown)


Keats (approved):
We should really have that logic in a single place

schilkp:
> We should really have that logic in a single place

Agreed - I was thinking something along the lines of a "SearchIndexGenerator" trait, with `add_section`, `add_page` etc funcs? That way the iteration/filtering logic can be in one place and drive whatever generator is active?

Happy to refactor this if you want - here or in a second pr :+1:

Keats:
I don't think we need a trait, we can probably just re-use the struct Item<'a> from that file and pass a Vec of that to each backend?

PR #3085, opened 2026-01-17
Separate search index item collection and serialization.

schilkp:
Previously, all search index styles manually gathered items which should be included from the Library. This lead to subtly different inclusion rules. For instance, the `elasticlunr` search index would include the sub-pages of redirect sections, while `fuse` did not.

This patch centralizes the collection into a single routine which is used for all search indices, making sure they behave the same.

This replaces #3079.

---

Sanity check:

## Code changes

### DIFF
--- a/components/search/src/elasticlunr.rs
+++ b/components/search/src/elasticlunr.rs
@@ -1,11 +1,10 @@
 use config::{Config, Search};
-use content::{Library, Section};
-use elasticlunr::{Index, IndexBuilder, lang};
+use content::Library;
+use elasticlunr::{IndexBuilder, lang};
 use errors::{Result, bail};
-use time::OffsetDateTime;
 use time::format_description::well_known::Rfc3339;
 
-use crate::clean_and_truncate_body;
+use crate::{IndexItem, clean_and_truncate_body, collect_index_items};
 
 pub const ELASTICLUNR_JS: &str = include_str!("elasticlunr.min.js");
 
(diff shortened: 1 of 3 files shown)


components/search/src/elasticlunr.rs
```diff
@@ -40,37 +39,30 @@ fn path_tokenizer(text: &str) -> Vec<String> {
...
 
-fn fill_index(
-    search_config: &Search,
-    title: &Option<String>,
-    description: &Option<String>,
-    datetime: &Option<OffsetDateTime>,
-    path: &str,
-    content: &str,
-) -> Vec<String> {
+fn fill_index(search_config: &Search, item: &crate::IndexItem) -> Vec<String> {
```
Keats: any reason here and in fuse to not pass by value? It's not like we will use it elsewhere?

schilkp: Updated to pass by value here.

Not sure what you mean by "in fuse" - there is no helper function in that module which could take an item by value?

Keats: I read it wrong sorry

components/search/src/elasticlunr.rs
```diff
@@ -40,37 +39,30 @@ fn path_tokenizer(text: &str) -> Vec<String> {
...
     let mut row = vec![];
 
     if search_config.include_title {
-        row.push(title.clone().unwrap_or_default());
+        row.push(item.title.clone().unwrap_or_default());
     }
 
     if search_config.include_description {
-        row.push(description.clone().unwrap_or_default());
+        row.push(item.description.clone().unwrap_or_default());
```
Keats, line 50: do you still need the clone()?

schilkp: As far as I can see yes: the elasticlunr search index generator assembles every row into a `Vec<String>` as every row contains a mixture of strings contained in Library and owned strings created by `date.format(),
path.to_string(), clean_and_truncate_body(),` ...

The clone creates owned strings for the entries of item that are not changed.

I kept this behavior the same. Avoiding the cloning would entail re-working the whole search index generator.

Or am I missing something?

Keats: since we don't use the fields afterwards we should be able to just do `item.description.unwrap_or_default()` etc I think?

schilkp: Sorry - I should stop replying while sleepy. 
I meant to write "as far as I can see *no*:"

`item.description` is a `&'a Option<String>`, as it is a reference to the real `Option<String>` inside the actual library.

`row` is a `Vec<String>`. It cannot be a `Vec<&'a String>` because it also holds strings that are directly owned (results from `date.format()`, `clean_and_truncate_body()`, ..). 

This means to insert `item.description` into `row`, I need to take ownership of it by cloning.

Without `clone()`:

```
error[E0507]: cannot move out of `*item.description` which is behind a shared reference
    --> components/search/src/elasticlunr.rs:50:18
     |
  50 |         row.push(item.description.unwrap_or_default());
     |                  ^^^^^^^^^^^^^^^^ ------------------- `*item.description` moved due to this method call
     |                  |
     |                  help: consider calling `.as_ref()` or `.as_mut()` to borrow the type's contents
     |                  move occurs because `*item.description` has type `std::option::Option<std::string::String>`, which does not implement the `Copy` trait
     |
```

With `as_ref` and without `clone`:

```
error[E0308]: mismatched types
    --> components/search/src/elasticlunr.rs:50:18
     |
  50 |         row.push(item.description.as_ref().unwrap_or_default());
     |             ---- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `String`, found `&String`
     |             |
     |             arguments to this method are incorrect
     |
```

Or am I misunderstanding what you are asking?

Keats: > Sorry - I should stop replying while sleepy.

Same, I forgot they were references.

PR #3089, opened 2026-01-22
Next version

Keats:


Stebe25:
While trying out the next branch, I noticed a couple of things:

get_url for static files is broken in multilingual sites (e.g. path="style.css"): zola prepends "lang/" to the path if the page/section lang is not the default language.

Components sometimes wrap their output (or parts of it) in `<pre><code> </code></pre>`, if there is a tab before the html tag; e.g. after "%}" or empty lines:
```
{% component wraps_in_pre_code() %}
	<p>tab => wrapped in pre-code</p>
<p>no tab  => not wrapped</p>
	<p>tab => not wrapped</p>

	<p>tab => wrapped in pre-code</p>
{% if x %}{% endif %}
	<p>tab => wrapped in pre-code</p>
{% if x %}{% endif -%}
	<p>tab => not wrapped</p>
{% endcomponent %}
```

Macros had access to the config and lang variables, while components do not (config and lang are therefore not available in templates included in components).

Keats:
Thanks for trying!

> get_url for static files is broken in multilingual sites (e.g. path="style.css"): zola prepends "lang/" to the path if the page/section lang is not the default language.

I just pushed a commit to fix that, good catch!

> Components sometimes wrap their output (or parts of it) in <pre><code> </code></pre>, if there is a tab before the html tag; e.g. after "%}" or empty lines:

That's sadly a CommonMark "feature". Things that are indented by 4 spaces are considered a code block. I'm very tempted to fork pulldown-cmark and remove that as it just causes issues non stop.

> Macros had access to the config and lang variables, while components do not (config and lang are therefore not available in templates included in components).

Do you mean shortcodes? You can pass it explicitly to the components that need it eg `{{ <component config lang /> }}` assuming the components parameters are called config/lang.

Stebe25:
> That's sadly a CommonMark "feature". Things that are indented by 4 spaces are considered a code block. I'm very tempted to fork pulldown-cmark and remove that as it just causes issues non stop.

Thanks, now that I know that, I can simply change my tab spaces or simply use -%}, but could be quite annoying for someone who doesn't know that.

> 
> > Macros had access to the config and lang variables, while components do not (config and lang are therefore not available in templates included in components).
> 
> Do you mean shortcodes? You can pass it explicitly to the components that need it eg `{{ <component config lang /> }}` assuming the components parameters are called config/lang.

I haven't used components much, because I ran into the CommonMark "feature", but from what I've seen, they turned out really nice! I mean shortcodes too, but macros also had access to these variables without explicitly passing them. Just thought to mention it, because it was useful if you don't have to pass config and lang around (almost) everywhere.
For most use cases lang is probably not that important anymore since the trans function no longer needs the currently mandatory "lang=lang" to get the translation (in multilingual sites). The num_format filter still needs a locale argument though.

I forgot to mention that `{{ __tera_context }}` no longer works. And the filters num_format and markdown are "unknown".

Keats:
> macros also had access to these variables without explicitly passing them

Hmm no macros were the same as components, you wouldn't get anything outside of what is defined in the macro parameter

> I forgot to mention that {{ __tera_context }} no longer works. 

:confused: there's even a test for that! do you have a sample template where it fails?

Stebe25:
> > I forgot to mention that {{ __tera_context }} no longer works.
> 
> 😕 there's even a test for that! do you have a sample template where it fails?

Sure, works on version <= 0.22, but fails for 0.23.
[tera_context.zip](https://github.com/user-attachments/files/25665577/tera_context.zip)

Keats:
Ok found the issue: https://github.com/Keats/tera2/pull/103
I'll publish a new alpha later today

ARitz-Cracker:
Hey, noticed that tera v2 was released recently, just wanted to say congrats! 🎉

Keats:
Thanks!

PR #3090, opened 2026-01-22
Docs: Add instructions for deployment with Cloudflare Workers

matoutech:
Cloudflare added instructions for migrating from Pages to Workers. Both products exist at the moment but it sounds like Pages is being phased out. See https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/

Sanity check:


Keats:
Thanks!

PR #3065, opened 2026-01-07
Add internal resolution for assets in `get_url` function

SamWilsn:
This is a rebase & squash of #2726 (by @ZzMzaw) on top of the latest `next` branch.

PR #3094, opened 2026-01-27
Make `@/` links resolve for assets too

SamWilsn:
An alternative (and incomplete) implementation of https://github.com/getzola/zola/pull/3065.

I've done some light testing with languages, paths, slugs, dates, etc. but comparing the complexity of this pull request against #3065, I'm pretty sure I'm missing something.

Anyway, wanted to get some early feedback before going too far down the rabbit hole!

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -269,7 +269,7 @@ impl Page {
                 path.push(filename);
                 path = path
                     .strip_prefix(base_path.join("content"))
-                    .expect("Should be able to stripe prefix")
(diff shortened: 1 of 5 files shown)


Keats:
I'm in a middle of a huge refactor to use tera2 so it's probably better to hold out a bit until this gets in the next branch. This is the kind of stuff I want to make sure is fixed in 0.23

Keats:
Closing in favour of https://github.com/getzola/zola/pull/3199
I've ended up fixing it while looking at the PR list so I can get 0.23 released soon

PR #3106, opened 2026-02-04
docs: Update Gentoo documentation

zyxhere:
Sanity check:

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -51,9 +51,9 @@ $ sudo dpkg -i zola_<version>_amd64_debian_<debian_version>.deb
 
 Zola is available via [GURU](https://wiki.gentoo.org/wiki/Project:GURU).
 
+See the end user documentation for how to [enable](https://wiki.gentoo.org/wiki/Project:GURU/Information_for_End_Users#Adding_the_GURU_repository) the GURU repository. Afterwards simply run:
+
 ```sh
-$ sudo eselect repository enable guru
-$ sudo emaint sync --repo guru
 $ sudo emerge --ask www-apps/zola
 ```
 


docs/content/documentation/getting-started/installation.md
```diff
@@ -51,12 +51,26 @@ $ sudo dpkg -i zola_<version>_amd64_debian_<debian_version>.deb
...
 
+See the end user documentation for how to [enable](https://wiki.gentoo.org/wiki/Project:GURU/Information_for_End_Users#Adding_the_GURU_repository) the GURU repository. Afterwards simply run:
+
 ```sh
-$ sudo eselect repository enable guru
-$ sudo emaint sync --repo guru
 $ sudo emerge --ask www-apps/zola
 ```
 
+To further optimize the build you can use `RUSTFLAGS` environmental variable to enable [LTO](https://doc.rust-lang.org/beta/rustc/linker-plugin-lto.html) and use [target-cpu](https://rust-lang.github.io/packed_simd/perf-guide/target-feature/rustflags.html#target-cpu) to optimize for the local machine
```
Keats: isn't LTO already enabled by default?

docs/content/documentation/getting-started/installation.md
```diff
@@ -51,12 +51,26 @@ $ sudo dpkg -i zola_<version>_amd64_debian_<debian_version>.deb
...
+
+```sh
+# You will need to use either lld or mold as as they are the only linkers that can understand LLVM's IR.
+$ sudo RUSTFLAGS="-C target-cpu=native -C linker-plugin-lto -C link-arg=-fuse-ld=lld" emerge --ask www-apps/zola
+```
+
+> In case of mold you might also need to pass `-C link-arg=-flto` if the compilation fails with LTO related errors.
+
+Using `-C target-cpu=native` means that rust will optimize zola for the specific
+CPU model being used to build it thus making it non-portable to other machines if they have a different CPU model. Alternatively you can use `target-cpu=x86-64-v3` to optimize for the [v3 microarchitecure](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) level of x86 making it portable to any CPU with x86-64-v3 support.
```
Keats: Do you see a significant perf change by using target-cpu=native with zola?

docs/content/documentation/getting-started/installation.md
zyxhere: We disable it downstream https://github.com/gentoo/guru/blob/a0e97391c1d39eb63de7739763e83989f3ae1e9d/www-apps/zola/zola-0.22.1.ebuild#L640 since release profile enables stripping which we want to do ourselves.

LTO in Cargo.toml only enables LTO for zola not every crate. The options in `$RUSTFLAGS` pass down to every single crate. linker-plugin-lto also causes the bundled C/C++ libraries (onigurami, libwebp-sys) to optimize with LTO.

docs/content/documentation/getting-started/installation.md
zyxhere: Good question I'll have to benchmark that (would using `time` be sufficient?). I'll also see if using thin LTO vs full LTO vs linker-plugin-lto causes performance/size differences

Keats:
Hyperfine is probably better?

zyxhere:
On Wed, 2026-02-04 at 03:07 -0800, Vincent Prouillet wrote:
> Keats left a comment (getzola/zola#3106)
> Hyperfine is probably better?

I have waited about 15 minutes and nothings happening with hyperfine.
How long does it usually take for the benchmark to even start?
```
 hyperfine 'zola --version'
```
The above works but zola serve doesn't.

zyxhere:
On Wed, 2026-02-04 at 17:11 +0500, zyxhere💭 wrote:
> On Wed, 2026-02-04 at 03:07 -0800, Vincent Prouillet wrote:
> > Keats left a comment (getzola/zola#3106)
> > Hyperfine is probably better?
> 
> I have waited about 15 minutes and nothings happening with hyperfine.
> How long does it usually take for the benchmark to even start?
> ```
>  hyperfine 'zola --version'
> ```
> The above works but zola serve doesn't.

Oh wait I think its not working because serve assumes theres a terminal
present to show the local ip address where it is serving? And hyperfine
spawns a non interactive (I think?) or some other weird thing is
happening?.

Well nevermind a script like the following is working fine.
```
  #!/bin/sh

  $(zola serve)
```

Keats:
you wan to use zola build, zola serve is a webserver that doesn't end until stopped

zyxhere:
On Wed, 2026-02-04 at 04:26 -0800, Vincent Prouillet wrote:
> Keats left a comment (getzola/zola#3106)
> you wan to use zola build, zola serve is a webserver that doesn't end until stopped

Duh... thanks now its working. Now lets try different benchmarks

zyxhere:
Not much difference
```sh
#  hyperfine --warmup 3 '/var/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build' '/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build' '/run/portage/www-apps/zola-0.22.1/image/usr/bin/zola build'
Benchmark 1: /var/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build
  Time (mean ± σ):     638.2 ms ±  13.5 ms    [User: 1046.8 ms, System: 171.1 ms]
  Range (min … max):   617.6 ms … 657.6 ms    10 runs
 
Benchmark 2: /tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build
  Time (mean ± σ):     631.1 ms ±  17.2 ms    [User: 1041.2 ms, System: 176.6 ms]
  Range (min … max):   604.4 ms … 661.0 ms    10 runs
 
Benchmark 3: /run/portage/www-apps/zola-0.22.1/image/usr/bin/zola build
  Time (mean ± σ):     612.9 ms ±  14.9 ms    [User: 1017.9 ms, System: 167.2 ms]
  Range (min … max):   584.1 ms … 635.5 ms    10 runs
 
Summary
  /run/portage/www-apps/zola-0.22.1/image/usr/bin/zola build ran
    1.03 ± 0.04 times faster than /tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build
    1.04 ± 0.03 times faster than /var/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola build
    ```
    
    Thee /var/tmp is vanilla zola, /tmp is with target-cpu=native, /run is with native and -C linker-plugin-lto.
Also heres the binary size:
```sh
# du /var/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola /tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola /run/portage/www-apps/zola-0.22.1/image/usr/bin/zola
43900	/var/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola
44256	/tmp/portage/www-apps/zola-0.22.1/image/usr/bin/zola
43064	/run/portage/www-apps/zola-0.22.1/image/usr/bin/zola
```

Keats:
Not sure it's worth the mention then

docs/content/documentation/getting-started/installation.md
```diff
@@ -51,9 +51,9 @@ $ sudo dpkg -i zola_<version>_amd64_debian_<debian_version>.deb
 
 Zola is available via [GURU](https://wiki.gentoo.org/wiki/Project:GURU).
 
+See the end user documentation for how to [enable](https://wiki.gentoo.org/wiki/Project:GURU/Information_for_End_Users#Adding_the_GURU_repository) the GURU repository. Afterwards simply run:
+
 ```sh
-$ sudo eselect repository enable guru
-$ sudo emaint sync --repo guru
```
Keats, line 56: what's the reason for removing that? The docs say the same thing except you can also add a custom text file

zyxhere: To encourage the use of `emerge` only. emaint works but remembering two seperate commands for syncing the repository isn't optimal

Keats:
Thanks!

PR #3117, opened 2026-02-28
register missing num_format and markdown filters in tera

Stebe25:

Sanity check:

## Code changes


Keats (approved):
They will be replaced in https://github.com/getzola/zola/blob/next/components/site/src/tpls.rs#L5 but yeah that's too late for the initial load so we could even use a placeholder fn but that works too.

Thanks!

PR #3125, opened 2026-03-12
feat: exclude individual pages from RSS/Atom feeds

### LINKED ISSUE #3124: Exclude individual pages from RSS/Atom feeds
Sections support `generate_feeds` in their front matter — could this be added to pages as well?

Pages like "About me" or "Impressum" shouldn't appear in feeds but currently require workarounds to exclude. A page-level `generate_feeds = false` would solve this and be consistent with the existing section front matter.

sassman:
- Allow individual pages to opt out of RSS/Atom feeds via `generate_feeds = false`
- Consistent with the existing section-level `generate_feeds` field
- Defaults to `true`, so existing sites are unaffected
- Closes #3124


Keats:
The naming doesn't work here, we can't generate a feed for a single page. Something like include_in_feeds would be better. The PR also needs to be done on the next branch and will have lots of conflicts otherwise.

PR #3128, opened 2026-03-18
feat: exclude individual pages from RSS/Atom feeds

### LINKED ISSUE #3124: Exclude individual pages from RSS/Atom feeds

sassman:
Follow-up to #3125, closes #3124.

- pages can opt out of all feeds via `include_in_feeds = false` in front matter
- defaults to true, so existing behavior is unchanged
- expose as template variable for conditional rendering


components/site/src/feeds.rs
```diff
@@ -36,8 +36,11 @@ pub fn prepare_feed(
     feed_limit: Option<usize>,
     cache: &RenderCache,
 ) -> FeedData {
-    let mut pages: Vec<_> =
-        all_pages.iter().copied().filter(|p| p.meta.date.is_some() && p.meta.render).collect();
+    let mut pages: Vec<_> = all_pages
+        .iter()
+        .copied()
+        .filter(|p| p.meta.date.is_some() && p.meta.include_in_feeds)
```
Keats: you dropped the `p.meta.render` check

Keats (changes requested):

sassman:
Thanks for bringing this over the line 🙏

PR #3139, opened 2026-04-10
Add minimal default theme to zola init

### LINKED ISSUE #3137: Improving zola init - thoughts?
## The Problem

Right now `zola init` creates empty directories. You run it, then immediately hit an error when you try `zola serve` because there are no templates. Every new user has to manually create `base.html`, `index.html`, `page.html`, etc. before they can see anything.

This is a pretty rough first experience compared to other SSGs.

## What I'm Thinking

Generate some basic files so the site actually works out of the box:

- `.gitignore` (public/, .DS_Store, etc.)
- `README.md` with basic commands
- `content/_index.md` 
- Minimal templates: `base.html`, `index.html`, `page.html`, `section.html`
- Simple `style.css` (nothing fancy, just readable defaults)

Then optionally add a prompt asking what type of site:
- **Blog** - adds `content/posts/` with an example post, taxonomies configured
- **Docs** - adds `content/docs/` with weight-based sorting, search enabled
- **Portfolio** - adds about page and projects section
- **Minimal** - just the basics (current behavior)

Could also support `zola init myblog --template blog` for non-interactive use.

## Why This Helps

- New users get a working site immediately
- Templates show how things work (better than empty files)
- Still minimal - we're talking ~50 lines of CSS, basic HTML structure
- Backward compatible if we keep a minimal option

## Questions

1. Does this fit with Zola's philosophy? I know you value simplicity.

2. Should templates be super bare-bones or include minimal styling?

3. Better as opt-in (`--scaffold` flag) or default behavior?

4. Would you rather keep this external (separate template repo)?

## Alternatives

If built-in templates feel wrong:
- Just add `.gitignore` and `README.md`, link to docs for the rest
- Create a separate `zola-templates` repo people can clone
- Improve the docs with copy-paste starter templates

## Prior Art

Most SSGs do something like this - Hugo, Jekyll, 11ty all generate working starter projects. But I get that Zola might want to stay more minimal.

---

Happy to implement this if it sounds useful. Just wanted to check first before diving in. Thoughts?

nooscraft:
## Summary

Fixes #3137.

`zola init` currently creates empty directories, causing immediate errors when running `zola serve`. This PR adds a minimal default theme so new users have a working site out of the box.

## What changes

Running `zola init` now generates:

- `.gitignore` (ignores `public/`, `.DS_Store`)
- `README.md` with basic commands
- `content/_index.md` - homepage content
- `templates/base.html`, `index.html`, `page.html`, `section.html` - basic templates
- `static/style.css` - ~150 lines of minimal CSS using system fonts

That's it. No dependencies, no JavaScript, no web fonts.

## Why this approach

Per @Keats's guidance in #3137, following 11ty's minimal style: enough to look intentional and not broken, but easy to replace entirely.

## Backward compatibility

Fully backward compatible. All existing tests pass. The `--force` flag works as before. Only adds files; no behavior changes.

---

**Note to @ReduxFlakes and @daudix:** Sorry for the confusion earlier. The original description was way too verbose. Hope this version is clearer.


src/cmd/init.rs
```diff
@@ -25,6 +25,201 @@ theme = "catppuccin-mocha"
...
+    overflow-x: auto;
+    margin-bottom: 1rem;
+}
+
+pre code {
+    background: none;
+    padding: 0;
+}
+"#;
+
```
Keats, line 36: let's put them in files, i haven't reviewed the output yet

Keats (changes requested):

src/cmd/init.rs
nooscraft: Updated!

PR #3149, opened 2026-04-30
Add `text_direction()` Tera function

### LINKED ISSUE #3144: Request: Report language directionality (LTR/RTL) as global Tera variable
### Summary

It would be wonderful if Zola exposed a `{{ dir }}` variable alongside the existing `{{ lang }}` variable through the global Tera context. This proposed `{{ dir }}` variable would have a value of either `"ltr"` or `"rtl"`, and would be usable with the [global `dir=` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir). Basically, this variable would expose whether the site's active language is predominantly left-to-right (LTR) or right-to-left (RTL) orthography.

### Motivation

Some Zola themes have hard-coded lists of which languages are left-to-right and which are right-to-left, and adjust their HTML structure accordingly ([example 1](https://github.com/welpo/tabi/blob/17029f0/templates/base.html#L19), [example 2](https://github.com/getzola/zola/issues/2594)). Keeping these arrays up-to-date over time, separately in each theme, doesn't sound like a great state of affairs, IMO.

Since Zola already validates the Unicode language code by [parsing it into a `unic_langid::LanguageIdentifier` instance (see `components/config/src/config/languages.rs#L116-125`](https://github.com/getzola/zola/blob/5d2d821/components/config/src/config/languages.rs#L116-L125)), I believe we could trivially compute `{{ dir }}` based on the return value of [`.character_direction()`](https://docs.rs/unic-langid/latest/unic_langid/struct.LanguageIdentifier.html#method.character_direction). Probably in a manner like so:

`LanguageIdentifier::character_direction()` | `{{ dir }}` value
---|---
`CharacterDirection::LTR` | `"ltr"`
`CharacterDirection::RTL` | `"rtl"`
`CharacterDirection::TTB` | `"ltr"`[^1]

[^1]: The `dir=` HTML attribute only deals with _horizontal_ text directionality and only supports `"ltr"` or `"rtl"`. Most i18n tutorials I've found online recommend setting `dir="ltr"` for top-to-bottom scripts, e.g. Mongolian and Japanese, because top-to-bottom character direction + **left-to-right paragraph direction** is the most common arrangement for these languages.

ebkalderon:
Closes #3144.

Please let me know if you would like me to rebase this work on top of the `next` branch, @Keats. Thanks so much for your input thus far! :heart:

### DIFF
--- a/components/templates/src/functions/i18n.rs
+++ b/components/templates/src/functions/i18n.rs
@@ -29,6 +29,47 @@ impl Function<TeraResult<String>> for Trans {
     }
 }
 
+#[derive(Debug)]
+pub struct TextDirection {
+    config: Config,
+}
+
+impl TextDirection {
+    pub fn new(config: Config) -> Self {
+        Self { config }
+    }
+}
+
+impl Function<TeraResult<String>> for TextDirection {
+    fn call(&self, kwargs: Kwargs, state: &State) -> TeraResult<String> {
+        use unic_langid::{CharacterDirection, LanguageIdentifier};
+
+        let lang: String = kwargs
+            .get("lang")?
+            .or(state.get("lang")?)
+            .unwrap_or_else(|| self.config.default_language.clone());
+
+        let lang: LanguageIdentifier = lang
+            .parse()
+            .map_err(|_| {
+                Error::message(format!("`text_direction`: Language `{lang}` is not a valid Unicode Language Identifier (see http://unicode.org/reports/tr35/#Unicode_language_identifier)"))
+            })?;
+
+        let dir = match lang.character_direction() {
+            CharacterDirection::LTR => "ltr",
+            CharacterDirection::RTL => "rtl",
+            // String must be compatible with the [`dir=` global HTML attribute][dir], so we only deal
+            // with horizontal text directionality (only "ltr" or "rtl"). Traditional Mongolian is
+            // written top-to-buttom with lines progressing left-to-right. Therefore, report "ltr" here.
+            //
+            // [dir]: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir
+            CharacterDirection::TTB => "ltr",
+        };
+
+        Ok(dir.to_string())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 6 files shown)


docs/content/documentation/templates/overview.md
```diff
@@ -527,6 +527,17 @@ URLs are cached based on the URL, and data files are cached based on the file mo
...
+### `text_direction`
+Gets the horizontal text directionality for the `default_language`, the `lang`uage given or the active language:
+
+```jinja
+{{/* text_direction() */}}
+{{/* text_direction(lang="fr") */}}
+{{/* text_direction(lang=lang) */}}
+```
+
+Returns the string literal `ltr` or `rtl`, which may be passed directly into the [`dir` global HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir).
```
ebkalderon, line 595: To aid code review, I'd like to call out the end-user documentation for this new function.

I chose to make the `lang` parameter behave identically in `text_direction()` as with `trans()`. I can totally change this, though, if you think it could be done better!

Keats:
You will need to do it on the next branch

ebkalderon:
Sounds good to me! I'll rebase onto the `next` branch and fix any merge conflicts.

ebkalderon:
Okay, I think I've managed to cleanly rebase all my changes onto the `next` branch. Ready for review!

docs/content/documentation/templates/overview.md
Keats: that seems fine to me

Keats (approved):

components/templates/src/functions/i18n.rs
```diff
@@ -29,6 +29,47 @@ impl Function<TeraResult<String>> for Trans {
...
+            .map_err(|_| {
+                Error::message(format!("`text_direction`: Language `{lang}` is not a valid Unicode Language Identifier (see http://unicode.org/reports/tr35/#Unicode_language_identifier)"))
+            })?;
+
+        let dir = match lang.character_direction() {
+            CharacterDirection::LTR => "ltr",
+            CharacterDirection::RTL => "rtl",
+            // String must be compatible with the [`dir=` global HTML attribute][dir], so we only deal
+            // with horizontal text directionality (only "ltr" or "rtl"). Traditional Mongolian is
+            // written top-to-buttom with lines progressing left-to-right. Therefore, report "ltr" here.
```
ebkalderon, line 63: Uh oh... I somehow missed this unfortunate `buttom` typo when I initially wrote this. Sorry about that, @Keats! :joy:

PR #3154, opened 2026-05-05
fix(templates): match get_taxonomy_term by slug as fallback

### LINKED ISSUE #2829: `get_taxonomy_term` does not handle terms with special characters
# Bug Report

## Environment

Zola version: 0.20.0

## Expected Behavior
The get_taxonomy_term function should correctly retrieve terms that contain special characters, such as accents.

## Current Behavior
When passing a term with special characters (e.g., "acción") to get_taxonomy_term, Zola returns the following error:
```
Error: Reaseon: `get_taxonomy_term` received an unknown term: acción
```
I attempted to normalize the term using `[slugify](https://keats.github.io/tera/docs/#slugify)`, but the issue persists.

## Step to reproduce

 1. Create a taxonomy (e.g., tags) containing a term with special characters (acción).
 2. Use get_taxonomy_term(kind="tags", term="acción") in a 
 3. Run Zola and observe the error.

This issue was initially discussed in the Zola forum: [Conditionally displaying RSS Feed icons for tags based on feed availability](https://zola.discourse.group/t/conditionally-displaying-rss-feed-icons-for-tags-based-on-feed-availability/2530)

SAY-5:
Closes #2829.

`get_taxonomy_term` only compared the user-provided term against `TaxonomyTerm.name`, so terms whose characters get slugified (accents, spaces, etc.) couldn't be looked up by the slug form that templates often have on hand. The lookup now falls back to slugifying the input and matching against `TaxonomyTerm.slug`, mirroring how `get_taxonomy_url` already works.

A regression test using `acción` / `accion` covers both lookup paths.

### DIFF
--- a/components/templates/src/functions/taxonomy.rs
+++ b/components/templates/src/functions/taxonomy.rs
@@ -355,13 +355,18 @@ mod tests {
         let tag =
             TaxonomyTerm::new("Programming", &config.default_language, &taxo_config, &[], &config);
         let tag_fr = TaxonomyTerm::new("Programmation", "fr", &taxo_config_fr, &[], &config);
+        // Regression for https://github.com/getzola/zola/issues/2829: terms
+        // whose names get slugified (accents, spaces, etc.) must be
+        // retrievable via the slug.
+        let tag_accent =
+            TaxonomyTerm::new("acción", &config.default_language, &taxo_config, &[], &config);
         let tags = Taxonomy {
             kind: taxo_config,
             lang: config.default_language.clone(),
             slug: "tags".to_string(),
             path: "/tags/".to_string(),
             permalink: "https://vincent.is/tags/".to_string(),
-            items: vec![tag],
+            items: vec![tag, tag_accent],
         };
         let tags_fr = Taxonomy {
             kind: taxo_config_fr,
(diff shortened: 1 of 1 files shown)


Keats:
You need to do the PR on the `next` branch

SAY-5:
Retargeted to `next` branch. Thanks @keats!

Heads up: `next` already implements the slug fallback in `GetTaxonomyTerm::call`, so the rebased PR now contains only the regression test for accented terms (closes #2829).

components/templates/src/functions/taxonomy.rs
```diff
@@ -421,4 +421,47 @@ mod tests {
...
+
+        // Lookup by the slug works.
+        let kwargs =
+            Kwargs::from([("kind", Value::from("tags")), ("term", Value::from("accion"))]);
+        let ctx = Context::new();
+        let res = get_taxonomy_term.call(kwargs, &State::new(&ctx)).unwrap();
+        let res_obj = res.as_map().unwrap();
+        assert_eq!(res_obj.get(&"name".into()).unwrap().as_str().unwrap(), "acción");
+        assert_eq!(res_obj.get(&"slug".into()).unwrap().as_str().unwrap(), "accion");
+    }
```
Keats, line 437: i think this test can be integrated in the one above? All the setup is already done there

SAY-5:
Consolidated as suggested. Thanks @Keats!

Keats:
Thanks!

PR #3156, opened 2026-05-09
Improve file creation and removal detection, especially on Windows

knutaf:
A bug fix in the `notify` crate exposed an issue in the way Zola handles file changes, especially on Windows. `notify` was intending to be squashing `ModifyKind::Any` events following a `Create` event, and fixed that behavior with https://github.com/notify-rs/notify/pull/701. When [zola updated from `notify-debouncer-full` v0.5 to v0.6](https://github.com/getzola/zola/commit/95ba13adc85e4a79c75d7e21907d366cf2dc5f1a) it pulled in this bug fix and exposed this issue.

In `notify-debouncer-full` v0.5, the following sequence of events could get through:

```
DebouncedEvent { event: Event { kind: Create(Any), paths: ["c:\\test\\blah.txt"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { t: 1743965.5073826s } }
DebouncedEvent { event: Event { kind: Modify(Any), paths: ["c:\\test\\blah.txt"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { t: 1743965.5096777s } }
```

In v0.6 and onwards, you'll only see the `Create` event, due to `notify`'s deliberate squashing.

That's a problem for Zola, because it's looking for `Create(File)` and `Create(Folder)` events. It is *not* looking for `Create(Any)`. `notify` does not produce the more granular `Create` events on all platforms, notably Windows, which does not have a reliable way of determining the type of item that changed. Some editors, such as VIM, save files in a way that produces the Create+Modify sequence of events, which on Windows is reduced to just a `Create(Any)` and so is missed by Zola's event filtering.

This PR fixes a few related issues in this area, starting with which event types are considered relevant:
- `Create(Any)` and `Create(Other)` are produced by some platforms
- likewise `RemoveKind(Any)` and `RemoveKind(Other)`

The `match` statement that checks different event types now explicitly mentions all of the event variants, both ones that Zola cares about and ones it does not, so that upon `notify` crate update we'll get a compile error if any variants are added or removed.

`Remove` events were completely ignored, and are now being fixed. Previously, due to debouncing (i.e. delaying) handling of events, by the time we handle a `Remove` event, the file path has typically already been deleted, and so the code that was ignoring unknown file paths would ignore that event. Now Zola doesn't check for unknown paths on remove events.

I added debug logging for event handling under a new trace category. As an example, to enable it via environment variable while also keeping the program's normal output verbosity for other areas, you'd set `RUST_LOG=info,file_change=debug`.

### DIFF
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -36,27 +36,37 @@ pub type MeaningfulEvent = (PathBuf, PathBuf, SimpleFileSystemEventKind);
 /// return our internal simplified representation. For events we don't care about,
 /// return `None`.
 fn get_relevant_event_kind(event_kind: &EventKind) -> Option<SimpleFileSystemEventKind> {
+    // Prefer having match arms for all specific event types, so that on a notify crate upgrade, we're forced to
+    // evaluate any changes and decide what corresponding mapping changes should happen.
     match event_kind {
-        // Nova on macOS reports this as it's final event on change
+        // Nova on macOS reports this as its final event on change
         EventKind::Modify(ModifyKind::Name(RenameMode::Any)) => {
             Some(SimpleFileSystemEventKind::Modify)
         }
-        EventKind::Create(CreateKind::File) | EventKind::Create(CreateKind::Folder) => {
+        // Windows only reports `Any`. Some filesystems also report `Other` in some situations.
+        EventKind::Create(CreateKind::Any | CreateKind::File | CreateKind::Folder | CreateKind::Other) => {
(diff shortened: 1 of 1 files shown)


Keats:
LGTM, just needs cargo fmt

knutaf:
> LGTM, just needs cargo fmt

I pushed an update with cargo fmt. Could you approve so that it can run through checks? Thanks.

Keats:
Thanks!

Issue #2361, opened 2023-11-21
Feature request: build a non distroless docker image


ogarcia:
Having a distroless image is interesting for people who simply want to run zola directly, but if you want to use zola on a CI/CD system you need the image to contain a shell.

I think you could make two variants of the image, by default a distroless one and then another one based on Debian (to give an example).

You would have to modify the Dockerfile to make the from a variable:

<details>
<summary>Dockerfile</summary>

```Dockerfile
ARG DISTRO=gcr.io/distroless/cc

FROM rust:slim AS builder

RUN apt-get update -y && \
  apt-get install -y make g++ libssl-dev && \
  rustup target add x86_64-unknown-linux-gnu

WORKDIR /app
COPY . .

RUN cargo build --release --target x86_64-unknown-linux-gnu


FROM ${DISTRO}
COPY --from=builder /app/target/x86_64-unknown-linux-gnu/release/zola /bin/zola
ENTRYPOINT [ "/bin/zola" ]
```

</details>

And add these two additional steps in the CI:

<details>
<summary>CI steps</summary>

```yaml
      - name: Extract metadata (tags, labels) for Docker
        id: meta
        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
        with:
          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
          flavor: latest=false
          tags: |
            type=ref,enable=true,priority=600,prefix=debian-,suffix=,event=tag

      - name: Build and push Docker image
        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
        with:
          build-args: |
            DISTRO=debian:stable
          context: .
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
```

</details>

Keats:
Isn't it easier to download the binary from the releases in CI?

ogarcia:
> Isn't it easier to download the binary from the releases in CI?

That is what I am currently doing. But if an _official image_ with shell is available you can use the image cache to store it and therefore only download it once.

LouDou:
I agree with this suggestion; doing downloads in CI containers is unnecessary boilerplate IMO, it would be nice to use a prebuilt CI image with shell included.

EDIT: reference to gitlab not supporting shell-less CI images: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/26501

[...]

PR #3163, opened 2026-05-23
feat(container): restructure Containerfile with three build targets

ogarcia:
This change slightly modifies how the Zola container is created so that it supports two variants: one distroless and the other based on Alpine Linux closing #2361

Making this change means that Zola is compiled within the container itself in a separate thread instead of being downloaded from releases. This is done to simplify the overall process of creating containers, and since it is only built in releases, it is not a change that will place an excessive load on the system (only _recompiles_ amd64). It is also necessary because the arm64 target with `musl` does not compile in Release-build.

Could we explore doing it the other way? Yes, but IMHO this solution is better because it is self-contained and not dependent.

I didn't use the “next” branch because it's separate from the Zola code, and to make sure the build went smoothly, I decided to use a more stable branch.

That said, feel free to tell me to go to hell. :joy:


ogarcia:
One small detail I forgot to mention. If you want to see how it works, I ran a test using a hypothetical new version, v0.22.3. The GitHub action run is [here](https://github.com/ogarcia/zola/actions/runs/26332149935), and the result is [here](https://github.com/ogarcia/zola/pkgs/container/zola).

Keats:
Sorry for the late reply.  How does it work for https://github.com/ogarcia/zola/pkgs/container/zola/886791326?tag=alpine-amd64 ? Which version of Zola does it includes? Always the `latest`?

ogarcia:
This workflow is triggered **only when a new version is released**, and what it does exactly is compile Zola from the code in that tag and include it in both container flavors: one that is distroless and contains only the binary (which is static), and the other that uses Alpine (which is useful for CI).

Since it compiles both an amd64 and an arm64 version, that's why you have the tags `distroless-amd64`, `distroless-arm64`, `alpine-amd64` and `alpine-arm64` so you can put it all together:
- `distroless-amd64`, `distroless-arm64`: `vX.X.X`
- `alpine-amd64`, `alpine-arm64`: `vX.X.X-alpine`

These tags (`distroless-amd64`, `distroless-arm64`, `alpine-amd64`, `alpine-arm64`) do _move_ between versions and always points to the latest version, but that doesn’t affect what’s already been published. When you tag a specific version, even if you point to `distroless-amd64` and `distroless-arm64`, that’s done for simplicity; internally, they track the hashes and always point to them.

It doesn't publish a `latest` tag because, from what I've seen, you didn't want that tag. It only publishes the tag based on the `vX.X.X` version.

Containerfile
```diff
@@ -0,0 +1,51 @@
...
+COPY . .
+
+RUN cargo build --release && \
+    cp target/release/zola zola && \
+    ./zola --version
+
+FROM docker.io/alpine:3.21 AS musl-libs
+RUN mkdir -p /out && \
+    cp /lib/ld-musl-$(uname -m).so.1 /out/ && \
+    cp /lib/libc.musl-$(uname -m).so.1 /out/
```
Keats, line 20: Since we use musl now and use scratch does HTTPS request still work? (eg link checker)

ogarcia: It works perfectly because you're using `rustls`, which already embeds the certificates. You'd only need to include external certificates if we switched to OpenSSL, but that's not the case with musl 😉

Keats:
Thanks!

Keats:
So that does break the default image:

```
  
  thread 'main' (1) panicked at components/templates/src/functions/load_data.rs:246:18:
  reqwest client build: reqwest::Error { kind: Builder, source: General("No CA certificates were loaded from the system") }
```

I'll have to release 0.23.1 to fix it :(

ogarcia:
Whoops! That's really strange because with `rusttls`, the certificates are supposed to be embedded in the binary. In fact, I remember trying it, and it worked fine.

Well, it's clear that no matter how many times you test something, the move to production always throws you for a loop...

PR #3164, opened 2026-05-24
docs: Fix docs for higher / lower when sorting by date

timdcrain:
The documentation (and some doc comments) state that `lower` corresponds to earlier pages and `higher` to later pages when sorting by date. This is the opposite of what the code actually does.

Sanity check:


Keats:
Thanks! Not sure how that managed to stay like that so long

PR #3183, opened 2026-07-06
fix(content): slugify ancestor directory components in page paths

### LINKED ISSUE #3114: Orphaned pages appear not to have paths slugified
In my website, there is one root `_index`, and there are many subdirectories of `content/` underneath that.
Each of these subdirectories has an `index.md` and a `changelog.md`.

What I've found is that `slugify.paths` set to `on`, will slugify only the path of the `index.md`, and not the `changelog.md`. So if the subdirectory name is, for example, `chim_climbing`, the `index.md` becomes `index.html` in `public/chim-climbing`, and the changelog becomes `index.html` in `public/chim_climbing`, the unmodified directory name prior to slugification.

I glanced over the source for a few moments and `page.rs` doesn't obviously appear as though it could be missing slugifying anything, but, it struck me as really odd and interesting that there's an explicit branch for when the page name is `index`. I think the other branches fail somehow and don't get slugified at all. 

Please lemme know if I can help somehow! I'd like to have my URLs all slugified, but, since this has broken my changelog pages, for now I've just set `slugify.paths = off`.

Pablosinyores:
## Summary

With `slugify.paths` enabled, only part of a page's URL was slugified. A page's own slug is slugified, but the ancestor directory names (`file.components`) were joined into the path verbatim.

The effect, from #3114: in a directory `chim_climbing/` containing `index.md` and `changelog.md`, the colocated `index.md` (whose directory name becomes its slug) renders at `/chim-climbing/`, but the sibling `changelog.md` renders at `/chim_climbing/changelog/` — the parent directory keeps its raw, un-slugified name. The same applies to any non-slug-safe ancestor directory (including deeper ancestors of an `index.md`).

## Fix

Slugify each ancestor component with the configured strategy when building the page path, matching how the slug itself is already handled. `SlugifyStrategy::Off` returns names unchanged, so behaviour is unchanged when path slugification is disabled.

## Testing

Added `can_slugify_ancestor_directories_in_path` in `components/content/src/page.rs`: a non-index page under `content/chim_climbing/` with `slugify.paths = On` now resolves to `/chim-climbing/changelog/`. Verified it fails on the previous code (`/chim_climbing/changelog/`) and passes with the fix. Full `content` crate suite green (146 tests), `cargo fmt` and `clippy` clean.

Closes #3114

### DIFF
--- a/components/content/src/page.rs
+++ b/components/content/src/page.rs
@@ -148,7 +148,19 @@ impl Page {
                     page.slug.clone()
                 }
             } else {
-                format!("{}/{}", page.file.components.join("/"), page.slug)
+                // Ancestor directory names must be slugified the same way as the
+                // slug itself. Otherwise a non-index page keeps its raw parent
+                // directory name (e.g. `chim_climbing/changelog`) while the sibling
+                // `index.md`, whose directory becomes the slug, is slugified
+                // (`chim-climbing`). See #3114.
+                let components = page
+                    .file
+                    .components
+                    .iter()
+                    .map(|c| slugify_paths(c, config.slugify.paths))
+                    .collect::<Vec<_>>()
+                    .join("/");
+                format!("{}/{}", components, page.slug)
             };
 
             if page.lang != config.default_language {
@@ -287,6 +299,29 @@ Hello world"#;
         assert_eq!(page.path, "/posts/intro/hello-world/");
         assert_eq!(page.components, vec!["posts", "intro", "hello-world"]);
         assert_eq!(page.permalink, "http://hello.com/posts/intro/hello-world/");
+    }
+
+    #[test]
+    fn can_slugify_ancestor_directories_in_path() {
+        // A non-index page in a directory whose name is not slug-safe must have
+        // that ancestor directory slugified, matching how the sibling `index.md`
+        // (whose directory becomes the slug) is already handled. See #3114.
+        let content = r#"
+    +++
+    +++
+    Hello world"#;
+        let mut config = Config::default();
+        config.slugify.paths = SlugifyStrategy::On;
+        let res = Page::parse(
+            Path::new("content/chim_climbing/changelog.md"),
+            content,
+            &config,
+            &PathBuf::new(),
+        );
+        assert!(res.is_ok());
+        let page = res.unwrap();
+        assert_eq!(page.path, "/chim-climbing/changelog/");
+        assert_eq!(page.components, vec!["chim-climbing", "changelog"]);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


components/content/src/page.rs
```diff
@@ -148,7 +148,19 @@ impl Page {
                     page.slug.clone()
                 }
             } else {
-                format!("{}/{}", page.file.components.join("/"), page.slug)
+                // Ancestor directory names must be slugified the same way as the
+                // slug itself. Otherwise a non-index page keeps its raw parent
+                // directory name (e.g. `chim_climbing/changelog`) while the sibling
+                // `index.md`, whose directory becomes the slug, is slugified
+                // (`chim-climbing`). See #3114.
```
Keats: no need for the full comment, just point to the GH issue

components/content/src/page.rs
```diff
@@ -289,6 +301,29 @@ Hello world"#;
         assert_eq!(page.permalink, "http://hello.com/posts/intro/hello-world/");
     }
 
+    #[test]
+    fn can_slugify_ancestor_directories_in_path() {
+        // A non-index page in a directory whose name is not slug-safe must have
+        // that ancestor directory slugified, matching how the sibling `index.md`
+        // (whose directory becomes the slug) is already handled. See #3114.
+        let content = r#"
```
Keats, line 303: same as above, just point to the gh issue, the test name + content is enough

Keats (changes requested):

Pablosinyores:
Trimmed both down to just reference #3114.

Keats:
Re-reading the code, I'm fairly sure this will potentially break a lot of sites since that's changing the slugification of components which hasn't been done. Lots of 404 with that patch I think?

Pablosinyores:
You're right - verified it. With `slugify.paths = "on"`, a section `blog_posts/` containing a page `my_page.md`:

- before: page at `/blog_posts/my-page/` (sits under its section `/blog_posts/`)
- with this patch: page at `/blog-posts/my-page/`, but the section stays `/blog_posts/` - so the page no longer sits under its own section. That's the 404s.

The actual bug is narrower - orphaned pages only (#3114 is a dir with `index.md` + `changelog.md` and no `_index.md`): the index becomes the slug so it's slugified, but the sibling's parent component isn't. Sections build their path separately and don't slugify components at all, so touching the shared page-path code breaks the section case.

Do you want it scoped to orphaned pages only, or sections + pages slugifying components consistently (which would change existing section URLs)? I'll redo it either way.

Keats:
I think it's a WONTFIX?

Pablosinyores:
Agreed, WONTFIX makes sense. Sections don't slugify their path components at all, so fixing only the page side moves pages out from under their own section - and scoping it to orphaned pages isn't clean in page.rs, since a page doesn't know at parse time whether its ancestor dirs are sections. Not worth the breakage for the narrow case. Closing this; #3114 stays open if anyone wants to do both sides together.

PR #3184, opened 2026-07-06
fix(serve): rebuild the whole site when anchor-link.html changes

### LINKED ISSUE #2940: Live-reload doesn't handle editing anchor-link.html
# Bug Report

## Environment

Zola version: `0.21.0`

## Expected Behavior

When updating `templates/anchor-link.html` while `zola serve` is running, my Web page should reflect the updated template.

## Current Behavior

While `zola serve` is running, editing `templates/anchor-link.html` (while having `insert_anchor_links` set in `config.toml` doesn't get reflected into the live server, despite Zola actually noticing that the template was recompiled. Killing and restarting it does rebuild properly. Reproducible with and without the `--fast` flag.

## Step to reproduce

1. Run `zola serve`
2. Edit `templates/anchor-link.html`
3. Verify that the change was indeed taken into consideration by Zola:
    ```
    Change detected @ 2025-07-27 21:33:05
    -> Template file(s) changed /home/ccjmne/git/blog/templates/anchor-link.html
    Reloading only template
    Done in 4ms.
    ```
4. Your page reloads but looks the same
5. Ctrl-C kill `zola serve`
6. Restart `zola serve`
7. Refresh your page and see your upgraded `anchor-link.html` template in action

Pablosinyores:
## Summary

Editing `templates/anchor-link.html` while `zola serve` is running does not show up in the browser until the server is restarted, even though Zola logs that it reloaded the template (#2940).

The reason, as noted by @Keats on the issue: reloading only recompiles the Tera template registry, but `anchor-link.html` is consumed during **Markdown rendering** (heading anchors), which is baked into the already-rendered page HTML. A template-registry reload doesn't re-render the Markdown, so the live pages stay stale.

## Fix

In the `serve` watcher's `Templates` change handler, detect whether any changed template is one that affects Markdown rendering (currently `anchor-link.html`) and trigger a full site rebuild for it, instead of the template-only reload. All other template changes keep the existing reload-only fast path.

The classification is a small pure helper, `template_change_requires_full_rebuild`, so it's unit-tested directly.

## Testing

Added `anchor_link_template_requires_full_rebuild` in `src/fs_utils.rs` (anchor-link.html -> full rebuild; ordinary templates -> reload-only). `cargo test -p zola fs_utils` green, `fmt` and `clippy` clean. Also verified manually: with `insert_anchor_links` set, editing `anchor-link.html` under `zola serve` now reflects in the served page without a restart.

Closes #2940

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -69,6 +69,9 @@ enum WatchMode {
 const METHOD_NOT_ALLOWED_TEXT: &[u8] = b"Method Not Allowed";
 const NOT_FOUND_TEXT: &[u8] = b"Not Found";
 
+// Templates used while rendering Markdown: reloading the registry isn't enough. See #2940.
+const ALWAYS_FULL_REBUILD: &[&str] = &["anchor-link.html"];
+
 // This is dist/livereload.min.js from the LiveReload.js v3.2.4 release
 const LIVE_RELOAD: &str = include_str!("livereload.js");
 
@@ -827,10 +830,6 @@ pub fn serve(
                                 .join(", ");
                             log::info!("-> Template file(s) changed {combined_paths}");
 
-                            // These are used while rendering Markdown, so a
-                            // registry reload isn't enough - the pages are
-                            // already rendered and need a full rebuild. See #2940.
-                            const ALWAYS_FULL_REBUILD: &[&str] = &["anchor-link.html"];
                             let needs_full_rebuild = full_paths.iter().any(|p| {
                                 p.file_name()
                                     .and_then(|n| n.to_str())
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -56,9 +56,7 @@ use site::sass::compile_sass;
 use site::{BuildMode, SITE_CONTENT, Site};
 use utils::fs::{clean_site_output_folder, copy_file, create_directory};
 
-use crate::fs_utils::{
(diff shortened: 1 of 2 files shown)


Keats (commented):
That's too complex, you can just have an array of ALWAYS_FULL_REBUILD and inline everything. No need for tests either

Pablosinyores:
Done - inlined it as an `ALWAYS_FULL_REBUILD` array in the watcher and dropped the helper and its test.

src/cmd/serve.rs
```diff
@@ -827,8 +827,24 @@ pub fn serve(
                                 .join(", ");
                             log::info!("-> Template file(s) changed {combined_paths}");
 
-                            log::info!("Reloading only template");
-                            reload_templates(&mut site)
+                            // These are used while rendering Markdown, so a
+                            // registry reload isn't enough - the pages are
+                            // already rendered and need a full rebuild. See #2940.
+                            const ALWAYS_FULL_REBUILD: &[&str] = &["anchor-link.html"];
```
Keats: move that at the top of the file so it's easy to see

Keats (changes requested):

Pablosinyores:
Moved to the top with the other consts. Re-checked the behaviour after the move: editing anchor-link.html under `zola serve` still logs "Rebuilding the site" and the served page picks up the change, and an ordinary template edit still logs "Reloading only template".

PR #3190, opened 2026-07-09
Adds C compiler requirement to "From source" instructions.

### LINKED ISSUE #3186: User needs a C Compiler available on system to build from source.
# Documentation issue

## Summary

Pretty straight-forward really: the user must have a C compiler on their system before they're able to build Zola from source. This isn't a problem for most OSes, but some minimal installs (like Ubuntu server, basic NixOS, AlpineOS, etc) don't come with a C compiler pre-packaged.

## Proposed solution

Add something like this to the _Build from Source_ instructions:

```
To build Zola from source, you will need: 

- [Rust and Cargo](https://www.rust-lang.org/).
- Any C compiler
```

I'm not sure if _any_ C compiler will actually do, or if the compiler needs specific functions/features/whatever. I just know that the one that comes with `build-essential` on Debian/Ubuntu does the trick.

Happy to throw a PR together for this. Just wanted to raise it and make sure it's _actually_ an issue worth fixing before I dump a new PR in your lap.

johnnymatthews:
Closes #3186.

Sanity check:

### DIFF
--- a/docs/content/documentation/getting-started/installation.md
+++ b/docs/content/documentation/getting-started/installation.md
@@ -232,8 +232,10 @@ Zola does not work in PowerShell ISE.
 
 ## From source
 
-To build Zola from source, you will need to have [Rust and Cargo](https://www.rust-lang.org/)
-installed.
+To build Zola from source, you will need: 
+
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

PR #3194, opened 2026-07-13
docs: note hard_link_static also applies to colocated assets

### LINKED ISSUE #3176: Avoid copying colocated assets in each locale
I find convenient to place images in a blog post's folder. However the images are copied in the filesystem in each locale's folder, wasting some space in both the storage and the precious filesystem memory cache.

There could be an option to prevent this by using symbolic links instead of copies. An option because maybe some filesystems or webservers don't like symbolic links, or people may have reasons not to want links.

If the page is available in the default locale, then the file is stored there. Else, another locale may be picked at random, I guess.

Pablosinyores:
Per your call on #3176 ("let's just update the docs").

`hard_link_static` already covers colocated page/section assets, not just the `static` directory (`copy_assets` passes it through to `copy_file_if_needed`), but the docs only mentioned `static`. This notes it in the Asset colocation section and broadens the config option description, so someone hitting the "assets duplicated across localized outputs" case from #3176 knows `hard_link_static = true` deduplicates them.

Docs only, on `next`. Closes #3176.

### DIFF
--- a/docs/content/documentation/content/overview.md
+++ b/docs/content/documentation/content/overview.md
@@ -44,6 +44,12 @@ assets, such as images or spreadsheets. Zola supports this pattern out of the bo
 All non-Markdown files you add in a page/section directory will be copied alongside the generated page when the site is
 built, which allows us to use a relative path to access them.
 
+By default these files are copied, so an asset shared across several outputs (for example the same image in multiple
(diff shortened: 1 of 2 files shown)


Keats (approved):

Keats:
Thanks!

PR #3197, opened 2026-07-19
Image filter

Keats:


components/imageproc/src/helpers.rs
```diff
@@ -152,11 +153,16 @@ pub fn get_processed_filename(
...
+    filter: FilterType,
 ) -> String {
     let mut hasher = DefaultHasher::new();
     hasher.write(input_src.as_ref());
     op.hash(&mut hasher);
     format.hash(&mut hasher);
+    // Only hash the filter if it's not the default.
+    if filter != FilterType::Lanczos3 {
+        filter.hash(&mut hasher);
+    }
```
lpulley, line 165: This seems questionable to me... if the default ever changes, won't the existing outputs be quietly kept, despite having a non-default filter? Surely it's cheap to just hash the filter type always?

Keats: It was to avoid changing the existent processed images

lpulley: Right... yeah. 👍

PR #3202, opened 2026-07-22
Prevent template file lookups outside authorized roots

danielcadev:
Sanity check:

## Code changes

## Summary

- Validate every resolved template file candidate against its authorized root.
- Preserve lookups from the site, `static`, `content`, theme, and external output directories.
- Add regression coverage for traversal through fallback search paths.

## Root cause

`search_for_file` enforced directory containment only when the initial candidate under the site root existed. When that candidate did not exist, fallback candidates under `static`, `content`, the output directory, or theme assets were returned without the same validation.

A path containing parent-directory components could therefore resolve through a fallback root to an existing file outside the authorized directory. Functions such as `load_data` would then read that file with the permissions of the Zola process.

This is the same vulnerability class previously addressed for `zola serve` in #2258, but in the template file lookup path.

## Fix

Associate every search directory with its authorized root and validate each existing candidate before returning it. The output directory remains independently authorized so absolute external `--output-dir` configurations continue to work.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p templates --offline --locked` — 86 passed
- `cargo test --all --offline --locked`
- `cargo build --all --offline --locked`

Regression tests cover traversal through an alternate search path, ordinary static-file resolution, and an external output directory.

### DIFF
--- a/components/templates/src/helpers.rs
+++ b/components/templates/src/helpers.rs
@@ -22,10 +22,16 @@ pub fn search_for_file(
     theme: &Option<String>,
     output_path: &Path,
 ) -> Result<Option<(PathBuf, String)>> {
-    let mut search_paths =
-        vec![base_path.join("static"), base_path.join("content"), base_path.join(output_path)];
+    let output_search_path = base_path.join(output_path);
+    let mut search_paths = vec![
+        (base_path.to_path_buf(), base_path.to_path_buf()),
+        (base_path.join("static"), base_path.to_path_buf()),
+        (base_path.join("content"), base_path.to_path_buf()),
+        (output_search_path.clone(), output_search_path),
+    ];
     if let Some(t) = theme {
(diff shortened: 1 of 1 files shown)


Keats:
It's more likely to break existing workflows than solving a security issue.

danielcadev:
Thanks. I verified the compatibility concern against the official v0.22.1 release. External symlinked asset directories are a real existing workflow, so I agree that the current patch is too broad.

There is still a confinement risk when a third-party theme or template can influence the file lookup, because fallback candidates may resolve outside the site root. This also appears inconsistent with the documentation stating that paths outside the Zola directory produce an error.

Addressing that safely would require a compatibility-aware design, such as an opt-in strict mode, rather than changing the default globally. I’m fine leaving this PR closed. Thanks for the review.

PR #3203, opened 2026-07-25
Document code fence filenames

### LINKED ISSUE #2867: Document setting a name for code blocks
# Documentation issue

## Summary
In https://www.getzola.org/documentation/content/syntax-highlighting/#annotations, it explains that I can use a `name` annotation "to specify a name the code block is associated with". However when doing this, superficially nothing happens.

When looking at the generated html, all it does it add a `data-name="filename"` tag to the affected HTML elements. These need to be processed with CSS, but I can't find any instructions on how to do this, nor does any of the themes I've tried out seem to support it yet (which is understandable given the novelty of the feature).

## Proposed solution
Explain somewhere how these tags work, with some basic CSS examples on how to use them in themes, or maybe even make it work out of the box by default somehow.

ColumbusLabs:
## Summary

Document how the `name` code-fence attribute becomes `data-name` on the
generated `<code>` element, and provide a small CSS example for displaying the
filename.

Fixes #2867.

Sanity check:

## Validation

- built the documentation site with the current Zola binary
- confirmed the generated HTML contains the documented `data-name` markup and CSS
- `git diff --check origin/master...HEAD`

Assisted by OpenAI Codex; the source, generated markup, and documentation build were reviewed and validated by me.

### DIFF
--- a/docs/content/documentation/content/syntax-highlighting.md
+++ b/docs/content/documentation/content/syntax-highlighting.md
@@ -146,6 +146,33 @@ highlight(code);
 ```
 ````
 
+Zola adds the name to the generated `<code>` element as a `data-name`
+attribute. The example above produces HTML with the following structure
+(highlighting markup and styles are omitted):
+
+```html
+<pre class="giallo">
+  <code data-lang="rust" data-name="mod.rs">...</code>
+</pre>
+```
+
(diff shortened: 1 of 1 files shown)


Keats:
Fixed in a commit in the next branch to take into account the latest giallo features. Thanks though!

PR #3207, opened 2026-07-28
Fix all the tera docs links by removing docs from the URL

tristanphease:
Hello, was browsing the docs and got some 404 links when going to https://keats.github.io/tera/docs#built-in-functions and realised the correct link was the one without the docs on the end. e.g. https://keats.github.io/tera/#built-in-functions

This PR updates them to correct values.
Note that the macros one doesn't find the anchor link on the docs, I think that one is changed to component, i.e. https://keats.github.io/tera/#components but I don't know enough about it to update it so just changing the link for now.
Also the templates one doesn't find the anchor link, not sure what link this one should resolve to.
The other ones all link correctly.

Sanity check:


Keats:
Thanks!

PR #3215, opened 2026-08-05
0.23: site feed fails - needs to be migrated to Tera v2

### LINKED ISSUE #3214: 0.23.1: rss feed generation fails
# Bug Report

## Environment

Zola version: 0.23.1

## Expected Behavior
Site-level rss.xml builds correctly.

## Current Behavior
`zola build` fails:
```
ERROR Failed to render feed 'rss.xml' --> __zola_builtins/rss.xml:6:18 Variable section is not defined
```
This is because `{% if section.title %}` is a strict error in Tera v2 if `section` is not defined (it was falsy in v1).

## Step to reproduce
1. `generate_feeds = true` in `config.toml`
2. `feed_filenames = ["rss.xml"]` in `config.toml`
3. Run `zola build`

Avanta8:
# Bug Report

## Environment

Zola version: 0.23.1

## Expected Behavior
Site-level rss.xml builds correctly.

## Current Behavior
`zola build` fails:
```
ERROR Failed to render feed 'rss.xml' --> __zola_builtins/rss.xml:6:18 Variable section is not defined
```
This is because `{% if section.title %}` is a strict error in Tera v2 if `section` is not defined (it was falsy in v1).
https://github.com/Keats/tera/blob/master/MIGRATION.md#behaviour-changes.

## Step to reproduce
1. `generate_feeds = true` in `config.toml`
2. `feed_filenames = ["rss.xml"]` in `config.toml`
3. Run `zola build`

### DIFF
--- a/components/templates/src/builtins/rss.xml
+++ b/components/templates/src/builtins/rss.xml
@@ -3,7 +3,7 @@
     <channel>
       <title>{{ config.title }}
         {%- if term %} - {{ term.name }}
-        {%- elif section.title %} - {{ section.title }}
+        {%- elif section and section.title %} - {{ section.title }}
         {%- endif -%}
       </title>
       <link>


Keats:
> This is because {% if section %} is a strict error in Tera v2 (it was falsy in v1): https://github.com/Keats/tera/blob/master/MIGRATION.md#behaviour-changes.

Nope that's not an error in v2 either

components/templates/src/builtins/rss.xml
```diff
@@ -2,12 +2,12 @@
 <rss xmlns:atom="http://www.w3.org/2005/Atom" version="2.0">
     <channel>
       <title>{{ config.title }}
-        {%- if term %} - {{ term.name }}
-        {%- elif section.title %} - {{ section.title }}
```
Keats, line 6: this line is the issue, it needs `{%- elif section and section.title %}` like atom.xml
Only this line needs to be changed

Keats (dismissed):

components/templates/src/builtins/rss.xml
Avanta8: Ah yes, my mistake - thanks for pointing it out. Have updated the PR now.

Keats:
Hm can hard reset the branch to next and add your change? Looks like some conflict there

Or maybe just squash your commits

Avanta8:
> Hm can hard reset the branch to next and add your change? Looks like some conflict there
> 
> Or maybe just squash your commits

Yep, updated it now!

Keats:
Thanks!

PR #3218, opened 2026-08-06
docs: fix incorrectly escaped code blocks in docs

spoonmilk:
Hi folks,

This resolves an issue I saw in the docs where code blocks weren't being correctly escaped due to {%- endraw -%} stripping off the newline before the closing ``` fence. The fix here was to replace it with {%- endraw %}, which doesn't strip the newline.

Fixed using grep/sed. Verified manually (using my eyes). 

N/A currently open.

Yes! All of them with the old endraw.


Keats:
Thanks!

Issue #3226, opened 2026-08-09
Panic when rendering body for section with subsections


beanaroo:
# Bug Report

I'm trying Zola for the first time. Porting existing docs over to a fresh site.

Trying to add components to the root page and experiencing a panic when including any kind of templating `{{`/`{%` in the top `_index.html` section body. Subsection bodies (with no additional subsections) are able to render just fine.

```plain
content/
├── _index.md
├── framework/
│   └── _index.md
└── reference/
    ├── _index.md	
    ├── google/
    │   └── _index.md
    └── azure/
        └── _index.md
```

## Environment

Zola version: 0.23.2

## Expected Behavior
Rendering section successfully

## Current Behavior

`RUST_BACKTRACE=full zola build` panics with:

```
thread '<unnamed>' panicked at components/content/src/ser.rs:165:38:
no entry found for key
stack backtrace:
   0:        0x101c8be10 - __mh_execute_header
   1:        0x100fec4a0 - __mh_execute_header
   2:        0x101c8a578 - __mh_execute_header
   3:        0x101c8a1c8 - __mh_execute_header
   4:        0x101ca92dc - __mh_execute_header
   5:        0x101ca9298 - __mh_execute_header
   6:        0x101ca97b8 - __mh_execute_header
   7:        0x101fe24ac - __mh_execute_header
   8:        0x101fe2704 - __mh_execute_header
   9:        0x100fde808 - __mh_execute_header
  10:        0x101bebe88 - __mh_execute_header
  11:        0x101bf0d78 - __mh_execute_header
  12:        0x101bf16d0 - __mh_execute_header
  13:        0x10203ebf0 - __mh_execute_header
  14:        0x101a32510 - __mh_execute_header
  15:        0x101a36050 - __mh_execute_header
  16:        0x101caa340 - __mh_execute_header
  17:        0x18c229c58 - __pthread_cond_wait
```

## Step to reproduce

**LLM assisted**: https://github.com/beanaroo/zola-subsection-panic

inklesspen:
I'm also experiencing this issue (which is wonderfully opaque; I'm very thankful to beanaroo for producing an informative report).

PR #3227, opened 2026-08-09
fix: panic when templating a section body that has subsections

beanaroo:
Fixes  https://github.com/getzola/zola/issues/3226

Sanity check:

## Code changes

* [x] Was a LLM used?

> [!NOTE]
> I am very new to Rust. While I've tried my best to understand the problem and the fix, two separate LLMs were used to identify, address, review and test the issue. The below description is me rewriting the LLM's explanation as I understand it.

---

### Root cause

`SerializingSection::new` looked up each entry of `section.subsections` with `library.sections[p]`. On a full build those lookups always miss:

1. `populate_sections` fills in `section.subsections` with real paths.
2. `render_markdown` then templates every section body in parallel. To do that it takes `Arc::make_mut(&mut self.library)`, so it cannot also lend the real library to the renderer, and passes a `Library::default()` instead.

So the paths are present but the map they point into is empty, and the index panics with `no entry found for key`. Any section body that trips `needs_templating` and has at least one subsection hits this.

The `serve --fast` path did not panic, because `add_section` rendered through `self.renderer()`, which has the real, fully populated library.

### What this does

Two changes:

- `SerializingSection::new` uses `.get()` in a `filter_map`, so a subsection that cannot be resolved is skipped rather than fatal. This is how the neighbouring fields already behave when the library is not queryable: `pages` arrives as `Vec::new()`, and `RenderCache` already used `filter_map` for the same lookup.
- All content templating now shares one always-empty `CONTENT_LIBRARY`. `render_markdown` uses it in place of its local dummy, and `renderer()` becomes `content_renderer()`, which no longer hands over the real library.
  - Its only two callers, `add_page` and `add_section`, are content templating, and `Renderer.library` is read only by `render_page_content` and `render_section_content`, so template rendering is untouched while build and serve now produce the same output.
  - The alternative, giving `render_markdown` the real library, would mean cloning the whole library on every build just to satisfy the borrow the parallel loop needs.

### Tests

Both were confirmed to fail without the fix:

- `can_render_section_with_subsections` covers the panic.
- `content_templating_never_sees_the_library` builds the real `test_site`, so the library is fully populated, then re-renders a section through `add_section(.., true)` as `serve --fast` does, and asserts that a loop over `section.subsections` in the body produces nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.23.3 (unreleased)
 
+- Fix panic when templating a section body that has subsections
+
 ## 0.23.2 (2026-08-07)
 
 - Fix dual class CSS generation
--- a/components/content/src/ser.rs
+++ b/components/content/src/ser.rs
@@ -162,7 +162,7 @@ impl<'a> SerializingSection<'a> {
         let subsections: Vec<&str> = section
             .subsections
             .iter()
-            .map(|p| library.sections[p].file.relative.as_str())
+            .filter_map(|p| library.sections.get(p).map(|s| s.file.relative.as_str()))
             .collect();
         let backlinks = find_backlinks(&section.file.relative, library);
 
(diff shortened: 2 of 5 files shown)


Keats:
Thanks, the LLM was on the right path but got a bit too eager trying to fix an edge case that isn't super important. I've pushed a simpler fix on the next branch.

beanaroo:
Thank you @Keats. Learning a lot!

Issue #3204, opened 2026-07-26
feature request: Translated pages reuse default metadata


bemyak:
I'm using Zola to help a creator to maintain a website. Currently is has 868 pages in two languages. One constant source of frustration and bugs is that the entire markdown preface have to be repeated for each language. An example:

```markdown
+++
title = "Summer Flame"
weight = 456

[extra]
cover = "summerflame.png"
faction = "desert"
flavor = "A spark of a cheerful fire that fills the world with mysticism and light."
types = ["spirit", "warden"]

[extra.abilities]
"Summer night" = "lighter than day."

[taxonomies]
tags = ["summer"]
related = ["mustikka", "flowerbug", "tuli", "seer", "jinni", "salttree", "loimuri"]
+++
```

A translated page needs an overwrite just for `title`, `flavor` and `abilities` fields; the rest is always identical.

What do you think if there would be per-section (or global as well?) setting `inherit_metadata`, which would make translated pages inherit all preface fields from the default language page?

I believe this will be useful to all Zola users, since at allows omitting fields that are commonly repeated in translated pages, e.g. `weight`, `date`, `slug`, etc.

I can work on the implementation if approved.

bemyak:
@Keats, friendly bump on this FR 😊

Keats:
Hey sorry I had the issue open but didn't reply yet. 
I think it's fine to add, although I don't know exactly where to stop inheriting things? eg do we copy all data/taxonomies by default?

bemyak:
Well, I can only speak from the perspective on my own Zola-based projects (personal blog and the aforementioned creator's website), but taxonomies should also be included.

For example, if I have an English blog post with tags `how-to, diy, guide`, I wouldn't want to duplicate it for another language. So if I later decide to add a new one to that list, the translated page will also be updated, which is desirable. If there's some tag, which needs translation, then the entire list should be duplicated in the translation. Seems reasonable 🤷 

I will start working on this, let's see if it turns out good or not.

PR #3239, opened 2026-08-11
Add inherit_metadata setting for translated pages

bemyak:
This PR implements https://github.com/getzola/zola/issues/3204

There were two ways of doing this:
1. Implement merging for already parsed pages. This would require making fields of `PageFrontMatter` into Optionals, which users can observe in serialized template. Also, the merge function must be updated every time a new field is added to PageFrontMatter. Empty arrays become indistinguishable from absent (e.g. for `authors = []`). Meh.
2. Merge the front matter after TOML parsing, but before being parsed as a page. This is much less impactful way, which I chose here. The drawback is that pages now need to carry the raw frontmatter.

The performance impact for an average size frontmatter (~10 fields) seems to be 45ms for 500 pages, so <1ms per page, part of which is newly rendered taxonomy pages. This is paid only when the feature is enabled, no change by default.

<details>
<summary>Sanity check</summary>

## Code changes

* [X] Was a LLM used? If so, for what? Yes, for writing tests, performing benchmarks and pondering ideas about feature integration, help with iterating over the code. I understand the submitted changes.

</details>


Keats:
I think 1 is better?

> Also, the merge function must be updated every time a new field is added to PageFrontMatter.

Make it a destructing pattern without `..` and skipping it would be a compile error

> Empty arrays become indistinguishable from absent (e.g. for authors = []). Meh

Make those Option so we can distinguish it

> which users can observe in serialized template

We wouldn't change what is serialized, it's another struct.

PR #3249, opened 2026-08-16
Update the Debian install docs to the pkg.haus APT archive

barnumbirr:
The barnumbirr/zola-debian per-release .deb flow this section documented has been superseded: the packaging moved into the pkghaus org and releases now publish to the [pkg.haus](https://pkg.haus) APT archive (Debian stable, testing and unstable, amd64/arm64, built from source at release tags, signed). This updates the section to the current install path.


Keats:
Thanks!

PR #3252, opened 2026-08-17
docs: add harvis deployment guide

fordnox:
Adds a deployment guide for [harvis](https://harvis.dev), following the one-file-per-provider pattern in `docs/content/documentation/deployment/`.

harvis is zero-setup static hosting: `npx harvis public` after `zola build` uploads the output and prints a live URL — no account or config. Free tier for small sites.

Disclosure: I'm the maintainer of harvis.

### DIFF
--- a/docs/content/documentation/deployment/harvis.md
+++ b/docs/content/documentation/deployment/harvis.md
@@ -1,0 +1,22 @@
++++
+title = "harvis"
+weight = 70
++++
+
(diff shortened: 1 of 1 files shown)


Keats:
Thanks for the PR but let's not add every single host/personal startup out there in the docs

fordnox:
thanks for feedback

PR #3258, opened 2026-08-21
[RFC] Merging page and section

Keats:


Keats:
cc @bemyak  cc @Hugo-Trentesaux

rfcs/text/001-page-section.md
```diff
@@ -0,0 +1,152 @@
...
+https://gohugo.io/content-management/page-bundles/
+https://gohugo.io/content-management/page-resources/
+
+
+
+## Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- I like the `_index.md` name since it means the files ends up at top of folder usually, some people don't and
+prefer `index.md` but that conflicts with colocated assets
```
bemyak: I've been thinking about this just recently. What do you think if we allow both `index.md` and `_index.md` to be used interchangeably and error out if both exist? This way users can always use `_index.md`. It's pretty handy if you have a lot of assets attached to your page, makes finding the page easier.

Keats: What would be the default template for `index.md` then? The underscore makes it clear

jdobonb: I am completely in favor of mandatory underscore prefixes for files that have special meaning for Zola.

I think it leads to more consistency with the SCSS naming convention (where the underscore also has a special meaning) and it can feel familiar for some JS frameworks users.

As a side note, I would love to see default templates like `page.html`to also be prefixed. I see myself putting all my templating inside a `templates/custom` directory to ensure no collision with the well-known ones, but I guess it's an unreasonable breaking change.

bemyak: > What would be the default template for `index.md` then

I think it should be `page.html` for both. It will require one-line fix for established users, but it will be much less confusing for the newcomers, which I believe is more important in the long run. By confusion I mean it'll be difficult to explain "yeah, we have only pages and sub-pages, BUT if you create this file with this special name here, then the notion of sections suddenly pops-up, because it's how things used to be back in the day" :shrug: 

But it's your call of course, I'll be fine with it either way.

Keats: I'll keep the page and section concept. Hugo is using that as well (+ same page/section.html default template) so it should be familiar to the majority + no changes from before Zola.
I'll update the doc a bit but I would still keep the section name in the docs

Keats:
Text updated. I'm fairly sure about keeping the _index.md and page.html/section.html at this point.

rfcs/text/001-page-section.md
```diff
@@ -0,0 +1,167 @@
...
+We keep the current `ser.rs` file to ensure the `page`/`section` variables stay the same.
+We keep an enum of `Kind` to differentiate between regular pages and section/collection pages.
+If we allow slugs on sections, we do need to ensure all the descendant permalinks are generated correctly.
+
+The `Library` will merge page and sections and we need to ensure `transparent` and the new equivalent of `populate_sections` still work.
+There should be no regression for parallelization of rendering or for `zola serve`.
+
+We do not want a single `item.children` property instead of having both `.{pages,subsections}` because sorting pages
+and subsection together does not really make sense currently, `sort_by` would not apply to sections.
+The `$NAME.{pages,subsections}` fields become a `Vec<Page>`, not a `Vec<&'a str>`: only for the new variables, not for the
```
Keats, line 102: The Vec<Page> part I need to check if it can be an issue perf wise or logic wise.

wakamex:
I ran into one case that seems relevant to this RFC while building content-derived navigation for a Zola docs theme. A section can contain ordinary pages and child sections in one authored order, but templates receive them as separate `pages` and `subsections` lists. Removing arbitrary distinctions between nodes sounds like the most general solution.

Why not go further down this path, in this RFC or a follow-up, and make every content node a `Page`? `self.pages` could contain every direct child `Page`, including directory pages, with `kind` distinguishing something like `Leaf` and `Branch`. A subsection would then be a relationship, not a
separate kind or collection. Exposing the existing `weight` on the common `Page` object would give themes one native value for ordering nodes of any kind, with no special handling.

The same cleanup seems possible for `index.md` and `_index.md`. If every content node is a `Page`, `index.md` could represent the page for a directory whether it has child pages, colocated assets, or both. `_index.md` could remain as a deprecated compatibility alias, with using both names in one directory treated as an error. The migration could be a follow-up, but it would be awesome to remove these two almost identically named files that do different things, currently requiring arcane knowledge to distinguish.

If `pages` and `subsections` remain separate, could the common `Page` template object at least expose its existing `weight`? A theme could then merge and sort the lists without duplicating each weight under `extra`, and Zola would not need to change `sort_by`.

Keats:
> Why not go further down this path, in this RFC or a follow-up, and make every content node a Page? self.pages could contain every direct child Page, including directory pages, with kind distinguishing something like Leaf and Branch.

So this makes sense when you're sorting things by weight. Sorting by the other things like date, a section might not have a meaningful value so you might have to just make stuff up.
It looks like Hugo exposes 3 vecs on a page: `.Pages` (all children including sections), `.RegularPages` (normal pages only) and `.Sections` (our subsections). I'd need to check how they handle the sorting.

> The same cleanup seems possible for index.md and _index.md

I personally wouldn't touch that. Or maybe as a much much later RFC but you'd have to convince me.

> could the common Page template object at least expose its existing weight?

Sure

bemyak:
> It looks like Hugo exposes 3 vecs on a page: .Pages (all children including sections), .RegularPages (normal pages only) and .Sections (our subsections). I'd need to check how they handle the sorting.

This is one of the reasons I don't use Hugo :) These nuances are hard for me to remember, I had to check the docs every time.

The whole point of the RFC was to stick with one entity to make users' life easier. Simplifying internal representation and fixing some bugs along the way was a nice side effect.

In general, I think implementing this as-is would help a lot. I would still use the chance to clear the `_index.md` vs `index.md` dichotomy. I don't like how similar they are, but at the same time annoyingly different :slightly_smiling_face: 

Also, in the RFC as written, get_page doesn't seem to check kind — am I reading it right that it can return sections too? If so, that's the single getter I was hoping for.

Keats:
> Also, in the RFC as written, get_page doesn't seem to check kind — am I reading it right that it can return sections too? 

Yes it would be able to return them since everything is a page and we wouldn't add the same check we do for get_section

Keats:
My issues with merging pages and section into a single field are:

1. How to sort sections? By weight it makes sense and it's the same logic as most documentation site but what about dates. Something like
```
blog/
├── _index.md        sort_by = "date"
├── post-a.md        date = 2026-08-25
├── post-b.md        date = 2026-08-20
└── archive/
    └── _index.md
```
Where does the archive go? Do we have to add a date just to make it work?

2. How to handle subsections in pagination? Are sections part of the pagination? In the above example with `paginate_by = 2` would there be a page with archive? We can't ignore them in the templates because pagination is calculated before template rendering. I guess we could exclude sections from pagination?

The only way I think this can work is if:
a. Sections don't participate in sort_by applied to the regular pages and are excluded from pagination
b. You can define `weight` without `sort_by`
c. You have 2 fields on the `current` item: `pages` (the same as currently) and `children` which contains both pages and subsection but sorted by weight. So you split sorted by users from the "tree" view.
In some cases this mean you might end up with some pages with both a `date` and a `weight` for example. I don't know if it would happen frequently though.

3. Transparent sections: we do not want those to appear as children of the parent structurally, so we cannot just show the current pages as a single field otherwise and represent structural children. Pages passed via `transparent` are shown as if they were direct children which they are not. We'd still need pages/sections or pages/children like above.


Overall I think merging into a single field does not make sense because it doesn't just represent all the direct children, it's a collection (maybe sorted, with lower/higher), that can be paginated and doesn't really work well when mixed. 


IMO replacing `_index.md` with `index.md` would be a much worse outcome overall so let's not discuss it in this RFC.

bemyak:
I have some answers to these, but the more I think about it, the more it's apparent that it's not the Zola's way. I'm working on my overengineered SSG'ish thing, and steering Zola that way isn't a good idea :slightly_smiling_face: 

I agree with your points. Pagination must be done pre-render, because it affects page paths. We _could_ process pages in hierarchical order, but it overcomplicates things with little benefit. So, if pagination is pre-render so should be sorting, since it should be done prior.

No objections anymore from my side, I think this will be a good step forward!

PR #3261, opened 2026-08-23
Build render cache before rendering Markdown

### LINKED ISSUE #3260: `get_taxonomy_url` fails in Markdown components
# Bug Report

## Environment

Zola version: `0.23.4` and current `master`

## Expected Behavior

`get_taxonomy_url` should work in components used from Markdown content.

## Current Behavior

It fails for an existing taxonomy:

```
`get_taxonomy_url` received an unknown taxonomy as kind: categories
```

## Step to reproduce

```jinja2
{% component taxonomy_url() %}
{{ get_taxonomy_url(kind="categories", term="a-category") }}
{% endcomponent %}
{{ <taxonomy_url /> }}
```

ccjmne:
Content components can call `get_page`, `get_section`, and taxonomy functions while Markdown is rendered. Those functions captured the initial empty `RenderCache`, however, because the cache was only populated after Markdown rendering completed.

Build the cache once from parsed content before registering the early global functions, then retain the existing rebuild after Markdown rendering so final templates receive rendered content and backlinks.

Add an integration assertion that resolves a taxonomy URL from a content component.

Fixes #3260.

## Code changes

* [x] Was a LLM used? If so, for what?
  GitHub Copilot to figure out how to make the smallest change, and to ascertain that this was indeed a regression in the first place.

### DIFF
--- a/components/site/src/lib.rs
+++ b/components/site/src/lib.rs
@@ -343,6 +343,8 @@ impl Site {
         // taxonomy Tera fns are loaded in `register_early_global_fns`
         // so we do need to populate it first.
         self.populate_taxonomies()?;
+        // Build the cache for Tera functions used in Markdown content.
+        Arc::make_mut(&mut self.cache).build(&self.library, &self.taxonomies, &self.tera);
         tpls::register_early_global_fns(self);
         self.render_markdown()?;
         Arc::make_mut(&mut self.library).fill_backlinks();
--- a/components/site/tests/site.rs
+++ b/components/site/tests/site.rs
@@ -144,6 +144,11 @@ fn can_build_site_without_live_reload() {
     assert!(file_contains!(public, "posts/python/index.html", "Basic shortcode"));
     assert!(file_contains!(public, "posts/python/index.html", "Arrrh Bob"));
     assert!(file_contains!(public, "posts/python/index.html", "Arrrh Bob_Sponge"));
+    assert!(file_contains!(
+        public,
+        "posts/python/index.html",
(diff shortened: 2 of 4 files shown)


components/site/src/lib.rs
```diff
@@ -343,6 +343,8 @@ impl Site {
         // taxonomy Tera fns are loaded in `register_early_global_fns`
         // so we do need to populate it first.
         self.populate_taxonomies()?;
+        // Build the cache for Tera functions used in Markdown content.
+        Arc::make_mut(&mut self.cache).build(&self.library, &self.taxonomies, &self.tera);
```
Keats, line 347: We don't want to rebuild the cache just to get the taxo url functions working. Are there others that are not working for you?

I explicitly don't _really_ want to expose the get_page/get_section because it's not fully filled yet.

Keats (changes requested):

components/site/src/lib.rs
ccjmne: Aaah yeah... that makes sense. Oops.

I only need `get_taxonomy_url`, which was available in shortcodes before 0.23. I’m not confident enough in the preferred internal design to revise this correctly myself...

I ran some LLM trying to implement a simple way to restore that more specific behaviour, and the resulting patch does do the job. It's even actually quite elegant, as far as I can follow, but I just am not familiar enough with the codebase to stand behind it 100% :-/

Would it still be useful if I pushed it for review?

Keats:
Fixed in `next` branch. For now scoped to get_taxonomy_url as I'm not sure about exposing the rest while it's processing.

PR #3116, opened 2026-02-27
Add support for wikilinks

Keats:
Close #2104

PR #3268, opened 2026-09-02
Resolve qualified wikilinks to colocated assets

wakamex:
As discussed in #3116, this PR lets wikilinks such as `[[source.pdf]]` resolve to colocated assets already addressable through Zola's `@/` links.

Asset lookup accepts a bare filename when it is unique. `[[guides/source.pdf]]` remains available as the exact content-root-relative form, and duplicate filenames produce an ambiguity error with qualified candidates. Non-Markdown assets still require their extension.

Unlike the earlier output-target version I linked in #3116, the resolver stores content and asset paths in one typed target index rather than storing emitted URLs. A resolved asset path passes through Zola's existing language-aware `@/` lookup. Asset wikilinks do not create backlinks because they point to assets rather than Markdown content.

The existing resolver and wikilink tests cover exact paths, unique filenames, ambiguous filenames, required extensions, excluded unrendered owners, and backlink behavior. The existing multilingual asset fixture verifies bare-filename links with the emitted English and French URLs.

AI was used for writing the code. GPT-5.6-Sol w/ high effort and a touch of sass. Reviewed by human.

No documentation exists to update yet on the wikilinks branch.

### DIFF
--- a/components/markdown/src/wikilinks.rs
+++ b/components/markdown/src/wikilinks.rs
@@ -6,24 +6,40 @@ pub enum WikilinkError {
     Ambiguous { candidates: Vec<String> },
 }
 
-/// Resolves content wikilinks by source path, output alias, or bare stem.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum WikilinkTarget {
+    Content(String),
+    Asset(String),
+}
+
+impl WikilinkTarget {
+    fn identity(&self) -> &str {
+        match self {
+            Self::Content(path) => identity(path),
+            Self::Asset(path) => path,
+        }
+    }
+}
+
+/// Resolves content wikilinks by source path, output alias, or bare stem, and asset wikilinks by
+/// path or bare filename.
 ///
-/// For each target, the resolver indexes three forms that point back to the full source path:
+/// Content targets are indexed by three forms that point back to the full source path:
 ///
 /// 1. The source path without its `.md` extension, such as `docs/overview`.
 /// 2. Each existing Zola alias, with surrounding slashes removed, such as `overview-old`.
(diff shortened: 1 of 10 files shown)


Keats:
> Asset lookup uses the exact content-root-relative asset path without adding bare-filename matching: [[guides/source.pdf]] resolves, but [[source.pdf]] does not.

Why not? I _think_ (but haven't checked) Obsidian works that way

wakamex:
Good point, I think it was mostly for PR brevity. I checked Obsidian, Foam, and Quartz, and they resolve a bare asset filename when it is unique. I updated this so `[[source.pdf]]` works when unique. And it reports duplicates using the new disambiguation path during build. Exact paths such as `[[guides/source.pdf]]` still work, and URL generation remains language-aware.

components/markdown/src/wikilinks.rs
```diff
@@ -24,6 +25,9 @@ pub struct WikilinkResolver {
     paths: AHashMap<String, Vec<usize>>,
     aliases: AHashMap<String, Vec<usize>>,
     stems: AHashMap<String, Vec<usize>>,
+    assets: Vec<String>,
+    asset_paths: AHashMap<String, usize>,
+    asset_names: AHashMap<String, Vec<usize>>,
```
Keats: I don't understand why it's split. It could be stored in the same place as the other permalinks and you just have something like:

```
#[derive(Clone, Debug)]
enum Kind { Content, Asset }

#[derive(Clone, Debug)]
struct Target { path: String, kind: Kind }

...
sources: Vec<Target>
```
(probably rename sources)

and you have a single `resolve` fn that works for everything

Keats (changes requested):

Keats:
Thanks!

PR #3270, opened 2026-09-05
Resolve wikilinks to localized taxonomy terms

wakamex:
This adds taxonomy-term wikilinks: `[[tags/evidence]]` and an unambiguous `[[evidence]]` select the current language, then fall back to the default language. Missing destinations and ambiguous names use the existing errors. Only rendered terms are indexed, and these links preserve fragments without creating content backlinks.

I register each taxonomy/term slug identity once in the existing resolver and keep localized destinations in a separate map. The Markdown filter also honors `lang` in taxonomy templates, where there is no `page` or `section` object.

All 166 Markdown, site, and template tests pass, with one existing test ignored. The multilingual fixture covers authored pages and taxonomy list/term templates; it caught English URLs on French taxonomy output before the filter fix. Formatting and workspace compilation checks pass.

[RFC #3258](https://github.com/getzola/zola/pull/3258) would consolidate the page and section rendering paths touched here. A later extension that gives taxonomy terms common page identity and language-aware lookup could remove this adapter while preserving the link syntax and resolution behavior.


Keats:
I don't think we want to add that, it's not supported by normal links either. You need to call `get_taxonomy_url` instead.

PR #3275, opened 2026-09-11
Replace grass 0.13 with accent-sass 0.15

jomiq:
[`grass 0.13` does not support the "new" color functions](https://github.com/connorskees/grass/issues/105). This causes compilation to fail when using frameworks as Bulma, Pico and probably others too. 

[`accent-sass`](https://github.com/zoosky/accent-sass) supports these functions. 

`accent-sass` is a recent fork of `grass` and as such is a straight drop-in replacement. 

## Included in this PR
* Replaces references to `grass`
* Update `sass.md`
* `Cargo.lock` updated


Keats:
Hmm the changes are fully vibecoded I'll need to spend some time looking at it

Keats:
Lol 12k more loc in 3 days, it was already something like +25k compared to grass when I checked. At least Sass has a big test suite so it should be fairly correct? Who knows.
I'll wait a bit more

jomiq:
Look, I don't want to promote slop, close and delete if you want. 

Do you know if there is any work done on Grass to fix the color functions issue?

Keats:
grass is not maintained anymore afaik. I just saw https://github.com/momiji-rs/sasso from crates.io which seems also LLM generated.

For something like Sass I'm kinda ok with something vibecoded since it has a big test suite and is low risk

jomiq:
I took a peek at sasso. My two cents is that accent-sass is vibing in a slightly more sensible fashion. You could take this PR as a low-risk stopgap.

PR #3282, opened 2026-09-14
fix(templates): normalize UNC device prefix in glob_base for network drives

### LINKED ISSUE #3229: bug: template not found
# Bug Report

## Environment

Zola version: 0.23.2

## Expected Behavior
The site should build with templates.

## Current Behavior
```
ERROR Failed to build the site
ERROR Failed to render page '\\?\C:\path\to\website\content\foo.md'
ERROR Reason: Template `foo.html` not found
```

I believe this is a Windows issue, specifically UNC paths with `\\?\`, but that's a guess based on the error message (and asking Claude).

## Steps to reproduce
On a Windows machine, 
- Initiate an empty zola project.
- Add `foo.md` to `content/` with `template = "foo.html"`.
- Add `foo.html` to `templates/` with skeleton html.
- Run `zola build`.

TrueFurina:
Fixes #3229 (the part still reproducible in 0.23.4 on UNC / network-drive paths).

## Problem

`glob_base` in `components/templates/src/lib.rs` relies on `dunce::simplified` to strip
the Windows verbatim prefix so that glob patterns match. That works for local drives
(`\\?\C:\site` -> `C:\site`), but `dunce` deliberately leaves network paths in their
verbatim UNC form `\\?\UNC\server\share\site` untouched, so the constructed glob pattern
never matches and every template under `templates/` is reported missing
(`Template \`foo.html\` not found`). This matches the reports in #3229: project on a
network/mapped drive fails, same project on a local drive works.

## Fix

`glob_base` now additionally rewrites the `\\?\UNC\` prefix to the plain `\\server\share`
form, which globs handle correctly. Local-drive behavior via `dunce::simplified` is
unchanged, and non-Windows paths are untouched.

## Verification

- Standalone harness exercising `glob_base`'s exact logic against `dunce 1.0.5`: 4/4 passed
  (local verbatim drive path, `\\?\UNC\server\share` network path, plain unix path, relative path)
- Root cause independently confirmed by kbvernon in #3229 (`dunce::simplified()` does not
  modify `\\?\UNC\...` paths)
- Docs/comment-only behavioral scope: single function in `components/templates`


Keats:
Can you reproduce the bug on a windows setup? I'm wondering if the issue is not just using `/` in https://github.com/getzola/zola/blob/master/components/templates/src/lib.rs#L107 instead of Path::join() and  dunce was an incorrect fix. We'd need some kind of test either way somewhere

TrueFurina:
Friendly ping - this is still open on my side and mergeable as far as I can tell. Happy to address any remaining feedback if something is blocking it. Thanks!

Keats:
Did you see my comment above?

TrueFurina:
Sorry for the generic ping earlier — I should have responded to your review directly. You're right, and I've reworked the branch (`57d3055`) along that line:

1. **Glob patterns are now built with `Path::join`** instead of `format!("{}/templates/...")` string formatting — both call sites in `load_tera`. That's the root cause you pointed at: a hard-coded `/` glued onto an already-platform-shaped base path.

2. **The UNC conversion stays, but as a separate testable function** (`strip_unc_prefix`), because `Path::join` alone can't fix a `\\?\UNC\server\share` base — `dunce` never strips that prefix, so the glob still sees an invalid path even with correct separators.

3. **Added unit tests** for `strip_unc_prefix` (UNC → `\\server\share`, drive paths pass through, plain `\\server\share` passes through) — pure string logic, so they run on any platform, not just Windows.

I couldn't reproduce on a full network-share setup locally, but the failing shape is `\\?\UNC\server\share` + `/`-joined glob, which matches the mapped-drive symptom. If the Windows CI job is green with the new tests, does that cover the test requirement you had in mind, or would you prefer an integration-style test around `load_tera` as well?

Keats:
Do we still need `strip_unc_prefix` ? Ideally we would have a test for load_tera with a problematic path if that's the actual cause

